PR #4, opened 2015-10-15
Prevent non-fatal errors from closing the supervisor.

ljskernel:
connects to #59 

This patch fixes a couple of cases where `log.Fatalf` would cause the supervisor
to exit immediately when it's inappropriate to do so. `log.Fatalf` and
co. should not be used unless the supervisor utterly cannot run without
whatever's being checked being in place.

This patch also adjusts the code that relied on the module's values being
created here being available, they now check and send an appropriate error
message if they're not there.

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/4)


lifeeth:
Checked by manually making sure the dbus wasnt available - gosuper did not die.

PR #5, opened 2015-10-15
Don't bind mount docker.sock

pcarranzav:
It doesn't exist on devices that use rce - plus using the host docker/rce doesn't make sense, users should use docker-in-docker. Supervisor would delete any user-created images anyways.


lifeeth:
LGTM

PR #8, opened 2015-10-28
Implement several update strategies

pcarranzav:
* Break up the update function into manageable parts
* Refactor the special variables into the RESIN_SUPERVISOR_ namespace
* Use three update strategies: normal, kill before download, hand over
* Do not log anything in case of a locked app

Needs discussion: Env Var names and namespaces are TBD.

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/8)

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -270,7 +270,9 @@ application.lockUpdates = lockUpdates = do ->
 				.catch ENOENT, _.noop
 				.catch (err) ->
 					release()
-					throw new Error("Updates are locked: #{err.message}")
+					err = new Error("Updates are locked: #{err.message}")
+					err.isLocked = true
+					throw err
 		.disposer (release) ->
 			Promise.try ->
 				lockFile.unlockAsync(lockName) if force != true
@@ -297,11 +299,13 @@ apiPollInterval = (val) ->
 	application.poll()
 
 specialActionEnvVars =
-	'RESIN_OVERRIDE_LOCK': null
-	'RESIN_VPN_CONTROL': utils.vpnControl
-	'RESIN_CONNECTIVITY_CHECK': utils.connectivityCheck
-	'RESIN_POLL_INTERVAL': apiPollInterval
-	'RESIN_LOG_CONTROL': utils.resinLogControl
+	'RESIN_SUPERVISOR_UPDATE_STRATEGY': null
+	'RESIN_SUPERVISOR_HANDOVER_TIMEOUT': null
+	'RESIN_SUPERVISOR_OVERRIDE_LOCK': null
+	'RESIN_SUPERVISOR_VPN_CONTROL': utils.vpnControl
+	'RESIN_SUPERVISOR_CONNECTIVITY_CHECK': utils.connectivityCheck
+	'RESIN_SUPERVISOR_POLL_INTERVAL': apiPollInterval
+	'RESIN_SUPERVISOR_LOG_CONTROL': utils.resinLogControl
 
 executedSpecialActionEnvVars = {}
 
@@ -322,6 +326,17 @@ wrapAsError = (err) ->
 	return err if _.isError(err)
 	return new Error(err.message ? err)
 
+selectAndKill = (appId) ->
+	knex('app').select().where({ appId })
+	.then ([ app ]) ->
+		if !app?
+			throw new Error('App not found')
+		kill(app)
+
+waitToKill = (app, timeout) ->
+	# TO-DO
+	# Wait for app to signal it's ready to die, or timeout to complete (if it is defined and not-empty)
+
 UPDATE_IDLE = 0
 UPDATE_UPDATING = 1
 UPDATE_REQUIRED = 2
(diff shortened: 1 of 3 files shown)


src/application.coffee
```diff
@@ -270,7 +274,9 @@ application.lockUpdates = lockUpdates = do ->
 				.catch ENOENT, _.noop
 				.catch (err) ->
 					release()
-					throw new Error("Updates are locked: #{err.message}")
+					err = new Error("Updates are locked: #{err.message}")
+					err.isLocked = true
```
Page-: It would be nicer to use a `typed-error` here and `throw new UpdatesLockedError("Updates are locked: #{err.message}")` or similar

src/application.coffee
```diff
@@ -322,6 +331,28 @@ wrapAsError = (err) ->
 	return err if _.isError(err)
 	return new Error(err.message ? err)
 
+selectAndKill = (appId) ->
```
Page-: This could be:
```coffee
	knex('app').select().where({ appId })
	.then ([ app ]) ->
		if !app?
			throw new Error('App not found')
		return app
```
and be used as `select(appId).then(kill)` or for https://github.com/resin-io/resin-supervisor/pull/8/files#diff-3e54a04880f9ac9cf4ef339b4c931ef6R392 as `select(localApp.appId).then(fetch)  if needsDownload`

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
 	forceNext: false
 	intervalHandle: null
 
+updateStrategies =
+	'normal-update': (localApp, app, needsDownload, force, timeout) ->
```
Page-: It'd be nice to name this based on its behaviour like for the others, that way the decision of what's "normal" is within the api and can be changed more easily if need be.  Not to mention it acts as documentation for this bit of code.

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
...
+		.then ->
+			Promise.using lockUpdates(localApp, force), ->
+				logSystemEvent(logTypes.updateApp, app) if localApp.imageId == app.imageId
+				selectAndKill(localApp.appId)
+				.then ->
+					start(app)
+			.catch (err) ->
+				logSystemEvent(logTypes.updateAppError, app, err) unless err.isLocked?
+				throw err
+	'kill-before-download': (localApp, app, needsDownload, force, timeout) ->
```
Page-: It feels like the signature for these functions might be better as an options object, ie `({ localApp, app, needsDownload, force, timeout })`

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
 	forceNext: false
 	intervalHandle: null
 
+updateStrategies =
```
Page-: There's a fair chunk of duplicated code between these strategies, I'm not sure it's particularly avoidable though since it's largely plumbing, except maybe a shared `.catch` handler.  Basically just keep it in mind, although I'm not sure you can do much about it

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
...
+	'normal-update': (localApp, app, needsDownload, force, timeout) ->
+		Promise.try ->
+			fetch(app) if needsDownload
+		.then ->
+			Promise.using lockUpdates(localApp, force), ->
+				logSystemEvent(logTypes.updateApp, app) if localApp.imageId == app.imageId
```
Page-: Where this log happens seems to vary between each update strategy, so I'm not sure what stage of an update in general it's meant to signify?

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
...
+				waitToKill(localApp, timeout)
+			.then ->
+				kill(localApp)
+		.catch (err) ->
+			logSystemEvent(logTypes.updateAppError, app, err) unless err.isLocked?
+			throw err
+
+
+updateUsingStrategy = (strategy, localApp, app, needsDownload, force, timeout) ->
+	if strategy not in _.keys(updateStrategies)
```
Page-: `if not _.has(updateStrategies, strategy)`

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
...
+		strategy = 'normal-update'
```
Page-: Maybe give a warning in the case of defaulting?

src/application.coffee
```diff
@@ -332,6 +363,122 @@ updateStatus =
...
+			apikey: apiKey
+
+getEnvAndFormatRemoteApps = (deviceId, remoteApps, uuid, apiKey) ->
+	Promise.map remoteApps, (app) ->
+		getEnvironment(app.id, deviceId, apiKey)
+		.then (environment) ->
+			app.environment_variable = environment
+			utils.extendEnvVars(app.environment_variable, uuid)
+		.then (env) ->
+			fullEnv = env
```
Page-: It would make more sense to just do `.then (fullEnv) ->`

src/application.coffee
pcarranzav: Hmm to the local logs? Cause sending to the dashboard logs sounds like it'd be noisy for users that aren't using this feature.

src/application.coffee
pcarranzav: It's done immediately before we affect the user's app, that is, before we kill or start (whichever comes first)

src/application.coffee
pcarranzav: Yup, the thing is what changes is mainly the order in which things and done... but true, something to keep in mind and possibly DRY out if we add more strategies in the future...

pcarranzav:
@Page- comments addressed, and I've tested the handover strategy (that's the trickiest one) in my devenv. Will test the other two now.

pcarranzav:
Tested the two other strategies now, and they work :)

src/application.coffee
Page-: ok

Page-:
Reviewed 2 of 2 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

_[Makefile, line 42 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UGrkc90VMkQ6DR2am:-K3UGrkc90VMkQ6DR2an:-1751225298) ([raw file](https://github.com/resin-io/resin-supervisor/blob/f483f93211401a934165bf028e5827146b311c32/Makefile#L42)):_
You can do
```
cp 01_nodoc tools/dind/config/
touch -t 7805200000
```
instead, in order to stop the cache invalidations (nice catch btw, I was wondering about it myself but didn't dig into why)

---

_[src/application.coffee, line 388 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UL9WlfCntyE3fEhZ_:-K3UL9WlfCntyE3fEhZa:-2080092748) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L388)):_
This should probably be inside the `Promise.using` to match `download-then-kill` in not showing an `Updating application...` with no matching finish/failure (in the case of updates being locked)

---

_[src/application.coffee, line 498 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UI87do8f9SzTfgbAq:-K3UI87do8f9SzTfgbAr:289878296) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L498)):_
Can this be a `Promise.join` please

---

_[src/application.coffee, line 503 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UIGQ9vJR1MBB4FtsA:-K3UIGQ9vJR1MBB4FtsB:-764373702) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L503)):_
This could be `.then ([ [ value: apiKey ], [ value: uuid ], apps ]) ->`

---

_[src/application.coffee, line 540 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UIZzbf5iWTYvJiUhM:-K3UIZzc_o70tEI2ymMf:-899770962) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L540)):_
selectAndKill doesn't exist anymore

---

_[src/application.coffee, line 573 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UIjvtyui6qaAapAm1:-K3UIjvtyui6qaAapAm2:1696767409) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L573)):_
Debug?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8)_

pcarranzav:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

_[src/application.coffee, line 540 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UIZzbf5iWTYvJiUhM:-K3WvMEzSz_YghEqMiQC:-276414660) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L540)):_
Nice catch :S

---

_[src/application.coffee, line 573 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UIjvtyui6qaAapAm1:-K3Wv4HMIQ0X0-sP5v6U:-1865212441) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L573)):_
Initially yes, but I think it will actually prove useful if something fails on a device.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8)_

Page-:
Reviewed 1 of 2 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[src/application.coffee, line 573 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3UIjvtyui6qaAapAm1:-K3ZF2R_Ho0MW2x-rh8i:1592432539) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8258d679a87d885ae1d729bbd7e30fa1aa5bfb3/src/application.coffee#L573)):_
I think it'd be more useful to do `_.each(failures, (err) -> console.error('Error: ', err, err.stack))` since `console.trace` will print the stack trace of this line (not that useful) along with the error message (useful, but without the stack trace could be cryptic)

---

_[src/application.coffee, line 491 \[r4\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3ZFdfZErzKj0ns8Qzs:-K3ZFdfZErzKj0ns8Qzt:722450461) ([raw file](https://github.com/resin-io/resin-supervisor/blob/59ac3cfb28e909679b447143f65c56da89239aee/src/application.coffee#L491)):_
I think you should be able to add `.get(0).get('value')` and make the calling sites even nicer

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8)_

Page-:
Reviewed 1 of 1 files at r5.

---

_[src/application.coffee, line 502 \[r5\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8#-K3pEeiKP6hhwK2fhoD5:-K3pEeiKP6hhwK2fhoD6:1929829889) ([raw file](https://github.com/resin-io/resin-supervisor/blob/d8dcb652dbe34e237ebb5fce17444904cb5e46d6/src/application.coffee#L502)):_
If you put this on the line before you can drop one level of indentation in the body here

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8)_

Page-:
Reviewed 1 of 1 files at r6.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/8)_

PR #9, opened 2015-10-28
Use random secret for logs channel

pcarranzav:
And fix reporting of new API secret upon regeneration.

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -13,7 +13,7 @@ module.exports = (application) ->
 	api = express()
 	api.use(bodyParser())
 	api.use (req, res, next) ->
-		utils.getOrGenerateSecret('apiSecret')
+		utils.getOrGenerateSecret('api')
 		.then (secret) ->
 			if req.query.apikey is secret
 				next()
@@ -129,7 +129,7 @@ module.exports = (application) ->
 	# Expires the supervisor's API key and generates a new one.
 	# It also communicates the new key to the Resin API.
 	api.post '/v1/regenerate-api-key', (req, res) ->
-		utils.newSecret('apiSecret')
+		utils.newSecret('api')
 		.then (secret) ->
 			device.updateState(api_secret: secret)
 			res.status(200).send(secret)
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -13,7 +13,7 @@ module.exports = (application) ->
 	api = express()
 	api.use(bodyParser())
 	api.use (req, res, next) ->
-		utils.getOrGenerateApiSecret()
+		utils.getOrGenerateSecret('apiSecret')
 		.then (secret) ->
 			if req.query.apikey is secret
 				next()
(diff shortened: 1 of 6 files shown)


src/utils.coffee
```diff
@@ -102,27 +102,28 @@ exports.connectivityCheck = _.once ->
 				blink.pattern.start(networkPattern)
 
 
-apiSecretPromise = null
-generateApiSecret = ->
+secretPromises = {}
+generateSecret = (name) ->
 	Promise.try ->
-		return config.forceApiSecret ? randomHexString.generate()
+		return config.forceApiSecret if name == 'apiSecret' && config.forceApiSecret?
```
petrosagg: You made the function generic, but this if in here feels dirty. A separate specific function for `'apiSecret'` a way to resolve this.

src/utils.coffee
```diff
@@ -102,27 +102,28 @@ exports.connectivityCheck = _.once ->
...
+		knex('config').update(secretInDB).where(key: name)
 		.then (affectedRows) ->
 			knex('config').insert(secretInDB) if affectedRows == 0
 		.return(newSecret)
 
-exports.newApiSecret = newApiSecret = ->
-	apiSecretPromise ?= Promise.resolve()
-	apiSecretPromise = apiSecretPromise.then ->
-		generateApiSecret()
+exports.newSecret = newSecret = (name) ->
```
Page-: This local `newSecret` define doesn't seem to be used anywhere?

src/utils.coffee
Page-: Alternatively you could have
```coffee
config.forceSecret =
	apiSecret: '...'
	other: '...'

return config.forceSecret[name] if config.forceSecret[name]?
```

src/utils.coffee
```diff
@@ -131,7 +132,7 @@ exports.extendEnvVars = (env, uuid) ->
 		RESIN_SUPERVISOR_ADDRESS: "http://#{host}:#{config.listenPort}"
 		RESIN_SUPERVISOR_HOST: host
 		RESIN_SUPERVISOR_PORT: config.listenPort
-		RESIN_SUPERVISOR_API_KEY: exports.getOrGenerateApiSecret()
+		RESIN_SUPERVISOR_API_KEY: exports.getOrGenerateSecret('apiSecret')
```
Page-: This would make more sense as just `'api'`, since the fact that it's a secret is already clear from the function name

pcarranzav: The thing is the same string is also used as a key to store the secret in the DB, and I'd rather keep it explicit there.

Page-: `knex('config').update(secretInDB).where(key: "#{name}Secret")`?

pcarranzav: Fair enough :grin:

pcarranzav: @Page- done.

src/api.coffee
```diff
@@ -129,9 +129,9 @@ module.exports = (application) ->
 	# Expires the supervisor's API key and generates a new one.
 	# It also communicates the new key to the Resin API.
 	api.post '/v1/regenerate-api-key', (req, res) ->
-		utils.newApiSecret()
+		utils.newSecret('api')
 		.then (secret) ->
-			device.updateState(apikey: secret)
+			device.updateState(api_secret: secret)
```
Page-: Why did you change this property name? It needs to match up with the api

Page-: Hmm, I guess it's been changed so that it now matches up

Page-:
LGTM

src/api.coffee
pcarranzav: Yeah, in app.coffee we always had it as api_secret, I messed up when I added this endpoint

PR #70, opened 2016-01-08
Revert "Make config.json writes atomic"

pcarranzav:
Reverts resin-io/resin-supervisor#64

config.json is bind mounted (the file, and not its directory), so this doesn't work and breaks provisioning. Need to think an alternative way of doing this.

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/70)


lorenzo-stoakes:
LGTM

PR #89, opened 2016-01-27
Quickfix: use String .split instead of _.split when getting Host OS version

pcarranzav:
lodash 3 doesn't have the _.split function :S

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/89)


abresas:
LGTM

Issue #91, opened 2016-02-03
Use default config values when env var is an empty string


pcarranzav:
Need to add checkValidKey for endpoint and port values

pcarranzav:
(Or checkInt if appropriate)

pcarranzav:
(in [config.coffee](https://github.com/resin-io/resin-supervisor/blob/master/src/config.coffee#L19))

PR #94, opened 2016-02-12
Check for valid strings or ints in all config values

pcarranzav:
connects to #91 

When the host OS passes us the config env vars, it does so after [using jq on config.json](https://github.com/resin-os/meta-resin/blob/master/meta-resin-common/recipes-support/resin-conf/resin-conf/resin-vars#L44-L56), so values that aren't present can appear as a string that says "null". We add some validation so that we correctly use the default values in this case.

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/94)

### DIFF
--- a/src/config.coffee
+++ b/src/config.coffee
@@ -7,38 +7,39 @@ checkInt = (s) ->
 		return
 	return i
 
-checkValidKey = (s) ->
-	# Make sure `s` exists and is not an empty string, or 'null'.
-	if !s? or s == 'null' or s == ''
+checkString = (s) ->
+	# Make sure `s` exists and is not an empty string, or 'null' or 'undefined'.
+	# This might happen if the parsing of config.json on the host using jq is wrong (it is buggy in some versions).
+	if !s? or s == 'null' or s == 'undefined' or s == ''
 		return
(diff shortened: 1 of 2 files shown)


src/config.coffee
```diff
@@ -7,38 +7,38 @@ checkInt = (s) ->
 		return
 	return i
 
-checkValidKey = (s) ->
+checkString = (s) ->
 	# Make sure `s` exists and is not an empty string, or 'null'.
-	if !s? or s == 'null' or s == ''
+	if !s? or s == 'null' or or s == 'undefined' or s == ''
```
abresas: is "undefined" also a possible string? 

I'd also prefer we have some sort of explanation that this is due to a bug in OS or something, but that's up to you.

LGTM otherwise.

pcarranzav: I really don't know, but I wanted to proactively protect from possible jq weirdness.
I'll add the comment and merge, cheers

abresas:
LGTM

pcarranzav:
Tested with dind, correctly falls back to defaults, and correctly uses the env var values if set.

PR #96, opened 2016-02-15
Remove quotes in OS version

pcarranzav:
connects to #95 

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/96)

### DIFF
--- a/src/device.coffee
+++ b/src/device.coffee
@@ -186,7 +186,8 @@ exports.getOSVersion = ->
 		for line in lines
 			[ key, val ] = line.split('=')
 			releaseItems[_.trim(key)] = _.trim(val)
-		return releaseItems['PRETTY_NAME']
+		# Remove enclosing quotes: http://stackoverflow.com/a/19156197/2549019
+		return releaseItems['PRETTY_NAME'].replace(/^"(.+(?="$))"$/, '$1')
 	.catch (err) ->
 		console.log("Could not get OS Version: ", err, err.stack)
 		return undefined
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Remove quotes in OS version [Pablo]
+
 # v1.5.0
 
 * Add support for delta image download [petrosagg and Pablo]


src/device.coffee
```diff
@@ -186,7 +186,8 @@ exports.getOSVersion = ->
 		for line in lines
 			[ key, val ] = line.split('=')
 			releaseItems[_.trim(key)] = _.trim(val)
-		return releaseItems['PRETTY_NAME']
+		# Remove enclosing quotes: http://stackoverflow.com/a/19156197/2549019
+		return releaseItems['PRETTY_NAME'].replace(/^"(.+(?="$))"$/, '$1')
```
abresas: this is extremely overkill.

you can do `_.trim(val, ' \t"')` on the above line, or `.match(/^"(.*)"$/)[1]` here

pcarranzav: The first would fail if the string was something like `Ubuntu 18.04 "cool cactus"`, the second one works but the stackoverflow answer mentioned reports this one is slightly faster.

pcarranzav: (Ubuntu doesn't really report it like that, bug just in case...)

abresas: trim removes from start and end only

pcarranzav: Exactly, so in the cool cactus case, it would remove the one at the end and not in the middle, leaving an orphaned poor little quote there.

abresas: Yours does exactly the same! And I think it should too...

> console.log('Ubuntu "foo"'.replace(/^"(.+(?="$))"$/, '$1'))
Ubuntu "foo"

abresas: I just reread your comment, my bad, deleted previous

abresas: Your examples made me think I'm not exactly sure why that file has quotes and what the supposed syntax is, are there escaped quotes inside the quotes for example?

In any case, LGTM if you know that the syntax either has quotes at start and end or you don't care.

pcarranzav: I've only seen two cases: we either have quotes at start and end that we want remove, or we don't have quotes at all. I'm just using this to protect against cases of the string having quotes elsewhere, which we wouldn't want to remove. Especially since users can do their own OS builds.

Issue #32, opened 2015-11-30
Supervisor endpoint to check for pending updates


pcarranzav:
To complement the update lock feature, it would be good if apps could query the supervisor to see if there's a pending update ready to be applied, so that users can opt-in to updates.

PR #97, opened 2016-02-15
Add endpoint to get device state

pcarranzav:
connects to #32

[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/97)

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -136,4 +136,7 @@ module.exports = (application) ->
 		.catch (err) ->
 			res.status(503).send(err?.message or err or 'Unknown error')
 
+	api.get '/v1/device', (req, res) ->
+		res.json(device.getState())
+
 	return api
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -578,11 +578,13 @@ application.update = update = (force) ->
 			throw new Error(joinErrorMessages(failures)) if failures.length > 0
 		.then ->
 			updateStatus.failed = 0
+			device.setUpdatePending(false)
 			# We cleanup here as we want a point when we have a consistent apps/images state, rather than potentially at a
 			# point where we might clean up an image we still want.
 			dockerUtils.cleanupContainersAndImages()
 		.catch (err) ->
 			updateStatus.failed++
+			device.setUpdatePending(true)
 			if updateStatus.state is UPDATE_REQUIRED
 				console.log('Updating failed, but there is already another update scheduled immediately: ', err)
 				return
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -144,6 +144,7 @@ fetch = (app) ->
 		.then ->
 			logSystemEvent(logTypes.downloadAppSuccess, app)
 			device.updateState(status: 'Idle', download_progress: null)
+			device.setUpdateState(update_downloaded: true)
 			docker.getImage(app.imageId).inspectAsync()
 		.catch (err) ->
 			logSystemEvent(logTypes.downloadAppError, app, err)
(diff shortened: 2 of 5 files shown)


abresas:
I would recommend a closure with `do ->` around getState and updateState
to limit the targetState variable scope.

But it's not a deal-breaker, so LGTM.

src/api.coffee
```diff
@@ -136,4 +136,8 @@ module.exports = (application) ->
 		.catch (err) ->
 			res.status(503).send(err?.message or err or 'Unknown error')
 
+	api.get '/v1/device', (req, res) ->
+		state = device.getState()
+		res.status(200).send(JSON.stringify(state))
```
abresas: why not `res.json(state)` ?
it sets 200 by default and also the correct headers, and stringifies for you.

pcarranzav: @abresas cheers, done

src/device.coffee
```diff
@@ -128,13 +128,11 @@ exports.getDeviceType = do ->
...
 
-# Calling this function updates the local device state, which is then used to synchronise
-# the remote device state, repeating any failed updates until successfully synchronised.
-# This function will also optimise updates by merging multiple updates and only sending the latest state.
-exports.updateState = do ->
+do ->
 	applyPromise = Promise.resolve()
 	targetState = {}
 	actualState = {}
+	updatePending = false
```
abresas: isn't this the same variable as updateStatus.state in src/application.coffee? :/

pcarranzav: Not exactly - that one is tri-state and refers to the application.update function itself (works as a sort of self-mutex). This one explicitly gives a true/false response to whether there's an update that should be applied but the supervisor can't apply it, specifically what users want when using this together with the update lock.

Moreover, we can't require application.coffee here since we'd have circular requires.

pcarranzav: It would be equivalent to updateStatus.failed != 0, but we still have the circular require issue.

src/application.coffee
```diff
@@ -578,11 +578,13 @@ application.update = update = (force) ->
 			throw new Error(joinErrorMessages(failures)) if failures.length > 0
 		.then ->
 			updateStatus.failed = 0
+			device.setUpdatePending(false)
 			# We cleanup here as we want a point when we have a consistent apps/images state, rather than potentially at a
 			# point where we might clean up an image we still want.
 			dockerUtils.cleanupContainersAndImages()
 		.catch (err) ->
 			updateStatus.failed++
+			device.setUpdatePending(true)
```
Page-: It feels like maybe this should be at the start of an update? Since the update is pending until it's completed? (or is this meant to show something else?)

pcarranzav: The use case for this is a user opting in for an update when the update lock is set. So the supervisor will download the update, try to apply it and fail, then set this flag so the user knows there's an update pending installation.

pcarranzav:
Tested with dind, seems to work quite nicely.

Page-:
Reviewed 1 of 1 files at r3, 1 of 2 files at r4, 2 of 3 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/97)*

abresas:
Reviewed 1 of 1 files at r3, 1 of 2 files at r4, 2 of 3 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/device.coffee, line 135 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/97#-KAasbwE4-Dky2aREfA7:-KAasbwE4-Dky2aREfA8:-340954147) ([raw file](https://github.com/resin-io/resin-supervisor/blob/c9dceafb40febd08f8c193c12e6590b20907c8c6/src/device.coffee#L135)):*
isn't this variable the same as updateStatus.state in src/application.coffee? :/

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/97)*

pcarranzav:
*[src/device.coffee, line 135 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/97#-KAasbwE4-Dky2aREfA7:-KApKxfwk3p68VJ7T6yc:-117983141) ([raw file](https://github.com/resin-io/resin-supervisor/blob/c9dceafb40febd08f8c193c12e6590b20907c8c6/src/device.coffee#L135)):*
No, updateStatus.state refers to the application.update  function being running or not. These flags are set at different stages *within* that function, and have different meanings. And even if it were the same variable, we wouldn't be able to access it from here.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/97)*

Issue #44, opened 2015-12-10
Rewrite the whole supervisor  in Go


pcarranzav:

pcarranzav:
I'll mark this as "blocked" since we're currently re-designing the architecture for this.
Still hoping much of the code from the associated PR will be useful anyways.

pcarranzav:
Closing, there's several changes in the supervisor's future but they do not include a switch to Go anymore.

PR #98, opened 2016-02-22
Some small changes to make things more idiomatic

discordianfish:
[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/98)

### DIFF
--- a/gosuper/device/device.go
+++ b/gosuper/device/device.go
@@ -7,10 +7,10 @@ import (
 	"sync"
 	"time"
 
-	"resin-supervisor/gosuper/config"
-	"resin-supervisor/gosuper/resin"
-	"resin-supervisor/gosuper/supermodels"
-	"resin-supervisor/gosuper/utils"
+	"github.com/resin-io/resin-supervisor/gosuper/config"
+	"github.com/resin-io/resin-supervisor/gosuper/resin"
+	"github.com/resin-io/resin-supervisor/gosuper/supermodels"
+	"github.com/resin-io/resin-supervisor/gosuper/utils"
 )
 
 var Uuid string
@@ -25,6 +25,7 @@ type Device struct {
 	bootstrapLock sync.Mutex
 	Config        config.UserConfig
 	DbConfig      *supermodels.Config
+	SuperConfig   config.SupervisorConfig
 }
 
 func (dev Device) readConfigAndEnsureUuid(superConfig config.SupervisorConfig) (uuid string, err error) {
@@ -38,7 +39,7 @@ func (dev Device) readConfigAndEnsureUuid(superConfig config.SupervisorConfig) (
 	}
 	if err != nil {
 		time.Sleep(time.Duration(superConfig.BootstrapRetryDelay) * time.Millisecond)
-		return readConfigAndEnsureUuid(superConfig)
+		return dev.readConfigAndEnsureUuid(superConfig)
 	}
 	return
 }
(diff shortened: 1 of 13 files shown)


discordianfish:
Now that I tried to compile the code I realize that nothing is working yet. Did I use the wrong branch? There are syntax and logical errors all over the place. Probably work in progress?
Guess in this state it doesn't make sense to submit PRs, correct?

pcarranzav:
Hey @discordianfish - thanks for submitting this!
Indeed, this is WIP, and I'm still writing most of it so there's bound to be errors all around. I'd say it's early for stylistic/idiomatic PRs, but PRs that implement functionality (i.e. some of the packages) would be more than welcome :)

gosuper/gosuper/api.go
```diff
@@ -124,23 +125,37 @@ func PurgeHandler(writer http.ResponseWriter, request *http.Request) {
 
 	if appId, err := parsePurgeBody(request); err != nil {
 		sendBadRequest("Invalid request")
-	} else if appId == "" {
+		return
+	}
+	if appId == "" {
```
lorenzo-stoakes: This and other cases will fail due to `appId` not being available in the scope. A variable declared like `foo` in:

```go
if foo := bar; ... {
}
```

Are in scope only for the if block AND any else blocks. This is why the `if`/`else if` was used here to limit the scope of these variables.

Your code here will not compile because of `appId` not being declared.

discordianfish:
@pcarranzav Right now this doesn't even compile, so it's hard to figure out in which direction the non functional parts should go.

I would recommend to remove all pieces that don't work so we end up with something that compiles and iterate from there on.

lorenzo-stoakes:
@discordianfish your changes additionally will not compile, so regardless you would need to significantly change this patch for it to work regardless of changes on the WIP.

discordianfish:
@lorenzo-stoakes I know, I wanted to change this but realize there are compile errors all over the place so I closed this. I fixed a few locally but realize that there is a lot missing to make it work. Maybe someone forget to push some changes? Just try to compile it and you'll see what I mean.

lorenzo-stoakes:
Sure @discordianfish and it's a good point, it is pretty difficult to work with something that doesn't compile :) I'd say stick with the main branch for now for contribs while we sort this one out. Also I don't mean to be overly critical here - we are definitely very open to contributions!

discordianfish:
Actually I got it compiling now

gosuper/device/device.go
```diff
@@ -7,10 +7,10 @@ import (
 	"sync"
 	"time"
 
-	"resin-supervisor/gosuper/config"
-	"resin-supervisor/gosuper/resin"
-	"resin-supervisor/gosuper/supermodels"
-	"resin-supervisor/gosuper/utils"
+	"github.com/resin-io/resin-supervisor/gosuper/config"
```
pcarranzav: This isn't necessary if you use `make ARCH=amd64 gosuper` to compile

pcarranzav:
I'm actually working on the missing bits and pieces so that it builds correctly, please give me a while to get it to a usable state.

gosuper/device/device.go
discordianfish: @pcarranzav Ah okay, but why do you do the rewriting? I would also strongly recommend using the go native vendoring: https://golang.org/s/go15vendor over Godeps (which is IMO a PITA) ;)

discordianfish:
Sure! Maybe I can help you guys with this soon full time :)

gosuper/gosuper/api.go
discordianfish: Let's move this discussion to #44

Issue #101, opened 2016-03-02
Reduce noise in logs


lekkas:
Reduce noise by not logging handled exceptions.

pcarranzav:
:+1:
The main place where we log exceptions that are handled is in [dockerUtils.cleanupContainersAndImages](https://github.com/resin-io/resin-supervisor/blob/master/src/docker-utils.coffee#L195)

Think you can handle this one @lekkas ?

lekkas:
Sure, thanks for the tip @pcarranzav ! I'm already working on a separate branch for this (see https://github.com/resin-io/resin-supervisor/commit/a9c5a1ac7cf2178047619557869939283da1e3ee) . It's still a WIP, I'll revisit it since I went berzerk with removing logs and should probably keep more that I initially thought :)

pcarranzav:
Cool @lekkas - could  you rename the branch to start with the issue number so that waffle picks  it up as "In Progress"?

PR #102, opened 2016-03-04
Reduce noise in logs

lekkas:
[Review on Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/102)

connects to #101

### DIFF
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -191,8 +191,7 @@ do ->
 				docker.getContainer(containerInfo.Id).removeAsync()
 				.then ->
 					console.log('Deleted container:', containerInfo.Id, containerInfo.Image)
-				.catch (err) ->
-					console.log('Error deleting container:', containerInfo.Id, containerInfo.Image, err)
+				.catch(_.noop)
 			.then ->
 				# And then clean up the images, as long as we aren't currently trying to fetch any.
 				return if imagesBeingFetched > 0
@@ -204,8 +203,7 @@ do ->
 						docker.getImage(tag).removeAsync()
 						.then ->
 							console.log('Deleted image:', tag, image.Id, image.RepoTags)
-						.catch (err) ->
-							console.log('Error deleting image:', tag, image.Id, image.RepoTags, err)
+						.catch(_.noop)
 
 	containerHasExited = (id) ->
 		docker.getContainer(id).inspectAsync()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Reduce noise in logs [Kostas]
+
 # v1.6.0
 
 * Add endpoint to get device state [Pablo]


lekkas:
@pcarranzav Do you think we should also consider removing [this](https://github.com/resin-io/resin-supervisor/commit/b550708e983f8f91efea44d4d05e2da6d0ada896#diff-e616ed67248b05ac46c3606a0781fb9cL195) ? I have no easy way of checking if this is also polluting supervisor logs in real devices.

pcarranzav:
@lekkas yes, we should remove that one too

src/docker-utils.coffee
```diff
@@ -204,8 +204,7 @@ do ->
 						docker.getImage(tag).removeAsync()
 						.then ->
 							console.log('Deleted image:', tag, image.Id, image.RepoTags)
-						.catch (err) ->
-							console.log('Error deleting image:', tag, image.Id, image.RepoTags, err)
+						.catch (err) -> # Ignore image delete error
```
pcarranzav: No need for `err` here.
Also might look nicer by defining:
```
ignore  = ->
```
somewhere above,and then here (and in the other error to ignore) do this:
```
.catch(ignore)
```

lekkas: I'd choose the empty `catch ->` version over defining a new `ignore` function, which I find a bit verbose. What do you think?

pcarranzav: I just feel it makes it easier to read ([example](https://github.com/resin-io/resin-wifi-connect/blob/master/src/app.coffee#L149)), but it's up to you.

pcarranzav:
LGTM

src/docker-utils.coffee
Page-: You can also use `_.noop` to be a bit more English-language, and reusing a single noop function helps the optimiser a bit as well

pcarranzav: +1 on using `_.noop`

pcarranzav:
connects to #101

pcarranzav:
@lekkas could you add a "connects to  #101" in the PR opening comment? That way waffle picks it up and associates the PR with the issue.

src/docker-utils.coffee
lekkas: `_.noop` it is

pcarranzav:
Reviewed 1 of 2 files at r1, 1 of 1 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/102)*

pcarranzav:
@Page- we need your approval in the convo on reviewable please :)

Page-:
Reviewed 1 of 2 files at r1, 1 of 1 files at r3.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/resin-io/resin-supervisor/102)*

Issue #106, opened 2016-03-10
Env var to specify container restart policy


pcarranzav:
Add a `RESIN_APP_RESTART_POLICY` env var that can take the values specified in: https://docs.docker.com/engine/reference/run/#restart-policies-restart
This should be passed to docker when running the container.

This will allow users to make their app auto-restart on exit.

stdexcept:
Could we make the default to auto-restart?

craig-mulligan:
@stdexcept If you set it as an application wide environment variable, all devices within that application will be set to auto-restart. Would that suffice?

stdexcept:
That would suffice, as long as it's not required for every device individually.

craig-mulligan:
@stdexcept Yip, you'd just need to set it once on an app level and the devices would automatically pick up the new config.

stdexcept:
Hi, 

Can this be prioritised higher? It seems like a ten minute fix but the implications are quite big. We really needed this months ago.

alexandrosm:
Hi @stdexcept, we don't believe this to be a 10-minute fix (otherwise we would have done it) but happy to take a PR.

pcarranzav:
@stdexcept the reason this is not a ten minute fix is that it requires an upgrade to docker in the host OS that still hasn't made it to production. We're working on it.

shaunmulligan:
@pcarranzav is this gonna be merged and tested now with the new docker version?

PR #120, opened 2016-03-28
Container restart policy specification.

abresas:
connects to #106 

Allow users to set container restart policy using environment variables.

RESIN_APP_RESTART_POLICY sets the name of the policy, and
if policy is "on-failure", optionally, RESIN_APP_RESTART_RETRIES
sets the maximum number of retries.

More information on docker docs:
https://docs.docker.com/engine/reference/run/#restart-policies-restart

One major change we introduce here is that the default policy is set
to always while we used to have the default "no".

We validate the arguments and pass retries parameter only for the case
of "on-failure" as specified in Docker API as of v1.19.
We could let docker handle the arguments directly, gaining
forwards-compatibility with any new features, but I opted
for an implementation that is as well-defined as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/120)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Add restart policies and change default to auto-restart [Aleksis]
+
 # v1.10.1
 
 * Switch to docker-delta library to use deltas v2 [petrosagg]
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -228,11 +228,13 @@ application.start = start = (app) ->
 			if portList?
 				portList.forEach (port) ->
 					ports[port + '/tcp'] = [ HostPort: port ]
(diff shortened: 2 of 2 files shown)


pcarranzav:
Code-wise LGTM, but I'm not sure if we should merge this until devices have a version of docker that supports this, mainly because we'll have a behaviour change dependent on something external to the supervisor (i.e. restart always by default will start happening when docker gets updated and not with a supervisor version bump)

pcarranzav:
@abresas I guess we can add this now? imo it's only missing a changelog entry.

pcarranzav:
Reviewed 1 of 1 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/120)*

PR #134, opened 2016-04-27
Add alpine supervisor build info

nghiant2710:
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/134)


Page-:
Is there a reason we aren't directly switching the default to alpine? I'd rather not have to deal with testing both debian based and alpine based supervisors if we don't have to

nghiant2710:
@Page- Alpine based supervisors need more testing to make sure they work properly.
- [ ] Need to test on amd64 devices (nuc) since I don't have that board.
- [ ] Need to test all supervisor endpoints (using SDK), I only tested with functions on dashboard.
I will spend more time to test this but It would be great if someone from our team who has a nuc can test amd64 alpine supervisor.
When alpine based supervisors are fully tested, we can retire debian based ones.

One more thing, we still have to keep armel debian supervisor since alpine doesn't support that architecture atm.

nghiant2710:
@Page- @pcarranzav jenkins build results here:
https://jenkins.dev.resin.io/job/armhf-alpine-supervisor-gh/21/
https://jenkins.dev.resin.io/job/i386-alpine-supervisor-gh/4/
https://jenkins.dev.resin.io/job/amd64-alpine-supervisor-gh/10/

I've not added these jobs to supervisor master job since it might delay other build. Will add after this PR merged

nghiant2710:
@Page- @pcarranzav jenkins build results here:
https://jenkins.dev.resin.io/job/armhf-alpine-supervisor-gh/21/
https://jenkins.dev.resin.io/job/i386-alpine-supervisor-gh/4/
https://jenkins.dev.resin.io/job/amd64-alpine-supervisor-gh/10/

I've not added these jobs to supervisor master job since it might delay other build. Will add after this PR merged

package.json
```diff
@@ -25,7 +25,7 @@
     "resin-register-device": "^2.0.0",
     "request-progress": "^0.3.1",
     "rwlock": "^5.0.0",
-    "sqlite3": "~3.0.4",
+    "sqlite3": "3.0.9",
```
Page-: You took away the `~`/`^`?

Dockerfile.amd64-alpine
```diff
@@ -0,0 +1,42 @@
...
+COPY config/supervisor/ /etc/supervisor/
+
+# Install dependencies
+WORKDIR /app
+COPY package.json postinstall.sh /app/
+RUN set -x \
+	&& buildDeps='g++ sqlite-dev sqlite-libs make' \
+	&& apk add --update $buildDeps \
+	&& JOBS=MAX npm install --unsafe-perm --production --no-optional \
+	&& cd /app/node_modules/sqlite3 && ./node_modules/.bin/node-pre-gyp install --build-from-source --sqlite=/usr/lib \
```
Page-: Why is this needed?

entry.alpine.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+export DBUS_SYSTEM_BUS_ADDRESS="unix:path=/mnt/root/run/dbus/system_bus_socket"
+
+/usr/bin/supervisord -c /etc/supervisor/supervisord.conf
+
+supervisorctl -c /etc/supervisor/supervisord.conf start resin-supervisor
+supervisorctl -c /etc/supervisor/supervisord.conf start go-supervisor
+
+tail -f /var/log/supervisor/supervisord.log
```
Page-: We don't have this in our standard images do we?  I'm wondering as I think this will cause docker logs to just grow and grow by default, as it doesn't do log rotation or anything that I'm aware of

entry.alpine.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+set -e
+
+[ -d /dev/net ] ||
+    mkdir -p /dev/net
+[ -c /dev/net/tun ] ||
+    mknod /dev/net/tun c 10 200
+
+cd /app
```
Page-: Aren't we already in `/app` due to the `WORKDIR` command?

Dockerfile.amd64-alpine
pcarranzav: There's an incompatibility between the libsqlite from debian and alpine, so the prebuilt binaries will not work here.

Page-: Is it possible to pass `--build-from-source` to `npm install`? If we can it'd be nice to do that to avoid any other incompatibilities that might happen as well

package.json
nghiant2710: Yup, only v3.0.9 contains the fix we need for amd64 alpine supervisor.

entry.alpine.sh
nghiant2710: This is from the standard images (debian supervisor). @pcarranzav can we improve this?

entry.alpine.sh
nghiant2710: yeah, this can be removed.

package.json
Page-: So `3.0.10`/`3.1.0` remove the fix?  Because both `~3.0.9` or `^3.0.9` will force `3.0.9` as the minimum version

entry.alpine.sh
Page-: Oh right, I thought this was our supervisor's log file, it's the supervisord log file in reality, this is ok (and needed to stop the container instantly exiting)

package.json
pcarranzav: Yeah apparently there was a patch that was then rolled back.

Page-: Well that's annoying, ok, a fixed version works then

Dockerfile.amd64-alpine
pcarranzav: Oh, apparently it's possible: https://github.com/mapbox/node-pre-gyp/issues/58#issuecomment-39261084
@nghiant2710 wanna give it a try?

entry.alpine.sh
nghiant2710: Removed!

Dockerfile.amd64-alpine
nghiant2710: @pcarranzav, I've tried and had no issue at all. PR updated!
Jenkins builds passed!
https://jenkins.dev.resin.io/job/amd64-alpine-supervisor-gh/13/
https://jenkins.dev.resin.io/job/armhf-alpine-supervisor-gh/25/
https://jenkins.dev.resin.io/job/i386-alpine-supervisor-gh/7/

Makefile
```diff
@@ -1,6 +1,6 @@
 DISABLE_CACHE = 'false'
 
-ARCH = rpi# rpi/amd64/i386/armv7hf/armel
+ARCH = rpi# rpi/amd64/i386/armv7hf/armel/armhf-alpine/amd64-alpine/i386-alpine
```
pcarranzav: @nghiant2710 can we use the same architecture here and have the alpine be passed as a separate parameter?
e.g. we use `ARCH=rpi` and `BASE_DISTRO=-alpine`, and then:
- Use `${ARCH}${BASE_DISTRO}` to choose the Dockerfile
- keep using `ARCH` for the image name
- append `BASE_DISTRO` to the tag

So that we have images that look like `resin/amd64-supervisor:1.7.0-alpine`.

This would correspond to what we had discussed privately with @Page- and @dfunckt

pcarranzav: We'd also have to use separate builds for rpi and armv7hf even though it's the same thing, but I don't see that as too much of a problem.

Dockerfile.armhf-alpine
```diff
@@ -0,0 +1,41 @@
...
+	&& npm dedupe \
+	&& npm cache clean \
+	&& rm -rf /tmp/* \
+	&& apk del $buildDeps \
+	&& rm -rf /var/cache/apk/*
+
+# Copy source
+COPY . /app/
+
+RUN mv -f /app/entry.alpine.sh /app/entry.sh
```
Page-: Can we instead change the `ENTRYPOINT` and `ln -sf /app/entry.sh /start` entries below to point to the alpine entry, rather than replacing the other

Makefile
nghiant2710: Sure, no problem at all

Dockerfile.armhf-alpine
nghiant2710: yeah, makes sense.
I've updated the PR

Makefile
nghiant2710: @pcarranzav, @Page- I've update PR. Distro moved to the tag.

pcarranzav:
:lgtm: but let's wait for @Page- 's approval too, and give this last version a test @nghiant2710

automation/jenkins_build.sh
```diff
@@ -11,36 +11,82 @@ docker pull resin/${ARCH}-supervisor:${ESCAPED_BRANCH_NAME} || docker pull resin
...
-	JOB_NAME=${JOB_NAME} \
-	DEPLOY_REGISTRY=registry.resinstaging.io/ \
-	PUBNUB_SUBSCRIBE_KEY=${PUBNUB_SUBSCRIBE_KEY} \
-	PUBNUB_PUBLISH_KEY=${PUBNUB_PUBLISH_KEY} \
-	MIXPANEL_TOKEN=${MIXPANEL_TOKEN} \
-	deploy
+if [ -z "$BASE_DISTRO" ]; then
+	# Debian
+	# Build the images
+	make SUPERVISOR_VERSION=${ESCAPED_BRANCH_NAME} \
```
Page-: Can you do:
```bash
MAKE_ARGS="ARCH=${ARCH} \
	JOB_NAME=${JOB_NAME} \
	BASE_DISTRO= \
	PUBNUB_SUBSCRIBE_KEY=${PUBNUB_SUBSCRIBE_KEY} \
	PUBNUB_PUBLISH_KEY=${PUBNUB_PUBLISH_KEY} \
	MIXPANEL_TOKEN=${MIXPANEL_TOKEN}"
```
and then just
```bash
make ${MAKE_ARGS} \
	SUPERVISOR_VERSION=${ESCAPED_BRANCH_NAME} \
	DEPLOY_REGISTRY= \
	deploy
```
to make it easier to see the difference between the different commands (and reduce duplication).

automation/jenkins_build.sh
```diff
@@ -11,36 +11,82 @@ docker pull resin/${ARCH}-supervisor:${ESCAPED_BRANCH_NAME} || docker pull resin
...
-	deploy
-make SUPERVISOR_VERSION=${VERSION} \
-	ARCH=${ARCH} \
-	JOB_NAME=${JOB_NAME} \
-	DEPLOY_REGISTRY=registry.resinstaging.io/ \
-	PUBNUB_SUBSCRIBE_KEY=${PUBNUB_SUBSCRIBE_KEY} \
-	PUBNUB_PUBLISH_KEY=${PUBNUB_PUBLISH_KEY} \
-	MIXPANEL_TOKEN=${MIXPANEL_TOKEN} \
-	deploy
+if [ -z "$BASE_DISTRO" ]; then
```
Page-: It should be fine to just do: `make ... BASE_DISTRO=${BASE_DISTRO} ...` I think, after all if `$BASE_DISTRO` is empty that's equivalent to doing `make ... BASE_DISTRO= ...` anyway

Makefile
```diff
@@ -9,13 +10,20 @@ JOB_NAME = 1
...
-SUPERVISOR_IMAGE=$(DEPLOY_REGISTRY)$(IMAGE)
-
 PUBNUB_SUBSCRIBE_KEY = sub-c-bananas
 PUBNUB_PUBLISH_KEY = pub-c-bananas
 MIXPANEL_TOKEN = bananasbananas
 
+ifeq ($(BASE_DISTRO),alpine)
+	IMAGE = "resin/$(ARCH)-supervisor:$(SUPERVISOR_VERSION)-$(BASE_DISTRO)"
+	DOCKERFILE = alpine.$(ARCH)
+else
```
Page-: I think it would be better to change this to a not exists check, as it stands if I specify `BASE_DISTRO=fedora` for instance then it will just silently use a debian base without any warning/error

automation/jenkins_build.sh
nghiant2710: Hmm, it didn't remember why I need to check the base_distro here but you're right. Will remove now

automation/jenkins_build.sh
nghiant2710: Updated!

Makefile
nghiant2710: @Page- this is exactly what you want. We need to check if it's `alpine` or not to put `alpine` in the tag. Otherwise, it uses debian.

Page-: My problem is that you can say `make BASE_DISTRO=fedora` or `make BASE_DISTRO=alpien` and it will just silently use debian with no warning or anything, which would be very confusing when you're expecting it to be `fedora` or `alpine` (but typoed)

nghiant2710: Added a check with info message, if `BASE_DISTRO` exists then build alpine supervisor, otherwise debian supervisor will be built.

Page-:
SGTM

PR #138, opened 2016-05-02
Stop all apps before rebooting or shutting down

### LINKED ISSUE #133: Stop the app before rebooting or shutting down
This would allow apps to catch a SIGTERM and clean up properly before the shutdown.

Front logo [Front conversations](https://app.frontapp.com/open/top_q14)

pcarranzav:
Closes #133 

(Still untested)

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -47,25 +47,51 @@ module.exports = (application) ->
 		application.update(req.body.force)
 		res.sendStatus(204)
 
-	unparsedRouter.post '/v1/reboot', (req, res) ->
-		new Promise (resolve, reject) ->
-			application.logSystemMessage('Rebooting', {}, 'Reboot')
-			utils.gosuper.post('/v1/reboot')
-			.on('error', reject)
-			.on('response', -> resolve())
-			.pipe(res)
+	parsedRouter.post '/v1/reboot', (req, res) ->
+		force = req.body.force
+		Promise.map utils.getKnexApps(), (theApp) ->
+			Promise.using application.lockUpdates(theApp.appId, force), ->
+				# There's a slight chance the app changed after the previous select
+				# So we fetch it again now the lock is acquired
+				utils.getKnexApp(theApp.appId)
+				.then (app) ->
+					application.kill(app, removeContainer: false) if app?
+		.then ->
+			new Promise (resolve, reject) ->
+				application.logSystemMessage('Rebooting', {}, 'Reboot')
+				utils.gosuper.post('/v1/reboot')
+				.on('error', reject)
+				.on('response', -> resolve())
+				.pipe(res)
 		.catch (err) ->
-			res.status(503).send(err?.message or err or 'Unknown error')
+			if err instanceof application.UpdatesLockedError
+				status = 423
+			else
+				status = 500
+			res.status(status).send(err?.message or err or 'Unknown error')
 
-	unparsedRouter.post '/v1/shutdown', (req, res) ->
-		new Promise (resolve, reject) ->
-			application.logSystemMessage('Shutting down', {}, 'Shutdown')
-			utils.gosuper.post('/v1/shutdown')
-			.on('error', reject)
-			.on('response', -> resolve())
-			.pipe(res)
+	parsedRouter.post '/v1/shutdown', (req, res) ->
+		force = req.body.force
+		Promise.map utils.getKnexApps(), (theApp) ->
+			Promise.using application.lockUpdates(theApp.appId, force), ->
+				# There's a slight chance the app changed after the previous select
+				# So we fetch it again now the lock is acquired
+				utils.getKnexApp(theApp.appId)
+				.then (app) ->
+					application.kill(app, removeContainer: false) if app?
+		.then ->
+			new Promise (resolve, reject) ->
+				application.logSystemMessage('Shutting down', {}, 'Shutdown')
+				utils.gosuper.post('/v1/shutdown')
+				.on('error', reject)
+				.on('response', -> resolve())
+				.pipe(res)
 		.catch (err) ->
-			res.status(503).send(err?.message or err or 'Unknown error')
+			if err instanceof application.UpdatesLockedError
+				status = 423
+			else
+				status = 500
+			res.status(status).send(err?.message or err or 'Unknown error')
 
 	parsedRouter.post '/v1/purge', (req, res) ->
 		appId = req.body.appId
@@ -130,7 +156,7 @@ module.exports = (application) ->
 		Promise.using application.lockUpdates(appId, force), ->
 			utils.getKnexApp(appId)
 			.tap (app) ->
-				application.kill(app, true, false)
+				application.kill(app, removeContainer: false)
 			.then (app) ->
 				res.json(_.pick(app, 'containerId'))
 		.catch utils.AppNotFoundError, (e) ->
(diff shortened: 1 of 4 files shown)


Page-:
If updates are locked then a reboot/shutdown attempt will fail right?

pcarranzav:
@Page- indeed

Page-:
Ok, this definitely changes the experience significantly, @shaunmulligan what are your thoughts?  I think at the very least if we keep this that there needs to be an option to force the shutdown/reboot

pcarranzav:
Yeah, that makes sense. Would also like to hear Shaun's thoughts.

shaunmulligan:
Yeah. We could have the UI warn them that their app has a lock and ask them if they wish to override. On the API we just have force:true option.

pcarranzav:
@Page- @shaunmulligan I added a `force` option (and rebased the branch)

src/api.coffee
```diff
@@ -47,15 +47,41 @@ module.exports = (application) ->
...
 
-	unparsedRouter.post '/v1/reboot', (req, res) ->
+	parsedRouter.post '/v1/reboot', (req, res) ->
+		force = req.body.force
 		utils.mixpanelTrack('Reboot')
-		request.post(config.gosuperAddress + '/v1/reboot')
-		.pipe(res)
+		knex('app').select()
+		.then (apps) ->
+			Promise.map apps, (theApp) ->
```
Page-: Doesn't `knex` return a bluebird promise you can `map` on directly?

src/api.coffee
```diff
@@ -47,15 +47,41 @@ module.exports = (application) ->
...
+				Promise.using application.lockUpdates(theApp.appId, force), ->
+					# There's a slight chance the app changed after the previous select
+					# So we fetch it again now the lock is acquired
+					knex('app').select().where({ appId: theApp.appId })
+					.then ([ app ]) ->
+						application.kill(app, true, false) if app?
```
Page-: It'd be really nice to convert `application.kill` to use an options object instead of these booleans, because I have no idea what they denote from this context

src/api.coffee
```diff
@@ -47,15 +47,41 @@ module.exports = (application) ->
...
+		.then ->
+			request.post(config.gosuperAddress + '/v1/reboot')
+			.pipe(res)
```
Page-: I think this stream needs an error handler, otherwise it could crash the supervisor

shaunmulligan:
@pcarranzav can we make some connecting tickets on UI, API, docs (anywhere else?)

pcarranzav:
@shaunmulligan: I created resin-io/resin-ui#206
On the API no changes are needed.
I think the docs change should be ticketed once the UI is PRd at least (to have a better idea of how it will look).

src/api.coffee
pcarranzav: Done.

src/api.coffee
pcarranzav: Done.

src/api.coffee
pcarranzav: Done.

src/application.coffee
```diff
@@ -103,8 +103,12 @@ logSpecialAction = (action, value, success) ->
 	logSystemMessage(msg, {}, "Apply special action #{if success then "success" else "in progress"}")
 
 application = {}
+application.UpdatesLockedError = UpdatesLockedError
 
-application.kill = kill = (app, updateDB = true, removeContainer = true) ->
+application.kill = kill = (app, opts = {}) ->
+	{ updateDB, removeContainer } = opts
```
Page-: This can be:
```coffee
{ updateDB = true, removeContainer = true } = opts
```

pcarranzav: Oh, nice

pcarranzav: Done.

src/api.coffee
```diff
@@ -47,15 +48,53 @@ module.exports = (application) ->
...
+			Promise.using application.lockUpdates(theApp.appId, force), ->
+				# There's a slight chance the app changed after the previous select
+				# So we fetch it again now the lock is acquired
+				knex('app').select().where({ appId: theApp.appId })
+				.then ([ app ]) ->
+					application.kill(app, removeContainer: false) if app?
+		.then ->
+			new Promise (resolve, reject) ->
+				request.post(config.gosuperAddress + '/v1/shutdown')
+				.on 'error', reject
```
Page-: I'd prefer `.on('error', reject)`, since when I first read this I thought it was `.on 'error', -> reject` due to the lack of parens

src/api.coffee
```diff
@@ -47,15 +48,53 @@ module.exports = (application) ->
...
+					application.kill(app, removeContainer: false) if app?
+		.then ->
+			new Promise (resolve, reject) ->
+				request.post(config.gosuperAddress + '/v1/reboot')
+				.on 'error', reject
+				.on 'response', -> resolve()
+				.pipe(res)
+		.catch (err) ->
+			if err instanceof application.UpdatesLockedError
+				status = 503
```
Page-: I think maybe 403 is a better status code?

pcarranzav: How about 423 Locked ?

src/application.coffee
pcarranzav: Just realized this doesn't work on CoffeeScript 1.9. We need to upgrade it, but it should be done separately.

Page-: https://github.com/resin-io/resin-supervisor/pull/251

src/api.coffee
Page-: Works for me, I didn't spot that one

src/application.coffee
```diff
@@ -132,7 +133,8 @@ logSpecialAction = (action, value, success) ->
 		msg = "Applying config variable #{action} = #{value}"
 	logSystemMessage(msg, {}, "Apply special action #{if success then "success" else "in progress"}")
 
-application.kill = kill = (app, updateDB = true, removeContainer = true) ->
+application.kill = kill = (app, opts = {}) ->
+	{ updateDB = true, removeContainer = true } = opts
```
Page-: It's also possible to put this in the function declaration if you want

src/utils.coffee
```diff
@@ -225,6 +225,9 @@ exports.getKnexApp = (appId, columns) ->
 			throw new AppNotFoundError('App not found')
 		return app
 
+exports.getKnexApps = (columns) ->
```
Page-: None of the callers actually use the `columns` argument, should they?

src/application.coffee
pcarranzav: Good point, done.

src/application.coffee
```diff
@@ -132,7 +133,7 @@ logSpecialAction = (action, value, success) ->
 		msg = "Applying config variable #{action} = #{value}"
 	logSystemMessage(msg, {}, "Apply special action #{if success then "success" else "in progress"}")
 
-application.kill = kill = (app, updateDB = true, removeContainer = true) ->
+application.kill = kill = (app, { updateDB = true, removeContainer = true }) ->
```
Page-: You still need the `= {}` ie `{ updateDB = true, removeContainer = true } = {}` otherwise it will throw when trying to dereference a null/undefined opts arguments

pcarranzav: Done

PR #139, opened 2016-05-02
Increase delta request timeout to 15 minutes

pcarranzav:
connects to #125 

15 minutes should be enough for the delta server to calculate most diffs, without being to large in case of a real failure, but we can discuss it.

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/139)


Page-:
LGTM

Issue #124, opened 2016-04-07
Add an application id env var to the container env vars


Page-:

pcarranzav:
Cheers, moved to #25

pcarranzav:
Reopening to track progress of exposing this single var

PR #141, opened 2016-05-09
Add RESIN_APP_ID variable

pcarranzav:
connects to #124

This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/141)

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -468,7 +468,7 @@ getEnvAndFormatRemoteApps = (deviceId, remoteApps, uuid, apiKey) ->
 		getEnvironment(app.id, deviceId, apiKey)
 		.then (environment) ->
 			app.environment_variable = environment
-			utils.extendEnvVars(app.environment_variable, uuid)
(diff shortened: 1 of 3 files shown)


src/application.coffee
```diff
@@ -468,7 +468,7 @@ getEnvAndFormatRemoteApps = (deviceId, remoteApps, uuid, apiKey) ->
 		getEnvironment(app.id, deviceId, apiKey)
 		.then (environment) ->
 			app.environment_variable = environment
-			utils.extendEnvVars(app.environment_variable, uuid)
+			utils.extendEnvVars(app.environment_variable, uuid, app.id)
```
Page-: It might be worth passing the entire app object, to pave the way for exposing other attributes of the app in future

Page-:
LGTM

src/application.coffee
pcarranzav: True, but in any case is an easy change for when we do that

Issue #152, opened 2016-05-27
Log useful output to stderr/stdout


lekkas:
At the moment supervisor prints out logs to stdout by tail'ing `/var/log/supervisor/supervisord.log`

We'd prefer to have this changed and print out more useful info in the container's stdout/stderr instead. For instance, tail'ing the logs of `/var/log/resin_supervisor_stdout.log` and/or `/var/log/resin_supervisor_stderr.log` would be more useful, since these files also include api_keys.

Ideally, extra care should be taken to prevent api_keys from printed out.

The reason behind this is to be able to use a logging container (e.g. Logspout) that will pick the stdout/stderr of the supervisor container and stream it to a remote logging server.

PR #153, opened 2016-05-28
Log useful supervisor info to stdout/stderr

lekkas:
Connects to #152

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/153)

### DIFF
--- a/src/utils.coffee
+++ b/src/utils.coffee
@@ -36,6 +36,16 @@ exports.mixpanelTrack = (event, properties = {}) ->
 			message: properties.error.message
 			stack: properties.error.stack
 
+	# Don't log private env vars (e.g. api keys)
+	if properties?.app?.env?
+		try
+			{ env } = properties.app
+			env = JSON.parse(env) if _.isString(env)
+			safeEnv = _.omit(env, config.privateAppEnvVars)
+			properties.app.env = JSON.stringify(safeEnv)
+		catch
+			properties.app.env = 'Fully hidden due to error in selective hiding'
+
 	console.log('Event:', event, JSON.stringify(properties))
 	# Mutation is bad, and it should feel bad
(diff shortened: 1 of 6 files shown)


src/utils.coffee
```diff
@@ -36,6 +36,16 @@ exports.mixpanelTrack = (event, properties = {}) ->
...
 
+	# Don't log private env vars (e.g. api keys)
+	if properties?.app?.env?
+		try
+			{ env } = properties.app
+			env = JSON.parse(env) if typeof env is 'string'
+			safeEnv = _.omit(env, config.privateAppEnvVars)
+			properties.app.env = JSON.stringify(safeEnv)
+		catch
+			_.noop
```
Page-: It'd be better to just drop the `catch` part altogether, ie
```coffee
try
	throw new Error('This will be ignored')
```

src/utils.coffee
```diff
@@ -36,6 +36,16 @@ exports.mixpanelTrack = (event, properties = {}) ->
 			message: properties.error.message
 			stack: properties.error.stack
 
+	# Don't log private env vars (e.g. api keys)
+	if properties?.app?.env?
+		try
+			{ env } = properties.app
+			env = JSON.parse(env) if typeof env is 'string'
```
Page-: Can you use `_.isString(env)` instead

src/utils.coffee
Page-: Although maybe if we hit an error checking for env vars to hide we should just set `properties.app.env = 'Fully hidden due to error in selective hiding'` or something like that

Page-:
Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/153)*

pcarranzav:
:lgtm:

PR #161, opened 2016-06-07
Use sh for postinstall as alpine has no bash

pcarranzav:
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/161)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "resin-supervisor",
   "version": "1.11.0",
   "scripts": {
-    "postinstall": "bash postinstall.sh",
+    "postinstall": "sh postinstall.sh",
     "start": "./entry.sh"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "resin-supervisor",
   "version": "1.11.0",
   "scripts": {
-    "postinstall": "bash postinstall.sh",
+    "postinstall": "sh postinstall.sh",
```
Page-: Can/should this be `./postinstall.sh`? So that it uses the hashbang of the script itself?

pcarranzav: We can do it, but it requires chmodding the script in all Dockerfiles...

Page-: Is the committed script not marked as executable already?

pcarranzav: It is, but we usually add the chmod for things we want to have as executable - it is safer in case, for  instance, someone ever decides to develop on Windows and messes it up in the git tree unknowingly. I like this pattern.

Page-:
SGTM

PR #162, opened 2016-06-09
Add missing quote in tag.

lorenzo-stoakes:
The error was being silently ignored by the marshaler resulting in userId getting exported as UserId.

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/162)


lifeeth:
LGTM

PR #165, opened 2016-06-10
Add listener for container events and reattach on restart

pcarranzav:
connects to #106

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/165)

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -14,6 +14,7 @@ lockFile = Promise.promisifyAll(require('lockfile'))
 bootstrap = require './bootstrap'
 TypedError = require 'typed-error'
 fs = Promise.promisifyAll(require('fs'))
+JSONStream = require 'JSONStream'
 
 class UpdatesLockedError extends TypedError
 
@@ -66,6 +67,14 @@ logTypes =
 	updateAppError:
 		eventName: 'Application update error'
 		humanName: 'Failed to update application'
+
+	appExit:
+		eventName: 'Application exit'
+		humanName: 'Application exited'
+
+	appRestart:
+		eventName: 'Application restart'
+		humanName: 'Restarting application'
 
 logSystemMessage = (message, obj, eventName) ->
 	logger.log({ message, isSystem: true })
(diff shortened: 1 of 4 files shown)


src/application.coffee
```diff
@@ -644,7 +653,28 @@ application.update = update = (force) ->
 			# Set the updating as finished in its own block, so it never has to worry about other code stopping this.
 			updateStatus.state = UPDATE_IDLE
 
+listenToEvents = ->
+	docker.getEventsAsync()
+	.then (stream) ->
```
Page-: You probably need to handle errors on this stream, to stop the process potentially dying

src/application.coffee
```diff
@@ -644,7 +653,28 @@ application.update = update = (force) ->
...
+		parser = JSONStream.parse()
```
Page-: You probably need to handle errors on this parser stream, to stop the process potentially dying

src/application.coffee
```diff
@@ -644,7 +653,28 @@ application.update = update = (force) ->
...
+		stream.pipe(parser)
+		parser.on 'data', (data) ->
```
Page-: @petrosagg will probably talk about some horrors of using `.on 'data'`, I can't remember the full details myself though

src/application.coffee
```diff
@@ -644,7 +653,28 @@ application.update = update = (force) ->
...
+			if data?.Type == 'container'
```
Page-: You could also check for `data.status in `[ 'die', 'restart' ]` to avoid needing to do the db query when it's not an event we care about.  In fact is it possible to specify to `docker events` that we only want certain types of events? That would be ideal

src/application.coffee
```diff
@@ -644,7 +653,28 @@ application.update = update = (force) ->
...
+				knex('app').select().where({ containerId: data.id })
+				.then ([ app ]) ->
+					if app?
+						if data.status == 'die'
+							logSystemEvent(appExit, app)
+						else if data.status == 'restart'
+							logSystemEvent(appRestart, app)
+							logger.attach(app)
```
Page-: It might be nicer to handle the initial logger attachment here as well, on a `start` event - that way all that logic is in one place (and would be easier to separate out more fully so the logger/etc module just hooks into certain events)

pcarranzav: It's a bit complicated, cause the event might get fired before we save `containerId` to the  DB. I'd rather make this change as unobtrusive as possible and in any case leave that for a later refactoring PR.

src/application.coffee
pcarranzav: Done.

src/application.coffee
Page-: Ok, that makes sense - I hadn't considered the race condition there

Page-:
SGTM

PR #180, opened 2016-06-23
Fix entry.sh when DOCKER_ROOT isn't set

Page-:
@pcarranzav @petrosagg

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/180)

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -25,9 +25,14 @@ fi
 
 export DBUS_SYSTEM_BUS_ADDRESS="unix:path=/mnt/root/run/dbus/system_bus_socket"
 
-# If docker data directory isn't mounted in the default path, symlink it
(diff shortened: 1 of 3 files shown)


entry.sh
```diff
@@ -25,9 +25,18 @@ fi
 
 export DBUS_SYSTEM_BUS_ADDRESS="unix:path=/mnt/root/run/dbus/system_bus_socket"
 
+# If DOCKER_ROOT isn't set then default it
+if [ -n "${DOCKER_ROOT}" ]; then
+	DOCKER_ROOT=/mnt/root/var/lib/docker
```
petrosagg: The default should be rce

petrosagg:
SGTM

PR #186, opened 2016-07-01
Add jenkins build info

nghiant2710:
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/186)

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -17,6 +17,17 @@ RUN apt-get update \
 		unzip \
 		wget \
 		xterm \
+		sudo \
 	&& rm -rf /var/lib/apt/lists/*
 
-WORKDIR /usr/src/yocto/
+RUN useradd -m builder \
+	&& mkdir /dest \
+	&& chown builder /dest
+
+USER builder
+
+ENV SOURCE_DIR /source
+ENV DEST_DIR /dest
+
+COPY build.sh /
+CMD ./build.sh
--- a/base-image/automation/jenkins-build.sh
+++ b/base-image/automation/jenkins-build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+date=$(date +'%Y%m%d' -u)
+REPO='resin/amd64-supervisor-base'
+
+git submodule update --init --recursive
+rm -rf dest
+mkdir dest
+
+docker build -t supervisor-base-builder .
+docker run --rm \
+	-v `pwd`:/source \
+	-v `pwd`/dest:/dest \
+	supervisor-base-builder
+if [ -f dest/rootfs.tar.bz2 ]; then
+	cat dest/rootfs.tar.bz2 | docker import - $REPO:$date
+	docker tag -f $REPO:$date $REPO:latest
+	docker push $REPO
+else
+	echo "rootfs is missing!"
+	exit 1
+fi
--- a/base-image/automation/jenkins-build.sh
+++ b/base-image/automation/jenkins-build.sh
@@ -16,7 +16,7 @@ docker run --rm \
 	-v `pwd`/dest:/dest \
 	supervisor-base-builder
 if [ -f dest/rootfs.tar.gz ]; then
-	cat dest/rootfs.tar.gz | docker import - $REPO:$date
+	docker import dest/rootfs.tar.gz $REPO:$date
 	docker tag -f $REPO:$date $REPO:latest
 	docker push $REPO
 else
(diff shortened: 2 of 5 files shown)


base-image/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+BUILD_DIR='/home/builder/tmp'
+
+mkdir -p $BUILD_DIR
+cp -r $SOURCE_DIR/* $BUILD_DIR/
+cd $BUILD_DIR
+source poky/oe-init-build-env build
+bitbake -k core-image-minimal
+cd tmp/deploy/images/
+cp $(find . -name *.rootfs.tar.bz2) $DEST_DIR/rootfs.tar.bz2
```
petrosagg: There is a symlink with a predictable name that you can use by adding the `--dereference` flag in `cp`

base-image/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -o errexit
+
+BUILD_DIR='/home/builder/tmp'
+
+mkdir -p $BUILD_DIR
+cp -r $SOURCE_DIR/* $BUILD_DIR/
+cd $BUILD_DIR
+source poky/oe-init-build-env build
+bitbake -k core-image-minimal
```
petrosagg: Why do we need the `-k` flag?

base-image/Dockerfile
```diff
@@ -17,6 +17,18 @@ RUN apt-get update \
 		unzip \
 		wget \
 		xterm \
+		sudo \
 	&& rm -rf /var/lib/apt/lists/*
 
-WORKDIR /usr/src/yocto/
+RUN useradd -m  -G sudo builder \
+	&& su - -c "echo '%sudo ALL=(ALL) ALL' >> /etc/sudoers" \
```
petrosagg: Where do we use sudo?

base-image/automation/jenkins-build.sh
```diff
@@ -0,0 +1,25 @@
...
+git submodule update --init --recursive
+rm -rf dest
+mkdir dest
+
+docker build -t supervisor-base-builder .
+docker run --rm \
+	-v `pwd`:/source \
+	-v `pwd`/dest:/dest \
+	supervisor-base-builder
+if [ -f dest/rootfs.tar.bz2 ]; then
```
petrosagg: Let's switch to gzip or no compression

base-image/automation/jenkins-build.sh
```diff
@@ -0,0 +1,25 @@
...
+	cat dest/rootfs.tar.bz2 | docker import - $REPO:$date
```
petrosagg: NIT: this could be a direct parameter of docker import instead of piping stdin

base-image/build.sh
nghiant2710: I think this way will work out of the box since there is only one file that its name ends with `.rootfs.tar.bz2` and we just need to find it.
We can predict the filename but this script will become way more complicated than it should be since rootfs files are inside subdirectory of `deploy/images` (qemuXXX, .i.e qemux86-64 for amd64 iirc). So we need to include that qemu thing in the build script to predict the name and path to rootfs file...

base-image/build.sh
nghiant2710: `-k` means `continue as much as possible after an error`
http://www.linuxcertif.com/man/1/bitbake/

base-image/Dockerfile
nghiant2710: bitbake won't run as root. So I create `builder` user and add it to sudo group

base-image/automation/jenkins-build.sh
nghiant2710: Updated!

base-image/automation/jenkins-build.sh
nghiant2710: updated!

base-image/Dockerfile
petrosagg: The first part I understand, but we don't need the sudo part

base-image/build.sh
petrosagg: right, so that will make builds that fail last longer. Why is this desirable?

nghiant2710: I had some issues with sanity check from yocto and this flag gets rid of them but we have a clean build now so I will remove it

base-image/Dockerfile
nghiant2710: removed!

base-image/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+BUILD_DIR='/home/builder/tmp'
+
+mkdir -p $BUILD_DIR
+cp -r $SOURCE_DIR/* $BUILD_DIR/
+cd $BUILD_DIR
+source poky/oe-init-build-env build
+bitbake core-image-minimal
+cd tmp/deploy/images/
+cp $(find . -name *.rootfs.tar.gz) $DEST_DIR/rootfs.tar.gz
```
petrosagg: You can use `--dereference` and copy using the symlink

nghiant2710: @petrosagg, updated!

Issue #188, opened 2016-07-04
Use --build-arg docker option instead of sed to configure dropbear in dind


lekkas:
`sed -i` is incompatible between OS X / Linux and it seems to be the only thing that breaks between the two OS'es so far.

Since version 1.9.0 `docker build` supports passing build-time parameters with [--build-arg]( https://docs.docker.com/engine/reference/commandline/build/#/set-build-time-variables-build-arg) , so we can use it instead of sed.

PR #190, opened 2016-07-04
dind: make sed work on both GNU and BSD

lekkas:
The alternative here is to simply instruct users install a GNU-compatible version of `sed` on their OS X development machine.

Connects to #188 

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/190)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,9 +73,12 @@ endif
 clean:
 	-rm Dockerfile
 
+# Make 'sed' work on both GNU and BSD
+supervisor-dind: TMP_SED := $(shell mktemp)
 supervisor-dind:
-	sed -i 's/\(ENV PASSWORDLESS_DROPBEAR\).*/\1 ${PASSWORDLESS_DROPBEAR}/' tools/dind/Dockerfile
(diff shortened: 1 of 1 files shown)


Page-:
Is there really no `-i` equivalent on mac? :(

lekkas:
I couldn't find a one-line combination for `-i / --in-place` that would work on both :/ 

The `-i` equivalent has small subtle differences between the GNU/BSD versions:

* `--in-place` is only supported on GNU/Linux
* `sed -i ''` and `sed -i=''` work on OS X but not on Linux (the first is invalid, the second creates a backup file with `=` appended to the original file name.
* `sed -i''` and `sed -i` work on Linux but not on OS X (both are invalid formats)

lekkas:
But I just noticed that since docker 1.9.0 `docker build` supports a `[--build-arg](https://docs.docker.com/engine/reference/commandline/build/#/set-build-time-variables-build-arg)` parameter for this exact problem that I was working around with sed in the first place

So I guess we can close this PR, remove sed altogether and use that option instead, since our development environment is already updated to 1.9.1

PR #200, opened 2016-07-08
use shared folders (downloads and sstate) in yocto build

nghiant2710:
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/200)

### DIFF
--- a/base-image/build.sh
+++ b/base-image/build.sh
@@ -3,10 +3,11 @@
 set -o errexit
 
 BUILD_DIR='/home/builder/tmp'
-
-mkdir -p $BUILD_DIR
-cp -r $SOURCE_DIR/* $BUILD_DIR/
-cd $BUILD_DIR
-source oe-core/oe-init-build-env build bitbake
-MACHINE=$TARGET_MACHINE bitbake core-image-minimal
-cp --dereference tmp-glibc/deploy/images/$TARGET_MACHINE/core-image-minimal-$TARGET_MACHINE.tar.gz  $DEST_DIR/rootfs.tar.gz
+groupadd -g $BUILDER_GID builder
+useradd -m -u $BUILDER_UID -g $BUILDER_GID builder
+sudo -H -u builder /bin/bash -c "mkdir -p $BUILD_DIR \
+	&& cp -r $SOURCE_DIR/* $BUILD_DIR/ \
+	&& cd $BUILD_DIR \
+	&& source oe-core/oe-init-build-env build bitbake \
+	&& DL_DIR=$SHARED_DOWNLOADS SSTATE_DIR=$SHARED_SSTATE MACHINE=$TARGET_MACHINE $BUILD_DIR/bitbake/bin/bitbake core-image-minimal"
+cp --dereference $BUILD_DIR/build/tmp-glibc/deploy/images/$TARGET_MACHINE/core-image-minimal-$TARGET_MACHINE.tar.gz  $DEST_DIR/rootfs.tar.gz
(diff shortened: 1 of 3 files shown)


base-image/build.sh
```diff
@@ -7,6 +7,11 @@ BUILD_DIR='/home/builder/tmp'
 mkdir -p $BUILD_DIR
 cp -r $SOURCE_DIR/* $BUILD_DIR/
 cd $BUILD_DIR
-source oe-core/oe-init-build-env build bitbake
-MACHINE=$TARGET_MACHINE bitbake core-image-minimal
-cp --dereference tmp-glibc/deploy/images/$TARGET_MACHINE/core-image-minimal-$TARGET_MACHINE.tar.gz  $DEST_DIR/rootfs.tar.gz
+sed -i "s#.*DL_DIR ?=.*#DL_DIR ?= \"$SHARED_DOWNLOADS\"#g" build/conf/local.conf
+sed -i "s#.*SSTATE_DIR ?=.*#SSTATE_DIR ?= \"$SHARED_SSTATE\"#g" build/conf/local.conf
```
petrosagg: Can't we set them in the same manner as `MACHINE`? By prefixing the bitbake command?

base-image/build.sh
```diff
@@ -7,6 +7,11 @@ BUILD_DIR='/home/builder/tmp'
...
+useradd -m builder
+chown builder -R $BUILD_DIR $DEST_DIR $SHARED_DOWNLOADS $SHARED_SSTATE
```
petrosagg: This is shared between a lot of jobs. You shouldn't recursively touch these directories.

Would the whole thing be simpler if it ran as root in the container?

base-image/build.sh
nghiant2710: I think we can, will give it a try

base-image/build.sh
nghiant2710: it's running as root in the container but bitbake won't run as root so we still need this user.
I will configure uid for this user the same with host so we don't need to touch this way

petrosagg: The user is not always the same. I've head issues when running as `petrosagg` on the build server because I don't have the default uid `1000`

petrosagg: We probably need something like this https://github.com/resin-os/resin-yocto-scripts/blob/master/automation/prepare-and-start.sh#L38-L42

nghiant2710: ah, right. Thanks

base-image/build.sh
nghiant2710: Updated!

base-image/build.sh
nghiant2710: Updated!

base-image/build.sh
```diff
@@ -3,10 +3,12 @@
...
 BUILD_DIR='/home/builder/tmp'
-
-mkdir -p $BUILD_DIR
-cp -r $SOURCE_DIR/* $BUILD_DIR/
-cd $BUILD_DIR
-source oe-core/oe-init-build-env build bitbake
-MACHINE=$TARGET_MACHINE bitbake core-image-minimal
-cp --dereference tmp-glibc/deploy/images/$TARGET_MACHINE/core-image-minimal-$TARGET_MACHINE.tar.gz  $DEST_DIR/rootfs.tar.gz
+groupadd -g $BUILDER_GID builder
+groupadd docker
```
petrosagg: we don't need docker in there

nghiant2710: removed

petrosagg:
SGTM

Issue #211, opened 2016-07-18
Updates fail in non-btrfs when looking for imageRootDir


pcarranzav:
Introduced with #205

Breaks the dind supervisor.

PR #212, opened 2016-07-18
Default to not bind mounting kmod if container distro can't be found

pcarranzav:
connects to #211 
@petrosagg @Page-

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/212)


Page-:
SGTM

PR #218, opened 2016-07-21
Fix double-printed logs after container restart

### LINKED ISSUE #217: Logs appear twice after container restart

lekkas:
Connects to #217 

Ref: https://github.com/resin-io-modules/resin-sync/issues/14

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/218)

### DIFF
--- a/src/lib/logger.coffee
+++ b/src/lib/logger.coffee
@@ -3,6 +3,7 @@ Docker = require 'dockerode'
 PUBNUB = require 'pubnub'
 Promise = require 'bluebird'
 es = require 'event-stream'
+Lock = require 'rwlock'
 
 disableLogs = false
 
@@ -52,17 +53,25 @@ exports.log = ->
 	publish(arguments...)
 
 do ->
+	_lock = new Lock()
+	_writeLock = Promise.promisify(_lock.async.writeLock)
+	loggerLock = (containerId) ->
+		_writeLock(containerId)
+		.disposer (release) ->
+			release()
+
 	attached = {}
 	exports.attach = (app) ->
-		if !attached[app.containerId]
-			dockerPromise.then (docker) ->
-				docker.getContainer(app.containerId)
-				.attachAsync({ stream: true, stdout: true, stderr: true, tty: true })
-				.then (stream) ->
-					attached[app.containerId] = true
-					stream.pipe(es.split())
-					.on('data', publish)
-					.on 'error', ->
-						attached[app.containerId] = false
-					.on 'end', ->
-						attached[app.containerId] = false
+		Promise.using loggerLock(app.containerId), ->
+			if !attached[app.containerId]
+				dockerPromise.then (docker) ->
+					docker.getContainer(app.containerId)
+					.attachAsync({ stream: true, stdout: true, stderr: true, tty: true })
+					.then (stream) ->
+						attached[app.containerId] = true
+						stream.pipe(es.split())
+						.on('data', publish)
+						.on 'error', ->
+							attached[app.containerId] = false
+						.on 'end', ->
+							attached[app.containerId] = false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Fix duplicate logs issue [Kostas]
 * **[Breaking]** Do not bind mount /run/dbus to /run/dbus [Pablo]
 * Default to not bind mounting kmod if container distro can't be found [Pablo]
 * Use log-timestamp to add timestamps to logs [Pablo]


pcarranzav:
Good catch @lekkas - however:
* This will break logs when the container is restarted by docker. The logger.attach there is needed for that. Instead what should be done is somehow mark that `appHasDied[app.containerId]=false` when restarting via the supervisor endpoints.
* Please use "connects to  #issue" instead of "Closes #issue" in the PR opening comment for proper Waffle tracking.

lekkas:
Ah right, I totally missed that. I'll look at it more closely because the underlying issue here appears to be a synchronisation one - both `docker.getEventsAsync()`and `application.start()` enter `attach()` and wait for the inner docker promises so neither manages to lock the other out.

Fixed the 'closes issue' as well.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218#-:-KNDCVbsfVT1WUqvooZZ:-730951431)*

pcarranzav:
A better solution would be to ensure inside `logger.coffee` that we only attach once to each container.

pcarranzav:
But actually... we currently do [here](https://github.com/resin-io/resin-supervisor/blob/master/src/lib/logger.coffee#L57)

lekkas:
Well, this lock doesn't work as expected - because both tasks will pass this check and then wait on the promises (e.g. `docker.getContainer(app.containerId)`) . Then both tasks will enter our critical section at the same time, which will result in the duplicate logs.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218#-:-KNDLej_QJ33PUeCF2sE:-1707431794)*

pcarranzav:
Right, so we need to add a rwlock there

lekkas:
Added

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218#-:-KNDWdxPyz8zwUU-IJCd:-502253829)*

src/lib/logger.coffee
```diff
@@ -52,17 +53,25 @@ exports.log = ->
 	publish(arguments...)
 
 do ->
+	_lock = new Lock()
```
emirotin: I've recently seen the same logic in a different PR (though now it seems to be gone) — worth moving to utils module?

pcarranzav: Yeah, we have a bunch of rwlocks in several parts of the supervisor.
The one in logger was for a different purpose and is no longer necessary. I think it's fine to have this here.

pcarranzav:
Please add a changelog entry. Other than that, LGTM :+1:

Page-:
Reviewed 1 of 1 files at r1, 1 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218)*

lekkas:
retest this please

lekkas:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---

*[src/lib/logger.coffee, line 56 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218#-KNDawPUMp_SUP3Ikmbl-r2-56:-KNGTJtRCDjkLMZtQRVW:676605422) ([raw file](https://github.com/resin-io/resin-supervisor/blob/efd3ed884950bcd065f9b1b6fd7937b46bd836a0/src/lib/logger.coffee#L56)):*
<details><summary><i>Previously, pcarranzav (Pablo Carranza Vélez) wrote…</i></summary>
> Yeah, we have a bunch of rwlocks in several parts of the supervisor.
> 
> The one in logger was for a different purpose and is no longer necessary. I think it's fine to have this here.
</details>
Cool

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218)*

lekkas:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/lib/logger.coffee, line 56 \[r2\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218#-KNDawPUMp_SUP3Ikmbl-r2-56:-KNGTsLo2M6RbGi8uOQJ:-499146009) ([raw file](https://github.com/resin-io/resin-supervisor/blob/efd3ed884950bcd065f9b1b6fd7937b46bd836a0/src/lib/logger.coffee#L56)):*
<details><summary><i>Previously, lekkas (Kostas Lekkas) wrote…</i></summary>
> Cool
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/218)*

PR #198, opened 2016-07-07
Open Embedded base image

petrosagg:
This PR adds all the infrastructure needed to build a `supervisor-base` image that is built using open embedded, instead of using a Dockerfile based on alpine or debian

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/198)

PR #223, opened 2016-07-28
oe-images build system integration

petrosagg:
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/223)


package.json
```diff
@@ -12,7 +11,6 @@
     "bluebird": "^2.9.24",
     "body-parser": "^1.12.0",
     "buffer-equal-constant-time": "^1.0.1",
-    "coffee-script": "~1.9.1",
```
Page-: This should be moved to `devDependencies` instead

Dockerfile.build.template
```diff
@@ -0,0 +1,37 @@
...
+	&& npm install -g coffee-script
+
+COPY package.json /usr/src/app/
+
+RUN JOBS=MAX npm install --unsafe-perm --production --no-optional \
+	&& npm dedupe
+
+COPY src /usr/src/app/src
+
+RUN coffee -c src
```
Page-: This seems like it should be `npm run build` or similar instead

Dockerfile.build.template
```diff
@@ -0,0 +1,37 @@
+# Build nodejs dependencies
+FROM resin/%%ARCH%%-node:6.2-slim
```
Page-: Can this be 6.3 instead?

Dockerfile.build.template
```diff
@@ -0,0 +1,37 @@
...
+
+RUN apt-get update \
+	&& apt-get install -y \
+		g++ \
+		libsqlite3-dev \
+		make \
+		python \
+		rsync \
+	&& rm -rf /var/lib/apt/lists/ \
+	&& npm install -g coffee-script
```
Page-: I'd like the version of this to be specified in the package.json ideally, but at the very least it needs to be specified here to avoid picking up breaking changes by accident

automation/jenkins_build.sh
```diff
@@ -12,14 +12,12 @@ docker pull resin/${ARCH}-supervisor:${ESCAPED_BRANCH_NAME} || docker pull resin
...
 MAKE_ARGS="ARCH=${ARCH} \
-    JOB_NAME=${JOB_NAME} \
-    BASE_DISTRO=${BASE_DISTRO} \
     PUBNUB_SUBSCRIBE_KEY=${PUBNUB_SUBSCRIBE_KEY} \
     PUBNUB_PUBLISH_KEY=${PUBNUB_PUBLISH_KEY} \
     MIXPANEL_TOKEN=${MIXPANEL_TOKEN}"
 
-make ${MAKE_ARGS} \
-    lint
+# Disabled until this is merged in npm https://github.com/npm/npm/pull/13257
```
Page-: Why is this an issue, I can't see why the `make lint` command would hit the different devices/filesystems issues?

Dockerfile.runtime.template
```diff
@@ -0,0 +1,41 @@
...
+	&& ln -s /lib /lib64
+
+ENV DOCKER_COMPOSE_VERSION 1.7.1
+
+ENV DOCKER_COMPOSE_SHA256_amd64 37df85ee18bf0e2a8d71cbfb8198b1c06cc388f19118be7bdfc4d6db112af834
+ENV DOCKER_COMPOSE_SHA256_i386 b926fd9a2a9d89358f1353867706f94558a62caaf3aa72bf10bcbbe31e1a44f0
+ENV DOCKER_COMPOSE_SHA256_armhf 3f0b8c69c66a2daa5fbb0c127cb76ca95d7125827a9c43dd3c36f9bc2ed6e0e5
+ENV DOCKER_COMPOSE_SHA256_armel a1025fed97536e2698798ea277a014ec5e1eae816a8cf3155ecbe9679e3e7bac
+
+RUN set -x \
```
Page-: Do both `wget` and `tar` come from busybox or could we save size by doing the extraction elsewhere and then copying in here, removing those deps?

Dockerfile.runtime.template
```diff
@@ -0,0 +1,41 @@
+# Minimal runtime image
+FROM resin/%%ARCH%%-supervisor-base
```
Page-: What is in supervisor-base? Where does it come from?

Dockerfile.runtime.template
```diff
@@ -0,0 +1,41 @@
...
+
+ENV DOCKER_COMPOSE_SHA256_amd64 37df85ee18bf0e2a8d71cbfb8198b1c06cc388f19118be7bdfc4d6db112af834
+ENV DOCKER_COMPOSE_SHA256_i386 b926fd9a2a9d89358f1353867706f94558a62caaf3aa72bf10bcbbe31e1a44f0
+ENV DOCKER_COMPOSE_SHA256_armhf 3f0b8c69c66a2daa5fbb0c127cb76ca95d7125827a9c43dd3c36f9bc2ed6e0e5
+ENV DOCKER_COMPOSE_SHA256_armel a1025fed97536e2698798ea277a014ec5e1eae816a8cf3155ecbe9679e3e7bac
+
+RUN set -x \
+	&& if [ $ARCH == "rpi" ]; then export ARCH=armf; fi \
+	&& wget http://resin-packages.s3.amazonaws.com/docker-compose/${DOCKER_COMPOSE_VERSION}/docker-compose-linux-${ARCH}-${DOCKER_COMPOSE_VERSION}.tar.gz \
+	&& echo "${DOCKER_COMPOSE_SHA256_%%ARCH%%}  docker-compose-linux-${ARCH}-${DOCKER_COMPOSE_VERSION}.tar.gz" | sha256sum -c \
```
Page-: There's a double space here

pcarranzav: the double space is actually needed by sha256sum

Dockerfile.runtime.template
pcarranzav: They are indeed both from busybox

Dockerfile.runtime.template
pcarranzav: See #198

Dockerfile.runtime.template
Page-: Wtf, why?

Dockerfile.runtime.template
Page-: Ok

Dockerfile.runtime.template
pcarranzav: No idea, but: https://github.com/gliderlabs/docker-alpine/issues/174#issuecomment-222951567

Dockerfile.runtime.template
Page-: Ok, that explains why I couldn't find it, although I'm still not sure precisely what's in it (I have a feeling a high level of yocto is needed to know the full details) but at least I know the idea behind it is the minimal busybox + nodejs :)

Dockerfile.runtime.template
pcarranzav: Also related: http://unix.stackexchange.com/a/139892

Page-: That is really really weird, it's worth adding a comment as to the reasoning imo because I just assumed it was a type

Dockerfile.build.template
Page-: Or 6.4-slim/6-slim, since we're on that now

petrosagg: The node version here has to match the node version in the oe-image. I don't want to use slim for this because it's a build image and I'd rather install as less packages as possible

Dockerfile.build.template
petrosagg: Will fix

Dockerfile.runtime.template
petrosagg: This `RUN` will be moved to the build image and it will use proper wget and tar.

Dockerfile.build.template
Page-: @petrosagg you're already using a slim image, I wasn't suggesting any change there.

As for matching the oe-image version, there definitely needs to be a comment to that effect here, otherwise I can easily see someone changing it without realising that it needs to stay in sync.  For that matter what is the process for upgrading the oe-image version, because in that case I'd suggest updating both to pick up the fixes

petrosagg: I added a comment. I'll update the version in a separate set of PRs. i'd like to get this moving

Dockerfile.build.template
petrosagg: done

package.json
petrosagg: done

Makefile
```diff
@@ -136,54 +131,86 @@ refresh-supervisor-src:
...
 
-supervisor: gosuper
-	cp Dockerfile.$(DOCKERFILE) Dockerfile
-	echo "ENV VERSION "`jq -r .version package.json` >> Dockerfile
-	echo "ENV DEFAULT_PUBNUB_PUBLISH_KEY $(PUBNUB_PUBLISH_KEY)" >> Dockerfile
-	echo "ENV DEFAULT_PUBNUB_SUBSCRIBE_KEY $(PUBNUB_SUBSCRIBE_KEY)" >> Dockerfile
-	echo "ENV DEFAULT_MIXPANEL_TOKEN $(MIXPANEL_TOKEN)" >> Dockerfile
+supervisor: nodesuper gosuper
+	sed 's/%%ARCH%%/$(ARCH)/g' Dockerfile.runtime.template > Dockerfile.runtime.$(ARCH)
+	echo "ENV VERSION $(shell jq -r .version package.json) \\" >> Dockerfile.runtime.$(ARCH)
```
Page-: Would these make sense to be build args?

petrosagg: Well, with build args I'd have to put the following in the Dockerfile and still handle it in the Makefile. For the purposes of minimising size and layer cound I think I'll leave it here

```Dockerfile
ARG VERSION
ENV VERSION $VERSION
```

Page-: OK

petrosagg:
Review status: 0 of 29 files reviewed at latest revision, 6 unresolved discussions.

---

*[Dockerfile.runtime.template, line 19 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVj-r1-19:-KR8lDPm0ae8OEKzRghf:b-9m2jh9) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/Dockerfile.runtime.template#L19)):*
<details><summary><i>Previously, pcarranzav (Pablo Carranza Vélez) wrote…</i></summary>
> Also related: http://unix.stackexchange.com/a/139892
</details>
since we use this method everywhere, especially in base images, and any attempt to delete the space will result to a failed build I'll leave as-is. it's one of the linux quirks one has to know

---

*[automation/jenkins_build.sh, line 19 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVY-r1-19:-KR8m0SPBXpO-M2kff8v:b-80yvw8) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/automation/jenkins_build.sh#L19)):*
<details><summary><i>Previously, Page- wrote…</i></summary>
> Why is this an issue, I can't see why the `make lint` command would hit the different devices/filesystems issues?
</details>
I'm not sure why this happens on our build server but it does

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

pcarranzav:
Review status: 0 of 29 files reviewed at latest revision, 7 unresolved discussions.

---

*[Dockerfile.build.template, line 50 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA-Jmh0g9oWpMvYNX3:-KRA-Jmlm7wavzX69mXP:blgchun) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/Dockerfile.build.template#L50)):*
> ```
> RUN npm run build
> 
> # Re-install just production modules. This should be fast due to npm cache from previous layer
> ```

Looks like this actually does trigger a rebuild of several modules (I actually see a few Jenkins builds stuck building mkfifo in this step)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

Page-:
Reviewed 29 of 29 files at r1, 7 of 7 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[CHANGELOG.md, line 5 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5AebVd1_U2SNLhns:-KRA5AebVd1_U2SNLhnt:b-1p177h) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/CHANGELOG.md#L5)):*
> ```Markdown
> * Switch to go version 1.6 [petrosagg]
> * Print all logs to stdout of container instead of file [petrosagg]
> * Upgrade nodejs to version 6.5 [petrosagg]
> ```

This should state 6.3?

---

*[automation/jenkins_build.sh, line 19 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVY-r1-19:-KNsOhkMmBu-o2GoOu6I:bbuh7nw) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/automation/jenkins_build.sh#L19)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> I'm not sure why this happens on our build server but it does
</details>
I guess it's related to the `npm install resin-lint`, I think it'd make sense to remove this part altogether and instead change `RUN npm run build``to `RUN npm run lint && npm run build` in the build Dockerfile, that way we always have the linter run automatically and at a point we have all the dependencies already available

---

*[gosuper/Dockerfile, line 2 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5myH2foRE8Mjk0Vw:-KRA5myH2foRE8Mjk0Vx:b-oipwte) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/gosuper/Dockerfile#L2)):*
> ```Dockerfile
> # Build golang supervisor
> FROM golang:1.6
> ```

Is it possible to jump to golang 1.7, it's meant to result in much smaller binaries (which fits perfectly with the aims of this PR) - I don't mind waiting until a later PR though, but would like to know the feasibility

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

petrosagg:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[CHANGELOG.md, line 5 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5AebVd1_U2SNLhns:-KRA78iaqEz_g4BNwlST:brndm33) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/CHANGELOG.md#L5)):*
<details><summary><i>Previously, Page- wrote…</i></summary>
> This should state 6.3?
</details>
oops, yeah

---

*[Dockerfile.build.template, line 50 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA-Jmh0g9oWpMvYNX3:-KRA7U84BflQGaUnkUoR:b-eur6qe) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/Dockerfile.build.template#L50)):*
<details><summary><i>Previously, pcarranzav (Pablo Carranza Vélez) wrote…</i></summary>
> Looks like this actually does trigger a rebuild of several modules (I actually see a few Jenkins builds stuck building mkfifo in this step)
</details>
It does, yeah.. but I couldn't find a better way to do it. Ideas are welcome

---

*[automation/jenkins_build.sh, line 19 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVY-r1-19:-KRA8rtJLn9E1F8mnwsG:bp9l1) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/automation/jenkins_build.sh#L19)):*
<details><summary><i>Previously, Page- wrote…</i></summary>
> I guess it's related to the `npm install resin-lint`, I think it'd make sense to remove this part altogether and instead change `RUN npm run build``to `RUN npm run lint && npm run build` in the build Dockerfile, that way we always have the linter run automatically and at a point we have all the dependencies already available
</details>
Right. I'll move it there

---

*[gosuper/Dockerfile, line 2 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5myH2foRE8Mjk0Vw:-KRA8F4vlT3ekJKAMgbC:bh9pol6) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/gosuper/Dockerfile#L2)):*
<details><summary><i>Previously, Page- wrote…</i></summary>
> Is it possible to jump to golang 1.7, it's meant to result in much smaller binaries (which fits perfectly with the aims of this PR) - I don't mind waiting until a later PR though, but would like to know the feasibility
</details>
I did try that and it compiled without errors but it only became 0.1MB smaller. So go 1.7 makes smaller binaries but less compressible ones. I reverted it because I wans't sure if things could break with 1.7

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

Page-:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[gosuper/Dockerfile, line 2 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5myH2foRE8Mjk0Vw:-KRBBWZ04GwyK_v-Ce-_:b-lz30fx) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/gosuper/Dockerfile#L2)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> I did try that and it compiled without errors but it only became 0.1MB smaller. So go 1.7 makes smaller binaries but less compressible ones. I reverted it because I wans't sure if things could break with 1.7
</details>
I expect it makes sense to just bump to 1.7, since we're doing a version bump anyway and will have to test the bump works - the smaller uncompressed size also has benefits in memory usage right? Due to the uncompressed in-memory parts being smaller

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

petrosagg:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[gosuper/Dockerfile, line 2 \[r3\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5myH2foRE8Mjk0Vw:-KRD4R6WMLFT307QwkWX:b-q47wj3) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/gosuper/Dockerfile#L2)):*
<details><summary><i>Previously, Page- wrote…</i></summary>
> I expect it makes sense to just bump to 1.7, since we're doing a version bump anyway and will have to test the bump works - the smaller uncompressed size also has benefits in memory usage right? Due to the uncompressed in-memory parts being smaller
</details>
I'm fine with adding it as long as @pcarranzav thinks it's ok to do it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

petrosagg:
Review status: 25 of 29 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[Dockerfile.build.template, line 14 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVi-r1-14:-KRD6HF5-_pwHmM324RU:b-896fix) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/Dockerfile.build.template#L14)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> Will fix
</details>
Done.

---

*[Dockerfile.runtime.template, line 16 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVj-r1-16:-KRD6Jx1vn50niY-9rz3:b-896fix) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/Dockerfile.runtime.template#L16)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> This `RUN` will be moved to the build image and it will use proper wget and tar. 
</details>
Done.

---

*[automation/jenkins_build.sh, line 19 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KNlNCPDUp1L3NfmboVY-r1-19:-KRD6OW7nEotwO2PJbQV:b-896fix) ([raw file](https://github.com/resin-io/resin-supervisor/blob/b681af2b057ba009ee09d30546902b9aaac6a967/automation/jenkins_build.sh#L19)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> Right. I'll move it there
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

petrosagg:
resolved conflicts and rebased on top of master

Page-:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

petrosagg:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[CHANGELOG.md, line 5 at r3](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5AebVd1_U2SNLhns:-KRaIdtGDbHNHFdKpU_7:b-rjchgi) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/CHANGELOG.md#L5)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> oops, yeah
</details>
I changed the base image to 6.5 so now the changelog is accurate :D

---

*[gosuper/Dockerfile, line 2 at r3](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5myH2foRE8Mjk0Vw:-KRaHYXm-L38eZRrZBt-:bibnof5) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/gosuper/Dockerfile#L2)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> I'm fine with adding it as long as @pcarranzav thinks it's ok to do it
</details>
Since there was a non-trivial error when we moved to golang 1.6 I won't include the 1.7 bump in this PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

Page-:
Reviewed 4 of 4 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[CHANGELOG.md, line 5 at r3](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5AebVd1_U2SNLhns:-KRbdUdtuHcmDuF1Lx9j:b-f6zusb) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/CHANGELOG.md#L5)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> I changed the base image to 6.5 so now the changelog is accurate :D
</details>
Even better :D  Although now the changelog needs to be moved to the top, as it's not part of v2.1.1 - apart from that it's good to go imo

---

*[gosuper/Dockerfile, line 2 at r3](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223#-KRA5myH2foRE8Mjk0Vw:-KRbdulZWTptV7F1xqE3:bqog80t) ([raw file](https://github.com/resin-io/resin-supervisor/blob/986880cac829923e0cf315980a32b379fca13144/gosuper/Dockerfile#L2)):*
<details><summary><i>Previously, petrosagg (Petros Angelatos) wrote…</i></summary>
> Since there was a non-trivial error when we moved to golang 1.6 I won't include the 1.7 bump in this PR
</details>
Agreed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/223)*

Dockerfile.runtime.template
```diff
@@ -0,0 +1,20 @@
+# Minimal runtime image
+FROM resin/%%ARCH%%-supervisor-base:20160913
+
+WORKDIR /usr/src/app
+
+COPY ./build/%%ARCH%%/src ./src
```
Page-: You could reduce the layers by changing this to:
```
COPY ./build/%%ARCH%%/src ./build/%%ARCH%%/node_modules ./build/%%ARCH%%/gosuper .
````

PR #224, opened 2016-07-30
Use the new device registration endpoint

### LINKED ISSUE #270: Use the new device registration endpoint

Page-:
Closes #270 
Depends on: https://github.com/resin-io-modules/resin-register-device/pull/22

### DIFF
--- a/src/app.coffee
+++ b/src/app.coffee
@@ -51,17 +51,17 @@ knex.init.then ->
 					)
 			)
 
+			updateIpAddr = ->
+				utils.gosuper.getAsync('/v1/ipaddr', { json: true })
+				.spread (response, body) ->
+					if response.statusCode == 200 && body.Data.IPAddresses?
+						device.updateState(
+							ip_address: body.Data.IPAddresses.join(' ')
+						)
+				.catch(_.noop)
+			console.log('Starting periodic check for IP addresses..')
+			setInterval(updateIpAddr, 30 * 1000) # Every 30s
+			updateIpAddr()
+
 		console.log('Starting Apps..')
 		application.initialize()
-
-		updateIpAddr = ->
-			utils.gosuper.getAsync('/v1/ipaddr', { json: true })
-			.spread (response, body) ->
-				if response.statusCode == 200 && body.Data.IPAddresses?
-					device.updateState(
-						ip_address: body.Data.IPAddresses.join(' ')
-					)
-			.catch(_.noop)
-		console.log('Starting periodic check for IP addresses..')
-		setInterval(updateIpAddr, 30 * 1000) # Every 30s
-		updateIpAddr()
--- a/src/bootstrap.coffee
+++ b/src/bootstrap.coffee
@@ -2,13 +2,18 @@ Promise = require 'bluebird'
 knex = require './db'
 utils = require './utils'
 deviceRegister = require 'resin-register-device'
+{ resinApi, request } = require './request'
 fs = Promise.promisifyAll(require('fs'))
 config = require './config'
 configPath = '/boot/config.json'
 appsPath  = '/boot/apps.json'
 _ = require 'lodash'
 deviceConfig = require './device-config'
+TypedError = require 'typed-error'
 userConfig = {}
+
+DuplicateUuidError = message: '"uuid" must be unique.'
+exports.ExchangeKeyError = class ExchangeKeyError extends TypedError
 
 bootstrapper = {}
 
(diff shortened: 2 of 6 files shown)


emirotin:
lgtm

pcarranzav:
Why do we want to move uuid generation to the server? We've purposefully had it on the device so far, and preloaded apps depend on that.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/224#-:-KOf5dipa3Y4j3xdtwNG:b-hs4j6h)*

alexandrosm:
We don't. We've decided against merging this. Let's make sure this doesn't
go through pls

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Change device registering so the server generates the UUID and returns a dedicated api key which we replace our provisioning key with. [Page]
```
pcarranzav: This description is no longer valid

src/bootstrap.coffee
```diff
@@ -32,20 +31,21 @@ bootstrap = ->
...
+			deviceApiKey: userConfig.deviceApiKey
+			provisioningApiKey: userConfig.apiKey
+			apiEndpoint: config.apiEndpoint
+		)
+		.then ({ id }) ->
 			userConfig.registered_at = Date.now()
-			userConfig.deviceId = device.id
+			userConfig.deviceId = id
+			# Delete the provisioning key now.
+			delete userConfig.apiKey
```
pcarranzav: For dependent devices, we'll probably need to keep the provisioning key, unless the device key works to provision (and query) dependent devices.
See: https://github.com/resin-io/resin-supervisor/pull/238/files#diff-1cad6517feeffa5e7c8b889dc57f761aR35

Page-: Keeping the provisioning key around would destroy the point of this PR :P  The device key can have permissions extended to create additional devices that depend on it though, although that'd be part of the API PR

CHANGELOG.md
Page-: Fixed

pcarranzav:
@Page- what's the status of this?
The code LGTM (though it needs a rebase), but I'd like to make sure that:
* The deviceApiKey has permissions to provision and manage dependent devices
* This PR https://github.com/resin-os/meta-resin/pull/223 gets merged first, so that we don't break the OS. We should also keep in mind that this won't be backwards compatible in the sense that it will break with an older OS (VPN won't be able to connect, which is critical).

Page-:
I've rebased this @pcarranzav and it should be good to review again now

src/app.coffee
```diff
@@ -32,6 +31,8 @@ knex.init.then ->
 
 		bootstrap.done
 		.then ->
+			# The device module communicates with the api and as such needs bootstrapping to complete
+			device = require './device'
```
pcarranzav: The parts that communicate with the API already happen after `bootstrap.done`, i.e. [getID](https://github.com/resin-io/resin-supervisor/blob/master/src/device.coffee#L21), and then [applyState](https://github.com/resin-io/resin-supervisor/blob/master/src/device.coffee#L182) which depends on getID.
So it should be fine to leave the require where it was, unless I'm missing something else?

Page-: It seems like it should be ok to do now, but I remember it causing issues before I moved it - it's likely something changed in the meantime

src/bootstrap.coffee
```diff
@@ -11,9 +10,6 @@ _ = require 'lodash'
 deviceConfig = require './device-config'
 userConfig = {}
 
-DuplicateUuidError = (err) ->
```
pcarranzav: Looks like we won't be catching this anymore... do we want to handle this case in the API? (make the register endpoint idempotent if the uuid exists and it's a valid provisioning key?)

Page-: I've added handling for it now

src/bootstrap.coffee
```diff
@@ -34,27 +35,71 @@ loadPreloadedApps = ->
...
+			applicationId: userConfig.applicationId
+			uuid: userConfig.uuid
+			deviceType: userConfig.deviceType
+			deviceApiKey: userConfig.deviceApiKey
+			provisioningApiKey: userConfig.apiKey
+			apiEndpoint: config.apiEndpoint
+		)
 		.timeout(config.apiTimeout)
 		.catch DuplicateUuidError, ->
+			console.log('Hit duplcate UUID')
```
pcarranzav: Debug code?

src/bootstrap.coffee
```diff
@@ -101,6 +148,13 @@ bootstrapper.done = new Promise (resolve) ->
 	bootstrapper.doneBootstrapping = ->
 		bootstrapper.bootstrapped = true
 		resolve(userConfig)
+		# If we're still using an old api key we can try to exchange it for a valid device key
+		if userConfig.apiKey?
```
pcarranzav: This will be done after the `resolve()` above... shouldn't we exchange the key before continuing? Also, this doesn't write to config.json after deleting userConfig.apiKey, so the file will not be modified

src/bootstrap.coffee
Page-: Initially, but I then expanded on it so it's possible to see in the supervisor logs what's going on with registration - I'll improve this one's message though

src/bootstrap.coffee
Page-: Fixed

PR #227, opened 2016-08-03
bootstrap: if offlineMode is enabled, persist only the uuid

petrosagg:
If we don't persist the uuid then every time the supervisor starts it
will think it's a new device. This triggers a wipe of the local state
and also a re-load of the preloaded apps. This in turn causes multiple
instances of the preloaded apps to be left running.

Signed-off-by: Petros Angelatos <petrosagg@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/227)

### DIFF
--- a/src/bootstrap.coffee
+++ b/src/bootstrap.coffee
@@ -111,7 +111,10 @@ bootstrapper.startBootstrapping = ->
 		readConfigAndEnsureUUID()
 		.tap ->
 			loadPreloadedApps()
-		.tap ->
-			bootstrapOrRetry()
+		.tap (uuid) ->
+			if bootstrapper.offlineMode
+				return knex('config').insert({ key: 'uuid', value: uuid })
+			else
+				return bootstrapOrRetry()
 
 module.exports = bootstrapper


lekkas:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---

*[src/bootstrap.coffee, line 116 \[r1\]](https://reviewable.io:443/reviews/resin-io/resin-supervisor/227#-KODz5i_CciVC4tGFoCh:-KODz5iaTb2Ktg-K3BrZ:-393794773) ([raw file](https://github.com/resin-io/resin-supervisor/blob/dfd354a63ce4e5a6702fe4d50e9a410c31920bf4/src/bootstrap.coffee#L116)):*
> ```CoffeeScript
> 		.tap (uuid) ->
> 			if bootstrapper.offlineMode
> 				return knex('config').insert([{ key: 'uuid', value: uuid }])
> ```

nit: brackets not required since it's a single row

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/resin-io/resin-supervisor/227#-:-KOE0H2gag9H4qxcVxQq:1417088721)*

src/bootstrap.coffee
```diff
@@ -111,7 +111,10 @@ bootstrapper.startBootstrapping = ->
 		readConfigAndEnsureUUID()
 		.tap ->
 			loadPreloadedApps()
-		.tap ->
-			bootstrapOrRetry()
+		.tap (uuid) ->
+			if bootstrapper.offlineMode
+				return knex('config').insert({ key: 'uuid', value: uuid })
```
Page-: What uses this / where is it read when in offline mode?

petrosagg: https://github.com/resin-io/resin-supervisor/blob/master/src/bootstrap.coffee#L105-L109

Page-: Right, but that's before this point, so it's already too late?

petrosagg: No, `knex('config').select('value').where(key: 'uuid')` doesn't throw if the key isn't there. It just returns an empty array.

Page-: So to make sure I understand this change is only for that check, ie to avoid reading config.json for the uuid when the supervisor gets restarted whilst in offline mode?

petrosagg: Before the my change the following was happening:

`bootstrapper.startBootstrapping` is called every time the supervisor starts. Then it will call `readConfig` and also a `knex('config').select('value').where(key: 'uuid')`. If it finds the uuid in sqlite, `startBoostrapping` just resolves to its value. Otherwise, it calls `readConfigAndEnsureUUID`, then `loadPreloadedApps` and then `bootstrapOrRetry`.

The problem is that `bootstrapOrRetry` is responsible for eventually setting the uuid value in sqlite, but because it's in offline mode this never happens. And that means that every time the supervisor starts in offline mode it ends up in the never resolving `bootstrapOrRetry` and running `loadPreloadedApps` again and again.

PR #229, opened 2016-08-13
Add iptables rules to block requests to the supervisor API from all interfaces except vpn, docker and local

pcarranzav:
connects to #228 

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/229)

### DIFF
--- a/src/utils.coffee
+++ b/src/utils.coffee
@@ -11,6 +11,7 @@ randomHexString = require './lib/random-hex-string'
 request = Promise.promisifyAll require 'request'
 logger = require './lib/logger'
 TypedError = require 'typed-error'
+execAsync = Promise.promisify(require('child_process').exec)
 
 # Parses package.json and returns resin-supervisor's version
 version = require('../package.json').version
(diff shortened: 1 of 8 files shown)


pcarranzav:
@petrosagg @Page-

petrosagg:
SGTM

src/utils.coffee
```diff
@@ -279,3 +280,15 @@ exports.validateKeys = (options, validSet) ->
 		return if !options?
 		invalidKeys = _.keys(_.omit(options, validSet))
 		throw new Error("Using #{invalidKeys.join(', ')} is not allowed.") if !_.isEmpty(invalidKeys)
+
+checkAndAddIptablesRule = (rule) ->
+	execAsync("iptables -C #{rule}")
+	.catch ->
```
Page-: Can't `execAsync` "succeed" (go to a `.then`) but with a non-zero exit code?

src/utils.coffee
```diff
@@ -279,3 +280,15 @@ exports.validateKeys = (options, validSet) ->
...
+		execAsync("iptables -A #{rule}")
+
+exports.createIpTablesRules = ->
+	allowedInterfaces = ['tun0', 'docker0', 'lo']
+	Promise.each allowedInterfaces, (iface) ->
```
Page-: This could be a `map` in order to execute all the iptables rules in parallel, but I don't mind either way (maybe there are issues if they happen in parallel, but I wouldn't expect so)

src/utils.coffee
pcarranzav: It shouldn't: https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

> If a callback function is provided, it is called with the arguments (error, stdout, stderr). On success, error will be null. On error, error will be an instance of Error. The error.code property will be the exit code of the child process while error.signal will be set to the signal that terminated the process. Any exit code other than 0 is considered to be an error.

src/utils.coffee
pcarranzav: Hmm, it feels off to run the iptables calls in parallel, that's the reason I chose `.each. purposefully

src/utils.coffee
Page-: Ok, it's been a while since I've used it myself and just had memories of exit code checking so wanted to check :)

src/utils.coffee
Page-: Ok

Page-:
LGTM

PR #232, opened 2016-08-15
WIP: Allow using an HTTP header for auth

pcarranzav:
connects to #231

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/232)

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -17,9 +17,15 @@ module.exports = (application) ->
 	parsedRouter.use(bodyParser())
 
 	api.use (req, res, next) ->
+		queryKey = req.query.apikey
+		header = req.get('Authorization') ? ''
+		match = header.match(/^ApiKey (\w+)$/)
+		headerKey = match?[1]
 		utils.getOrGenerateSecret('api')
 		.then (secret) ->
-			if bufferEq(new Buffer(req.query.apikey), new Buffer(secret))
+			if queryKey? && bufferEq(new Buffer(queryKey), new Buffer(secret))
+				next()
+			else if headerKey? && bufferEq(new Buffer(headerKey), new Buffer(secret))
 				next()
 			else
 				res.sendStatus(401)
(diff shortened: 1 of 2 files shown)


pcarranzav:
@petrosagg @Page-

petrosagg:
SGTM

src/api.coffee
```diff
@@ -17,9 +17,15 @@ module.exports = (application) ->
 	parsedRouter.use(bodyParser())
 
 	api.use (req, res, next) ->
+		queryKey = req.query.apikey
+		header = req.get('Authorization')
+		match = header.match(/^ApiKey ([a-f\d]{64})$/)
```
Page-: I don't think it makes sense for this to check for a fixed length/characters, for instance it won't work with the dev api key of 'bananas'

pcarranzav: True, fixed

src/api.coffee
```diff
@@ -17,9 +17,15 @@ module.exports = (application) ->
 	parsedRouter.use(bodyParser())
 
 	api.use (req, res, next) ->
+		queryKey = req.query.apikey
+		header = req.get('Authorization')
+		match = header.match(/^ApiKey ([a-f\d]+)$/)
```
Page-: This still won't work for 'bananas', those troublesome `n`s, not to mention that `s` :P

pcarranzav: :banana: damn, now really fixed

src/api.coffee
```diff
@@ -17,9 +17,15 @@ module.exports = (application) ->
 	parsedRouter.use(bodyParser())
 
 	api.use (req, res, next) ->
+		queryKey = req.query.apikey ? ''
```
Page-: You check for `queryKey?` below, but this line sets a default so that one will never be nullish

pcarranzav: Today's clearly not my brightest day. Fixed.

Page-:
LGTM

PR #237, opened 2016-08-22
add archiving logic for commit endpoint

curcuz:
added archiving logic and `tar-fs` dependency to proxyvisor module

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/237)


lifeeth:
@resin-jenkins ok to test

src/proxyvisor.coffee
```diff
@@ -57,9 +60,13 @@ router.get '/v1/assets/:commit', (req, res) ->
 		return res.status(404).send('Not found') if !app
 		getAssetsPath(app.imageId)
 	.then (path) ->
-		getTarArchive(path)
-	.then (archivePath) ->
-		res.sendFile(archivePath)
+		dest = __dirname + '/' + app.imageId + '.tar'
```
pcarranzav: Might be better to use /tmp for this? Either that or /data

curcuz: `/tmp` feels safer for eventual recoveries from saturation

src/proxyvisor.coffee
```diff
@@ -57,9 +60,13 @@ router.get '/v1/assets/:commit', (req, res) ->
 		return res.status(404).send('Not found') if !app
 		getAssetsPath(app.imageId)
 	.then (path) ->
-		getTarArchive(path)
-	.then (archivePath) ->
-		res.sendFile(archivePath)
+		dest = '/tmp/' + app.imageId + '.tar'
```
pcarranzav: Also, the imageId might have slashes on it, so this might try to store in directories that don't exist. Maybe use the app.commit as a filename instead of imageId?

curcuz: that's interesting. It's clear that i made a wrong assumption around `imageId` property - will update

Page-:
add to whitelist

src/proxyvisor.coffee
```diff
@@ -57,9 +60,13 @@ router.get '/v1/assets/:commit', (req, res) ->
...
-		getTarArchive(path)
-	.then (archivePath) ->
-		res.sendFile(archivePath)
+		dest = '/tmp/' + app.commit + '.tar'
+		getTarArchive(path,dest)
+	.then (archive) ->
+		archive.on 'finish', ->
+			res.sendFile(dest)
+	.catch (err) ->
+		console.log err
```
Page-: This should do a `res.send` of some sort as well, otherwise no response will be sent and the request will just hang until it times out

src/proxyvisor.coffee
```diff
@@ -86,7 +102,7 @@ exports.fetchDependentApps = (appId, apiKey) ->
 			localImageIds = _.map(localDependentApps, 'imageId')
 			toDelete = _.difference(localImageIds, remoteImageIds)
 			toDownload = _.difference(remoteImageIds, localImageIds)
-			Promise.map toDownload, 
+			Promise.map toDownload,
```
Page-: This is invalid coffeescript?

PR #238, opened 2016-08-23
Dependent apps and devices

pcarranzav:
Closes #235 
Fixes #209 as the new DB structure and way we apply host config should solve the issue.
Closes #25 as we now expose additional env vars.

+ switch to using a specific device state endpoint

PR #244, opened 2016-08-29
Various details on proxyvisor

pcarranzav:
@Page- addressing your comments from #238 here since we're testing that branch on a few devices already.

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/244)


Page-:
SGTM

PR #245, opened 2016-09-01
device.coffee: Remove hardcode of resin-boot partitions's mount point

telphan:
@pcarranzav @agherzan @floion *@emirotin (Sensei)

This PR should go hand in had with this: https://github.com/resin-os/meta-resin/pull/251

Signed-off-by: Theodor Gherzan <theodor@gherzan.ro>

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/245)


pcarranzav:

src/device.coffee
```diff
@@ -44,10 +44,19 @@ exports.getID = do ->
 exports.reboot = ->
 	request.postAsync(config.gosuperAddress + '/v1/reboot')
 
+exec = require('child_process')
```
emirotin: this should go to the top section, and better be `{ execSync } = require('child_process ')` (then just use this bare function)

pcarranzav:
@telphan I see a few issues with this:
* This code will throw synchronously if it fails, crashing the supervisor
* The `exec` definition would better go above with the other requires (and can be combined with the `execAsync` require).
* It would be better to make this function asynchronous, using `execAsync`, returning a Promise and changing `setBootConfig` so that it fulfills the promise to get the mountpoint.
* I don't know if /proc/mounts will return the proper mountpoint inside the container, and if so, it probably still needs `/mnt/root/` to be prepended.

Welcome to coffeescript ;)

All in all, this approach will work if we do those changes, but I wonder if it would be better to pass the mount point as an environment variable to the supervisor (as we do with config.json), defaulting to the old value for backwards compatibility. In that case, we just add an entry for this in `config.coffee` and then use `config.bootMountPoint` here. Maybe @Page- or @petrosagg can weigh in.

src/device.coffee
```diff
@@ -44,10 +44,19 @@ exports.getID = do ->
...
+
+findMountPointByPartLabel = (label) ->
+  command = "grep $(blkid -L #{label}) /proc/mounts"
+  try
+    return (exec.execSync(command).toString().split " ")[1]
```
Page-: This should be `return exec.execSync(command).toString().split(' ')[1]`

src/device.coffee
```diff
@@ -44,10 +44,19 @@ exports.getID = do ->
...
+  catch err
+    throw new Error ("Could not determine mount point for partition #{label}: #{stderr}, ExitCode: #{error.code}", { error: body.Error }, 'Set log to display error')
```
Page-: The `Error` constructor only takes one argument, I'm not sure what the second two actually are either?

src/device.coffee
```diff
@@ -47,7 +47,7 @@ exports.reboot = ->
 exports.hostConfigEnvVarPrefix = 'RESIN_HOST_'
 bootConfigEnvVarPrefix = 'RESIN_HOST_CONFIG_'
 bootBlockDevice = '/dev/mmcblk0p1'
-bootMountPoint = '/mnt/root/boot'
+bootMountPoint = '/mnt/root/' + config.bootMountPoint
```
pcarranzav: Looks like the last `/` in `/mnt/root/` should be removed here

src/config.coffee
```diff
@@ -49,3 +49,4 @@ module.exports =
 		'RESIN_API_KEY'
 	]
 	dataPath: checkString(process.env.RESIN_DATA_PATH) ? '/resin-data'
+  bootMountPoint: checkString(process.env.BOOT_MOUNTPOINT) ? '/boot'
```
pcarranzav: Tabs, not spaces ;)

telphan: Fixing.

src/device.coffee
telphan: You would be correct.

pcarranzav:
:lgtm:

Issue #247, opened 2016-09-06
Allow the new named interface used in the HOST OS for VPN


lifeeth:
Relates to https://github.com/resin-os/meta-resin/pull/283

This needs to be deployed before https://github.com/resin-os/meta-resin/issues/282

PR #248, opened 2016-09-06
Add iptables rules to allow resin-vpn named interface to be used by VPN

lifeeth:
Connects to #247

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/248)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+
+* Add iptables rules to allow resin-vpn named interface to be used by VPN [Praneeth]
 * Updated to coffee-script ~1.10.0 [Page]
 * Catch exec format error and provide friendlier error message [Aleksis]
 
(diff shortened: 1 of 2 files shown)


pcarranzav:
Not  sure if we should conditionally allow tun0 or resin-vpn depending on which we use?
Otherwise LGTM

lifeeth:
@pcarranzav I think for a really robust setup we should parse the openvpn config and figure out what to allow. Older configs mean tun* unfortunately.

floion:
Just a heads-up, the resin-vpm changes have been merged in meta-resin

pcarranzav:
Okay, since time presses let's merge this and add the conditional check as soon as possible.
I still think we can treat old configs as tun0, as users using that interface are edge cases (which will be already solved from meta-resin). What do you think @lifeeth ?

lifeeth:
@pcarranzav sorry missed this - Sounds good. This is atleast as good as what was before.

PR #249, opened 2016-09-07
Use config and environment from API, allow configurable hook endpoints

pcarranzav:
Improvements to #238 
Untested

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/249)

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -150,8 +150,12 @@ fetch = (app, setDeviceUpdateState = true) ->
 		device.updateState(status: 'Downloading', download_progress: 0)
 
 		Promise.try ->
-			JSON.parse(app.env)
-		.then (env) ->
+			# In dependent apps env is replaced with config
+			if !app.env? and app.config?
+				env = JSON.parse(app.config)
+			else
+				env = JSON.parse(app.env)
+
 			if env['RESIN_SUPERVISOR_DELTA'] == '1'
 				dockerUtils.rsyncImageWithProgress(app.imageId, onProgress)
 			else
@@ -463,13 +467,12 @@ getRemoteState = (uuid, apiKey) ->
 		console.error("Failed to get state for device #{uuid}. #{err}")
 		throw err
 
+# TODO: Actually store and use app.environment and app.config separately
 parseEnvAndFormatRemoteApps = (remoteApps, uuid, apiKey) ->
 	appsWithEnv = _.mapValues remoteApps, (app, appId) ->
 		utils.extendEnvVars(app.environment, uuid, appId)
-		.then (fullEnv) ->
-			env = _.omit(fullEnv, _.keys(specialActionEnvVars))
-			env = _.omit env, (v, k) ->
-				_.startsWith(k, device.hostConfigEnvVarPrefix)
+		.then (env) ->
+			fullEnv = _.merge({}, env, app.config)
 			return {
 				appId
 				commit: app.commit
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -370,3 +370,7 @@ do ->
 			res.json(data)
 		.catch (err) ->
 			res.status(500).send(err?.message or err or 'Unknown error')
+
+	exports.getImageEnv = (id) ->
+		docker.getImage(id).inspectAsync()
+		.get('Config').get('Env')
--- a/src/proxyvisor.coffee
+++ b/src/proxyvisor.coffee
@@ -1,5 +1,6 @@
 Promise = require 'bluebird'
-{ docker } = require './docker-utils'
+dockerUtils = require './docker-utils'
+{ docker } = dockerUtils
 express = require 'express'
 fs = Promise.promisifyAll require 'fs'
 { resinApi } = require './request'
@@ -44,7 +45,7 @@ router.post '/v1/devices', (req, res) ->
 				user: userId
 				application: req.body.applicationId
 				uuid: uuid
-				device_type: req.body.deviceType or 'edge'
+				device_type: 'edge'
 				device: deviceId
 				registered_at: Math.floor(Date.now() / 1000)
 				logs_channel: logsChannel
@@ -82,7 +83,7 @@ router.get '/v1/devices/:uuid', (req, res) ->
 		console.error('Error on GET /v1/devices/:uuid:', err, err.stack)
 		res.status(503).send(err?.message or err or 'Unknown error')
 
-router.put '/v1/devices/:uuid/logs', (req, res) ->
+router.post '/v1/devices/:uuid/logs', (req, res) ->
 	uuid = req.params.uuid
 	m = {
 		message: req.body.message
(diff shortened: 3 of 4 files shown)


src/proxyvisor.coffee
```diff
@@ -187,24 +203,44 @@ exports.fetchAndSetTargetsForDependentApps = (state, fetchFn) ->
...
-		sendUpdate(device) if device.targetCommit != device.commit or not _.isEqual(device.targetEnv, device.env)
+	knex('dependentApp').select()
+	.map (app) ->
+		dockerUtils.getImageEnv(app.imageId)
+		.then (imageEnv) ->
+			return imageEnv.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS if imageEnv.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS?
+			return app.config.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS if app.config.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS?
+			return "#{config.proxyvisorHookReceiver}/v1/devices/"
+		.then (endpoint) ->
+			return {
```
Page-: You could also do this as
```coffee
.map (app) ->
	return Promise.props {
		appId: app.appId
		endpoint: dockerUtils.getImageEnv(app.imageId)...
	}
```

src/proxyvisor.coffee
```diff
@@ -187,24 +203,44 @@ exports.fetchAndSetTargetsForDependentApps = (state, fetchFn) ->
...
 exports.sendUpdates = ->
-	# Go through knex('dependentDevice') and sendUpdate if targetCommit or targetEnv differ from the current ones.
-	knex('dependentDevice').select()
-	.map (device) ->
-		sendUpdate(device) if device.targetCommit != device.commit or not _.isEqual(device.targetEnv, device.env)
+	knex('dependentApp').select()
+	.map (app) ->
+		dockerUtils.getImageEnv(app.imageId)
+		.then (imageEnv) ->
+			return imageEnv.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS if imageEnv.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS?
```
Page-: I'm not sure if it's nicer but this could be
```coffee
return imageEnv.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS ? app.config.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS ? "#{config.proxyvisorHookReceiver}/v1/devices/"
```
or
```coffee
return imageEnv.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS ?
	app.config.RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS ?
	"#{config.proxyvisorHookReceiver}/v1/devices/"
```

src/proxyvisor.coffee
```diff
@@ -1,5 +1,6 @@
 Promise = require 'bluebird'
-{ docker } = require './docker-utils'
+dockerUtils = require './docker-utils'
+docker = dockerUtils.docker
```
Page-: This can be
```coffee
{ docker } = dockerUtils
```

src/docker-utils.coffee
```diff
@@ -369,4 +369,9 @@ do ->
...
 			res.json(data)
 		.catch (err) ->
-			res.status(500).send(err?.message or err or 'Unknown error')
\ No newline at end of file
+			res.status(500).send(err?.message or err or 'Unknown error')
+
+	exports.getImageEnv = (id) ->
+		docker.getImage(id).inspectAsync()
+		.then (img) ->
+			return image.Config.Env
```
Page-: This can also be `.get('Config').get('Env')`

src/application.coffee
```diff
@@ -463,8 +467,10 @@ getRemoteState = (uuid, apiKey) ->
 		console.error("Failed to get state for device #{uuid}. #{err}")
 		throw err
 
+# TODO: Actually store and use app.environment and app.config separately
 parseEnvAndFormatRemoteApps = (remoteApps, uuid, apiKey) ->
 	appsWithEnv = _.mapValues remoteApps, (app, appId) ->
+		_.merge(app.environment, app.config)
```
Page-: I think this can be dropped, along with the:
```coffee
env = _.omit(fullEnv, _.keys(specialActionEnvVars))
env = _.omit env, (v, k) ->
	_.startsWith(k, device.hostConfigEnvVarPrefix)
```
below

Page-: Wait, nvm, it should be:
```coffee
utils.extendEnvVars(app.environment, uuid, appId)
.then (env) ->
	fullEnv = _.merge({}, env, app.config)
	return ...
```

src/application.coffee
```diff
@@ -150,8 +150,12 @@ fetch = (app, setDeviceUpdateState = true) ->
 		device.updateState(status: 'Downloading', download_progress: 0)
 
 		Promise.try ->
-			JSON.parse(app.env)
-		.then (env) ->
+			# In dependent apps env is replaced with config
```
Page-: In all apps they're replaced?

pcarranzav: This will receive apps processed in `parseEnvAndFormatRemoteApps`, so for non-dependent apps it will still be `env` (until we actually start using `config`)

src/application.coffee
```diff
@@ -463,10 +467,12 @@ getRemoteState = (uuid, apiKey) ->
...
 		throw err
 
+# TODO: Actually store and use app.environment and app.config separately
 parseEnvAndFormatRemoteApps = (remoteApps, uuid, apiKey) ->
 	appsWithEnv = _.mapValues remoteApps, (app, appId) ->
 		utils.extendEnvVars(app.environment, uuid, appId)
-		.then (fullEnv) ->
+		.then (extendedEnv) ->
+			fullEnv = _.merge({}, extendedEnv, app.config)
 			env = _.omit(fullEnv, _.keys(specialActionEnvVars))
```
Page-: You can drop these omits now and just use `extendedEnv`, since these omits are just trying to get rid of the things that are only in app.config - so if you use the version before you add the app.config stuff you'll be fine

pcarranzav: Oh, true, brilliant

src/application.coffee
Page-: If this is coming from `parseEnvAndFormatRemoteApps` then `app.env` won't include `RESIN_SUPERVISOR_DELTA`?

I'm ok in general with shortcuts like this if they're necessary to get it ready a bit sooner, but I would like them to be cleaned up shortly after, rather than hanging around forever (since the supervisor code is already fairly complex)

pcarranzav: It will because we switch it to the `fullEnv` before actually running the update. It's ugly.
So yeah, as soon as this is merged and I clear up the pipeline a bit I'll clean this mess up.

pcarranzav:
@Page- done

PR #251, opened 2016-09-08
Update to coffee-script ~1.10.0

Page-:
Connects to #250 

The only change in the generated code is that in try/catch blocks where the error is not caught (no catch block, or no var used) then the error object name is now `_error` rather than `error` or `undefined` (?!) - no actual semantic differences

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/251)


pcarranzav:
:lgtm:

pcarranzav:
(Don't know why I remembered us having some issue with coffee 1.10, but my memory must have failed)

PR #253, opened 2016-09-09
DRY up calls to gosuper API, and use empty Host header

pcarranzav:
connects to #252
The Host header will allow requests to continue working when we move to Go 1.6 or above.

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/253)

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -48,14 +48,24 @@ module.exports = (application) ->
 		res.sendStatus(204)
 
 	unparsedRouter.post '/v1/reboot', (req, res) ->
-		utils.mixpanelTrack('Reboot')
-		request.post(config.gosuperAddress + '/v1/reboot')
-		.pipe(res)
+		new Promise (resolve, reject) ->
+			utils.mixpanelTrack('Reboot')
+			utils.gosuper.post('/v1/reboot')
+			.on('error', reject)
+			.on('response', -> resolve())
+			.pipe(res)
+		.catch (err) ->
+			res.status(503).send(err?.message or err or 'Unknown error')
 
 	unparsedRouter.post '/v1/shutdown', (req, res) ->
-		utils.mixpanelTrack('Shutdown')
-		request.post(config.gosuperAddress + '/v1/shutdown')
-		.pipe(res)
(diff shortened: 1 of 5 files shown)


src/api.coffee
```diff
@@ -49,12 +49,12 @@ module.exports = (application) ->
 
 	unparsedRouter.post '/v1/reboot', (req, res) ->
 		utils.mixpanelTrack('Reboot')
-		request.post(config.gosuperAddress + '/v1/reboot')
+		utils.gosuper.post('/v1/reboot')
 		.pipe(res)
```
Page-: This should have an error handler, otherwise the supervisor could crash from a stream piping error.

src/api.coffee
```diff
@@ -49,12 +49,12 @@ module.exports = (application) ->
...
 
 	unparsedRouter.post '/v1/shutdown', (req, res) ->
 		utils.mixpanelTrack('Shutdown')
-		request.post(config.gosuperAddress + '/v1/shutdown')
+		utils.gosuper.post('/v1/shutdown')
 		.pipe(res)
```
Page-: This should have an error handler, otherwise the supervisor could crash from a stream piping error.

src/utils.coffee
```diff
@@ -177,10 +177,29 @@ exports.resinLogControl = (val) ->
...
 	console.log('Logs enabled: ' + val)
 
+gosuperRequest = (method, endpoint, options = {}, callback) ->
+	if _.isFunction(options)
+		callback = options
+		options = {}
+	options.method = method
+	options.url = config.gosuperAddress + endpoint
+	options.headers ?= {}
+	options.headers['Host'] = ''
```
Page-: Would it be better/nicer to use [request.defaults](https://github.com/request/request#requestdefaultsoptions) for this?

pcarranzav: Changed it  to use it @Page- 
I think we can probably use it better (e.g. have gosuperPost and gosuperGet be defaults on top of this), but it feels we don't win much and the result might be a bit convoluted.

Page-:
SGTM

PR #254, opened 2016-09-09
Bind mount linker in amd64 and libc in all platforms when we mount kmod

### LINKED ISSUE #221: host OS kmod linker is at a different path on amd64 platforms
Yocto puts the linker at `/lib/ld-linux-x86-64.so.2`, but debian puts it at `/lib64/ld-linux-x86-64.so.2`. Since we bind-mount the `kmod` binary from the host in the container, it fails to run on amd64 platforms.

The current workaround is to run `ln -s /lib64/ld-linux-x86-64.so.2 /lib/ld-linux-x86-64.so.2` in the container.

The supervisor could bind-mount the linker as well so that the workaround isn't needed

pcarranzav:
Fixes #221

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-supervisor/254)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Bind mount linker in amd64 platforms when we mount kmod [Pablo]
 * DRY up calls to gosuper API, and use empty Host header [Pablo]
 
 # v2.1.1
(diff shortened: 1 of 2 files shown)


pcarranzav:
Will be fixed on the base images.

PR #271, opened 2016-09-28
Fix disabling logs to display on newer OS with different service name

### LINKED ISSUE #239: Broken feature: tty-replacement disabling
Due to https://github.com/resin-os/meta-resin/commit/10e46c614e979bafdad0faf9c57870724232f6e0 this supervisor feature was broken. The tty-replacement.service was renamed to resin-info@tty1.service.

pcarranzav:
Fixes #239

### DIFF
--- a/gosuper/gosuper/api.go
+++ b/gosuper/gosuper/api.go
@@ -223,7 +223,8 @@ func VPNControl(writer http.ResponseWriter, request *http.Request) {
 //LogToDisplayControl is used to control tty-replacement service status with dbus
 func LogToDisplayControl(writer http.ResponseWriter, request *http.Request) {
 	sendResponse, sendError := responseSenders(writer)
-	serviceName := "tty-replacement.service"
+	serviceName := "resin-info@tty1.service"
+	serviceNameOld := "tty-replacement.service"
 	var body LogToDisplayBody
 	if err := parseJSONBody(&body, request); err != nil {
 		log.Println(err)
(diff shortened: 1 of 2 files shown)


pcarranzav:
(untested)

gosuper/gosuper/api.go
```diff
@@ -236,6 +237,15 @@ func LogToDisplayControl(writer http.ResponseWriter, request *http.Request) {
 		return
 	}
 
+	if loaded, err := systemd.Dbus.GetUnitProperty(serviceName, "LoadState"); err != nil {
+		sendError(fmt.Errorf("Unable to get log to display load status: %v", err))
+		return
+	} else if loaded.Value.String() == `"not-found"` {
```
Page-: Is there maybe an exit code that can be checked for, rather than a hardcoded string?

pcarranzav: We could use systremd's GetUnit dbus method, but go-systemd doesn't seem to support it. And we already check against a string for "active" below, so in any case this is consistent (though I agree the hardcoded strings are ugly).
BTW, I still haven't tested if this works like this, I'll check it now (it's an educated guess from what `systemctl status` returns).

pcarranzav:
Successfully tested on an rpi3.

gosuper/gosuper/api.go
Page-: Yeah, my worry is running this on a system that isn't using English strings, or just if systemd decides to reword the outputs - if there's no avoiding it then I can live with it, but would rather avoid it if possible

pcarranzav: True, but so far this action only makes sense on Resin OS, where we can keep track of things changing like that  (systemd versions changing, or use of locales)

PR #273, opened 2016-09-28
Ignore resinhup images and containers during cleanup

### LINKED ISSUE #144: Ignore resinhup images when deleting downloaded docker images
Resinhup images format:
https://github.com/resin-os/resinhup/blob/master/README.md

To be considered that in the future we might switch from internal registry to dockerhub,

pcarranzav:
Closes #144 

cc @agherzan

### DIFF
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -10,6 +10,7 @@ knex = require './db'
 Lock = require 'rwlock'
 utils = require './utils'
 rimraf = Promise.promisify(require('rimraf'))
+device = require './device'
 
 docker = new Docker(socketPath: config.dockerSocket)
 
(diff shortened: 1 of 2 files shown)


src/docker-utils.coffee
```diff
@@ -127,9 +129,10 @@ do ->
 					supervisorTags = _.flatten(supervisorTags)
 					appTags = _.flatten(appTags)
 					locallyCreatedTags = _.flatten(locallyCreatedTags)
-					return { images, supervisorTags, appTags, locallyCreatedTags }
+					resinhupTagRegex = new RegExp("resinhup/resinhup-#{deviceType}$")
```
Page-: I'm not particularly a fan of hardcoding this for a specific image, wouldn't this tie in with side-loading stuff? Maybe have resinhup be side-loaded and ran that way or something?

agherzan:
Nice

src/docker-utils.coffee
pcarranzav: That would require resinhup to have access to the supervisor API key, which it currently doesn't. It would also get us into an entanglement between the supervisor and resinhup which I'd like to avoid (as resinhup aggressively changes the underlying OS and will know best when to stop the supervisor, etc).
This hardcoding gives an easy win by allowing a shorter downtime, because the updater doesn't need to stop the supervisor while pulling the resinhup image.

pcarranzav:
Closing this since it grew old and there was no consent on whether to add the hardcoded check or not. Will leave the branch there, and reassess how we'll implement this.

PR #275, opened 2016-09-28
Update `dindctl logs` to use journalctl now that we send logs there

Page-:
Connect to #274


Page-:
It's your call @pcarranzav

pcarranzav:
then kill that changelog entry with fire @Page- :P

PR #279, opened 2016-09-29
On cleanup, force removal for images and containers, and remove container volumes

pcarranzav:
Connects to #278

### DIFF
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -145,7 +145,7 @@ do ->
 							return true
 						return containerHasExited(containerInfo.Id)
 				.map (containerInfo) ->
-					docker.getContainer(containerInfo.Id).removeAsync()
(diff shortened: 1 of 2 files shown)


src/docker-utils.coffee
```diff
@@ -145,7 +145,7 @@ do ->
 							return true
 						return containerHasExited(containerInfo.Id)
 				.map (containerInfo) ->
-					docker.getContainer(containerInfo.Id).removeAsync()
+					docker.getContainer(containerInfo.Id).removeAsync(v: true, force: true)
```
Page-: So adding the `v: true` here is huge, however one thing I wonder is how much this will help due to the `containerHasExited(containerInfo.Id)` check above (meaning it should be stopped and hence able to be removed anyway)

pcarranzav: That check will only happen for supervisor containers (see the check for supervisor tags right above)

PR #281, opened 2016-10-02
resumable delta downloads

### LINKED ISSUE #140: Delta fetch resume support
Adding the ability to resume a delta download at any point would improve things under poor network conditions massively.  Currently it is restarted from the beginning each time, and relies on the reduced size of most deltas to offset the lack of any resume support.

abresas:
Closes #140

### DIFF
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -10,6 +10,7 @@ knex = require './db'
 Lock = require 'rwlock'
 utils = require './utils'
 rimraf = Promise.promisify(require('rimraf'))
+resumable = require 'resumable-request'
 
 docker = new Docker(socketPath: config.dockerSocket)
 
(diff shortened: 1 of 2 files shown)


abresas:
@pcarranzav I don't understand the failure message. What gives?

pcarranzav:
@abresas it's an error on Jenkins, not related to the PR. Trying to get it fixed to retry the build

pcarranzav:
Have you tested this @abresas?

abresas:
@pcarranzav I haven't tested it with supervisor, no. Any tips on how to do it easily? Haven't done testing with supervisor for ages.

src/docker-utils.coffee
```diff
@@ -72,7 +73,7 @@ do ->
 				findSimilarImage(imgDest)
 			.then (imgSrc) ->
 				new Promise (resolve, reject) ->
-					progress request.get("#{config.deltaHost}/api/v2/delta?src=#{imgSrc}&dest=#{imgDest}", timeout: DELTA_REQUEST_TIMEOUT)
+					progress resumable(request, { url: "#{config.deltaHost}/api/v2/delta?src=#{imgSrc}&dest=#{imgDest}", timeout: DELTA_REQUEST_TIMEOUT })
```
pcarranzav: @abresas I just tested this and it breaks because the statusCode below (line 82) is now 206 instead of 200. Not sure if just testing for 206 should work?

pcarranzav:
I tested this allowing the 206 status and it works when there's no interruptions.
I tried interrupting the network and the result was an rsync exit status 30 ("timeout in data send/receive") so this needs a bit more investigation. It could be that rsync timed out while resumable-request was retrying, or that the data produced by the resumed request was invalid.

pcarranzav:
@petrosagg you said today you'd be working on resumable deltas - is it okay to close this PR? my understanding is your approach will be different, right?

pcarranzav:
Closing this since there's other ongoing work that should solve this issue
cc @petrosagg

PR #285, opened 2016-10-03
Update to lodash 4

### LINKED ISSUE #283: _.isNil doesn't exist in lodash 3.10, which breaks config.txt

Page-:
Fixes #283

### DIFF
--- a/src/proxyvisor.coffee
+++ b/src/proxyvisor.coffee
@@ -149,11 +149,11 @@ router.put '/v1/devices/:uuid', (req, res) ->
 			resinApi.patch
 				resource: 'device'
 				id: device.deviceId
-				body: _.pick({ status, is_online, commit }, isDefined)
+				body: _.pickBy({ status, is_online, commit }, isDefined)
 				customOptions:
 					apikey: apiKey
 			.then ->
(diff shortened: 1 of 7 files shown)


pcarranzav:
Fixes #283

Page-:
Yeah, tests are definitely needed, the way I checked was to go through the lodash changelog and searched for things list there in the code, and also went through the uses of lodash in the code and searched for changes to those uses in the lodash changelog, so it got checked in both ways to be a bit safer (I've found this method works pretty well for other places I've converted, but it's always possible that something could slip through)

src/proxyvisor.coffee
```diff
@@ -218,9 +218,9 @@ exports.fetchAndSetTargetsForDependentApps = (state, fetchFn, apiKey) ->
 		localApps = _.indexBy(localDependentApps, 'appId')
```
pcarranzav: @Page- this needs to be replaced by `_.keyBy(localDependentApps, 'appId')`

Page-: Fixed, I think I might have missed it in a merge

pcarranzav:
I've given this a bit of testing and it seems okay. I recommend some testing on your side to make sure, but looks good to merge.

PR #288, opened 2016-10-04
document dependent apps feature

### LINKED ISSUE #367: Document dependent devices
Add documentation for dependent devices/application supervisor endpoints and hooks

curcuz:
Closes #367

### DIFF
--- a/docs/dependent-apps.md
+++ b/docs/dependent-apps.md
@@ -1,0 +1,292 @@
+# Using the Resin Supervisor to manage dependent applications
+
+Since version 2.5.0 the Resin Supervisor can act as a proxy for dependent apps.
+
+Only Supervisors after version 2.5.0 have this functionality, and some of the endpoints appeared in later versions (we've noted it down where this is the case).
+
+## What is a dependent application
+
+A **dependent application** is a resin application that targets devices not capable of interacting directly with the Resin API - the reasons can be several, the most common are:
+
+- no direct Internet capabilities
+- not able to run Resin OS (being a microcontroller, for example)
+
+The **dependent application** is scoped under a resin application, which gets the definition of **gateway application**.
+
+The **gateway application** is responsible for detecting, provisioning and managing **dependent devices** belonging to one of its **dependent applications**. This is possible leveraging a new set of endpoints exposed by the Resin Supervisor.
+
+When a new version of the dependent application is git pushed, the supervisor will download the docker image and expose the assets in one of the endpoints detailed below. It is then the gateway application (i.e. the user app that is run by the supervisor) that is responsible for ensuring those assets get deployed to the dependent devices, using the provided endpoints to perform the management.
+
+A dependent application follows the same development cycle of a conventional resin application:
+
+- it binds to your git workspace via the **resin remote**
+- it consists in a Docker application
+- it offers the same environment and configuration variables management
+
+There are some differences:
+
+- it does not support Dockerfile templating
+- the Dockerfile must target either an `x86` or `amd64` base image
+- the actual firmware/business logic must be stored in the `/assets` folder within the built docker image.
+  - You can either just `COPY` a pre-built artifact in that folder, or build your artifact at push time and then store it in the `/assets` folder.
+- **a dependent application Docker image is only used to build, package and deliver the firmware on the dependent device via resin-supervisor - it won't be run at any point.**
+
+## How a dependent application works
+
+### Endpoints
+
+The supervisor exposes a REST API to interact with the dependent applications and dependent devices models that come from the Resin API - it also allows using a set of hooks to have push functionality, both documented below.
+
+# HTTP API reference
+
+## Applications
+
+### GET /v1/dependent-apps
+Dependent Applications List
+
+**Example**
+
+```bash
+curl -X GET $RESIN_SUPERVISOR_ADDRESS/v1/dependent-apps?apikey=$RESIN_SUPERVISOR_API_KEY
+```
+
+**Response**
+`HTTP/1.1 200 OK`
+
+```javascript
+[
+     {
+         "id": 13015,
+         "device_type": "edge",
+         "name": "edgeApp1",
+         "commit": "d43bea5e16658e653088ce4b9a91b6606c3c2a0d",
+         "config": {}
+     },
+     {
+         "id": 13016,
+         "device_type": "edge",
+         "name": "edgeApp2",
+         "commit": "d0f6624d6410fa079159fa3ebe0d3af46753d75d",
+         "config": {}
+     }
+ ]
+```
+
+### GET /v1/dependent-apps/:appId/assets/:commit
+Dependent Application Updates Registry
+
+**Example**
+
+```bash
+curl -X GET $RESIN_SUPERVISOR_ADDRESS/v1/dependent-apps/<appId>/assets/<commit>?apikey=$RESIN_SUPERVISOR_API_KEY
+```
+
+**Response**
+`HTTP/1.1 200 OK`
+
+
+```none
+[application/x-tar] .tar
+```
+
+## Devices
+
+### GET /v1/devices
+Dependent Devices List
+
+**Example**
+
+```bash
+curl -X GET $RESIN_SUPERVISOR_ADDRESS/v1/devices?apikey=$RESIN_SUPERVISOR_API_KEY
+```
+
+**Response**
+`HTTP/1.1 200 OK`
+
+
+```javascript
+[
+     {
+         "id": 1,
+         "uuid": "5ae8cf6e062c033ea38435498ad9b487bcc714e9eab0fed0404ee56e397790",
+         "appId": 13015,
+         "device_type": "edge",
+         "logs_channel": "69f961abffaad1ff66031b29f712be4fb19e1bfabf1fee7a9ebfb5fa75a1fbdb",
+         "deviceId": "47270",
+         "is_online": null,
+         "name": "blue-sun",
+         "status": "Provisioned",
+         "download_progress": null,
+         "commit": "d43bea5e16658e693088ce4b9a91b6606c3c2a0d",
+         "targetCommit": "d43bea5e16653e653088ce4b9a91b6606c3c2a0d",
+         "environment":{},
+         "targetEnvironment":{},
+         "config":{},
+         "targetConfig":{"RESIN_SUPERVISOR_DELTA":"1"}
+     },
+     {
+         "id": 3,
+         "uuid": "8dc608765fd32665d49d218a7eb4657bc2ab8a56db06d2c57ef7c7e9a115da",
+         "appId": 13015,
+         "device_type": "edge",
+         "logs_channel": "d0244a90e8cd6e9a1ab410d3d599dea7f15110a6fe37b2a8fd69bb6ee0bce043",
+         "deviceId": "47318",
+         "is_online": null,
+         "name": "wild-paper",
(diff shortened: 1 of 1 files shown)


docs/dependent-apps.md
```diff
@@ -0,0 +1,277 @@
...
+- the Dockerfile must target a `x86` base image
+- the actual firmware/business logic must be stored in the `/assets` folder within the built docker image.
+
+  - You can either just `COPY` a pre-built artifact in that folder, or build your artifact at push time and then store it in the `/assets` folder.
+
+## How a dependent application works
+
+### Design
+
+Resin Supervisor exposes a REST interface when the data-flow goes from the Resin Application to the Resin API (from now on, called **pull** dataflow) and a **Hook** mechanism when the data-flow goes from the Resin API to the Resin Application (from now on, called **push** dataflow). The term **Hook** is meant as a convention of HTTP requests made from the Resin Supervisor to the gateway application
```
josephroberts: Can we change `resin application` to `gateway application` please and also standardise on either `resin API` or `resin supervisor`

docs/dependent-apps.md
```diff
@@ -0,0 +1,277 @@
...
+           "download_progress": null,
+           "commit": "d43bea5e16658e653088ce4b9a11b6606c3c2a0d",
+           "targetCommit": "d43bea5e16658e653088se4b9a91b6606c3c2a0d",
+           "env": null,
+           "targetEnv": null
+       }
+```
+
+### POST /v1/devices/:uuid/logs
+Dependent Dependent Device Log
```
josephroberts: I think we can remove the first `Dependent` here?

docs/dependent-apps.md
```diff
@@ -0,0 +1,277 @@
...
+
+```bash
+curl -H "Content-Type: application/json" -X PUT /
+http://127.0.0.1:1337/v1/devices/<uuid>/restart
+```
+**Response**
+`HTTP/1.1 200 OK`
+
+### PUT /v1/devices/:uuid
+Dependent Device Update Notification
```
josephroberts: What are your thoughts on having a second response the gateway can send to stop the proxy-visor from continually polling this endpoint? In the case of the resin ENM its not necessary to continually poll and results in unnecessary writes to the `/data` partition

docs/dependent-apps.md
```diff
@@ -0,0 +1,277 @@
...
+
+**Example**
+
+```bash
+curl -H "Content-Type: application/json" -X PUT /
+--data '{"commit":" <commit>","environment": "<environment>"}' http://127.0.0.1:1337/v1/devices/<uuid>
+```
+**Response**
+`HTTP/1.1 200 OK`
```
curcuz: I don't think this is the right place to raise that question, since this PR aims to document the state of the art.

docs/dependent-apps.md
curcuz: why it would result in unecessary writes in the `/data` partition?

docs/dependent-apps.md
curcuz: yes

docs/dependent-apps.md
curcuz: resin API and resin supervisor are 2 different things

josephroberts: seems like these two statements conflict then? Probably worth explaining the difference anyway as its not clear to me

> a **Hook** mechanism when the data-flow goes from the Resin API to the Resin Application

> **Hook** is meant as a convention of HTTP requests made from the Resin Supervisor to the gateway application

docs/dependent-apps.md
josephroberts: Probably not the right place. but essentially I haven't implemented any DB caching so each time I get the request I write the target commit for the device into my DB which is stored on `/data`. Furthermore it could potentially be days before the micro comes online for an update and it seems wasteful to keep making the request given that the ENM has already told the proxy-visor it is going to handle it

curcuz: that seems very specific to your implementation, and not with this - will file an issue for that, which is not going to be that specific anyways

docs/dependent-apps.md
curcuz: Resin Supervisor acts as a proxy between the Gateway application and the Resin API for dependent applications. Will extend the explanation to make it more clear

pcarranzav:

docs/dependent-apps.md
```diff
@@ -0,0 +1,277 @@
...
+
+```bash
+curl -X GET $RESIN_SUPERVISOR_ADDRESS/v1/devices?apikey=$RESIN_SUPERVISOR_API_KEY
+```
+
+**Response**
+`HTTP/1.1 200 OK`
+
+
+```javascript
```
josephroberts: Missing `config` and `targetConfig`, `env` fields named incorrectly, e.g.
```
{"id":48097,"uuid":"3b1b4444c04870de8f22747ba5c68de6af5e01a3db1c3f7fcd43034b37fab5","appId":13829,"device_type":"edge","logs_channel":"d350a2d189329abfa1e47f9a210d17e740ff656921cfe59fd2a32aa6f9b6961d","deviceId":"48097","is_online":null,"name":"silent-moon","status":"Provisioned","download_progress":null
,"commit":null,"targetCommit":"dd61526c9aa37ea3b1fd50cd639d85e6274a3990","environment":{},"targetEnvironment":{},"config":{},"targetConfig":{"RESIN_SUPERVISOR_DELTA":"1"}}
```

docs/dependent-apps.md
```diff
@@ -0,0 +1,289 @@
+# Using the Resin Supervisor for managing dependent applications
```
pcarranzav: I think "Using the Resin Supervisor to manage dependent applications" sounds better, not sure if it's more grammatically correct?

docs/dependent-apps.md
```diff
@@ -0,0 +1,289 @@
...
+- the Dockerfile must target a `x86` base image
+- the actual firmware/business logic must be stored in the `/assets` folder within the built docker image.
+
+  - You can either just `COPY` a pre-built artifact in that folder, or build your artifact at push time and then store it in the `/assets` folder.
+
+## How a dependent application works
+
+### Design
+
+Resin Supervisor exposes a REST interface when the data-flow goes from the gateway application to the Resin API (from now on, called **pull** dataflow) and a **Hook** mechanism when the data-flow goes from the Resin API to the gateway application (from now on, called **push** dataflow). The term **Hook** is meant as a convention of HTTP requests made from the Resin Supervisor to the gateway application
```
pcarranzav: This description with data-flows feels a bit awkward. I'd reorganize this paragraph to just say that the supervisor  exposes a REST API to interact with the models that come from a Resin API, and allows using a set of hooks to have push functionality.
Moreover, users could want to completely ignore the hooks and work only through the API, since all the relevant information is there.

docs/dependent-apps.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Hook
+
+It's worth noting the **Hook** APIs described in this document consist in the opposite direction flow: It's the Resin Supervisor consuming a REST interface exposed by the gateway application. These methods are described as **what the Resin Supervisor expects to find on a push dataflow**
+
+### Hook configuration
+
+You can point the supervisor where to find the hook server via a configuration variable.
+
+- `RESIN_DEPENDENT_DEVICES_HOOK_ADDRESS` _(defaults to `http://0.0.0.0:1337/v1/devices/`)_
```
pcarranzav: Might be good to point out here that the supervisor will append the dependent device uuid to every hook request URL.

docs/dependent-apps.md
```diff
@@ -0,0 +1,289 @@
...
+**Example**
+
+```bash
+curl -H "Content-Type: application/json" -X POST --data '{"message":"detected movement","timestamp":1472142960}' /
+$RESIN_SUPERVISOR_ADDRESS/v1/devices/<uuid>/logs?apikey=$RESIN_SUPERVISOR_API_KEY
+```
+**Response**
+`HTTP/1.1 202 ACCEPTED`
+
+## Hook (the request the Resin Supervisor performs)
```
pcarranzav: Should be plural: "Hooks" and "requests"

pcarranzav: All of this should probably go together with the Hook configuration above, instead of having the API endpoints explained in the middle.

docs/dependent-apps.md
```diff
@@ -0,0 +1,289 @@
...
+A **dependent application** is a resin application that targets devices not capable of interacting directly with the Resin API - the reasons can be several, the most common are:
+
+- no direct Internet capabilities
+- not able to run Resin OS (being a microcontroller, for example)
+
+The **dependent application** is scoped under a resin application, which gets the definition of **gateway application**.
+
+The **gateway application** is responsible for detecting, provisioning and managing **dependent devices** belonging to one of its **dependent applications**. This is possible leveraging a new set of endpoints exposed by the Resin Supervisor.
+
+A dependent application follows the same development cycle of a conventional resin application:
```
lekkas: It's also mentioned below that a dependent application:

> consists in a Docker application
> the actual firmware/business logic must be stored in the `/assets` folder within the built docker image.

So if I understood this correctly, the resulting dependent application Docker image is only used to build, package and deliver the firmware in the `/assets` folder on the device, is that correct? If so, we might want to express this more clearly in the doc and stress that, in contrast with both conventional and gateway applications, a `git push` for a dependent app will not result in a new running container on the device.

docs/dependent-apps.md
```diff
@@ -0,0 +1,288 @@
...
+
+- it binds to your git workspace via the **resin remote**
+- it consists in a Docker application
+- it offers the same environment and configuration variables management
+
+There are some differences:
+
+- it does not support Dockerfile templating
+- the Dockerfile must target a `x86` base image
```
pcarranzav: I thought amd64 worked too?

docs/dependent-apps.md
pcarranzav: This one hasn't been addressed yet?

docs/dependent-apps.md
```diff
@@ -0,0 +1,288 @@
...
+## What is a dependent application
+
+A **dependent application** is a resin application that targets devices not capable of interacting directly with the Resin API - the reasons can be several, the most common are:
+
+- no direct Internet capabilities
+- not able to run Resin OS (being a microcontroller, for example)
+
+The **dependent application** is scoped under a resin application, which gets the definition of **gateway application**.
+
+The **gateway application** is responsible for detecting, provisioning and managing **dependent devices** belonging to one of its **dependent applications**. This is possible leveraging a new set of endpoints exposed by the Resin Supervisor.
```
pcarranzav: I'd add something to make the responsibilities clearer, like:
> When a new version of the dependent application is `git push`ed, the supervisor will download the docker image and expose the assets in one of the endpoints detailed below. It is then the gateway application (i.e. the user app that is run by the supervisor) that is responsible for ensuring those assets get deployed to the dependent devices, using the provided endpoints to perform the management.

It doesn't cover the *whole* flow, but at least it hints at the user what things we're not doing for the device management.

docs/dependent-apps.md
curcuz: addressed in latest commit

josephroberts:
@curcuz LGTM. It would be really nice to show the error responses and what they mean as well as the success responses.

docs/dependent-apps.md
curcuz: It is now on L256

PR #294, opened 2016-10-04
Use getRegistryAndName from docker-toolbelt 1.2.0

### LINKED ISSUE #293: Image name parsing for supervisor images in cleanup is broken
And docker-progress 2.2.0 surfaced the bug causing all update cycles to fail

pcarranzav:
Fixes #293 

Depends on https://github.com/resin-io-modules/docker-toolbelt/pull/2 being merged and published as 1.2.0.

### DIFF
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -1,5 +1,5 @@
 Docker = require 'docker-toolbelt'
-{ getRegistryAndName, DockerProgress } = require 'docker-progress'
+{ DockerProgress } = require 'docker-progress'
 Promise = require 'bluebird'
 progress = require 'request-progress'
 dockerDelta = require 'docker-delta'
@@ -46,7 +46,7 @@ findSimilarImage = (repoTag) ->
 DELTA_REQUEST_TIMEOUT = 15 * 60 * 1000
 
 getRepoAndTag = (image) ->
-	getRegistryAndName(image)
+	docker.getRegistryAndName(image)
 	.then ({ registry, imageName, tagName }) ->
 		registry = registry.toString().replace(':443', '')
 		return { repo: "#{registry}/#{imageName}", tag: tagName }
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -101,27 +101,32 @@ do ->
 		Promise.using readLockImages(), ->
 			dockerProgress.pull(image, onProgress)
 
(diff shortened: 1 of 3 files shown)


pcarranzav:
@Page- I also added a normalization to all image tags (otherwise this caused the supervisor to kill itself, but it also makes the cleanup more robust). Still approve?

src/docker-utils.coffee
```diff
@@ -101,27 +101,30 @@ do ->
 		Promise.using readLockImages(), ->
 			dockerProgress.pull(image, onProgress)
 
-	supervisorTag = config.supervisorImage
```
Page-: I think it makes sense to keep this but as `supervisorTag = normalizeRepoTag(config.supervisorImage)` to avoid repeating the normalisation work each time (it might also be worth making the normalisation/splitting stuff return synchronously, as it doesn't actually need anything async)

src/docker-utils.coffee
```diff
@@ -101,27 +101,30 @@ do ->
...
-	if !/:/g.test(supervisorTag)
-		# If there is no tag then mark it as latest
-		supervisorTag += ':latest'
+	normalizeRepoTag = (image) ->
```
Page-: It probably makes sense to have this in the docker-toolbelt?

pcarranzav: It does, but I'd rather have it here for now (`buildRepoTag` is used below in other places) and split it out later, so that we can fix this bug quickly.

src/docker-utils.coffee
pcarranzav: Kept it as a promise, but made it so that it's only computed once

pcarranzav: (The getRegistryAndName can throw, so it does make sense that it's wrapped as a Promise)

Page-: Well it definitely needs to be error checked, but having something synchronous wrapped in a promise does mean you have to deal with async stuff when you don't necessarily have to

src/docker-utils.coffee
Page-: I mean I think it'd make sense for buildRepoTag to be in docker to also be in docker-toolbelt?

pcarranzav: Yeah, I'll move that out soon, just couldn't have peace of mind until this bug was fixed

PR #300, opened 2016-10-05
Improvements on config and dependent device handling

### LINKED ISSUE #291: Updates should not be marked as failed if the proxyvisor hook didn't respond

### LINKED ISSUE #298: appId should be an int when hitting the dependent devices hook

### LINKED ISSUE #299: Setting the dependent devices hook address on a config var doesn't seem to work
But it works when the variable is set on the parent app's Dockerfile...

pcarranzav:
* Store config vars when there's nothing else to update
* Do not mark an update as failed if the hook failed
* When hitting the dependent devices hook, send appId as int

Fixes #299 
Fixes #298 
Fixes #291

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -91,6 +91,16 @@ logTypes =
 		eventName: 'Application restart'
 		humanName: 'Restarting application'
 
+	updateAppConfig:
+		eventName: 'Application config update'
+		humanName: 'Updating config for application'
+	updateAppConfigSuccess:
+		eventName: 'Application config updated'
+		humanName: 'Updated config for application'
+	updateAppConfigError:
+		eventName: 'Application config update error'
+		humanName: 'Failed to update config for application'
+
 application = {}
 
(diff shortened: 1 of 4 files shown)


src/application.coffee
```diff
@@ -636,6 +650,18 @@ application.update = update = (force) ->
...
+							# It can notoriously affect setting dep. devices hook address
+							# if nothing else changes in the app.
+							# So we just save them.
+							app = remoteApps[appId]
+							logSystemEvent(logTypes.updateAppConfig, app)
+							knex('app').update(app).where({ appId })
+							.then ->
+								logSystemEvent(logTypes.updateAppConfigSuccess, app)
+							.catch (err) ->
+								logSystemEvent(logTypes.updateAppConfigError, app, err)
```
Page-: I think this still needs to throw the error for later handling?

pcarranzav: @Page- done

PR #306, opened 2016-10-06
Add validation to dependent device provisioning

### LINKED ISSUE #305: Dependent devices can be provisioned without appId
Which causes mayhem on the sqlite db

pcarranzav:
Fixes #305

### DIFF
--- a/src/proxyvisor.coffee
+++ b/src/proxyvisor.coffee
@@ -48,6 +48,10 @@ router.get '/v1/devices', (req, res) ->
 		res.status(503).send(err?.message or err or 'Unknown error')
 
 router.post '/v1/devices', (req, res) ->
+	appId = req.body.appId
+	if !appId? or _.isNaN(parseInt(appId)) or parseInt(appId) <= 0
+		res.status(400).send('appId must be a positive integer')
+		return
 	Promise.join(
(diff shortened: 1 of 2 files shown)


src/proxyvisor.coffee
```diff
@@ -48,6 +48,10 @@ router.get '/v1/devices', (req, res) ->
 		res.status(503).send(err?.message or err or 'Unknown error')
 
 router.post '/v1/devices', (req, res) ->
+	appId = req.body.appId
+	if !appId? or _.isNaN(parseInt(appId)) or parseInt(appId) <= 0
```
curcuz: validating a null or <=0 appId on the supervisor side is good, but should't this request resulted in a KO from the API?

Page-: I expect the api didn't actually provision the device, and also didn't return a provisioned device, the status code is unfortunately a bit weird because the request technically did succeed within the permissions you have.  It makes more sense if you think of GET devices, which returns all the devices you have access to, even if that turns out to be none and that is correct - for POST devices it does the equivalent, it creates all of the devices you have permission to create even if that is none, however the expectation in this case is that it should be a 4xx error code if it doesn't create all of them

PR #307, opened 2016-10-06
Fix docker utils getImageEnv by correctly parsing the returned array

pcarranzav:


src/docker-utils.coffee
```diff
@@ -384,6 +384,9 @@ do ->
 	exports.getImageEnv = (id) ->
 		docker.getImage(id).inspectAsync()
 		.get('Config').get('Env')
+		.then (env) ->
+			# env is an array of strings that say 'key=value'
+			return _.mapValues(_.keyBy(_.map(env, (el) -> el.split('=')), '0'), '1')
```
Page-:
```coffee
_(env)
.invokeMap('split', '=')
.fromPairs()
.value()
```

pcarranzav: Nice, I was looking at the lodash docs for a way to do this...
Done.

PR #308, opened 2016-10-06
Always pull the base images when building

pcarranzav:


Page-:
It probably makes sense to add `--pull` to the gosuper build as well

pcarranzav:
Well that one uses the official golang so it should be more stable, it felt like overkill to always pull there

pcarranzav:
Though actually scratch that, it does make sense as it should incorporate bugfixes.  Will do.

Page-:
It only pulls if there's changes btw, so if there's no changes it's essentially a noop (other than checking the id matches)

pcarranzav:
True, done

PR #309, opened 2016-10-07
Improvements in proxyvisor

### LINKED ISSUE #290: Ability to change behaviour of the Dependent Device Update hook
Currently the Dependent Device Update Hook is continually polled until the dependant device is updated. This could go on for a long time until the dependant device comes online and is able to be updated.

I propose responding with `200` to disable the polling behaviour and `202 to keep the same behaviour.

### LINKED ISSUE #292: Unable to delete a dependant device
When deleting a dep device from the dashboard it gets removed from the UI but not from the local DB, furthermore it does not call the DELETE device enm hook

pcarranzav:
Closes #290 
Closes #292

* Better parameter handling in PUT /v1/devices/:uuid
* An update hook response of 200 will cause the proxyvisor to stop pinging the hook
* Allow deleting dependent apps and devices
* Implement delete dependent device hook
* Omit some fields when responding with a device object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+* Better parameter handling in PUT /v1/devices/:uuid [Pablo]
+* An update hook response of 200 will cause the proxyvisor to stop pinging the hook [Pablo]
+* Allow deleting dependent apps and devices [Pablo]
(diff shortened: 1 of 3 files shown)


pcarranzav:
Changelog will be added after approval

PR #313, opened 2016-10-10
Avoid installing npm dependencies twice

pcarranzav:
Instead of installing all dependencies, then clearing node_modules and running a production install, we do it the other way round: first install production deps, then install dev deps, build and then clean them up using `npm prune`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+* Avoid installing npm dependencies twice [Pablo]
 * Updated to bluebird 3 [Page]
 * Better parameter handling in PUT /v1/devices/:uuid [Pablo]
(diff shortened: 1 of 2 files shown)


Page-:
So I checked the difference between npm ls of before and after, and before has less deps, basically the changes after this PR include optional dependencies (presumably the `npm install --only=dev` installs optional modules and then prune doesn't remove them)

pcarranzav:
I was wondering why the image is slightly larger. I'll try to avoid installing optionals and check again.

pcarranzav:
@Page- @petrosagg `diff <(docker run registry.resinstaging.io/resin/node-supervisor-amd64:master npm ls) <(docker run registry.resinstaging.io/resin/node-supervisor-amd64:dont-npm-install-twice npm ls)` shows no output now.

PR #315, opened 2016-10-10
Update knex to ~0.12.3

Page-:
Connects to #314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
+* Updated knex to ~0.12.3
+
 # v2.6.1
 
-Do not use parsed JSON to compare dep. device state [Pablo]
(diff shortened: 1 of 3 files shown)


Page-:
It worked when I connected the supervisor to my devenv, but I didn't test everything (eg dependent device stuff) - although looking at the changelog the only breaking change I can see that might affect us is that `undefined` variables are no longer set to null in `insert`/`update` statements.

As for the dependencies, that was just me testing to make sure all the copies of bluebird + lodash are finally deduplicated with this PR (they are :) ). I'll those changes though, as they don't actually matter for the final result due to the `~`/`^` resolving

pcarranzav:
Sounds good, okay to merge imo

PR #317, opened 2016-10-13
Properly set device name, and set undefined arguments to setConfig as null

pcarranzav:
The recent knex update causes that when we try to update a single value as undefined it will cause an sqlite error.
This also surfaced the fact that we're actually always setting the device name with the same value instead of using the value returned from the API.

### DIFF
--- a/src/utils.coffee
+++ b/src/utils.coffee
@@ -150,7 +150,7 @@ exports.getConfig = getConfig = (key) ->
 	.then ([ conf ]) ->
 		return conf?.value
 
-exports.setConfig = (key, value) ->
+exports.setConfig = (key, value = null) ->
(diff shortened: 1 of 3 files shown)


src/utils.coffee
```diff
@@ -151,6 +151,7 @@ exports.getConfig = getConfig = (key) ->
 		return conf?.value
 
 exports.setConfig = (key, value) ->
+	value = null if _.isUndefined(value)
```
Page-: This can be `exports.setConfig = (key, value = null) ->`, it'll set value to `null` if it's either `undefined` or `null`, which is equivalent

pcarranzav: Good point, done @Page-

PR #318, opened 2016-10-13
Do not throw an exception if the LED file can't be opened

pcarranzav:


pcarranzav:
Just realized this was not the problem I was seeing in some logs.

PR #319, opened 2016-10-16
Makefile: removed the deprecated force flag

zozo123:
Since docker 1.10 the -f option of docker tag
is obsolete. In docker 1.12 it is completely
removed.

https://docs.docker.com/engine/deprecated/#/f-flag-on-docker-tag

Signed-off-by: Yossi Eliaz <yossi@resin.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,18 @@ DEPLOY_REGISTRY =
 
 SUPERVISOR_VERSION = master
 
+DOCKER_VERSION:=$(shell docker version --format '{{.Server.Version}}')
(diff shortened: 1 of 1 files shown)


Page-:
add to whitelist

Page-:
I don't think this can be removed yet, as it won't work on many versions of docker

pcarranzav:
As Page says, we can't do this yet, at least until Jenkins has docker > 1.10.

zozo123:
@pcarranzav @Page- Please see the new and hopefully better solution.

zozo123:
Fixed.

PR #320, opened 2016-10-20
AUFS support and dependent app assets improvement

pcarranzav:
* Use appId in dependent app assets tar path, and only create the tar if it doesn't exist
* Support AUFS by upgrading node-docker-delta to 1.0.0 and docker-toolbelt to 1.3.0

Depends on resin-io-modules/docker-toolbelt#4 (should be published as 1.3.0) and resin-io/node-docker-delta#11 (should be published as 1.0.0).


src/proxyvisor.coffee
```diff
@@ -182,17 +177,18 @@ router.put '/v1/devices/:uuid', (req, res) ->
 		console.error("Error on #{req.method} #{url.parse(req.url).pathname}", err, err.stack)
 		res.status(503).send(err?.message or err or 'Unknown error')
 
-tarPath = ({ commit }) ->
-	return '/tmp/' + commit + '.tar'
+tarPath = ({ appId, commit }) ->
+	return '/tmp/' appId + '-' + commit + '.tar'
```
Page-: This is missing a `+`

pcarranzav: Oops, fixed

src/proxyvisor.coffee
```diff
@@ -182,17 +177,18 @@ router.put '/v1/devices/:uuid', (req, res) ->
 		console.error("Error on #{req.method} #{url.parse(req.url).pathname}", err, err.stack)
 		res.status(503).send(err?.message or err or 'Unknown error')
 
-tarPath = ({ commit }) ->
-	return '/tmp/' + commit + '.tar'
+tarPath = ({ appId, commit }) ->
+	return '/tmp/' + appId + '-' + commit + '.tar'
```
petrosagg: This should be `/var/tmp` otherwise you're risking us properly mounting `/tmp` as tmpfs in the future and then storing everything in RAM

pcarranzav: Created #329 for this.

PR #325, opened 2016-10-21
Fix multiple update loops appearing after a long period of updates failing.

Page-:
This was getting annoying when I was developing and had multiple update loops flooding the logs, making it hard to look at a single attempt in isolation

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -445,6 +445,7 @@ waitToKill = (app, timeout) ->
 UPDATE_IDLE = 0
 UPDATE_UPDATING = 1
 UPDATE_REQUIRED = 2
+UPDATE_SCHEDULED = 3
 
 updateStatus =
 	state: UPDATE_IDLE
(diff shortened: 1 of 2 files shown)


src/application.coffee
```diff
@@ -581,12 +582,20 @@ compareForUpdate = (localApps, remoteApps) ->
...
 	return { toBeRemoved, toBeDownloaded, toBeInstalled, toBeUpdated, appsWithUpdatedConfigs, allAppIds }
 
-application.update = update = (force) ->
-	if updateStatus.state isnt UPDATE_IDLE
-		# Mark an update required after the current.
-		updateStatus.forceNext = force
-		updateStatus.state = UPDATE_REQUIRED
-		return
+application.update = update = (force, scheduled = false) ->
+	console.log('hit update', _.omit(updateStatus, 'intervalHandle'))
```
pcarranzav: Debug code?

Page-: Yeah, thought I'd removed it all, but apparently not - fixed

PR #326, opened 2016-10-23
Log whether deltas are being used when downloading an app

### LINKED ISSUE #304: System logs should report when deltas are enabled

pcarranzav:
Closes #304

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -36,6 +36,9 @@ logTypes =
 	downloadApp:
 		eventName: 'Application download'
 		humanName: 'Downloading application'
+	downloadAppDelta:
+		eventName: 'Application delta download'
+		humanName: 'Downloading delta for application'
 	downloadAppSuccess:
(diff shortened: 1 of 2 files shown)


src/application.coffee
```diff
@@ -36,6 +36,9 @@ logTypes =
 	downloadApp:
 		eventName: 'Application download'
 		humanName: 'Downloading application'
+	downloadAppDelta:
+		eventName: 'Application delta download'
```
pcarranzav: I just merged this :/ 
I was thinking if it just says "Application download" we'll know it's a regular download from now on... but I guess we could PR that

Page-: Yeah, the problem is that we only send the event, not the supervisor version, so unless we do a bunch of work external to mixpanel matching up supervisor versions to uuids to events (and if there's a supervisor/host os update track when that was) then we cannot be sure which type of "application download" event we're seeing

PR #336, opened 2016-10-28
Add Promise timeouts to all external requests

### LINKED ISSUE #19: For every blocking operation there should be a well-defined timeout

### LINKED ISSUE #69: Update attempts can hang after a failure
The last thing logged in relation to updating is:
`Scheduling another update attempt due to failure:  2000 [...]`
with the only potential exception being `Event: Update notification {}`, with no attempt to actually perform the update

Front logo [Front conversations](https://app.frontapp.com/open/top_rau)

pcarranzav:
Should solve issues when the request hangs after headers have been sent, which seemingly can happen on flaky connections.
Closes #19
Closes #69

### DIFF
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -44,6 +44,7 @@ findSimilarImage = (repoTag) ->
 		return 'resin/scratch'
 
 DELTA_REQUEST_TIMEOUT = 15 * 60 * 1000
+DELTA_TOTAL_TIMEOUT = 30 * 60 * 1000
 
 getRepoAndTag = (image) ->
 	docker.getRegistryAndName(image)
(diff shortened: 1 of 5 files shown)


src/docker-utils.coffee
```diff
@@ -90,6 +91,7 @@ do ->
 							.on('id', resolve)
 							.on('error', reject)
 					.on 'error', reject
+				.timeout(DELTA_TOTAL_TIMEOUT)
```
Page-: It would be nice if this could be a timeout based on the `res.pipe` not having any data flow for X period (along with a timeout for the data to start flowing in the first place)

pcarranzav: That should be covered by request's timeout (times out if the headers haven't been received in a certain time) and rsync's "data send/receive timeout" (which we have hit more than a few times).

PR #338, opened 2016-10-31
Properly delete dependent apps and their corresponding devices

### LINKED ISSUE #337: Dependent apps are not being deleted

pcarranzav:
Closes #337

### DIFF
--- a/src/proxyvisor.coffee
+++ b/src/proxyvisor.coffee
@@ -239,8 +239,7 @@ exports.fetchAndSetTargetsForDependentApps = (state, fetchFn, apiKey) ->
 			return app.commit? and app.imageId? and !_.some(localApps, imageId: app.imageId)
 		toBeRemoved = _.filter localApps, (app, appId) ->
 			return app.commit? and !_.some(remoteApps, imageId: app.imageId)
-		toBeDeletedFromDB = _.filter localApps, (app, appId) ->
-			return !remoteApps[appId]?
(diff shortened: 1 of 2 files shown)


src/proxyvisor.coffee
```diff
@@ -239,8 +239,7 @@ exports.fetchAndSetTargetsForDependentApps = (state, fetchFn, apiKey) ->
 			return app.commit? and app.imageId? and !_.some(localApps, imageId: app.imageId)
 		toBeRemoved = _.filter localApps, (app, appId) ->
 			return app.commit? and !_.some(remoteApps, imageId: app.imageId)
-		toBeDeletedFromDB = _.filter localApps, (app, appId) ->
-			return !remoteApps[appId]?
+		toBeDeletedFromDB = _.map(_.filter(localApps, (app, appId) -> !remoteApps[appId]?), 'appId')
```
Page-: This can be changed to:
```coffee
_(localApps).reject((app, appId) -> remoteApps[appId]?).map('appId').value()
```
The two main changes are `filter` -> `reject`, meaning we can drop the `!` (not a big deal, but still kinda nice), and to use the chaining method which allows shortcut fusion, meaning we only actually do one loop over the array rather than two (I think the chaining style also reads nicer as well)

pcarranzav: Nice, done.

PR #344, opened 2016-11-01
Provide a better error message when app is not found for purging

### LINKED ISSUE #343: Confusing purging error while moving a device to another app
1. provision a device in app A -> device in running app A
2. move the device to app B
3. purge the app data => 'App not found' UI

In the log we also get:

```
01.11.16 11:29:39 [+0000] Purging /data
01.11.16 11:29:39 [+0000] Error purging /data: AppNotFoundError: App not found
```
It seems related to https://github.com/resin-io/hq/wiki/Scratch-Pad#why-does-data-disappear-when-i-move-a-device-between-applications but not really.

After around 5 minutes of waiting, killing, stopping, etc the purge feature is available while not throwing the above mentioned error.

There are at least two issues here:
1. we have a downtime of around 5 minutes when no code is running on the device
2. the purge is broken during this time and the error thrown is not helpful at all

As I know that this feature needs some rethinking, I would say that for now a better error would be really helpful and avoid user confusions.

pcarranzav:
Closes #343

### DIFF
--- a/src/api.coffee
+++ b/src/api.coffee
@@ -113,12 +113,15 @@ module.exports = (application) ->
 					.finally ->
 						application.start(app)
 		.catch (err) ->
+			status = 503
+			if err instanceof utils.AppNotFoundError
+				errMsg = "App not found: an app needs to be installed for purge to work.
+					If you've recently moved this device from another app,
+					please push an app and wait for it to be installed first."
+				err = new Error(errMsg)
+				status = 400
 			application.logSystemMessage("Error purging /data: #{err}", { appId, error: err }, 'Purge /data error')
-			throw err
-		.catch utils.AppNotFoundError, (e) ->
-			return res.status(400).send(e.message)
-		.catch (err) ->
-			res.status(503).send(err?.message or err or 'Unknown error')
+			res.status(status).send(err?.message or err or 'Unknown error')
 
 	unparsedRouter.post '/v1/tcp-ping', (req, res) ->
 		utils.disableCheck(false)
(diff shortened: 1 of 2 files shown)


src/api.coffee
```diff
@@ -113,12 +113,15 @@ module.exports = (application) ->
 					.finally ->
 						application.start(app)
 		.catch (err) ->
+			status = 503
+			if err instanceof utils.AppNotFoundError
+				errMsg = "App not found: an app needs to be installed for purge to work.
+					If you've recently moved this device from another app,
+					please push an app and wait for it to be installed first."
+				err = new Error(errMsg)
+				status = 400
```
Page-: Should this be 404?

pcarranzav: Yeah it's probably better, but we've been using 400 so far

src/api.coffee
```diff
@@ -113,12 +113,15 @@ module.exports = (application) ->
 					.finally ->
 						application.start(app)
 		.catch (err) ->
+			status = 503
+			if err instanceof utils.AppNotFoundError
+				errMsg = "App not found: an app needs to be installed for purge to work.
+					If you've recently moved this device from another app,
+					please push an app and wait for it to be installed first."
+				err = new Error(errMsg)
```
pcarranzav: Well we'll only hit this error in `utils.getKnexApp` so it's pretty easy to know where it comes from, and it general it relates to the `appId` being invalid. Also this is already merged...

Page-: I know it's merged, I just didn't get a chance to comment before it was :P

Ok, it works ok for now, it's just nice in general to keep stack traces intact for places where it might not be clear (I've had so many headaches trying to chase down where an error actually originates because of silenced stack traces in the past :(  )

src/api.coffee
Page-: Ok

PR #357, opened 2016-12-09
Bump the MAX_BASIS_DIRS limit for rsync.

michal-mazurek:


pcarranzav:

PR #360, opened 2016-12-15
Base image: remove deprecated -f in docker tag

pcarranzav:
We remove a -f flag used in the base image automation script
when tagging the base image as latest.

Docker 1.10 deprecated the use of -f when doing docker tag,
and it was removed in Docker 1.12. The base image is aways built
in machines with docker > 1.10, and this change allows them to be
built on Docker >= 1.12.

Fixes #361 
Signed-off-by: Pablo Carranza Velez <pablo@resin.io>

PR #363, opened 2016-12-15
Build the base image together with the supervisor

### LINKED ISSUE #356: Deltas fail with too many layers because rsync can't take enough link-dests
rsync can take up to 20 --link-dest arguments, and this has to be fixed by patching rsync in supervisor-base to allow more arguments.

### LINKED ISSUE #362: Base image builds are not done in sync with supervisor builds
Currently every PR that changes the base image has to be followed with a PR that bumps the base image version on the supervisor Dockerfile. We could use a hash of the base-image folder to detect changes and build on-demand.

pcarranzav:
This commit changes the automation scripts so that the base image
is built together with the rest of the supervisor. We use a hash of
the contents of the base-image folder as tag for the base image so
that unnecessary rebuilds are avoided.

The build scripts are also modified to always clean up the resulting base image
and the image used to build it. We use docker pull and push to enable caching.

Fixes #362 
Also fixes #360
Also fixes #356 because this will pick up the changes from #357 
Change-Type: patch

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -2,18 +2,27 @@
 set -e
 
 # Jenkins build steps
-ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
+export ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
+BASE_IMAGE_VERSION=$(tar -c --mtime='1970-01-01' --owner=0 --group=0 -f - base-image | md5sum | awk -F " " '{print $1}')
+export BASE_IMAGE_REPO=resin/$ARCH-supervisor-base
+export BASE_IMAGE_TAG=resin/$ARCH-supervisor-base:$BASE_IMAGE_VERSION
+
+# Try to pull the base image according to the contents of the base-image folder, otherwise build it
+docker pull $BASE_IMAGE_TAG || (cd base-image && bash -ex automation/jenkins-build.sh)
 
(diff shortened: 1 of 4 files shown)


automation/jenkins_build.sh
```diff
@@ -2,18 +2,26 @@
 set -e
 
 # Jenkins build steps
-ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
+export ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
+BASE_IMAGE_VERSION=$(tar -cf - base-image | md5sum | awk -F " " '{print $1}')
```
Page-: The hash will change whenever the metadata like mtime changes, which happens all the time as it's set to the time of checkout

automation/jenkins_build.sh
```diff
@@ -2,18 +2,26 @@
 set -e
 
 # Jenkins build steps
-ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
+export ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
+BASE_IMAGE_VERSION=$(tar -c --mtime='1970-01-01' --owner=0 --group=0 -f - base-image | md5sum | awk -F " " '{print $1}')
```
Page-: @petrosagg will be horrified you didn't use `1990-10-10` :P

petrosagg: you also need `--sort=name` so that the files have a reproducible order

PR #364, opened 2016-12-15
Start using versionist for versioning and changelog

pcarranzav:
We use the default versionist configuration.
Changelog has been edited to match the versionist format as much as possible.

Change-Type: patch


README.md
```diff
@@ -164,6 +164,18 @@ make ARCH=amd64 test-integration
 The tests will fail if the supervisor API is down - bear in mind that the supervisor image takes a while to start the actual supervisor program, so you might have to wait a few minutes between running the supervisor and testing it.
 The test expects the supervisor to be already running the application (so that the app is already on the SQLite database), so check the dashboard to see if the app has already downloaded.
 
+## Contributing
+
+If you're interested in contributing, that's awesome!
+
+Here's a few guidelines to make the process easier for everyone involved.
+
+* Every PR *should* have an associated issue, and the PR's opening comment should say "Fixes #issue" or "Closes #issue".
```
lekkas: Question: at some point we used the `connects to` waffle convention right? I guess we've dropped that for the supervisor?

pcarranzav: Yes, but we stopped using it after it started being treated as a module instead of a component, i.e. a supervisor version is set in meta-resin, so we can't treat supervisor versions as "Waiting for staging" or "Waiting for production"; merging a PR fixes an issue immediately and then it's up to whoever uses this module to update to the latest version.

PR #366, opened 2016-12-17
Use find and sort to make the base image hash more deterministic

pcarranzav:
We've been using tar directly on the base image folder, and this makes
the file ordering unpredictable between filesystems, so we use find and
sort so that the files are always tarred in the same order, producing the same
hash in different filesystems.

### DIFF
--- a/automation/jenkins_build.sh
+++ b/automation/jenkins_build.sh
@@ -3,7 +3,10 @@ set -e
 
 # Jenkins build steps
 export ESCAPED_BRANCH_NAME=$(echo $sourceBranch | sed 's/[^a-z0-9A-Z_.-]/-/g')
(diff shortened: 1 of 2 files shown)


pcarranzav:
It's still producing different hashes every time (I'm looking at the builds for different arches, and testing on my PC and a build server, and I can't find *why* the hash always differs). I'm still investigating.

pcarranzav:
Okay, I think I finally managed to make it deterministic across build servers. The order in which I gave tar the options mattered, *obviously*

pcarranzav:
Merging since I confirmed it works okay afaict. If @petrosagg has comments I'll address in another PR (otherwise all the Jenkins builds are taking ages in the meantime)

PR #370, opened 2016-12-20
[Breaking] Do not bind mount kmod if the host is not Resin OS 1.X

### LINKED ISSUE #359: If the OS is not ResinOS 1.X, the supervisor should not bind mount kmod
ResinOS 2.X will not have compressed modules anymore, and we can assume that the default for any other OS is also not to have them, so bind mounting kmod is not necessary (and can lead to a variety of problems).

pcarranzav:
Resin OS 2.X removes the use of compressed modules, which was the initial
motivation for us to bind mount kmod into user containers (as Debian distros
don't include support for compressed modules).

This is a breaking change, but we still keep bind mounting on devices that are
on 1.X to ensure we don't break apps currently relying on the feature.

Implementation note: some functions in device.coffee have been refactored to
extract (DRY) a memoization procedure for Promise-returning functions.
`device.getOSVersion()` now also memoizes its result.

Fixes #359
Change-Type: major

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -210,13 +210,15 @@ fetch = (app, setDeviceUpdateState = true) ->
 			throw err
 
 shouldMountKmod = (image) ->
-	Promise.using docker.imageRootDirMounted(image), (rootDir) ->
-		utils.getOSVersion(rootDir + '/etc/os-release')
-	.then (version) ->
-		return version? and (version.match(/^Debian/i) or version.match(/^Raspbian/i))
-	.catch (err) ->
-		console.error('Error getting app OS release: ', err)
-		return false
+	device.getOSVersion().then (osVersion) ->
+		return false if not /^Resin OS 1./.test(osVersion)
+		Promise.using docker.imageRootDirMounted(image), (rootDir) ->
+			utils.getOSVersion(rootDir + '/etc/os-release')
+		.then (version) ->
+			return version? and /^(Debian|Raspbian)/i.test(version)
+		.catch (err) ->
+			console.error('Error getting app OS release: ', err)
+			return false
 
 application.start = start = (app) ->
 	volumes = utils.defaultVolumes
--- a/src/device.coffee
+++ b/src/device.coffee
@@ -1,5 +1,6 @@
 _ = require 'lodash'
 Promise = require 'bluebird'
+memoizee = require 'memoizee'
 knex = require './db'
 utils = require './utils'
 { resinApi } = require './request'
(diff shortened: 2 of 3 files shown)


src/device.coffee
```diff
@@ -10,36 +10,36 @@ execAsync = Promise.promisify(require('child_process').exec)
...
-							uuid: uuid
-					customOptions:
-						apikey: apiKey
-				)
-				.timeout(config.apiTimeout)
-			.then (devices) ->
-				if devices.length is 0
-					throw new Error('Could not find this device?!')
-				return devices[0].id
+memoizePromise = (func) ->
```
Page-: This can be replaced by memoizee https://www.npmjs.com/package/memoizee#promise-returning-functions
```coffee
memoizePromise = _.partial(memoizee, _, promise: true)
```
which avoids having custom code for this stuff and can also be used for more advanced memoisation if we need it - but mainly it's just to reduce the amount of code we need to maintain here

src/application.coffee
```diff
@@ -210,13 +210,15 @@ fetch = (app, setDeviceUpdateState = true) ->
...
 shouldMountKmod = (image) ->
-	Promise.using docker.imageRootDirMounted(image), (rootDir) ->
-		utils.getOSVersion(rootDir + '/etc/os-release')
-	.then (version) ->
-		return version? and (version.match(/^Debian/i) or version.match(/^Raspbian/i))
-	.catch (err) ->
-		console.error('Error getting app OS release: ', err)
-		return false
+	device.getOSVersion().then (osVersion) ->
+		return false  if not osVersion.match(/^Resin OS 1./)
```
Page-: There's a double space here

src/application.coffee
```diff
@@ -210,13 +210,15 @@ fetch = (app, setDeviceUpdateState = true) ->
...
+		Promise.using docker.imageRootDirMounted(image), (rootDir) ->
+			utils.getOSVersion(rootDir + '/etc/os-release')
+		.then (version) ->
+			return version? and (version.match(/^Debian/i) or version.match(/^Raspbian/i))
```
Page-: It would be nicer to do `/^(Debian|Raspbian)/i.test(version)`:
* the `regex.test()` format as it just returns a boolean directly
* the `/^(Debian|Raspbian)/i` so we only need to one one regex

src/application.coffee
Page-: `/^Resin OS 1./.test(osVersion)` is nicer imo as it only returns a boolean, rather than a matches array

src/device.coffee
pcarranzav: Nice, done

src/application.coffee
pcarranzav: Fixed :S

src/application.coffee
pcarranzav: Done

src/application.coffee
pcarranzav: Done

PR #377, opened 2017-01-06
Add a delay between successive updates of device state to the API

### LINKED ISSUE #375: Requests to update the device state on the API should be throttled
Otherwise we can potentially make *a lot* of requests while downloading an image (updating the progress bar way too fast)

pcarranzav:
We add a 1s delay between requests to the API to apply state changes,
as this will throttle it to a point that it has a reasonable rate while
preventing too many unnecessary requests to the API.

Closes #375

### DIFF
--- a/src/device.coffee
+++ b/src/device.coffee
@@ -167,6 +167,8 @@ exports.getDeviceType = do ->
 				return configFromFile.deviceType
 
 do ->
+	APPLY_STATE_SUCCESS_DELAY = 1000
+	APPLY_STATE_RETRY_DELAY = 5000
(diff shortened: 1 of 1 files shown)


src/device.coffee
```diff
@@ -167,6 +167,8 @@ exports.getDeviceType = do ->
 				return configFromFile.deviceType
 
 do ->
+	APPLY_STATE_SUCCESS_DELAY = 50
```
Page-: Can we increase this to 1s, .05s is probably too small to have any noticeable effect

pcarranzav: @Page-, done

PR #379, opened 2017-01-16
Increase delta timeouts, and provide nicer message when the delta server times out

### LINKED ISSUE #378: The timeout for delta downloads is too small
Some images can take way more than 30 minutes to download... especially under harsh network conditions. We need to define a reasonable value.
Causes an ESOCKETTIMEOUT error when the timeout is hit, instead of continuing normally.

Front logo [Front conversations](https://app.frontapp.com/open/top_zrn)

pcarranzav:
Current delta timeouts are too limiting, so we increase the request timeout to 30 minutes which is big enough that
the server will time out first and we can provide a nice message letting the user know we'll retry; and we increase
the total timeout to 24 hours to account for really big deltas over slower connections (the rsync calls will time out anyways
if something else goes wrong, as they have a 5 minute I/O timeout).

The timeouts are now configurable with the RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT and RESIN_SUPERVISOR_DELTA_TOTAL_TIMEOUT
configuration variables.

Closes #378 
Change-Type: minor

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -196,7 +196,9 @@ fetch = (app, setDeviceUpdateState = true) ->
 
 			if conf['RESIN_SUPERVISOR_DELTA'] == '1'
 				logSystemEvent(logTypes.downloadAppDelta, app)
-				dockerUtils.rsyncImageWithProgress(app.imageId, onProgress)
+				requestTimeout = conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'] ? 30 * 60 * 1000
+				totalTimeout = conf['RESIN_SUPERVISOR_DELTA_TOTAL_TIMEOUT'] ? 24 * 60 * 60 * 1000
+				dockerUtils.rsyncImageWithProgress(app.imageId, { requestTimeout, totalTimeout }, onProgress)
 			else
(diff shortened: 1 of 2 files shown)


pcarranzav:
Happy to discuss the timeout values, but I think the range of connections speeds and image sizes we might encounter make it very hard to find a sound value for total download time.

shaunmulligan:
Is it not possible (probably a bit complicated) to do some kind of exponential back off?

pcarranzav:
You mean increase the timeout when it fails? It is indeed possible, and probably not so complicated (we already do something similar with the update cycle). But not sure if it would actually help? The timeout is supposed to be a last resource when something fails and hangs in a weird state.

Page-:
Is it worth letting these be configurable via config vars?  We could even have them hidden from the ui and just have them available for us to change the timeouts without requiring a code change/update (and could apply to a lot of hardcoded config in the supervisor), cc @petrosagg @alexandrosm @shaunmulligan

pcarranzav:
I did what @Page- suggested (cc @shaunmulligan)

pcarranzav:
Merging with broken tests cause they are failing for unrelated reasons

PR #380, opened 2017-01-16
Update versionist to v2.8.0

pcarranzav:
Also remove versionist.conf.js as it's no longer necessary (versionist now
defaults to its internal versionist.conf) and actually breaks.

Change-Type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,6 @@
   "devDependencies": {
     "coffee-script": "~1.11.0",
     "resin-lint": "^1.3.1",
-    "versionist": "^2.6.2"
+    "versionist": "^2.8.0"
   }
 }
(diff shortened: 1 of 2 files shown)


lurch:
> and actually breaks

Should a new bug be filed against versionist?

pcarranzav:
@lurch I was seeing some odd error where  it seemed like it couldn't find versionist.conf.js, but now it seems like it actually works, so it must have been something odd with my setup. I'll file it if I see it again, but so far it looks like it was my fault.

In any case, we can still safely remove the versionist.conf.js we have here.

pcarranzav:
(I changed the commit message to remove the "actually breaks")

PR #383, opened 2017-01-18
Issue #382: Validate delta timeout variables

### LINKED ISSUE #382: Delta timeout variables should be validated as integers

pcarranzav:
Also take validation functions into a module, and use that in all cases where
we need to check for an integer or string.
Closes #382 

### DIFF
--- a/src/lib/validation.coffee
+++ b/src/lib/validation.coffee
@@ -1,0 +1,17 @@
+exports.checkInt = (s, options = {}) ->
+	# Make sure `s` exists and is not an empty string.
+	if !s
+		return
+	i = parseInt(s, 10)
+	if isNaN(i)
+		return
+	if options.positive && i <= 0
+		return
+	return i
+
+exports.checkString = (s) ->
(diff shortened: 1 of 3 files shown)


brownjohnf:
I like it.

src/lib/validation.coffee
```diff
@@ -0,0 +1,17 @@
+exports.checkInt = (s, options = {}) ->
+	# Make sure `s` exists and is not an empty string.
+	if !s
+		return
+	i = parseInt(s, 10)
+	if isNaN(i)
```
Page-: Using `isFinite` probably makes more sense, to exclude Infinity

src/lib/validation.coffee
```diff
@@ -0,0 +1,17 @@
...
+		return
+	if options.positive && i <= 0
```
Page-: I think 0 would be a valid choice for the delays/timeouts, functionally it'd be indistinguishable from 1ms, but for setting 1ms to work and 0ms to end up as x seconds would be a bit weird

PR #387, opened 2017-01-27
Issue #386: Allow forcing updates when an update was already scheduled

### LINKED ISSUE #386: Using force on /v1/update fails if the update is already scheduled
From https://github.com/resin-io/resin-supervisor/blob/master/src/application.coffee#L604 - calling the update function when the update is already scheduled will just return, instead of really scheduling another update which would be necessary if the call was made with `force = true`.

pcarranzav:
The logic for updateStatus.forceNext is changed so that its value is checked when the scheduled update is run, instead
of when the update is scheduled. And when an update is already scheduled and a new request comes in,
we mark forceNext as true if the new request requires a force update.

Closes #386 
Change-Type: patch

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -604,14 +604,19 @@ application.update = update = (force, scheduled = false) ->
 		when UPDATE_SCHEDULED
 			if scheduled isnt true
 				# There's an update scheduled but it isn't this one, so just stop
+				# but if we have to force an update, do it in the one that is scheduled
+				updateStatus.forceNext or= force
 				return
 		when UPDATE_IDLE
 			# All good, carry on with the update.
 		else
 			# Mark an update required after the current in-progress update.
-			updateStatus.forceNext = force
+			updateStatus.forceNext or= force
 			updateStatus.state = UPDATE_REQUIRED
 			return
+
+	force or= updateStatus.forceNext
+	updateStatus.forceNext = false
 	updateStatus.state = UPDATE_UPDATING
 	bootstrap.done.then ->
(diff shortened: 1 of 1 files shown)


src/application.coffee
```diff
@@ -604,14 +604,19 @@ application.update = update = (force, scheduled = false) ->
 		when UPDATE_SCHEDULED
 			if scheduled isnt true
 				# There's an update scheduled but it isn't this one, so just stop
+				# but if we have to force an update, do it in the one that is scheduled
+				updateStatus.forceNext = updateStatus.forceNext or force
```
Page-: This can be `updateStatus.forceNext or= force`

src/application.coffee
```diff
@@ -604,14 +604,19 @@ application.update = update = (force, scheduled = false) ->
...
 				return
 		when UPDATE_IDLE
 			# All good, carry on with the update.
 		else
 			# Mark an update required after the current in-progress update.
-			updateStatus.forceNext = force
+			updateStatus.forceNext = updateStatus.forceNext or force
```
Page-: This can be `updateStatus.forceNext or= force`

src/application.coffee
```diff
@@ -604,14 +604,19 @@ application.update = update = (force, scheduled = false) ->
...
 			updateStatus.state = UPDATE_REQUIRED
 			return
+	if scheduled is true
```
Page-: Why this check? Do we care if it's scheduled or not for the sake of `forceNext`?

pcarranzav: Hmm actually we don't, true

PR #394, opened 2017-03-06
Issues #23 and #236: Use docker logs to get all logs from the container, including those before supervisor start

### LINKED ISSUE #23: Logs printed immediately after start up are lost due to race condition
This is probably due to the time spent between starting the container and attaching to the output

### LINKED ISSUE #236: Omitting logs on reboot
If I reboot the device (via the api) it omits the first two logs of my app. 

Logs after a restart

```
21.08.16 20:23:58 [+0100] Starting application 'registry.resin.io/microbeastlondon/def087039ac902b6df73f527dca5754d002638a5'
21.08.16 20:24:00 [+0100] Started application 'registry.resin.io/microbeastlondon/def087039ac902b6df73f527dca5754d002638a5'
21.08.16 20:24:00 [+0100] rendering image
21.08.16 20:24:04 [+0100] using "DejaVu Sans Mono-16", pixelsize=16.67 file=/usr/share/fonts/truetype/dejavu/DejaVuSansMono.ttf
```

logs after a reboot 

```
21.08.16 20:25:21 [+0100] Starting application 'registry.resin.io/microbeastlondon/def087039ac902b6df73f527dca5754d002638a5'
21.08.16 20:25:31 [+0100] Started application 'registry.resin.io/microbeastlondon/def087039ac902b6df73f527dca5754d002638a5'
```

pcarranzav:
Closes #23 
Closes #236 

We've been using docker attach, which only gives us the logs since we attach. This change allows getting the
full logs from the beginning.
We also use the timestamps that come with the logs from docker, as they will be more precise and are more relevant now
that we're getting previous logs from history.

Change-Type: patch

(Untested)

### DIFF
--- a/src/lib/logger.coffee
+++ b/src/lib/logger.coffee
@@ -97,12 +97,17 @@ do ->
 			if !attached[app.containerId]
 				dockerPromise.then (docker) ->
 					docker.getContainer(app.containerId)
-					.attachAsync({ stream: true, stdout: true, stderr: true, tty: true })
+					.logsAsync({ follow: true, stdout: true, stderr: true, timestamps: true })
 					.then (stream) ->
 						attached[app.containerId] = true
 						stream.pipe(es.split())
-						.on('data', publish)
-						.on 'error', ->
+						.on 'data', (logLine) ->
+							space = logLine.indexOf(' ')
+							if space > 0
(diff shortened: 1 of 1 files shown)


Page-:
@pcarranzav do you know if the docker logs persist across reboots, because I have a feeling that they do and that could cause a lot of previous logs to come in via this command

PR #395, opened 2017-03-07
Issue #20: Change the update lock to a temporary filesystem

### LINKED ISSUE #20: Ensure we don't kill a locked user's app if the supervisor crashes and restarts
Every time the supervisor starts, it clears all locks by doing unlockAndStart(app).
However, it should check whether the app container is already running and should have _some_ way to identify if the lock belongs to the app (and not to a previously crashed supervisor) - in this case, it should **not** unlock since it means the app is actually locked and in a critical section.

Front logo [Front conversations](https://app.frontapp.com/open/top_vqp)

pcarranzav:
Closes #20 
Connects to https://github.com/resin-io/hq/pull/683

The lock is now located at `/tmp/resin-supervisor/<appId>/` on the host, and `/tmp/resin/`
on the user container. The old lock location is supported only in Resin OS 1.X (and both locks are
taken in that case).

This fixes the race condition when the app is started before the supervisor, and takes a lock that is
cleared on supervisor startup.

Change-Type: major

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -6,6 +6,8 @@ set -o errexit
     mkdir -p /dev/net
 [ -c /dev/net/tun ] ||
     mknod /dev/net/tun c 10 200
+[ -d /mnt/root/tmp/resin-supervisor ] ||
+    mkdir -p /mnt/root/tmp/resin-supervisor
 
 mount -t tmpfs -o size=1m tmpfs /var/run/resin
 
--- a/docs/update-locking.md
+++ b/docs/update-locking.md
@@ -2,13 +2,27 @@
 
 Locking updates means that the Supervisor will not be able to kill your application. This is meant to be used at critical sections of your code where you don't want to be interrupted, or to control that updates are only installed at certain times.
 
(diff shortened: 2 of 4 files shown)


brownjohnf:
LGTM

entry.sh
```diff
@@ -6,6 +6,8 @@ set -o errexit
     mkdir -p /dev/net
 [ -c /dev/net/tun ] ||
     mknod /dev/net/tun c 10 200
+[ -d /mnt/root/tmp/resin-supervisor ] ||
+    mkdir -p /mnt/root/tmp/resin-supervisor
```
brownjohnf: Not necessarily related to this PR, but if you use the -p flag with mkdir it should be safe to run it whether the directory exists or not.

pcarranzav: Yeah, true, I'll fix it in a following PR

pcarranzav:
@brownjohnf yes, that's the idea. The fact that it *did* persist led to the supervisor having to clear it on startup, causing a race condition (#20). Now if the app always take the lock on startup it will be safe.

brownjohnf:
Ah okay. Sounds good then.

BrianAdams:
@pcarranzav Just clarifying your point above:

If we want the system to lock updates by default, we just have to have the container set the lock in `/tmp/resin/`.  On a reboot of the device, the container will start *before* the supervisor starts checking for updates.  Thus as long as the container always sets the lock, it will never be updated by  the supervisor unless something later clears the lock or overrides the lock via the resin dashboard.

Correct?

pcarranzav:
@BrianAdams correct. Just ensure that your container *always* takes the lock immediately on startup and the supervisor will not update it until the lock is released. The supervisor always ensures containers are started and only then polls for updates, so there is no race condition anymore.

PR #398, opened 2017-03-07
Allow setting the supervisor to a "local mode" which stops apps and prevents cleanup

pcarranzav:
Connects to https://github.com/resin-io/hq/issues/364
(we now allow any config variable to take different truthy or falsy values)

Connects to https://github.com/resin-io/hq/pull/688

A RESIN_SUPERVISOR_LOCAL_MODE variable is introduced. When this variable is "1", all apps
are stopped and the update cycle stops executing changes other than deviceConfig changes
and the proxyvisor.

Change-Type: minor

### DIFF
--- a/src/utils.coffee
+++ b/src/utils.coffee
@@ -230,7 +230,7 @@ exports.getKnexApp = (appId, columns) ->
 exports.getKnexApps = (columns) ->
 	knex('app').select(columns)
 
-exports.getOSVersion = (path) ->
+exports.getOSReleaseField = (path, field) ->
 	fs.readFileAsync(path)
 	.then (releaseData) ->
 		lines = releaseData.toString().split('\n')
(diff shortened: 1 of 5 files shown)


src/utils.coffee
```diff
@@ -179,13 +180,14 @@ exports.extendEnvVars = (env, uuid, appId, appName, commit) ->
 
 # Callback function to enable/disable tcp pings
 exports.enableConnectivityCheck = (val) ->
-	bool = val is 'false'
-	disableCheck(bool)
+	enabled = checkTruthy(val) ? true
```
brownjohnf: Won't this set enabled to null or undefined if checkTruthy returns false?

pcarranzav: Luckily no:
```
coffee> foo = false ? true
false
coffee> foo
false
```

brownjohnf: Okay.

PR #403, opened 2017-03-22
Issue #402: Ensure that app configs are non-null and valid json objects

### LINKED ISSUE #402: After a supervisor update, an app's config can be null causing a duplicated container
It can happen if a device updates from a supervisor that didn't have the config column.
This can cause a duplicated container to be started as the initial start fails with:
```
Error starting apps: TypeError: Cannot read property 'RESIN_APP_RESTART_POLICY' of null
```

The fix is to do something like
```
knex('app').update({ config: '{}' }).whereNull('config')
```
when initializing the DB in `db.coffee`.

pcarranzav:
Closes #402 
Connects to resin-io/hq#726

This prevents duplicated containers when updating from older supervisors before the config column
was introduced.

Change-Type: patch

### DIFF
--- a/src/db.coffee
+++ b/src/db.coffee
@@ -52,6 +52,9 @@ knex.init = Promise.all([
 				addColumn('app', 'appId', 'string')
 				addColumn('app', 'config', 'json')
 			]
+			.then ->
+				# When updating from older supervisors, config can be null
+				knex('app').update({ config: '{}' }).whereNull('config')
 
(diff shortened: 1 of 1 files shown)


brownjohnf:
LGTM

afitzek:
LGTM :+1:

pcarranzav:
@brownjohnf @afitzek could you please add an Approval? thanks!

PR #405, opened 2017-03-23
404 authentication for registry and delta

afitzek:
Not ready yet

### DIFF
--- a/src/application.coffee
+++ b/src/application.coffee
@@ -7,7 +7,7 @@ dockerUtils = require './docker-utils'
 Promise = require 'bluebird'
 utils = require './utils'
 logger = require './lib/logger'
-{ cachedResinApi } = require './request'
+{ cachedResinApi, request } = require './request'
 device = require './device'
 lockFile = Promise.promisifyAll(require('lockfile'))
 bootstrap = require './bootstrap'
@@ -214,15 +214,15 @@ fetch = (app, setDeviceUpdateState = true) ->
 
 		Promise.try ->
 			conf = JSON.parse(app.config)
-
-			if conf['RESIN_SUPERVISOR_DELTA'] == '1'
-				logSystemEvent(logTypes.downloadAppDelta, app)
-				requestTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'], positive: true) ? 30 * 60 * 1000
-				totalTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_TOTAL_TIMEOUT'], positive: true) ? 24 * 60 * 60 * 1000
-				dockerUtils.rsyncImageWithProgress(app.imageId, { requestTimeout, totalTimeout }, onProgress)
-			else
-				logSystemEvent(logTypes.downloadApp, app)
-				dockerUtils.fetchImageWithProgress(app.imageId, onProgress)
+			Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
+				if conf['RESIN_SUPERVISOR_DELTA'] == '1'
+					logSystemEvent(logTypes.downloadAppDelta, app)
+					requestTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'], positive: true) ? 30 * 60 * 1000
+					totalTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_TOTAL_TIMEOUT'], positive: true) ? 24 * 60 * 60 * 1000
+					dockerUtils.rsyncImageWithProgress(app.imageId, { requestTimeout, totalTimeout, uuid, apiKey }, onProgress)
+				else
+					logSystemEvent(logTypes.downloadApp, app)
+					dockerUtils.fetchImageWithProgress(app.imageId, onProgress, { uuid, apiKey })
 		.then ->
 			logSystemEvent(logTypes.downloadAppSuccess, app)
 			device.updateState(status: 'Idle', download_progress: null)
--- a/src/docker-utils.coffee
+++ b/src/docker-utils.coffee
@@ -62,7 +62,7 @@ do ->
 		.disposer (release) ->
 			release()
 
-	exports.rsyncImageWithProgress = (imgDest, { requestTimeout, totalTimeout, startFromEmpty = false }, onProgress) ->
+	exports.rsyncImageWithProgress = (imgDest, { requestTimeout, totalTimeout, uuid, apiKey, startFromEmpty = false }, onProgress) ->
 		Promise.using readLockImages(), ->
 			Promise.try ->
 				if startFromEmpty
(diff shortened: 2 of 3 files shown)


afitzek:
We need this PR https://github.com/resin-io/docker-progress/pull/23 and the new docker-progress version in the package.json

src/application.coffee
```diff
@@ -217,9 +218,27 @@ fetch = (app, setDeviceUpdateState = true) ->
 
 			if conf['RESIN_SUPERVISOR_DELTA'] == '1'
 				logSystemEvent(logTypes.downloadAppDelta, app)
-				requestTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'], positive: true) ? 30 * 60 * 1000
-				totalTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_TOTAL_TIMEOUT'], positive: true) ? 24 * 60 * 60 * 1000
-				dockerUtils.rsyncImageWithProgress(app.imageId, { requestTimeout, totalTimeout }, onProgress)
+
+				Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
+					appImageInfos = app.imageId.split('/')
```
pcarranzav: I think this should all be inside `dockerUtils.rsyncImageWithProgress`, changing it to receive an `authconfig` object in the options. Plus you can use `docker.getRegistryAndName(imageId)` in there to split the components of the imageId.

src/docker-utils.coffee
```diff
@@ -97,11 +107,17 @@ do ->
...
 
 	exports.fetchImageWithProgress = (image, onProgress) ->
 		Promise.using readLockImages(), ->
-			dockerProgress.pull(image, onProgress)
+			Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
+				dockerOptions =
+					authconfig:
+						username: 'd_' + uuid,
+						password: apiKey,
+						serveraddress: config.registryEndpoint
```
pcarranzav: I'd rather we avoid using `config.registryEndpoint` as we're not using it elsewhere and it could therefore be removed. Preferably we can get the registry from the `image` as you're doing in the deltas part.

pcarranzav:

src/application.coffee
```diff
@@ -217,9 +218,27 @@ fetch = (app, setDeviceUpdateState = true) ->
...
+					registryService = appImageInfos[0]
+					repositoryPath = appImageInfos.slice(1).join('/')
+					tokenEndpoint = "#{config.apiEndpoint}/registry/token" #TODO read from state endpoint
```
dfunckt: Heads up: the endpoint has been changed to `/auth/v1/token` but in any case let's please sync on this before merging this PR

afitzek: Got it. :)

src/docker-utils.coffee
```diff
@@ -218,7 +240,15 @@ do ->
...
-				dockerProgress.pull(repoTag, onProgress)
+				Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
+					docker.getRegistryAndName(image)
+					.then ({ registry, imageName, tagName }) ->
+						dockerOptions =
+							authconfig:
+								username: 'd_' + uuid,
+								password: apiKey,
+								serveraddress: config.registryEndpoint
+						dockerProgress.pull(repoTag, onProgress, registry)
```
pcarranzav: I think this should be `dockerProgress.pull(repoTag, onProgress, { authconfig })`?

src/docker-utils.coffee
```diff
@@ -218,7 +240,15 @@ do ->
...
 			.then ->
-				dockerProgress.pull(repoTag, onProgress)
+				Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
+					docker.getRegistryAndName(image)
+					.then ({ registry, imageName, tagName }) ->
+						dockerOptions =
+							authconfig:
+								username: 'd_' + uuid,
+								password: apiKey,
+								serveraddress: config.registryEndpoint
```
pcarranzav: This should be `serveraddress: registry` I think

src/application.coffee
```diff
@@ -17,6 +17,7 @@ JSONStream = require 'JSONStream'
 proxyvisor = require './proxyvisor'
 { checkInt, checkTruthy } = require './lib/validation'
 deviceConfig = require './device-config'
+config = require './config'
```
pcarranzav: This one is already required in line 5

src/docker-utils.coffee
```diff
@@ -62,46 +62,68 @@ do ->
...
 					docker.getImage(id).tagAsync({ repo, tag, force: true })
 			.catch dockerDelta.OutOfSyncError, (err) ->
 				console.log('Falling back to delta-from-empty')
-				exports.rsyncImageWithProgress(imgDest, { requestTimeout, totalTimeout, startFromEmpty: true }, onProgress)
+				exports.rsyncImageWithProgress(imgDest, { requestTimeout, totalTimeout, authOpts, startFromEmpty: true }, onProgress)
 
 	exports.fetchImageWithProgress = (image, onProgress) ->
 		Promise.using readLockImages(), ->
-			dockerProgress.pull(image, onProgress)
+			Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
```
pcarranzav: I think this should be done in `application.fetch` (you already do it in the deltas case, you could just do it always before checking for the RESIN_SUPERVISOR_DELTA config). This makes it more consistent: you always get the uuid and apikey there, and pass the credentials to either fetchImageWithProgress or rsyncImageWithProgress.

src/docker-utils.coffee
```diff
@@ -218,7 +240,15 @@ do ->
 			.then ([ img ]) ->
 				knex('image').insert({ repoTag }) if !img?
 			.then ->
-				dockerProgress.pull(repoTag, onProgress)
+				Promise.join utils.getConfig('apiKey'), utils.getConfig('uuid'), (apiKey, uuid) ->
```
pcarranzav: This is part of the "sideload" API, usually used for external images, and it will be changed soon. I think there's no need to add credentials to this.

afitzek:
Got it tested, with the local devenv.

src/docker-utils.coffee
```diff
@@ -62,46 +62,78 @@ do ->
...
-					.on 'error', reject
+					docker.getRegistryAndName(imgDest)
+					.then ({ registry, imageName, tagName }) ->
+						tokenEndpoint = "#{config.apiEndpoint}/auth/v1/token"
+						authOpts =
+							auth:
+								user: 'd_' + uuid
+								pass: apiKey
+								sendImmediately: true
+						request.get "#{tokenEndpoint}?service=#{registry}&scope=repository:#{imageName}:pull", authOpts, (err, res, body) ->
```
pcarranzav: Will this also work to authenticate the delta server to pull `imgSrc`? As it will have to be able to pull both the source and destination to compute the delta

afitzek: As long as imgSrc and imgDst are in the same repository it will, so it will allow the delta server to pull all different tags from one repository. Will the delta server calculate diffs of images from different repositories?

afitzek: just realized that we can hand over multiple repositories to the token endpoint. I'll fix that and use imgSrc and imgDst.

pcarranzav: Good, yeah, cause there's usually a different repo per commit

src/docker-utils.coffee
```diff
@@ -218,7 +250,7 @@ do ->
 			.then ([ img ]) ->
 				knex('image').insert({ repoTag }) if !img?
 			.then ->
-				dockerProgress.pull(repoTag, onProgress)
+				dockerProgress.pull(repoTag, onProgress, dockerOptions)
```
pcarranzav: `dockerOptions` here seems to be a leftover

afitzek: You are right. Going to remove that as well.

src/docker-utils.coffee
```diff
@@ -62,46 +62,86 @@ do ->
...
-							.on('error', reject)
-					.on 'error', reject
+					docker.getRegistryAndName(imgDest)
+					.then ({ registry, imageName, tagName }) ->
+
+						registryDst = registry
+						imageNameDst = imageName
+						tagNameDst = tagName
+
+						docker.getRegistryAndName(imgSrc)
```
pcarranzav: This could be done inside a `Promise.join` together with the other `docker.getRegistryAndName` above (but it's no big deal so we can address it in a later PR).

src/docker-utils.coffee
```diff
@@ -62,46 +62,86 @@ do ->
...
+						.then ({ registry, imageName, tagName }) ->
+							tokenEndpoint = "#{config.apiEndpoint}/auth/v1/token"
+							authOpts =
+								auth:
+									user: 'd_' + uuid
+									pass: apiKey
+									sendImmediately: true
+							url = "#{tokenEndpoint}?service=#{registryDst}&scope=repository:#{imageNameDst}:pull&scope=repository:#{imageName}:pull"
+							request.get url, authOpts, (err, res, body) ->
```
pcarranzav: Probably better to use `request.getAsync` and use a Promise instead of a callback

PR #406, opened 2017-03-23
Issue #230: Add a custom User-Agent header to all requests

### LINKED ISSUE #230: Custom user-agent on supervisor

pcarranzav:
Header is in the format Supervisor/X.Y.Z (Linux; Resin OS v2.A.B.revC; Dev) - omitting any fields
that are not available depending on the OS.

Closes #230 
Change-Type: minor

(Untested)

### DIFF
--- a/src/lib/os-release.coffee
+++ b/src/lib/os-release.coffee
@@ -1,0 +1,32 @@
+Promise = require 'bluebird'
+fs = require 'fs'
+_ = require 'lodash'
+
+exports.getOSReleaseField = (path, field) ->
+	try
+		releaseData = fs.readFileSync(path)
+		lines = releaseData.toString().split('\n')
+		releaseItems = {}
+		for line in lines
+			[ key, val ] = line.split('=')
+			releaseItems[_.trim(key)] = _.trim(val)
+		# Remove enclosing quotes: http://stackoverflow.com/a/19156197/2549019
+		return releaseItems[field].replace(/^"(.+(?="$))"$/, '$1')
+	catch err
+		console.log('Could not get OS release field: ', err, err.stack)
+		return undefined
+
+
+exports.getOSVersionSync = (path) ->
+	exports.getOSReleaseField(path, 'PRETTY_NAME')
+
+exports.getOSVersion = (path) ->
+	Promise.try ->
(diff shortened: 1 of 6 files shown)


src/lib/os-release.coffee
```diff
@@ -0,0 +1,32 @@
...
+exports.getOSReleaseField = (path, field) ->
+	try
+		releaseData = fs.readFileSync(path)
+		lines = releaseData.toString().split('\n')
+		releaseItems = {}
+		for line in lines
+			[ key, val ] = line.split('=')
+			releaseItems[_.trim(key)] = _.trim(val)
+		# Remove enclosing quotes: http://stackoverflow.com/a/19156197/2549019
+		return releaseItems[field].replace(/^"(.+(?="$))"$/, '$1')
```
dfunckt: I think the regex can be `/^"([^"]+)"$/`

src/request.coffee
```diff
@@ -3,10 +3,24 @@ PlatformAPI = require 'pinejs-client'
 Promise = require 'bluebird'
 request = require 'request'
 url = require 'url'
+osRelease = require './lib/os-release'
+
+osVersion = osRelease.getOSVersionSync(config.hostOsVersionPath)
+osVariant = osRelease.getOSVariantSync(config.hostOsVersionPath)
+supervisorVersion = require('./lib/supervisor-version').replace('v', '')
```
dfunckt: I don't know what values `process.env.SUPERVISOR_TAG_EXTRA` can have, but if it contains `v`s they'll be removed. Maybe use this regex instead `/^v/`?

pcarranzav: Oh, true, no need to replace at all

src/lib/os-release.coffee
pcarranzav: This on ealready does that (check the stackverflow link), and has been tested in production, so I'd say let's stick with it

src/request.coffee
pcarranzav: Fixed @dfunckt

src/lib/os-release.coffee
dfunckt: Agreed.

PR #421, opened 2017-04-04
Issue #420: Avoid supervisor crash without connection by properly memoizing promise-returning functions

### LINKED ISSUE #420: Supervisor crashes when it starts up without connection
In [`device.getID`](https://github.com/resin-io/resin-supervisor/blob/master/src/device.coffee#L19), memoization makes the function throw synchronously if the memoized function returns a rejected promise. 

This causes the `getID` function to throw a fatal error when the supervisor starts up without an internet connection.

It seems that changing `memoizePromise` to use `promise: 'then'` should solve the issue.

pcarranzav:
Closes #420 

device.getID caused a fatal error when connection was down, as the memoization with `promise: true` throws
synchronously. Changing memoizee to use `promise: 'then'` makes the memoization work as expected.

Change-Type: patch

### DIFF
--- a/src/device.coffee
+++ b/src/device.coffee
@@ -13,7 +13,8 @@ bootstrap = require './bootstrap'
 { checkTruthy } = require './lib/validation'
 osRelease = require './lib/os-release'
 
-memoizePromise = _.partial(memoizee, _, promise: true)
+# If we don't use promise: 'then', exceptions will crash the program
+memoizePromise = _.partial(memoizee, _, promise: 'then')
 
 
 exports.getID = memoizePromise ->


src/device.coffee
```diff
@@ -13,7 +13,7 @@ bootstrap = require './bootstrap'
 { checkTruthy } = require './lib/validation'
 osRelease = require './lib/os-release'
 
-memoizePromise = _.partial(memoizee, _, promise: true)
+memoizePromise = _.partial(memoizee, _, promise: 'then')
```
Page-: Can you add a comment explaining why this is needed

pcarranzav: @Page-, done

PR #428, opened 2017-04-25
Add aufs-tools to the dind supervisor for use when running on aufs

Page-:
Change-Type: patch


pcarranzav:
What's the purpose of this @Page-?

Page-:
docker tries to use aufs_link (or some similarly named program) when running on aufs and you need to install the aufs-tools package in order to have it available

PR #431, opened 2017-04-27
Only delete the provisioning key if the supervisor is running on an OS that supports using the deviceApiKey

pcarranzav:
This avoids problems when updating the supervisor on an older OS, where the VPN and other
host services still require config.json to have an apiKey field to authenticate.

Change-Type: patch

### DIFF
--- a/src/bootstrap.coffee
+++ b/src/bootstrap.coffee
@@ -10,6 +10,10 @@ appsPath  = '/boot/apps.json'
 _ = require 'lodash'
 deviceConfig = require './device-config'
 TypedError = require 'typed-error'
+osRelease = require './lib/os-release'
+semver = require 'semver'
+semverRegex = require('semver-regex')()
+
 userConfig = {}
 
 DuplicateUuidError = message: '"uuid" must be unique.'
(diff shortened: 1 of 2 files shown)


src/bootstrap.coffee
```diff
@@ -150,15 +154,20 @@ bootstrapper.done = new Promise (resolve) ->
...
-				knex('config').update(value: userConfig.deviceApiKey).where(key: 'apiKey')
-				.then ->
-					fs.writeFileAsync(configPath, JSON.stringify(userConfig))
+			Promise.join(
+				osRelease.getOSVersion(config.hostOSVersionPath)
+				exchangeKey()
+				(osVersion) ->
+					# Only delete the provisioning key if we're on a Resin OS version that supports using the deviceApiKey (2.0.2 and above)
+					# Otherwise VPN and other host services that use an API key will break.
+					hasDeviceApiKeySupport = /^Resin OS /.test(osVersion) and semver.gte(semverRegex.test(osVersion), '2.0.2')
```
Page-: I think we would want to delete the key for custom OS builds as well, since they should be based on an up to date enough meta-resin to work (and we don't want to have the case that custom os builds keep the provisioning key around forever)

pcarranzav: Yeah, I guess we can do that - after all if someone's using a custom build it's their responsibility to keep it updated / check for supervisor compatibility.  I'll change this.

PR #435, opened 2017-05-10
API improvements

josephroberts:


pcarranzav:

josephroberts:
@curcuz @pcarranzav @alexandrosm these are the changes I would like to make to the API.

The main points are:
* Added the notion of a local device ID so that the ENM does not need to store this mapping @Page- can we add this to the backend?
* Added an endpoint to get all devices that belong to an application
* The provision endpoint now takes the local device ID
* The info update endpoint now takes download progress
* Added an identify endpoint
* Renamed the restart endpoint to reboot
* Added a not implemented response to optional endpoints

Issue #444, opened 2017-06-05
Dependent device DB update


josephroberts:
Connects with https://github.com/resin-io/hq/pull/728

PR #445, opened 2017-06-05
Update dependent device DB

josephroberts:
Change-type: patch

Connects with #444


alexandrosm:
@josephroberts, I'm assuming you're updating the model, not the DB.

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

PR #446, opened 2017-06-05
Add support for aarch64

pcarranzav:
Change-Type: minor


pcarranzav:
@lurch apparently armv7 programs *may or may not* work on armv8 processors - I guess RPi3 may have better compatibility than other boards. I may be wrong on this, but we've been seeing problems on an arm64 board that *seem* to come from this incompatibility.
In any case this version will allow us to better make use of armv8 features and the 64-bit arch.
See: https://community.arm.com/processors/f/discussions/4798/armv8-backwards-compatibility-with-armv7

pcarranzav:
(we were using the armv7 supervisor on aarch64 rpi3)

lurch:
> (we were using the armv7 supervisor on aarch64 rpi3)

Ahh, that explains it! Thanks Pablo :)

agherzan:
For reference: https://github.com/resin-os/meta-resin/blob/500894f7179b733f7c771a5ed8deb221c9caab27/meta-resin-common/recipes-containers/docker-disk/docker-resin-supervisor-disk.bb#L8

pcarranzav:
(This depends on #448)

pcarranzav:
For this to work (or at least build), I had to update the Go dependencies (especially https://github.com/shirou/gopsutil which was broken on arm64 at the version we had). I switched to using Glide instead of Godep as Godep wasn't working so well and Glide gives a more npm-like management. We're now using the `vendor/` folder as is appropriate in newer Go versions.
Updated to Go 1.7 because some deps use the `context` library which was only added to the standard library in that version.

pcarranzav:
I recommend doing a commit-per-commit review, as the one where we switch to Glide changes an impossible amount of files.

gosuper/Dockerfile
```diff
@@ -6,12 +6,11 @@ RUN apt-get update \
 		build-essential \
 		curl \
 		rsync \
-		upx-ucl \
 	&& rm -rf /var/lib/apt/lists/
 
-ENV GOLANG_VERSION 1.6.3
+ENV GOLANG_VERSION 1.7.6
```
Page-: Does it make sense to jump straight to 1.8?

gosuper/Dockerfile
```diff
@@ -25,11 +24,24 @@ RUN mkdir /usr/src/go \
 	&& export GOROOT_BOOTSTRAP=/usr/local/go-bootstrap \
 	&& cp -r /usr/local/go /usr/local/go-bootstrap \
 	&& cd /usr/local/go/src \
-	&& patch -p2 -i /go-${GOLANG_VERSION}-patches/0001-Revert-runtime-check-and-fail-early-with-a-message-i.patch \
+	&& patch -p2 -i /go-${GOLANG_VERSION}-patches/0001-dont-fail-when-no-mmx.patch \
 	&& patch -p2 -i /go-${GOLANG_VERSION}-patches/0002-implement-atomic-quadword-ops-with-FILD-FISTP.patch \
 	&& ./make.bash \
 	&& rm -rf /usr/local/go-bootstrap
 
+ENV UPX_VERSION 3.94
```
Page-: What changes are we looking at between the packaged version and 3.94?

pcarranzav:
@Page- it is indeed a separate PR https://github.com/resin-io/resin-supervisor/pull/448 (and this depends on that one)

gosuper/Dockerfile
pcarranzav: Basically add support for arm64. From https://upx.github.io/upx-news.txt
```
Changes in 3.94 (12 May 2017):
  * Add support for arm64-linux (aka "aarch64").
  * Add support for --lzma compression on 64-bit PowerPC (Thierry Fauck).
  * For Mach, "upx -d" will unpack a prefix of the file (and warn).
  * Various improvements to the ELF formats.
  * bug fixes - see https://github.com/upx/upx/milestone/3
```

gosuper/Dockerfile
pcarranzav: Good point, done

pcarranzav:
@resin-jenkins test this please

lurch:
@hedss I wonder if it'd be worth there being a JenkinsTestProcBot that would start/restart a jenkins test if an appropriate label was added (a bit like the `ready-to-merge` VersionBot label), instead of only listening on comments, in order to cut down on the number of 'noisy' comments on a PR thread?
And yes, I realise I've just added more noise to this PR thread myself ;-)  (sorry Pablo)

hedss:
@lurch The only way to kick it like that would be to get the bot to talk directly to Jenkins and know which build this is (though this could be done in the bot config for the repo). However, adding a label, you're still going to get '<blah> added label', '<blah> removed label', etc. If there's a strong opinion though, I'm more than happy for an issue to be raised on the ProcBots repo (and always am, the more ideas, the better!). :)

lurch:
But 'added label' and 'removed label' physically take up less screen-space than a comment, and unlike adding comments to a PR they don't send notification emails. (AFAIK?)
@pcarranzav gave my idea the thumbs-up, so I'll add a ProcBots suggestion issue... :)

Page-:
I usually trigger a new one by clicked 'details' on the build and then hitting 'retry' - jenkins automatically updates the build status in that case

pcarranzav:
I usually retry on the Jenkins UI too, but sometimes it's easier to do it from GitHub when looking at the PR - having a label that does that would definitely make it cleaner imo

PR #448, opened 2017-06-07
Remove the undocumented and unused sideload and compose APIs

pcarranzav:
This allows us to also remove a few npm dependencies and the docker compose binary.

Change-Type: major


Dockerfile.build.template
```diff
@@ -11,32 +11,11 @@ RUN apt-get update \
...
-
-ENV DOCKER_COMPOSE_SHA256_amd64 37df85ee18bf0e2a8d71cbfb8198b1c06cc388f19118be7bdfc4d6db112af834
-ENV DOCKER_COMPOSE_SHA256_i386 b926fd9a2a9d89358f1353867706f94558a62caaf3aa72bf10bcbbe31e1a44f0
-ENV DOCKER_COMPOSE_SHA256_rpi 3f0b8c69c66a2daa5fbb0c127cb76ca95d7125827a9c43dd3c36f9bc2ed6e0e5
-ENV DOCKER_COMPOSE_SHA256_armv7hf 3f0b8c69c66a2daa5fbb0c127cb76ca95d7125827a9c43dd3c36f9bc2ed6e0e5
-ENV DOCKER_COMPOSE_SHA256_armel a1025fed97536e2698798ea277a014ec5e1eae816a8cf3155ecbe9679e3e7bac
-
-RUN set -x \
-	&& mkdir -p rootfs-overlay/usr/bin/ \
-	&& ln -s /lib rootfs-overlay/lib64 \
```
pcarranzav: @petrosagg do you remember what this was here for? Is this for docker compose or do we need the lib64 in the runtime image for something else?

src/db.coffee
```diff
@@ -56,19 +56,6 @@ knex.init = Promise.all([
 				# When updating from older supervisors, config can be null
 				knex('app').update({ config: '{}' }).whereNull('config')
 
-	knex.schema.hasTable('image')
```
Page-: I kinda wonder if we should add drop statements instead

Dockerfile.build.template
Page-: I have a feeling it's needed

petrosagg: I can't remember so if we remove it we'll have to retest. This symlink is only relevant (if at all) to amd64 builds

pcarranzav: Looks like it was indeed necessary for gosuper to work

lekkas:
afaik only https://github.com/resin-io-playground/logspout-scripts ever used this API, I've marked the project there as deprecated.

PR #449, opened 2017-06-07
Make sure the supervisor doesn't continue when rebooting

pcarranzav:
Otherwise the execution may continue for a bit, and for instance start a download that will
be interrupted shortly after.

### DIFF
--- a/src/device.coffee
+++ b/src/device.coffee
@@ -1,7 +1,6 @@
 _ = require 'lodash'
 Promise = require 'bluebird'
 memoizee = require 'memoizee'
-knex = require './db'
 utils = require './utils'
 { resinApi } = require './request'
 device = exports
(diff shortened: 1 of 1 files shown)


pcarranzav:
Ah hadn't thought about that, busybox will restart it... I guess we should just hang forever?

src/device.coffee
```diff
@@ -40,8 +39,16 @@ exports.getID = memoizePromise ->
 			throw new Error('Could not find this device?!')
 		return devices[0].id
 
+waitForReboot = ->
+	console.log('Waiting for reboot...')
+	Promise.delay(3000).then(waitForReboot)
```
Page-: We've had cases where a device tries to reboot and then never does, meaning this will just spam our device logs like crazy - it also doesn't stop the supervisor doing other stuff (which I'm guessing was the aim?)

pcarranzav: You're right, I'll close this while I think of a better solution (otherwise it'll get stale quickly)

PR #453, opened 2017-06-14
When apiEndpoint is not defined, work in offline mode (plus fix mixpanel init)

pcarranzav:
The logic to disable mixpanel initialization in offline mode was inverted :S causing mixpanel to *only* be initialized when in offline mode. This fixes it.

The supervisor uses an `API_ENDPOINT` environment variable to define what API to register to. Up to now this has been defaulted to `https://api.resin.io`.
(In Resin OS devices this environment variable ultimately comes from config.json).
This commit changes the behavior so that an empty value of that environment variable causes the supervisor to work in "offline mode", i.e. not connected to a remote server.
Basically only preloaded apps and the supervisor API work in this mode.

The config.json `supervisorOfflineMode` field still works for backwards compatibility, but we'll treat it as deprecated and it should be removed eventually.


src/utils.coffee
```diff
@@ -18,7 +18,7 @@ device = require './device'
 exports.supervisorVersion = require('./lib/supervisor-version')
 
 configJson = require('/boot/config.json')
-if Boolean(configJson.supervisorOfflineMode)
+if Boolean(config.apiEndpoint) and !Boolean(configJson.supervisorOfflineMode)
```
Page-: Would it make sense to move this `and` into the config.coffee file so we don't have to do the double check in every case and can instead use `if config.offlineMode`?

pcarranzav: The thing is the config.coffee file doesn't currently use config.json and I don't want to mix it up in this PR as it's gonna be messy. I'm working on a better fix (consolidate *all* configuration into a single module) but I want to get this in before (since we need it rather urgently).

Page-: Ok, because this is just asking for someone to mess up the check in the future, it really would be nice to have consolidation so the complexity can be kept in one place rather than spread around like this

PR #464, opened 2017-07-01
Use webpack to join all modules

pcarranzav:
Change-Type: patch


Dockerfile.build.template
```diff
@@ -19,20 +19,21 @@ RUN apt-get update \
 RUN mkdir -p rootfs-overlay && \
 	ln -s /lib rootfs-overlay/lib64
 
-COPY package.json /usr/src/app/
 
-# Install only the production modules
-RUN JOBS=MAX npm install --production --no-optional --unsafe-perm \
-	&& npm dedupe
 
+# Install only the production modules that have
```
petrosagg: Looks like this sentence is missing `native dependencies`

remove-hashbang-loader.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function (source) {
```
petrosagg: What is this for? Which of our files include a hashbang?

pcarranzav: Some of the dependencies (e.g. [JSONStream](https://github.com/dominictarr/JSONStream/blob/master/index.js)) are hybrid executable-library and have a `#! /usr/bin/env node` at the beginning of the file.

Dockerfile.build.template
pcarranzav: Good point, fixed (but with `C extensions`)

package.json
```diff
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "start": "./entry.sh",
-    "build": "coffee -c src",
+    "build": "coffee -c src && webpack -p",
```
Page-: `-p`? Can you use a long arg if there is one, as I have no idea off-hand what `-p` is

src/utils.coffee
```diff
@@ -17,7 +17,7 @@ device = require './device'
 
 exports.supervisorVersion = require('./lib/supervisor-version')
 
-configJson = require('/boot/config.json')
```
Page-: I think you can add this to the `externals` list instead

webpack.config.js
```diff
@@ -0,0 +1,29 @@
...
+	module: {
+		loaders: [{
+			test: /\.js$/,
+			use: require.resolve('./remove-hashbang-loader')
+		}]
+	},
+	externals: {
+		'mkfifo': 'commonjs mkfifo',
+		'sqlite3': 'commonjs sqlite3',
+		'mysql2': 'null',
```
Page-: What does the `'null'` do? I'm assuming it's special (I've come across null-loader in webpack 1)

package.json
Page-: Can we use `coffee-loader` in webpack to do the coffee compilation please, rather than having two separate build steps

Dockerfile.build.template
```diff
@@ -19,20 +19,21 @@ RUN apt-get update \
...
 	ln -s /lib rootfs-overlay/lib64
 
-COPY package.json /usr/src/app/
 
-# Install only the production modules
-RUN JOBS=MAX npm install --production --no-optional --unsafe-perm \
-	&& npm dedupe
 
+# Install only the production modules that have C extensions
+RUN JOBS=MAX npm install --production --no-optional --unsafe-perm sqlite3 mkfifo \
```
Page-: Should we change our dependencies to have all the modules that get built in as `devDependencies` and then the remaining ones we still have to install separately be the only ones in `dependencies`, that way we can avoid having a special dependencies list here

package.json
pcarranzav: Done

pcarranzav: Done

src/utils.coffee
pcarranzav: I'd rather change this to reading the file and parsing it - require feels  more like a hack here. In any case this will all change in #465

webpack.config.js
pcarranzav: The idea was to load this modules as `null` - but now I've changed to using null-loader (didn't know it existed, so thanks!)

Dockerfile.build.template
pcarranzav: Really good idea, done

src/utils.coffee
Page-: Ok

package.json
```diff
@@ -9,16 +9,25 @@
   },
   "scripts": {
     "start": "./entry.sh",
-    "build": "coffee -c src",
+    "build": "coffee -c src && webpack --optimize-minimize --define process.env.NODE_ENV=\"production\"",
```
Page-: Can't these options be put in `webpack.config.js` so we have all the webpack config in one place?

pcarranzav:
@Page- the uncompressed amd64 image goes from 54.2MB to 40.5MB, so we get a nice ~13MB reduction.

webpack.config.js
Page-: It still shows as `'null'`?

package.json
Page-: You should be able to remove the `coffee -c` now

webpack.config.js
pcarranzav: In the end this didn't work (webpack still tried and couldn't resolve the modules) - so I switched to a function that just usess `commonjs <modulename>` - this will cause the calls to require these modules to remain unchanged which is what we want after all.

package.json
pcarranzav: They can, but what the optimize-minimize option does is rather complex (they [require two plugins with some arguments](https://github.com/webpack/webpack/blob/57825153e531a079eb0dbd2528e47f4cbafcc7f2/bin/convert-argv.js#L443) so I'd rather leave this like this as it looks way simpler.

pcarranzav: God, webpack is hard to configure. I had to add the `extensions` field to webpack.config.js to make it work.

Page-: Ok

Page-: What about the `NODE_ENV` definition?

pcarranzav: Done for the NODE_ENV def

Dockerfile.build.template
```diff
@@ -21,18 +21,19 @@ RUN mkdir -p rootfs-overlay && \
 
 COPY package.json /usr/src/app/
 
-# Install only the production modules
+# Install only the production modules that have C extensions
 RUN JOBS=MAX npm install --production --no-optional --unsafe-perm \
-	&& npm dedupe
+	&& npm dedupe && cp -R node_modules node_modules_prod
```
Page-: It makes more sense imo to move the `cp -R` to the `RUN npm install` below, rather than persisting the copy in this layer

webpack.config.js
```diff
@@ -0,0 +1,69 @@
...
+]
+
+// We iterate over the node modules and mark all optional dependencies as external
+var dirs = fs.readdirSync('./node_modules')
+for (let dir of dirs){
+	try {
+		let packageJson = JSON.parse(fs.readFileSync('./node_modules/' + dir + '/package.json'))
+		if (packageJson.optionalDependencies != null){
+			for (let key in packageJson.optionalDependencies) {
+				externalModules.push(new RegExp('^' + key + '(/.*)?$'))
```
Page-: You might want to use `_.escapeRegExp(key)` just to make sure it doesn't have any special regex characters in it

webpack.config.js
```diff
@@ -0,0 +1,69 @@
...
+	'mysql',
+	'strong-oracle',
+	'oracle',
+	'oracledb',
+	'pg-query-stream'
+]
+
+// We iterate over the node modules and mark all optional dependencies as external
+var dirs = fs.readdirSync('./node_modules')
+for (let dir of dirs){
```
Page-: Missing a space before the `{`

webpack.config.js
```diff
@@ -0,0 +1,69 @@
...
+	'mariasql',
+	'mssql',
+	'mysql',
+	'strong-oracle',
+	'oracle',
+	'oracledb',
+	'pg-query-stream'
+]
+
+// We iterate over the node modules and mark all optional dependencies as external
```
Page-: We only iterate over the top level node modules, not any nested or namespaced modules - it should probably handle those as well

Page-: This doesn't handle the cases where dependency `a` marks `c` as optional but `b` marks it as required

webpack.config.js
Page-: You could build up an array of optional modules and then turn that into one regex, eg
```
optionalModules = optionalModules.concat(_.keys(packageJson.optionalDependencies))
...
externalModules.push(new RegExp('^' + optionalModules.map(_.escapeRegExp).join('|') + '(/.*)?$'))

File: webpack.config.js [hunk 12]
```diff
@@ -0,0 +1,69 @@
...
+				test: /\.coffee$/,
+				use: require.resolve('coffee-loader')
+			}
+		]
+	},
+	externals: (context, request, callback) => {
+		for (let m of externalModules) {
+			if ((typeof m === 'string' && m === request) || (m instanceof RegExp && m.test(request))) {
+				return callback(null, 'commonjs ' + request);
+			}
```
Page-: It might be worth adding an else that throws an error, rather than silently swallowing invalid `externalModules` entries

Dockerfile.build.template
pcarranzav: Done

webpack.config.js
pcarranzav: Good point, done

pcarranzav: Added a check for that

webpack.config.js
pcarranzav: Fixed

webpack.config.js
pcarranzav: :+1:

pcarranzav: Done

webpack.config.js
pcarranzav: Done

Dockerfile.build.template
```diff
@@ -21,18 +21,20 @@ RUN mkdir -p rootfs-overlay && \
 
 COPY package.json /usr/src/app/
 
-# Install only the production modules
+# Install only the production modules that have C extensions
 RUN JOBS=MAX npm install --production --no-optional --unsafe-perm \
 	&& npm dedupe
 
+COPY webpack.config.js remove-hashbang-loader.js /usr/src/app/
```
Page-: I think you can use
```coffee
COPY src webpack.config.js remove-hashbang-loader.js /usr/src/app/

File: webpack.config.js [hunk 14]
```diff
@@ -0,0 +1,93 @@
...
+		catch (e) {
+			continue;
+		}
+		if (packageJson.optionalDependencies != null){
+			for (let key in packageJson.optionalDependencies) {
+				maybeOptionalModules.push(key)
+			}
+		}
+		if (packageJson.dependencies != null){
+			for (let key in packageJson.dependencies) {
```
Page-: This can be
```coffee
requiredModules = requiredModules.concat(packageJson.dependencies)

File: webpack.config.js [hunk 15]
```diff
@@ -0,0 +1,93 @@
...
+			lookForOptionalDeps(internalNodeModules);
+		}
+		try {
+			packageJson = JSON.parse(fs.readFileSync(path.join(sourceDir, dir, '/package.json')));
+		}
+		catch (e) {
+			continue;
+		}
+		if (packageJson.optionalDependencies != null){
+			for (let key in packageJson.optionalDependencies) {
```
Page-: This can be
```coffee
maybeOptionalModules = maybeOptionalModules.concat(packageJson.optionalDependencies)

Dockerfile.build.template
pcarranzav: Doesn't really work, it treats src as a file that way

webpack.config.js
pcarranzav: Done (but using `_.keys`)

webpack.config.js
pcarranzav: Done (but using `_.keys`)

PR #465, opened 2017-07-01
WIP: Consolidate config, split out api binding, add state inference

pcarranzav:
This is a big work in progress to refactor a bunch of things in the supervisor to prepare for a maintainable multicontainer implementation.
This PR will:
1) Consolidate all of the supervisor configuration into a single config module - making the use of config.json and the config db table transparent.
2) Avoid duplication of config values between config.json and the db - config.json will be used as a data store that is modified atomically.
3) Split out all interactions with the Resin API into an "API Binder" module.
4) Add "current state inference", i.e. get the current running containers and device config directly from docker and the OS, instead of keeping current state in the db
5) Add unit testing to as many modules as possible. Probably not with complete coverage, but this should considerably improve the supervisor's reliability and help avoid regressions.

If you comment on this PR, please keep in mind it's a work in progress and will be evolving a lot. I just wanted to PR for visibility (and Jenkins builds).


pcarranzav:
Closing this as the PR will go directly from the `multicontainer` branch.

Issue #466, opened 2017-07-03
Replace the edge device type with generic-amd64


josephroberts:
The edge device type is being deprecated and replaced with a generic family of device types as part of dependent devices V2.

pcarranzav:
Fixed in #467

PR #467, opened 2017-07-03
Replace edge device type with generic-amd64

josephroberts:
Change-type: major

Connects to #466 

This PR deprecates the `edge` device type by making `generic-amd64` the default. 

I have added an optional `device_type` parameter to the `provision` endpoint so that users can still provision an `edge` device type if they need to.

I have removed the `device_type` returned by the `get /v1/dependent-apps` endpoint as it was hard coded and the supervisor has no way of knowing that the device type actually is.

This work will be superseded by the next proxyvisor version currently being developed, with the aim of completely removing the need for the user to specify a `device_type`, instead this will come from the new `state` endpoint.


pcarranzav:
@resin-jenkins test this please

src/proxyvisor.coffee
```diff
@@ -42,10 +42,13 @@ router.get '/v1/devices', (req, res) ->
 		res.status(503).send(err?.message or err or 'Unknown error')
 
 router.post '/v1/devices', (req, res) ->
-	appId = req.body.appId
+	{ appId, device_type } = req.body
+
 	if !appId? or _.isNaN(parseInt(appId)) or parseInt(appId) <= 0
 		res.status(400).send('appId must be a positive integer')
 		return
+	device_type = 'generic-amd64' if _.isUndefined(device_type)
```
pcarranzav: nit: I think this would be better as:
```
device_type = 'generic-amd64' if !device_type?
```

pcarranzav:
@josephroberts I'm running a final autohat test and then I'll merge this

PR #468, opened 2017-07-04
Fix semver comparison for OS version when determining if the device has deviceApiKey support

pcarranzav:
The current setup would cause the check to always fail - the consequence is not *that* bad since
the provisioning key still gets overwritten, but it's better to delete it if we can.

Change-Type: patch


src/bootstrap.coffee
```diff
@@ -169,7 +169,7 @@ bootstrapOrRetry = ->
 
 hasDeviceApiKeySupport = (osVersion) ->
 	try
-		!/^Resin OS /.test(osVersion) or semver.gte(semverRegex.test(osVersion), '2.0.2')
+		!/^Resin OS /.test(osVersion) or semver.gte(semverRegex.exec(osVersion)[0], '2.0.2')
```
dfunckt: This looks fine but might be better to let semver parse the version, i.e. with something like:

```
version = osVersion.replace(/^Resin OS /, '')
return false if version is osVersion # not Resin OS
return semver.valid(version) and semver.gte(version, '2.0.2')
```

dfunckt: Hmm, unless `osVersion` is something like "Resin OS v2.3.4 (dev)" in which case ignore my comment completely.

pcarranzav: Thanks - indeed I have to consider the case when the version has other things appended

PR #470, opened 2017-07-05
Refactor build process and switch to CircleCI

pcarranzav:
A rather big refactor of how the supervisor is built, to make it more usable and maintainable.
* Added a circle.yml file to build the supervisor on Circle
* Unified all the dockerfiles into a single multi-stage build
* Refactored the Makefile to reduce the amount of necessary variables to run a build, make use of the  multi-stage build, and document its usage
* Refactor jenkins_build.sh into a build.sh that is easier to use and is documented - and can also be used to build locally without pushing the artifacts.
* Refactor the dindctl script to take options instead of environment variables.
* Change the dind supervisor to use a local image instead of pulling an external one.
* Fix a problem in dind that breaks the host's ability to run terminals
* Add a deploy-to-resin.js that allows deploying to the Resin API as a supervisor_release. This is integrated into the CircleCI build if a master build is triggered with `DEPLOY_TO_RESIN=true`
* Add a script (automation/pr-to-meta-resin.sh) that automates creating PRs to meta-resin to bump the supervisor version. This one has to be run manually.


pcarranzav:
I think this is somewhat ready for review. The main things that I think are missing are:
* Add jobs for the missing architectures in circle.yml (for now I've only enabled i386 and amd64, but it's just a matter of copying the pattern)
* Give CircleCI credentials for `resin/` repos on dockerhub - @brownjohnf could you do this? I had tried with the DOCKER_PASSWORD you had set but I think it's outdated.
* Add `PUSH_IMAGES=true` to the calls to automation/build.sh on circle.yml, so that the builds actually push the resulting images - this will also make docker layer caching start working, so hopefully each build won't take three hours as they do now (since they constantly build the base image).

I might be missing something else though. #465, on the other hand, should add a decent amount of unit tests that we can integrate into the Circle builds too - ideally running the tests before going on to the full docker builds.

Page-:
@pcarranzav it would be huge if you could split this PR into separate commits - currently it's absolutely massive and very hard to review

pcarranzav:
@Page- done (as best as I could since the changes are a bit interdependent)

tools/dind/config/services/resin-supervisor-dind.service
```diff
@@ -9,7 +9,7 @@ WorkingDirectory=/usr/src/app
 EnvironmentFile=/usr/src/app/config/env
 EnvironmentFile=/usr/src/app/config/localenv
 ExecStartPre=/bin/bash -c 'if [ "${PRELOADED_IMAGE}" == "true" ]; then /usr/bin/docker pull $(jq ".[0].imageId" ${APPS_PATH}); fi'
-ExecStartPre=/usr/bin/docker pull ${SUPERVISOR_IMAGE}
+ExecStartPre=/usr/bin/docker load -i /usr/src/supervisor-image.tar
```
Page-: Can you use `--input` please

Dockerfile
```diff
@@ -0,0 +1,207 @@
...
+FROM resin/amd64-node:6.5-slim as amd64-node-base
+RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-start && chmod +x /usr/bin/cross-build-start
+RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-end && chmod +x /usr/bin/cross-build-end
+
+FROM resin/i386-node:6.5-slim as i386-node-base
+RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-start && chmod +x /usr/bin/cross-build-start
+RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-end && chmod +x /usr/bin/cross-build-end
+
+# Build nodejs dependencies
+FROM $ARCH-node-base as node
```
Page-: Does this pick `$ARCH` up from the build args or is there another mechanism

circle.yml
```diff
@@ -46,14 +46,42 @@ jobs:
       DOCKER_USERNAME: travisciresin
```
Page-: Would it makes sense to abstract the standard env stuff to a default env section of the yaml?

Makefile
```diff
@@ -120,7 +120,10 @@ ${DOCKERD_PROXY}:
 		touch ${DOCKERD_PROXY}; \
 	fi
 
-supervisor-dind: ${DOCKERD_PROXY}
+supervisor-tar:
+	docker save -o tools/dind/supervisor-image.tar $(SUPERVISOR_IMAGE)
```
Page-: Can you use `--output` please

Dockerfile
pcarranzav: Yup, it picks it from the build args

tools/dind/config/services/resin-supervisor-dind.service
pcarranzav: Done

Makefile
pcarranzav: Done

automation/Dockerfile
```diff
@@ -4,7 +4,13 @@ RUN apk add --no-cache \
...
 	git \
-	bash
+	bash \
+	nodejs \
+	nodejs-npm
+
 WORKDIR /src
+COPY ./automation/package.json /src/automation/package.json
+RUN cd automation \
+	&& npm install
```
Page-: Can you add the `JOBS=MAX` and `npm cache clean` etc please

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,95 @@
...
+	apiPrefix: PLATFORM_ENDPOINT,
+	passthrough: requestOpts
+});
+
+resinApi._request(_.extend({
+	url: apiEndpoint + '/config',
+	method: 'GET'
+}, resinApi.passthrough))
+.then(function (config) {
+	let deviceTypes = config.deviceTypes;
```
Page-:
```js
.then(({ deviceTypes }) => {

File: automation/deploy-to-resin.js [hunk 7]
```diff
@@ -0,0 +1,95 @@
...
+
+
+const PLATFORM_ENDPOINT = url.resolve(apiEndpoint, '/v2/')
+var resinApi = new PlatformAPI({
+	apiPrefix: PLATFORM_ENDPOINT,
+	passthrough: requestOpts
+});
+
+resinApi._request(_.extend({
+	url: apiEndpoint + '/config',
```
Page-: If all you want is the device-types you should use `/config/device-types`

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,95 @@
...
+	return Promise.mapSeries(deviceTypes, function (deviceType) {
+		if (deviceType.arch === arch) {
+			let image_name = 'registry.resinstaging.io/resin/' + arch + '-supervisor';
+			let supervisor_version = tag;
+			let device_type = deviceType.slug;
+			let customOptions = {}
+			if (_.isEmpty(apiToken)) {
+				customOptions.apikey = apikey
+			}
+			console.log('Deploying ' + supervisor_version + ' for ' + device_type);
```
Page-: You can use
```es6
console.log(`Deploying ${supervisor_version} for ${device_type}`)
```
to have string interpolation

automation/pr-to-meta-resin.sh
```diff
@@ -0,0 +1,84 @@
...
+	cat $THIS_FILE | awk '{if(/^#/)print;else exit}' | tail -n +2 | sed 's/\#//'
+	exit 1
+fi
+
+if [[ -z "$GITHUB_PASSWORD" || -z "$GITHUB_USERNAME" ]]; then
+	echo "GITHUB_USERNAME and GITHUB_PASSWORD are required"
+	exit 1
+fi
+
+USER=${USER:-`whoami`}
```
Page-:
```bash
USER=${USER:-$(whoami)}
```
is preferable to `` ` `` because it allows for nesting and handles a bunch of other situations without surprises (it's not necessary in this case, but it's still a good habit to get into)

Dockerfile
```diff
@@ -0,0 +1,207 @@
...
+
+# The node version here should match the version of the runtime image which is
+# specified in the base-image subdirectory in the project
+FROM resin/rpi-node:6.5-slim as rpi-node-base
+FROM resin/armv7hf-node:6.5-slim as armv7hf-node-base
+FROM resin/armel-node:6.5-slim as armel-node-base
+FROM resin/aarch64-node:6.5-slim as aarch64-node-base
+
+FROM resin/amd64-node:6.5-slim as amd64-node-base
+RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-start && chmod +x /usr/bin/cross-build-start
```
pcarranzav: That won't work as it would call the arm shell without emulator. We have to call cross-build-start in the `[ "command" ]` format so that it skips the shell.
My hope is that we'll have amd64 and i386 base images that already include a noop cross-build-start, but for now this is the only way. Good thing is this doesn't make it to the runtime image.

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,95 @@
+// Deploy a supervisor image as a supervisor_release in the Resin API
+//
+// Environment variables:
+// This program deploys only for device types where the architecture matches $ARCH.
+// It deploys to the API specified by $API_ENDPOINT and using a provided $API_TOKEN or $API_KEY
+// (if both are set, API_TOKEN is preferred).
+// The tag to deploy must be passed as $TAG.
+//
+var PlatformAPI = require('pinejs-client');
```
pcarranzav: Good point, done

pcarranzav: Done

automation/deploy-to-resin.js
pcarranzav: Nice, didn't know we had that one.  Done

automation/deploy-to-resin.js
pcarranzav: Done (no need for the {} now)

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,95 @@
...
+});
+
+resinApi._request(_.extend({
+	url: apiEndpoint + '/config',
+	method: 'GET'
+}, resinApi.passthrough))
+.then(function (config) {
+	let deviceTypes = config.deviceTypes;
+	// This is a critical step so we better do it serially
+	return Promise.mapSeries(deviceTypes, function (deviceType) {
```
pcarranzav: Done

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,95 @@
...
+		if (deviceType.arch === arch) {
+			let image_name = 'registry.resinstaging.io/resin/' + arch + '-supervisor';
+			let supervisor_version = tag;
+			let device_type = deviceType.slug;
+			let customOptions = {}
+			if (_.isEmpty(apiToken)) {
+				customOptions.apikey = apikey
```
pcarranzav: Done, I think

automation/deploy-to-resin.js
pcarranzav: Done

automation/pr-to-meta-resin.sh
pcarranzav: Done, and I'll read up about the difference

automation/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+	jq \
+	git \
+	bash \
+	nodejs \
+	nodejs-npm
+
+WORKDIR /src
+COPY ./automation/package.json /src/automation/package.json
+RUN cd automation \
+	&& npm install
```
Page-: Lost from last review:
> Can you add the `JOBS=MAX` and `npm cache clean` etc please

Dockerfile
Page-: Ok, that makes sense, can you combine the two `RUN`s into one instead then please

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,94 @@
...
+resinApi._request(_.extend({
+	url: apiEndpoint + '/config/device-types',
+	method: 'GET'
+}, resinApi.passthrough))
+.then( (deviceTypes) => {
+	// This is a critical step so we better do it serially
+	return Promise.mapSeries(deviceTypes, (deviceType) => {
+		if (deviceType.arch === arch) {
+			let image_name = `registry.resinstaging.io/resin/${arch}-supervisor`;
+			let supervisor_version = tag;
```
Page-: These should all be `const`s

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,94 @@
...
+			let customOptions = {};
+			if (_.isEmpty(apiToken)) {
+				customOptions.apikey = apikey;
+			}
+			console.log(`Deploying ${supervisor_version} for ${device_type}`);
+			return resinApi.post({
+				resource: 'supervisor_release',
+				body: {
+					image_name: image_name,
+					supervisor_version: supervisor_version,
```
Page-: This can be
```
body: {
	image_name,
	supervisor_version,
	device_type
}
```
although I think it would be better to just assign the object in-line rather than having js vars using a different naming style to our standard

automation/deploy-to-resin.js
```diff
@@ -0,0 +1,94 @@
...
+					device_type: device_type,
+					is_public: true
+				},
+				customOptions: customOptions
```
Page-: This can be just
```
customOptions

pcarranzav:
All done @Page-

pcarranzav:
@Page- it took a few more changes to actually make it work - I removed the docker-in-docker builder and adjusted a few details to improve the build times. Still approve?

PR #472, opened 2017-07-10
When the device is about to reboot or shutdown, close the API server and avoid applying updates

pcarranzav:
We mark when the device is rebooting and avoid some steps in the update cycle that change the device
state, similarly to when the device is in local mode, to avoid problems with non-atomic operations.
This doesn't solve *all* the potential scenarios of a reboot happening in the middle of an update, but at least
should prevent the case where we start an app container and reboot the device before saving the containerId, potentially
causing a duplicated container issue.

We also correct the API docs to reflect the 202 response when reboot or shutdown are successful.

Change-Type: patch


pcarranzav:
@Page- I took a different approach on this one. It doesn't cover all cases but should help us solve the issue presented here https://forums.resin.io/t/preloading-application-all-port-bindings-fail/1417/4
i.e. the problem when we apply config.txt changes, trigger a reboot and then the app is partially installed before the reboot

taeold:
Any update on this PR? I would love to see this PR merged to try out pre-loading on my app again!

src/api.coffee
```diff
@@ -59,10 +59,9 @@ module.exports = (application) ->
 		.then ->
 			new Promise (resolve, reject) ->
```
Page-: This `new Promise` is no longer needed?

src/api.coffee
```diff
@@ -82,10 +81,9 @@ module.exports = (application) ->
...
```
Page-: This `new Promise` is no longer needed?

src/device.coffee
```diff
@@ -40,8 +40,25 @@ exports.getID = memoizePromise ->
 			throw new Error('Could not find this device?!')
 		return devices[0].id
 
+exports.shuttingDown = false
```
Page-: Does it make sense to combine this and the offline mode stuff?

pcarranzav: They're different, as offline mode would still allow updates to be applied if they were set locally... this will change a lot with multicontainer anyways.

src/api.coffee
pcarranzav: Can't believe I missed this, thanks

PR #473, opened 2017-07-12
Update nodejs to v6.11.1

pcarranzav:
Change-Type: patch

Requires new base images (ARCH-node:6.11.1-slim) to be built.
Not sure if adding the nodejs_6.11.1.bb file there will work, let's see what Jenkins says. We might wanna PR the bump into meta-nodejs anyways.


pcarranzav:
Okay, that didn't work (adding the nodejs_6.11.1.bb there doesn't catch).
I'll have to fork meta-nodejs for now (until we can get a PR into the official one) to add it unless @agherzan knows a better way to do this.

pcarranzav:
Okay, so it looks like the hashseed vuln doesn't affect meta-nodejs builds since they're built without snapshot. I'll leave this open until 6.11.1 makes it into meta-nodejs upstream as it's still nice to have, but I won't merge with meta-nodejs from resin-playground.

pcarranzav:
Closing this as the build process has changed so this would need a big refactor in the rebase.
And we'll wait for https://github.com/imyller/meta-nodejs/pull/75 to update to node 8 directly.

PR #477, opened 2017-07-25
Update docker-delta to 1.1.1, docker-toolbelt to 3.0.1, docker-progress to 2.6.0 to add support for deltas and overlay2

pcarranzav:
This makes the Async suffix for docker functions unnecessary. It also allows us to remove dockerode as an
explicit dependency.

Change-Type: minor


src/lib/logger.coffee
```diff
@@ -20,10 +20,7 @@ initialised = new Promise (resolve) ->
 		resolve(config)
 
 dockerPromise = initialised.then (config) ->
-	docker = Promise.promisifyAll(new Docker(socketPath: config.dockerSocket))
-	# Hack dockerode to promisify internal classes' prototypes
-	Promise.promisifyAll(docker.getImage().constructor.prototype)
-	Promise.promisifyAll(docker.getContainer().constructor.prototype)
+	docker = new Docker(socketPath: config.dockerSocket)
```
Page-: Any reason we don't use the docker-utils exports docker here?

pcarranzav: Good question, I think there might have been a reason but I'll try it out

PR #479, opened 2017-07-26
Fix preloading and parsing of arguments in the dind supervisor

pcarranzav:
We now remove the quotes from the image name in the ExecStartPre of the supervisor dind service,
so that the images are properly pulled.

We also fix the parsing of arguments in dindctl so that it correctly parses all of the arguments (the last one was being missed if it didn't have a parameter).

Change-Type: patch


dindctl
```diff
@@ -115,12 +124,18 @@ function runDind {
 	if [ "$PRELOADED_IMAGE" = "true" ]; then
 		echo "Running with preloaded apps"
 	fi
+	image_available=$( (docker inspect $SUPERVISOR_IMAGE &> /dev/null && echo "true") || echo "false" )
+	if [ "$image_available" = "false" ]; then
```
Page-: I think you can change this to
```bash
if docker inspect $SUPERVISOR_IMAGE &> /dev/null; then
```

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "start": "./entry.sh",
     "build": "webpack --optimize-minimize",
+    "build-no-optimize": "webpack",
```
Page-: You could change to the following
```
"build": "webpack"
```
```
npm run build -- --optimize-minimize
```
but I think it's nice to have the default be the one we would usually want.

dindctl
```diff
@@ -46,13 +46,14 @@ SUPERVISOR_EXTRA_MOUNTS=""
 DIST_MOUNTED="false"
 DIND_IMAGE="resin-supervisor-dind"
 PRELOADED_IMAGE=""
+OPTIMIZE="true"
```
Page-: What do you think of an option where we run the supervisor directly from the coffee-script files in dindctl without needing to separately recompile? It would make debugging even easier, especially if sourcemaps are giving us the correct line numbers

tools/dind/config/services/resin-supervisor-dind.service
```diff
@@ -8,7 +8,7 @@ Before=openvpn-resin.service
 WorkingDirectory=/usr/src/app
 EnvironmentFile=/usr/src/app/config/env
 EnvironmentFile=/usr/src/app/config/localenv
-ExecStartPre=/bin/bash -c 'if [ "${PRELOADED_IMAGE}" == "true" ]; then /usr/bin/docker pull $(jq ".[0].imageId" ${APPS_PATH}); fi'
+ExecStartPre=/bin/bash -c 'if [ "${PRELOADED_IMAGE}" == "true" ]; then /usr/bin/docker pull $(jq ".[0].imageId" ${APPS_PATH} | sed \'s/"//g\'); fi'
```
Page-: Can you change this to
```
jq --raw-output ".[0].imageId" ${APPS_PATH}
```
please, which means that jq outputs the string contents by themselves rather than a JSON version (which may have escaped characters in, eg it will print `\n` rather than an actual new line)

pcarranzav: Oh nice, jq is like a box of chocolates

package.json
pcarranzav: Yeah, agreed that it's better to make optimization default

dindctl
pcarranzav: That's a nice idea, but I'd leave it for later as it involves a few more changes

dindctl
pcarranzav: Done (but it needed to be negated with `!`)

Page-: Right, for some reason I thought this was checking against true.. my bad

PR #481, opened 2017-07-28
Fix references in deploy-to-resin.js and use github credentials when pushing in pr-to-meta-resin.sh

pcarranzav:
Change-Type: patch


automation/pr-to-meta-resin.sh
```diff
@@ -48,12 +48,13 @@ docker-resin-supervisor-disk: Update to ${TAG}
 
 Changelog-Entry: Update supervisor to ${TAG}
 Change-Type: patch
-" <<EOF
+"
+
+	git push origin $HEAD <<EOF
```
Page-: I think you could also use
```
git push https://${GITHUB_USERNAME}:${GITHUB_PASSWORD}@github.com/resin-os/meta-resin.git $HEAD
```
or you could use an oauth token from https://github.com/settings/tokens as
```
git push https://$OAUTH_TOKEN:x-oauth-basic@github.com/name/repo.git $HEAD
```
which lets you limit the permissions a bit

Alternatively if you use one of those urls in the git clone I don't think you'll have to worry about providing the credentials in the other places of this script

pcarranzav: Nice one! I changed it to use it in the push cause when cloning then the pasword becomes visible when doing `git remote -v`.

I also realized that hub picks up `$GITHUB_USER` and `$GITHUB_PASSWORD` so I changed it to use that instead.

PR #483, opened 2017-08-02
Resumable deltas

dfunckt:
Applying a delta update consists of two parts:

1. The request to the delta server for the delta payload (an rsync batch file, plus some prepended Docker metadata). The response is a redirect to a URL that contains the delta (currently S3).
2. The request for the actual download of the delta. The response is streamed directly to rsync, which applies it onto the mounted root filesystem of the final image.

The first step may take a while as it may trigger the generation of the delta if the request is the first one for this combination of src/dest image and the images are large. If the request times out, either because of the delta server taking too long to respond or bad network, the Supervisor automatically schedules a retry to be performed after a while.

Currently, similar behaviour applies to the second step as well -- if the request fails, we immediately bail out and the Supervisor schedules a retry of the whole process (i.e. from step 1). But in this case it means we might have downloaded and applied some or most of the delta when a socket timeout occurs causing us to start all over again, wasting time and bandwidth.

This commit splits the process into the two discreet steps and improves the behaviour on the second step. Specifically:

- makes the Supervisor try to resume the delta download request several times before it bails out and starts the process all over again.
- removes arbitrary timeout which applied over the whole process and meant some deltas would never manage to be applied (because of large delta size and low network bandwidth).
- makes sure any launched rsync processes always exit and any opened streams consumed and closed.

Most of the improvements are in the two dependencies linked below -- `resumable-request` and `node-docker-delta` -- and this commit merely combines the updated versions of these modules.

Change-Type: minor
Connects-To: #140
Depends-On: https://github.com/resin-io/node-docker-delta/pull/19
Depends-On: https://github.com/resin-io-modules/resumable-request/pull/2
---- Autogenerated Waffleboard Connection: Connects to #140


src/request.coffee
```diff
@@ -16,12 +17,23 @@ if osVersion?
 	else
 		userAgent += " (Linux; #{osVersion})"
 
+# With these settings, the device must be unable to receive a single byte
+# from the network for a continuous period of 20 minutes before we give up.
+# (reqTimeout + retryInterval) * retryCount / 1000ms / 60sec ~> minutes
+DEFAULT_REQUEST_TIMEOUT = 30000 # ms
+DEFAULT_REQUEST_RETRY_INTERVAL = 10000 # ms
+DEFAULT_REQUEST_RETRY_COUNT = 30
```
CameronDiver: I'm not too familiar with the supervisor codebase, but do you think it's worth adding these to the `config` object used elsewhere?

dfunckt:
Tested the PR locally in a running supervisor and it works nicely.

pcarranzav:
@dfunckt did you test bringing the device offline and checking that the delta resumes correctly?

src/application.coffee
```diff
@@ -218,9 +218,15 @@ fetch = (app, setDeviceUpdateState = true) ->
...
 				if conf['RESIN_SUPERVISOR_DELTA'] == '1'
 					logSystemEvent(logTypes.downloadAppDelta, app)
-					requestTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'], positive: true) ? 30 * 60 * 1000
-					totalTimeout = checkInt(conf['RESIN_SUPERVISOR_DELTA_TOTAL_TIMEOUT'], positive: true) ? 24 * 60 * 60 * 1000
-					dockerUtils.rsyncImageWithProgress(app.imageId, { requestTimeout, totalTimeout, uuid, apiKey }, onProgress)
+					deltaOpts = {
+						uuid, apiKey
+						# use user-defined timeouts, but fallback to defaults if none is provided.
+						requestTimeout: checkInt(conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'], positive: true)
+						applyTimeout: checkInt(conf['RESIN_SUPERVISOR_DELTA_APPLY_TIMEOUT'], positive: true)
```
pcarranzav: Shouldn't this have a default timeout specified here?

dfunckt:
@pcarranzav yep, I did. As soon as the supervisor gets hold of the delta URL (the part before behaves as before, see PR description), it has very good chances to manage to download and apply the delta. In any case, the resume handling code lives inside resumable-request which I have tested thoroughly while developing the feature.

src/application.coffee
dfunckt: @pcarranzav, as it is now, if any of the options aren't supplied, the variables will be undefined and will cause `resumable-request` to fall back to defaults, which are set here: https://github.com/resin-io/resin-supervisor/pull/483/files#diff-5b6270202e7b362597737840968ab7fcR23

pcarranzav: Right, but that doesn't work for `applyTimeout`, because docker-delta doesn't have a default timeout - and I think we should set one to avoid problems with malformed deltas that cause rsync to hang

dfunckt: Oh right, I see now. Good catch 👍 

Any preference for a default? Something like 5 minutes (as it was at some point I think?) -- note that the timeout will start counting from the moment the stream to rsync closes, i.e. after download succeeds or fails.

dfunckt: Added a default rsync timeout of 6 minutes.

src/request.coffee
dfunckt: @CameronDiver sorry, I missed your comment. These are the absolute lowest-level defaults and can be overridden by code that uses this module and, in the case of deltas, by users themselves.

PR #486, opened 2017-08-18
Ensure /var/run/resin exists before trying to mount it as tmpfs

pcarranzav:
Fixes issues on systems where mount files if the target doesn't exist (seems to happen on
boards that run docker on overlayfs).

Change-Type: patch


pcarranzav:
Tested on Shaun's device, works fine, merging.

PR #489, opened 2017-08-24
Deploy all archs by default to Resin

dfunckt:
This makes the deploy-to-resin script deploy all supported archs by default, unless one is specified via the command line.

Change-Type: patch


dfunckt:
@pcarranzav good point -- updated.

pcarranzav:
Some tests fail due to an ongoing incident on CircleCI, I'm retrying them.

pcarranzav:
@dfunckt would you please rebase and merge after #491 is merged?

dfunckt:
@pcarranzav I will if I'm around but it seems tests take a while (and some fail too), so if I'm not and you need this before Monday please go ahead and merge.

PR #490, opened 2017-08-24
Prefer err.message when reporting errors from dockerode

pcarranzav:
...then err.json and err.reason.

Errors from docker-modem that are passed from dockerode can have a "json" or "reason" property,
but that is generally less descriptive than the more standard "message", and can show up in the logs
as `[object Object]`. This commit changes it so that we log err.message if it is non-empty, and otherwise
look for json and reason.

Change-Type: patch


pcarranzav:
@zozo123 do you have a device hitting this issue? cause I'd like to test this on a device before merging

pcarranzav:
Tested, looks like it works fine, will rebase and merge

PR #491, opened 2017-08-25
Remove a file from sqlite3 that took 30MB

pcarranzav:
node.dtps was big and useless.

See https://github.com/mapbox/node-sqlite3/issues/861

Change-Type: patch


Dockerfile
```diff
@@ -162,7 +163,8 @@ RUN find . -path '*/coverage/*' -o -path '*/test/*' -o -path '*/.nyc_output/*' \
 		-o -name '*.yml' \
 		-delete \
 	&& find . -type f -path '*/node_modules/sqlite3/deps*' -delete \
-	&& find . -type f -path '*/node_modules/knex/build*' -delete
+	&& find . -type f -path '*/node_modules/knex/build*' -delete \
+	&& rm -f node_modules/sqlite3/node.dtps
```
dfunckt: I think `rm` will fail if the file is not found -- might make sense to "catch" this case so it doesn't start failing suddenly.

pcarranzav: the `-f` makes it ignore non-existent files.
I just added a `-r` too cause it turned out it was a directory.

PR #492, opened 2017-08-25
Forward resume options

dfunckt:
This was missed from #483 and user-set resume options were ignored.

Change-Type: patch


src/docker-utils.coffee
```diff
@@ -52,9 +52,9 @@ getRepoAndTag = (image) ->
 			registry = ''
 		return { repo: "#{registry}#{imageName}", tag: tagName }
 
-applyDelta = (imgSrc, deltaUrl, { requestTimeout, applyTimeout, retryCount, retryInterval }, onProgress) ->
+applyDelta = (imgSrc, deltaUrl, requestTimeout, applyTimeout, resumeOpts, onProgress) ->
```
pcarranzav: I'd rather keep the function signature as `(imgSrc, deltaUrl, options, onProgress)`, i.e. making this `(imgSrc, deltaUrl, { requestTimeout, applyTimeout, resumOpts }, onProgress)`, what do you think?
Either that or keep the function definition as before and build `resumeOpts` inside this function.

dfunckt: Sounds good -- PR updated.

dfunckt:
@pcarranzav -- it would be nice to have this with the next OS release.

pcarranzav:
@dfunckt merging, will add to next release

dfunckt:
Thanks 👍

PR #498, opened 2017-09-12
Multicontainer proxyvisor

josephroberts:
@pcarranzav tested and with these 2 changes everything seems to work


src/application-manager.coffee
```diff
@@ -17,7 +17,7 @@ Images = require './compose/images'
 Networks = require './compose/networks'
 Volumes = require './compose/volumes'
 
-Proxyvisor = require './proxyvisor'
+Proxyvisor = require './proxyvisor/proxyvisor'
```
pcarranzav: If you rename proxyvisor.coffee into index.coffee then here you can still `require './proxyvisor'`

src/proxyvisor/proxyvisor-v1.coffee
```diff
@@ -0,0 +1,182 @@
...
+	validStringOrUndefined,
+	validObjectOrUndefined,
+	# tarDirectory,
+	# tarFilename,
+	tarPath,
+	getTarArchive,
+	# cleanupTars,
+	# formatTargetAsState,
+	# formatCurrentAsState,
+} = require './lib'
```
pcarranzav: I'd rather call this "utils"

src/proxyvisor/proxyvisor.coffee
```diff
@@ -1,232 +1,35 @@
...
+	# validObjectOrUndefined,
+	# tarDirectory,
+	# tarFilename,
+	# tarPath,
+	# getTarArchive,
+	cleanupTars,
+	formatTargetAsState,
+	formatCurrentAsState,
+} = require './lib'
+proxyvisorRouterV1 = require './proxyvisor-v1.coffee'
```
pcarranzav: Since proxyvisor-v1.coffee exports the class, you can directly do `ProxyvisorRouterV1 = require './proxyvisor-v1'` (also, no need for the .coffee).
Also, I'd prefer it if you kept the filename consistent with the class name, i.e. `proxyvisor-router-v1.coffee`, as we do in the other modules.

src/proxyvisor/proxyvisor.coffee
```diff
@@ -1,232 +1,35 @@
...
-					commit: app.commit
-					name: app.name
-					config: JSON.parse(app.config ? '{}')
-				}
-			.then (apps) ->
-				res.json(apps)
-			.catch (err) ->
-				console.error("Error on #{req.method} #{url.parse(req.url).pathname}", err, err.stack)
-				res.status(503).send(err?.message or err or 'Unknown error')
+		@router.use('/v1', new proxyvisorRouterV1.ProxyvisorRouterV1(@proxyvisor))
```
pcarranzav: Continuing from the comment above, here you can just use `new ProxyvisorRouterV1(@proxyvisor)` instead of `new proxyvisorRouterV1.ProxyvisorRouterV1(@proxyvisor)`

src/proxyvisor/proxyvisor.coffee
```diff
@@ -471,7 +274,7 @@ module.exports = class Proxyvisor
 					config: device.apps[appId].config
 					environment: device.apps[appId].environment
 				}
-				if device.targetCommit? and !_.isEqual(targetState, currentState) and !_.isEqual(targetState, @acknowledgedState[device.uuid])
+				if device.apps[appId].targetCommit? and !_.isEqual(targetState, currentState) and !_.isEqual(targetState, @acknowledgedState[device.uuid])
```
pcarranzav: Nice catch here :) I had found it when debugging on your device but forgot to change it

src/proxyvisor/proxyvisor-v1.coffee
```diff
@@ -0,0 +1,182 @@
...
+
+# Does this need to be a class? I tried without and routing worked but couldn't access any of the other objects e.g. @db
+module.exports = class ProxyvisorRouterV1
+	constructor: (@proxyvisor) ->
+		console.log('constructor')
+		{ @config, @logger, @db, @docker, @reportCurrentState } = @proxyvisor
+		@router = express.Router()
+		@router.use(bodyParser.urlencoded(extended: true))
```
pcarranzav: Since the parent router already uses bodyParser, there's no need to add it here

src/proxyvisor/proxyvisor-v1.coffee
```diff
@@ -0,0 +1,182 @@
...
+		@router.use(bodyParser.json())
+
+	@router.get '/test', (req, res) ->
```
pcarranzav: All of these need one more indentation level, as they're actually part of the constructor.

src/db.coffee
```diff
@@ -180,15 +180,15 @@ module.exports = class DB
...
 							t.string('device_type')
 							t.string('logs_channel')
 							t.string('deviceId')
 							t.boolean('is_online')
 							t.string('name')
 							t.string('status')
-							t.string('download_progress')
-							t.string('is_managed_by')
+							t.integer('download_progress')
+							t.string('device')
```
pcarranzav: Isn't this the same a s the `deviceId` above?

src/proxyvisor/utils.coffee
```diff
@@ -0,0 +1,78 @@
...
+		return "#{self.tarDirectory(appId)}/#{self.tarFilename(appId, commit)}"
+
+	getTarArchive: (source, destination) ->
+		fs.lstatAsync(destination)
+		.catch ->
+			mkdirp(path.dirname(destination))
+			.then ->
+				execAsync("tar -cvf '#{destination}' *", cwd: source)
+
+	cleanupTars: (appId, commit) ->
```
pcarranzav: Check the latest version of this function - there's been changes to this that are getting lost in the rebase

src/db.coffee
josephroberts: https://github.com/resin-io/resin-api/blob/master/src/models/resin.sbvr#L502-L505 is the model. I incorrectly thought it would translate into a field called `is_managed_by` on the DB but it actually translates to `device` as a FK pointing to the managing device

PR #499, opened 2017-09-14
Explicitly define the source for deltas and allow cross-app deltas

pcarranzav:
This commit changes the way the source for a delta is determined. We used to do it by comparing the available tags with the one we want and relying on the format that includes the app in the image name. Now we explicitly choose a delta source from the previous app version if we have one, and otherwise use the image from any available app - which will allow us to have a valid source when moving a device between apps.

For this to work consistently if there's an unexpected reboot, we now avoid deleting an app from the db until the full update has succeeded. Instead, we mark the app for deletion so that we still have the image stored after the reboot.

This commit also changes a .map to .mapSeries when iterating over appIds for removal/install/update - this avoids parallel treatment of apps which can cause inconsistencies in the status reported to the API.

Change-Type: patch


pcarranzav:
Untested, will proceed to test now

pcarranzav:
Now properly tested on an rpi3 and it works fine

PR #500, opened 2017-09-25
Mixpanel tunneling through resin API, and extra hiding of properties

pcarranzav:


izavits:
Let's wait to see if more events needs to be filtered before merging.
There is at least one more: https://github.com/resin-io/hq/pull/844#issuecomment-332126159

but I'm searching for all the cases

pcarranzav:
Added extra filtering for that case @izavits, waiting for your report on any other ones to filter - and then I guess we need @afitzek to confirm we're good.

izavits:
Thanks @pcarranzav. I'm compiling a list with the events and the properties that need filtering here: https://docs.google.com/spreadsheets/d/1L9c2C2yIVuMvsCbQZ4opZgF1sxIKNJpHwKEdzVyHRyY/edit#gid=0

Andreas can also check in case we need to hide more fields.
// Cc. @afitzek @petrosagg

pcarranzav:
Looking at that document, it seems like this PR already addresses all of them that relate to the supervisor :)

izavits:
@pcarranzav Yes, I believe that the supervisor events should be ok with this PR. However I would wait for another approval from the security team.

pcarranzav:
Agreed

izavits:
@pcarranzav hey Pablo. Please don't merge yet. We had a discussion with Petros and after all we will not hash the `commit` value and the `imageId`. What we need though is to hide the `api_secret` and the `RESIN_SUPERVISOR_API_KEY`

izavits:
@pcarranzav Yes please. Do not hide the commit, do not hide imageId, hide env, hide api_secret. Thank you!

> Cool, so I'll remove the hiding of commit and imageId, but keep hiding the full env and api_secret. Is that decided and settled or are there discussions pending? @izavits

izavits:
I accidentally closed the PR when I commented. Hope everything is ok now that I re-opened it @pcarranzav

pcarranzav:
done @izavits  :)
just need to test it

src/utils.coffee
```diff
@@ -40,15 +40,13 @@ exports.mixpanelTrack = (event, properties = {}) ->
...
-		try
-			{ env } = properties.app
-			env = JSON.parse(env) if _.isString(env)
-			safeEnv = _.omit(env, config.privateAppEnvVars)
-			properties.app.env = JSON.stringify(safeEnv)
-		catch
-			properties.app.env = 'Fully hidden due to error in selective hiding'
+	# Don't log env vars as they may include keys or other secrets
+	if properties?.app?
+		delete properties.app.env
```
petrosagg: I'd be safer to have a whitelist otherwise if we change this object the data will leak.

pcarranzav: hmm good point, good thing I hadn't merged this yet.
I'll change it to a whitelist, we'd just need to decide what elements of `app` and `stateDiff` are relevant for mixpanel.

pcarranzav: I propose picking these from app: `appId, imageId, commit, name`
And these from stateDiff (when the supervisor  fails to report current state): `status, download_progress, commit, os_version, supervisor_version, ip_address`

Make sense @petrosagg @pimterry @afitzek @izavits ? anything else you'd like to see or see removed?
This filters logs_channel (which is a sensitive field) and some fields that I think are irrelevant, like containerId, provisioning_progress, etc (I don't see how we'd use them on mixpanel).

Keep in mind much of these events will change soon with multicontainer (e.g. most of the things we currently associate with an "app" will now be associated with a "service", so we'll have "Service start" instead of "Application start", etc).

izavits: I don't think that `provisioning_progress` will be needed for analytics, nor containerId. I think those fields you mention are fine.

pcarranzav:
@petrosagg @izavits I changed the filtering to use a whitelist

src/utils.coffee
```diff
@@ -40,16 +48,13 @@ exports.mixpanelTrack = (event, properties = {}) ->
...
-			{ env } = properties.app
-			env = JSON.parse(env) if _.isString(env)
-			safeEnv = _.omit(env, config.privateAppEnvVars)
-			properties.app.env = JSON.stringify(safeEnv)
-		catch
-			properties.app.env = 'Fully hidden due to error in selective hiding'
-
+	# Filter properties to only send the whitelisted keys and values
+	properties = _.pick(properties, _.keys(mixpanelWhitelist))
+	properties = _.mapValues properties, (v, k) ->
```
petrosagg: We should use something like https://github.com/nemtsov/json-mask This is a very ad-hoc implementation that will break once we have more recursion

pcarranzav: @petrosagg done (but I still kept the mask defined with an array, otherwise it was an unreadable long line)

(didn't know json-mask, it's nice!)

PR #502, opened 2017-10-04
Update docker-progress

dfunckt:
The updated version adds more accurate progress reporting when the underlying engine is resin-os/docker (AKA Balaena)

Change-Type: patch


pcarranzav:
(and maybe even add "...to support more accurate progress reporting when the container engine is balaena")

PR #504, opened 2017-10-16
Delta improvements

dfunckt:
- Updates resumable-request to 1.0.1
- Updates docker-progress to 2.0.3
- Removes `DEFAULT_DELTA_APPLY_TIMEOUT`; it’s not needed anymore, docker-delta reliably tracks rsync.
- Properly end the update when applying the delta results in an error.

Depends on:

- https://github.com/resin-io-modules/resumable-request/pull/4
- https://github.com/resin-io/node-docker-delta/pull/22

Change-Type: patch


src/application.coffee
```diff
@@ -220,13 +218,12 @@ fetch = (app, { deltaSource, setDeviceUpdateState = true } = {}) ->
 				if conf['RESIN_SUPERVISOR_DELTA'] == '1'
 					logSystemEvent(logTypes.downloadAppDelta, app)
 					deltaOpts = {
-						uuid, apiKey
+						uuid, apiKey, deltaSource
 						# use user-defined timeouts, but fallback to defaults if none is provided.
 						requestTimeout: checkInt(conf['RESIN_SUPERVISOR_DELTA_REQUEST_TIMEOUT'], positive: true)
-						applyTimeout: checkInt(conf['RESIN_SUPERVISOR_DELTA_APPLY_TIMEOUT'], positive: true) ? DEFAULT_DELTA_APPLY_TIMEOUT
```
Page-: What happens now if `RESIN_SUPERVISOR_DELTA_APPLY_TIMEOUT` isn't provided, or isn't a number?

dfunckt: It’ll be undefined which will result in docker-delta’s default timeout be applied, which is 0, meaning “no timeout”. That’s the behaviour we want by default, given that any timeout value is just arbitrary. If however this causes issues in specific cases, the env var can be used to adjust it.

src/docker-utils.coffee
```diff
@@ -38,7 +38,7 @@ applyDelta = (imgSrc, deltaUrl, { requestTimeout, applyTimeout, resumeOpts }, on
 				deltaStream = dockerDelta.applyDelta(imgSrc, timeout: applyTimeout)
 				res.pipe(deltaStream)
 				.on('id', resolve)
-				.on('error', reject)
+				.on('error', req.destroy.bind(req))
```
pcarranzav: How does the Promise get rejected in this case? Does `req.destroy` make `req` fire an error event?

src/application.coffee
```diff
@@ -220,13 +218,12 @@ fetch = (app, { deltaSource, setDeviceUpdateState = true } = {}) ->
...
+						applyTimeout: checkInt(conf['RESIN_SUPERVISOR_DELTA_APPLY_TIMEOUT'], positive: true)
```
pcarranzav: If deltas now track rsync, is the applyTimeout still relevant or should it be removed altogether?

dfunckt: Now it applies as a timeout over the whole update (much like it was with the old deltas), and by default is 0 (due to falling back to using docker-delta's default) which means "no timeout". It'll be handy for any use case where there's need to limit the allowed time to update for whatever, but can't think of concrete use cases TBH. I'd be hesitant to remove it though.

src/docker-utils.coffee
dfunckt: Yeah, `req.destroy` will be called with an `Error` which will make `ResumableRequest` emit an error event, which will eventually reject the promise. That's the clearest way I could think of that allowed both `ResumableRequest` and the delta stream to be properly closed (and also kill rsync if needed).

src/application.coffee
pcarranzav: Yeah, true, I guess it's good to have it as a plan B in any case

src/application.coffee
Page-: Ok

PR #482, opened 2017-07-31
Multicontainer

pcarranzav:
This PR adds support for multicontainer applications, and refactors most of the supervisor. It is huge, but it fixes a variety of issues:

Closes #559 
Closes #531 
Closes #506 
Closes #487 
Closes #461
Closes #450 
Closes #439 
Closes #426 
Closes #423 
Closes #388 
Closes #358 
Closes #348 
Closes #226 
Closes #194 
Closes #115 
Closes #63 
Closes #52 
Closes #30 

Related specs:
Connects-to: https://github.com/resin-io/hq/pull/616
Connects-to: https://github.com/resin-io/hq/pull/734

Also connects-to:
* https://github.com/resin-io/hq/issues/1007
* https://github.com/resin-io/hq/issues/996 (image urls have changed)
* https://github.com/resin-io/hq/issues/991 (offline mode improved)
* https://github.com/resin-io/hq/issues/971
* https://github.com/resin-io/hq/issues/967 (containers with bridge networking)
* https://github.com/resin-io/hq/issues/948
* https://github.com/resin-io/hq/issues/850 (same as 226)
* https://github.com/resin-io/hq/issues/820
* https://github.com/resin-io/hq/issues/817 (config variables now are independent from app)
* https://github.com/resin-io/hq/issues/569
* https://github.com/resin-io/hq/issues/494
* https://github.com/resin-io/hq/issues/424 (labels to configure access to resin API and supervisor API)
* https://github.com/resin-io/hq/issues/383
* https://github.com/resin-io/hq/issues/307
* https://github.com/resin-io/hq/issues/226 (volumes will be deleted when an app is deleted)

It may connect to other issues, but that's what I could find so far.

PR #507, opened 2017-10-18
Better container management to avoid duplicated/spurious containers

pcarranzav:
Connects to https://github.com/resin-io/hq/issues/820

The objective of this PR is to improve how we manage containers by using a container name (that we craft before creating the container) to identify containers instead of using the containerId (which we have to store *after* creating the container, so it can't be atomic).

There's also a refactor of how containers are cleaned up to properly remove any containers that are not the running supervisor or the known app containers.

It's an intermediate step to what we'll do in #482 where we infer the current apps from the running containers without using the sqlite DB, but should solve most if not all of the ocurrences of spurious containers we've seen in the wild.

I have yet to test it to see if it works and produces the behavior we want.


src/api.coffee
```diff
@@ -185,13 +189,16 @@ module.exports = (application) ->
...
-				app.env = _.omit(JSON.parse(app.env), config.privateAppEnvVars)
-				# Don't return data that will be of no use to the user
-				res.json(app)
+				application.getContainerId(app)
+				.then (containerId) ->
+					# Don't return keys on the endpoint
+					app.env = _.omit(JSON.parse(app.env), config.privateAppEnvVars)
+					app.containerId = containerId
+					# Don't return data that will be of no use to the user
+					res.json(_.omit(app, [ 'containerName' ]))
```
Page-: I'd prefer a whitelist approach to rather than this blacklist approach (ie `_.pick`), also does the loss of the returned `containerId` matter? Technically it's a breaking change

pcarranzav: The whitelist is already implemented by selecting the columns in utils.getKnexApp above - here we just want to remove the containerName, which we purpusefully add to the selection only because we need it to get the containerId. I added the application.getContainerId function here precisely so that it's *not* a breaking change, we still send it.

pcarranzav:
@Page- I made a few fixes (some bugs that came up during testing) and added an extra commit. This last one fixes a somewhat unrelated potential problem. We can only do this fix after this PR's changes because we now "soft-delete" the app when starting it fails, that's why I included it here. Would you take another look?

src/api.coffee
Page-: Ok

PR #510, opened 2017-10-20
Ensure preloaded apps get the deviceApiKey in the env vars, and apps …

### LINKED ISSUE #457: Preloaded apps get restarted when the supervisor provisions
Apparently because the `RESIN_API_KEY` env var changes.

pcarranzav:
…never get the provisioning key, and improve detection of cases when the device has been pre-provisioned

It appears preloaded apps have been getting restarted because the "apiKey" configuration value was only available after provisioning succeeded. This change ensures the
deviceApiKey that the device will use is injected into the env vars of preloaded apps, ensuring the app is not restarted (unless provisioning fails and the uuid and deviceApiKey are
regenerated, but this should be rare).

We also ensure that whenever an app's RESIN_API_KEY env var is populated, it is *always* done with the deviceApiKey and never with the provisioning apiKey.

This change also improves the check for the DuplicateUuidError that can happen if a device has been provisioned but the API's response hasn't been persisted - the error message
returned from the API has been known to have a few variations (usually an extra dot at the end), so we now use _.startsWith instead of checking for equal strings to make the
supervisor still work under these variations.

Closes #457
Change-Type: patch


src/bootstrap.coffee
```diff
@@ -16,7 +16,7 @@ semverRegex = require('semver-regex')
 
 userConfig = {}
 
-DuplicateUuidError = message: '"uuid" must be unique.'
+DuplicateUuidError = (err) -> _.startsWith(err.message, '"uuid" must be unique')
```
Page-: Why the change? Also it should probably be in a commit by itself if it's unrelated to the api key stuff

src/application.coffee
```diff
@@ -880,7 +882,7 @@ application.initialize = ->
 	listenToEvents()
 	getAndApplyDeviceConfig()
 	.then ->
-		knex('app').whereNot(markedForDeletion: true).select()
+		knex('app').whereNot(markedForDeletion: true).orWhereNull('markedForDeletion').select()
```
Page-: This change seems unrelated, what is it for?

src/application.coffee
```diff
@@ -635,7 +635,7 @@ getRemoteState = (uuid, apiKey) ->
 # TODO: Actually store and use app.environment and app.config separately
 parseEnvAndFormatRemoteApps = (remoteApps, uuid, apiKey) ->
 	appsWithEnv = _.mapValues remoteApps, (app, appId) ->
-		utils.extendEnvVars(app.environment, uuid, appId, app.name, app.commit)
+		utils.extendEnvVars(app.environment, uuid, apiKey, appId, app.name, app.commit)
```
Page-: You use a passed in `apiKey` here but then do a `utils.getConfig('deviceApiKey')` below, is that correct?

src/bootstrap.coffee
```diff
@@ -40,9 +40,10 @@ loadPreloadedApps = ->
 		fs.readFileAsync(appsPath, 'utf8')
 		.then(JSON.parse)
 		.map (app) ->
-			utils.extendEnvVars(app.env, userConfig.uuid, app.appId, app.name, app.commit)
+			utils.extendEnvVars(app.env, userConfig.uuid, userConfig.deviceApiKey, app.appId, app.name, app.commit)
 			.then (extendedEnv) ->
 				app.env = JSON.stringify(extendedEnv)
+				app.markedForDeletion = false
```
Page-: What is this about?

pcarranzav: This is part of the second commit - the commit message explains the purpose. We use markedForDeletion to do soft deletes and recently discovered that since preloaded apps don't have this internal field, they don't get loaded. The change above (orWhereNull...) is done for the same reason.

Page-: Oh, I thought both commits had the same message when I skimmed the commit page, turns out they just share the same first three words :(

src/bootstrap.coffee
pcarranzav: It's explained in the commit message - the API has had a bug where it added an extra dot at the end of the message, and I want to make the supervisor resilient to that bug. Added it inside this commit since it's a small change that doesn't affect behavior much, but I guess I can move it into a separate commit.
It is still *sort of* related since if this check fails then the supervisor will keep retrying the provisioning instead of doing the key echxange to register the deviceApiKey.

src/application.coffee
Page-: Ok, I read the commit message properly this time and it makes sense, however doesn't `whereNot..true` find cases where it's null anyway?

pcarranzav: Explained below in your other question - part of the second commit which fixes a different bug for preloaded apps.

src/application.coffee
pcarranzav: Yes, I guess I'll call it deviceApiKey here for consistency.

src/bootstrap.coffee
Page-: Ok, I'd prefer it to be a separate commit, but otherwise I'm good with it, or even with `_.includes`

src/bootstrap.coffee
pcarranzav: lol, sorry for the confusion :)

src/application.coffee
pcarranzav: whereNot true doesn't match null values :(

src/bootstrap.coffee
pcarranzav: Moved to a separate commit

PR #512, opened 2017-10-24
Fixes #511, add support for new supervisor image for IOT2000 board

### LINKED ISSUE #511: Add a new supervisor image for IOT2000 device (Due to the LOCK prefix issue of CPU)
Details can be found here: https://www.flowdock.com/app/rulemotion/resin-devices/threads/HCCFkB1yVR4lLdgahmzZFiHFX6I

nghiant2710:
To fix the CPU bug with the LOCK prefix on the IOT2000, we need to create a new supervisor image which is quark specific.

@telphan can you please update the device so it will use this supervisor image instead after this PR is merged.


gosuper/build.sh
```diff
@@ -12,6 +12,9 @@ case "$ARCH" in
 'i386')
 	export GOARCH=386
 ;;
+'iot2000')
+	export GOARCH=386
```
petrosagg: we should also set `GO386=387`

pcarranzav: @petrosagg we already set it in the Dockerfile https://github.com/resin-io/resin-supervisor/blob/master/Dockerfile#L96

pcarranzav:
@nghiant2710 the armv7hf build failed for a connectivity issue, so a retry should work. But the iot2000 build failed because of what I commented above that is missing in the Dockerfile.

Dockerfile
```diff
@@ -115,6 +115,8 @@ FROM resin/i386-node:6.5-slim as i386-node-base
 RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-start && chmod +x /usr/bin/cross-build-start \
```
petrosagg: @pcarranzav why are you doing those two things here and later you do `RUN [ "cross-build-start" ]`?

pcarranzav: @petrosagg since the Dockerfile is the same for all architectures, and the ARM builds need a cross-build-start, we add a cross-build-start to i386 and amd64 that is a noop. We can't conditionally run cross-build-start because checking for the condition would need to be done in qemu (discussed in https://github.com/resin-io/resin-supervisor/pull/470#discussion_r127535215 ) - but if you have any other ideas I'd love to hear them.

PR #518, opened 2017-10-31
Use balena 17.06 and resin-base v2.9.2 for the docker-in-docker supervisor

pcarranzav:
Change-Type: patch


tools/dind/config/services/docker.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
```
dfunckt: You may want to copy unit files verbatim from https://github.com/resin-os/balena/tree/17.06-resin/contrib/init/systemd to a folder, that you never modify them, and add overrides in a separate file like you do for aufs.

tools/dind/Dockerfile
```diff
@@ -20,11 +20,8 @@ ARG PASSWORDLESS_DROPBEAR=false
...
-RUN apt-key adv --keyserver hkp://pgp.mit.edu:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D \
-	&& echo deb https://apt.dockerproject.org/repo debian-${RELEASE_NAME} main > /etc/apt/sources.list.d/docker.list \
-	&& apt-get update || true \
-	&& apt-get --force-yes install docker-engine=${DOCKER_VERSION}-0~${RELEASE_NAME} \
-	&& rm -rf /var/lib/apt/lists/*
+RUN curl -sL "https://github.com/resin-os/balena/releases/download/${BALENA_VERSION}/balena-${BALENA_VERSION}-x86_64.tar.gz" \
+	| tar xzv -C /usr/bin --strip-components=1
 
 ENV DIND_COMMIT 3b5fac462d21ca164b3778647420016315289034
 RUN curl -sL https://raw.githubusercontent.com/docker/docker/${DIND_COMMIT}/hack/dind > /usr/bin/dind \
```
dfunckt: Shouldn’t you also copy dind from balena?

pcarranzav: Good point, fixed

tools/dind/config/services/docker.service
dfunckt: Yeah, I think it's fine to match meta-resin, and maybe I should direct this comment there as Docker's "built-in" unit files have a couple of settings that might be of interest.

PR #521, opened 2017-11-03
.dockerignore: Ignore the .git directory in submodules

pcarranzav:
I'm seeing that docker build cache is getting invalidated in some cases, and the only
possible culprit I can think of is the .git directory in the base image submodules.

This change adds a wildcard in the dockerignore file to ignore .git in all directories, not just the root.

Change-Type: patch

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
 .dockerignore
-.git
+**/.git
 node_modules
 .editorconfig
 tools


pcarranzav:
Closing this since it really has no benefit (.git is only a text file that always has the same content), and this PR will cause a bunch of cache misses all over again.

PR #524, opened 2017-11-09
Use uglifyjs-webpack-plugin ^1.0.1 to optimize the supervisor code

pcarranzav:
We've been using UglifyJS 0.4.6 (the webpack default) so far, but this doesn't support ES6 and some dependency
updates are starting to cause builds to break (e.g. request/request#2772, which also happens to break
my builds in the multicontainer branch).

Here we switch to the latest uglifyjs-webpack-plugin which is designed for ES2015 support.

Change-Type: patch

Edited: originally this PR meant to use babel-minify instead of the newer uglifyjs, hence Page's comments below.


pcarranzav:
@Page- webpack by default depends on uglifyjs-webpack-plugin ^0.4.6, so to be able to use the newer one we'd have to use a forked webpack. See: https://github.com/webpack/webpack/issues/3673

Page-:
I'm pretty sure you can do it by adding the 1.x version to package.json and using the plugin via
```
var UglifyJSPlugin = require('uglifyjs-webpack-plugin')
...
plugins: [
	...
	new UglifyJSPlugin()
]
```

Mainly I'm just wondering if uglifyjs will give better results, and it should also hopefully be closer to what we currently do

pcarranzav:
@Page- thanks, that worked!  I had tried it before and it hadn't worked but I guess I must've done something wrong.

PR #528, opened 2017-11-21
update-locking: updating the documentation with extra  information and tools

imrehg:
Connects-to: https://github.com/resin-io/docs/issues/638


docs/update-locking.md
```diff
@@ -2,7 +2,7 @@
 
 Locking updates means that the Supervisor will not be able to kill your application. This is meant to be used at critical sections of your code where you don't want to be interrupted, or to control that updates are only installed at certain times.
 
-In order to do this, users can create a file called `resin-updates.lock` that will prevent the Supervisor from killing and restarting the app. As any other lockfile, the Supervisor itself will create such file before killing the app, so you should only create it in exclusive mode. This means: only create the lockfile if it doesn't already exist. Several tools exist to simplify this process, for example [npm/lockfile](https://github.com/npm/lockfile).
+In order to do this, users take the lock on a file called `resin-updates.lock` that will prevent the Supervisor from killing and restarting the app. As any other lockfile, the Supervisor itself will create such file before killing the app, so you should only create it in exclusive mode. This means: only create the lockfile if it doesn't already exist. Several tools exist to simplify this process, for example [npm/lockfile](https://github.com/npm/lockfile).
```
LucianBuzzo: `users take the lock on a file` is very awkward wording and doesn't make much sense to me. Can this be improved?

LucianBuzzo: Is `Creating a lock` more semantic?

docs/update-locking.md
```diff
@@ -18,7 +18,29 @@ In supervisors >= v4.0.0 and any OS that is not Resin OS 1.X, the old lock locat
 
 ### Taking the lock
 
-Using the above-mentioned library, the lock can be acquired like in this CoffeeScript example:
+There are many different tools and libraries to provide proper lockfile functionality. Here are a few examples, and you can search online for others in other programming languages not mentioned here explicitly.
+
+#### Shell
+
+One simple way to take the lock is using [lockfile](https://linux.die.net/man/1/lockfile) (available for example in Debian from the `procmail` package) or [flock](https://linux.die.net/man/1/flock) (available for example in Debian from the `linux-utils` package). Some example use are:
```
LucianBuzzo: From my experiments, simply using `touch /tmp/resin/resin-updates/lock` worked without any additional configuration.

pcarranzav: That only works *sometimes*, and we want to discourage it. When using `touch`, the supervisor may have created the lock previously, in which case the app will be killed and the user will think the lock was taken.

docs/update-locking.md
pcarranzav: `Creating a lock` suggests that creating the file is sufficient, I think what @imrehg wanted to make clearer is that the creation needs to happen with the O_EXCL flag to take the lock rather than just create the file in any way

pcarranzav: So maybe what we could do is clarify thet `take the lock` means `create the lockfile with O_EXCL and O_CREAT flags set`

LucianBuzzo: @pcarranzav I think that would definitely be an improvement. I think it's important that we don't make *too many* assumptions about the user's knowledge level here, any additional information/pointers we can provide is good. We certainly need to find a balance between verbosity and readability, however, the expression `take the lock` doesn't seem to be widely used or commonly known.

imrehg:
@LucianBuzzo @pcarranzav thanks for the feedback :tada: I've rewritten this, and definitely for the better. Let me know if there's any other issue (factual, readability, clarity....)

docs/update-locking.md
```diff
@@ -2,7 +2,7 @@
 
 Locking updates means that the Supervisor will not be able to kill your application. This is meant to be used at critical sections of your code where you don't want to be interrupted, or to control that updates are only installed at certain times.
```
LucianBuzzo: NIT: `or to control` would read better as `or to ensure`

docs/update-locking.md
```diff
@@ -2,7 +2,7 @@
...
 
-In order to do this, users can create a file called `resin-updates.lock` that will prevent the Supervisor from killing and restarting the app. As any other lockfile, the Supervisor itself will create such file before killing the app, so you should only create it in exclusive mode. This means: only create the lockfile if it doesn't already exist. Several tools exist to simplify this process, for example [npm/lockfile](https://github.com/npm/lockfile).
+In order to do this, users create a lockfile called `resin-updates.lock` in a way that it has exclusive access, which will prevent the Supervisor from killing and restarting the app. As any other lockfile, the Supervisor itself will create such file before killing the app, so you should only create it in exclusive mode. This means: only create the lockfile if it doesn't already exist. The exclusive access is achived by opening the lockfile with the [O_EXCL and O_CREAT flags](https://linux.die.net/man/3/open), and several tools exist to simplify this process with examples given [below][#creating-the-lockfile].
```
LucianBuzzo: NIT
- `users create a lockfile` --> `users can create a lockfile`
- `As any other lockfile` --> `As with any other lockfile`
- `This means: only create the lockfile if it doesn't already exist.` --> `This means that the lockfile should only be created if it doesn't already exist`
- `The exclusive access is achived` --> `The exclusive access is achieved`
- `[below][#creating-the-lockfile]` This doesn't appear to render correctly when viewing the document.

docs/update-locking.md
```diff
@@ -16,9 +16,33 @@ assumption that updates are locked (see [issue #20](https://github.com/resin-io/
 
 In supervisors >= v4.0.0 and any OS that is not Resin OS 1.X, the old lock location is completely ignored.
 
-### Taking the lock
+### Creating the lockfile
 
-Using the above-mentioned library, the lock can be acquired like in this CoffeeScript example:
+There are many different tools and libraries to provide proper lockfile functionality. Here are a few examples, and you can search online for others in other programming languages not mentioned here explicitly.
```
LucianBuzzo: NIT
- `libraries to provide proper lockfile functionality` --> `libraries that provide proper lockfile functionality`
- Being able to search online for information is implicit, so we probably don't need to mention it, I'd reword this section as:
```
There are many different tools and libraries to provide proper lockfile functionality and a few common examples are shown below.
```

docs/update-locking.md
```diff
@@ -16,9 +16,33 @@ assumption that updates are locked (see [issue #20](https://github.com/resin-io/
...
+
+__Note:__ Just creating the lockfile (for example by `touch /tmp/resin/resin-updates.lock`) is not adequate to prevent updates. That file way won't have the exclusive access flag set, and thus does not provide reliable locking.
```
LucianBuzzo: Nice explanation!

NIT
- `for example by touch` --> `for example by using touch`
- `That file way won't` --> `A file created in this way won't`

docs/update-locking.md
```diff
@@ -16,9 +16,33 @@ assumption that updates are locked (see [issue #20](https://github.com/resin-io/
...
+
+#### Shell
+
+One simple way to take the lock is using [lockfile](https://linux.die.net/man/1/lockfile) (available for example in Debian from the `procmail` package) or [flock](https://linux.die.net/man/1/flock) (available for example in Debian from the `linux-utils` package). Some example use are:
```
LucianBuzzo: NIT
I would restructure this `Shell` section so that the lockfile example is immediately below the lockfile description and the same with the flock example

docs/update-locking.md
```diff
@@ -16,9 +16,33 @@ assumption that updates are locked (see [issue #20](https://github.com/resin-io/
...
+flock /tmp/resin/resin-updates.lock -c '... (command)'
+```
+
+For more examples and explanation of the functionality, check the links to the specific tools above.
+
+#### Javascript and Coffeescript
+
+Using the [`lockfile` library](https://www.npmjs.com/package/lockfile), the lock can be acquired like in this CoffeeScript example:
 ```coffeescript
 	lockFile = require 'lockfile'
```
LucianBuzzo: NIT can we reduce the indent of this example by one column?

docs/update-locking.md
```diff
@@ -35,7 +59,16 @@ Using the above-mentioned library, the lock can be acquired like in this CoffeeS
 			throw err if err?
 ```
 
-There are other libraries you can use in different languages, for example [this Python library](http://pythonhosted.org/lockfile/lockfile.html#examples).
+#### Python
+
+In Python you can use the for example the [`lockfile` library](http://pythonhosted.org/lockfile/lockfile.html#examples)
```
LucianBuzzo: NIT
`In Python you can use the for example the` --> `In Python you can use the`

docs/update-locking.md
```diff
@@ -35,7 +59,16 @@ Using the above-mentioned library, the lock can be acquired like in this CoffeeS
...
+```python
+from lockfile import LockFile
+lock = LockFile("/tmp/resin/resin-updates.lock")
+with lock:
+    print lock.path, 'is locked.'
+```
+Check the link for more examples, and other Python libraries that provide locking.
```
LucianBuzzo: NIT you can remove the comma from this sentence

docs/update-locking.md
imrehg: Outside of update scope, but updated because it's nicer :)

docs/update-locking.md
imrehg: Updated, thanks!

docs/update-locking.md
imrehg: Updated, thanks!

docs/update-locking.md
imrehg: Updated

docs/update-locking.md
imrehg: Done!

docs/update-locking.md
imrehg: Done

docs/update-locking.md
imrehg: Done

docs/update-locking.md
imrehg: BTW, the `below` will work in the docs, when it is pulled in, the headers are rendered with anchors; not so in just markdown preview.

docs/update-locking.md
imrehg: Done

imrehg:
@LucianBuzzo addressed all the NIT, and this PR is much better for it! Let me know if there's anything more that pops to mind.

docs/update-locking.md
```diff
@@ -16,26 +16,59 @@ assumption that updates are locked (see [issue #20](https://github.com/resin-io/
 
 In supervisors >= v4.0.0 and any OS that is not Resin OS 1.X, the old lock location is completely ignored.
 
-### Taking the lock
+### Creating the lockfile
 
-Using the above-mentioned library, the lock can be acquired like in this CoffeeScript example:
+There are many different tools and libraries that provide proper lockfile functionality. There are many different tools and libraries to provide proper lockfile functionality and a few common examples are shown below.
```
pcarranzav: `There are many different tools and libraries that provide proper lockfile functionality` is repeated

docs/update-locking.md
```diff
@@ -16,26 +16,59 @@ assumption that updates are locked (see [issue #20](https://github.com/resin-io/
...
+In Python you can use the [`lockfile` library](http://pythonhosted.org/lockfile/lockfile.html#examples)
+```python
+from lockfile import LockFile
+lock = LockFile("/tmp/resin/resin-updates.lock")
+with lock:
+    print lock.path, 'is locked.'
+```
+Check the link for more examples and other Python libraries that provide locking.
 
 ### Overriding the lock
```
pcarranzav: While we're at it, could you change the name of the configuration variable below to `RESIN_SUPERVISOR_OVERRIDE_LOCK` and make it clear that it's a configuration variable instead  of an env var?

The old `RESIN_OVERRIDE_LOCK` is actually deprecated (though for now it still works for backwards-compat).

docs/update-locking.md
```diff
@@ -1,8 +1,8 @@
...
 
-In order to do this, users can create a file called `resin-updates.lock` that will prevent the Supervisor from killing and restarting the app. As any other lockfile, the Supervisor itself will create such file before killing the app, so you should only create it in exclusive mode. This means: only create the lockfile if it doesn't already exist. Several tools exist to simplify this process, for example [npm/lockfile](https://github.com/npm/lockfile).
+In order to do this, users can create a lockfile called `resin-updates.lock` in a way that it has exclusive access, which will prevent the Supervisor from killing and restarting the app. As with any other lockfile, the Supervisor itself will create such file before killing the app, so you should only create it in exclusive mode. This means that the lockfile should only be created if it doesn't already exist. The exclusive access is achieved by opening the lockfile with the [O_EXCL and O_CREAT flags](https://linux.die.net/man/3/open), and several tools exist to simplify this process with examples given [below](#creating-the-lockfile).
```
LucianBuzzo: NIT `will create such file` --> `will create such a file`

docs/update-locking.md
imrehg: Duh, sorry!

docs/update-locking.md
imrehg: @pcarranzav the value is set to `1` or `true`? I see that the multicontainer demo uses true, and I wonder, which is better to say.

imrehg:
@pcarranzav @LucianBuzzo addressed, sorry for the language :/

PR #530, opened 2017-11-29
Add a /v1/healthy endpoint that fails if the supervisor is unhealthy,…

pcarranzav:
… and a HEALTHCHECK command to the Dockerfile that uses it

We add an endpoint to the supervisor API that checks the following conditions to determine whether the supervisor is healthy:
* That the update cycle has run fully, in a time that's less than twice the poll interval. Unless we're downloading an image, in which case
we assume it's healthy (otherwise we'd get into the issue of determining a reasonable timeout for the image download, which is already done in a configurable way with delta options and the like).
* That the current state report to the Resin API hasn't failed more than 3 times. Unless the device has no connectivity, or the connectivity check is disabled, in which case we don't know
if the report failed simply because there's no network.
* That the gosuper component is working (since we periodically hit its API to get the IP addresses, we mark it as not working if this API call fails).

We need this endpoint to be unauthenticated for the docker daemon to be able to hit it (though, as the rest of the API, it is protected with iptables rules).

Change-Type: minor


src/application.coffee
```diff
@@ -121,6 +134,10 @@ application = {}
 application.UpdatesLockedError = UpdatesLockedError
 application.localMode = false
 
+application.healthy = ->
+	timeSinceLastCycle = (process.hrtime()[0] - updateStatus.lastFullUpdateCycle) * 1000
```
petrosagg: Why not `Date.now()`?

src/application.coffee
```diff
@@ -860,6 +869,7 @@ application.update = update = (force, scheduled = false) ->
 				else
 					updateStatus.state = UPDATE_IDLE
 			device.updateState(status: 'Idle')
+			updateStatus.lastFullUpdateCycle = process.hrtime()[0]
```
petrosagg: Same here

src/application.coffee
pcarranzav: Because process.hrtime will not change if ntp changes the clock

Dockerfile
```diff
@@ -209,4 +209,6 @@ ENV CONFIG_MOUNT_POINT=/boot/config.json \
 	DEFAULT_PUBNUB_SUBSCRIBE_KEY=$DEFAULT_PUBNUB_SUBSCRIBE_KEY \
 	DEFAULT_MIXPANEL_TOKEN=$DEFAULT_MIXPANEL_TOKEN
 
+HEALTHCHECK CMD wget -qO- http://127.0.0.1:${LISTEN_PORT:-48484}/v1/healthy || exit 1
```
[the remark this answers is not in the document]
pcarranzav: I added some parameters, do they look like sensible values? With a 5 minute interval and 3 retries, it will take 15 minutes to detect an unhealthy supervisor, which to me feels like a reasonable time to give it a chance to recover / avoid false positives. And the start period of 1 minute should be enough for everything to start up... what do you think?

src/application.coffee
pcarranzav: Same as above -  avoids problems with clock changes due to ntp

PR #532, opened 2017-12-08
Avoid stopping the VPN until a remote target state has been fetched, …

pcarranzav:
…and retry applying config variables when they fail

Change-Type: patch


src/application.coffee
```diff
@@ -519,14 +525,22 @@ executeSpecialActionsAndHostConfig = (conf, oldConf) ->
 			return _.startsWith(key, device.hostConfigConfigVarPrefix)
 		if !_.isEqual(hostConfigVars, oldHostConfigVars)
 			device.setHostConfig(hostConfigVars, oldHostConfigVars, logSystemMessage)
+			.then (changedHostConfig) ->
+				needsReboot = changedHostConfig
+				if changedHostConfig
+					_.forEach oldHostConfigVars, (val, key) ->
```
petrosagg:
```coffeescript
for own val, key in oldHostConfigVars
```

PR #535, opened 2017-12-11
Dockerfile: improve caching by installing devDependencies before copy…

pcarranzav:
…ing the code

This should help avoid having to run the npm install for devDependencies (which are actually most of the modules)
every time there's a change in the code.

Change-Type: patch

(Probably not gonna be a *huge* improvement, but might save us a few minutes per build, if docker's caching works more or less well...)


pcarranzav:
Confirmed that this can improve the build times quite a bit.

PR #539, opened 2017-12-14
Do the webpack build in an amd64 image to improve build times

pcarranzav:
Change-Type: patch

(Experiment to see if build times improve)


pcarranzav:
Turns out this is a pretty good improvement. With no changes in package.json, build time is less than 6 minutes. And with changes in package.json (i.e. npm install not cached), build time is 15 minutes - around half what it used to be.

Dockerfile
```diff
@@ -144,18 +174,7 @@ COPY package.json /usr/src/app/
 # Install only the production modules that have C extensions
 # Save the modules and then install devDependencies for build
```
dfunckt: I think this line needs to be removed now as it seems devDeps are installed in `noed-build`.

PR #545, opened 2018-01-13
Multicontainer logger

pcarranzav:
Work in progress to implement logging to the Resin API on the multicontainer supervisor.

During an experimental phase, logs can be switched between Pubnub and Resin with a config variable.

Logs can have an imageId attached, which the resin API uses to infer the image_install that each log relates to. Logs sent to Pubnub, instead, just have a `c` to identify the serviceId that they relate to.

(Still untested)

### DIFF
--- a/src/api-binder.coffee
+++ b/src/api-binder.coffee
@@ -60,11 +60,11 @@ module.exports = class APIBinder
 		@_writeLock('getTarget').disposer (release) ->
 			release()
 
-	init: (startServices = true) ->
-		@config.getMany([ 'offlineMode', 'resinApiEndpoint', 'bootstrapRetryDelay', 'currentApiKey' ])
-		.then ({ offlineMode, resinApiEndpoint, bootstrapRetryDelay, currentApiKey }) =>
+	initClient: =>
+		@config.getMany([ 'offlineMode', 'resinApiEndpoint', 'currentApiKey'  ])
+		.then ({ offlineMode, resinApiEndpoint, currentApiKey }) =>
 			if offlineMode
-				console.log('Offline Mode is set, skipping API binder initialization')
+				console.log('Offline Mode is set, skipping API client initialization')
 				return
 			baseUrl = url.resolve(resinApiEndpoint, '/v4/')
 			passthrough = _.cloneDeep(requestOpts)
@@ -78,7 +78,12 @@ module.exports = class APIBinder
 				apiPrefix: baseUrlLegacy
 				passthrough: passthrough
 			@cachedResinApi = @resinApi.clone({}, cache: {})
-			if !startServices
+
+	start: =>
+		@config.getMany([ 'offlineMode', 'bootstrapRetryDelay' ])
+		.then ({ offlineMode, bootstrapRetryDelay }) =>
+			if offlineMode
+				console.log('Offline Mode is set, skipping API binder initialization')
 				return
 			console.log('Ensuring device is provisioned')
 			@provisionDevice()
@@ -244,6 +249,27 @@ module.exports = class APIBinder
 				id: id
 				body: updatedFields
 			.timeout(conf.apiTimeout)
+
+	_sendLogsRequest: (uuid, data) =>
+		reqBody = _.map(data, (msg) -> _.mapKeys(msg, (v, k) -> _.snakeCase(k)))
+		@config.get('resinApiEndpoint')
+		.then (resinApiEndpoint) =>
+			endpoint = url.resolve(resinApiEndpoint, "/device/v2/#{uuid}/logs")
+			requestParams = _.extend
+				method: 'POST'
+				url: endpoint
+				body: reqBody
+			, @cachedResinApi.passthrough
+
+			@cachedResinApi._request(requestParams)
+
+	logDependent: (uuid, msg) =>
+		@_sendLogsRequest(uuid, [ msg ])
+
+	logBatch: (messages) =>
+		@config.get('uuid')
+		.then (uuid) =>
+			@_sendLogsRequest(uuid, messages)
 
 	# Creates the necessary config vars in the API to match the current device state,
 	# without overwriting any variables that are already set.
--- a/src/logger.coffee
+++ b/src/logger.coffee
@@ -5,36 +5,91 @@ es = require 'event-stream'
 Lock = require 'rwlock'
 { checkTruthy } = require './lib/validation'
 
-LOG_PUBLISH_INTERVAL = 110
+class NativeLoggerBackend
+	constructor: ->
+		@logPublishInterval = 1000
+		@maxLogLinesPerBatch = 10
+		@maxQueueSize = 100
+		@maxLineLength = 300
+		@publishQueue = []
+		@intervalHandle = null
+		@publishEnabled = false
+		@offlineMode = false
+		@shouldStop = false
 
-# Pubnub's message size limit is 32KB (unclear on whether it's KB or actually KiB,
-# but we'll be conservative). So we limit a log message to 2 bytes less to account
-# for the [ and ] in the array.
-MAX_LOG_BYTE_SIZE = 30000
-MAX_MESSAGE_INDEX = 9
+	publishLoop: =>
+		startTime = process.hrtime()
+		Promise.try =>
+			return if @offlineMode or !@publishEnabled or @publishQueue.length is 0
+			currentBatch = @publishQueue.splice(0, @maxLogLinesPerBatch)
+			# Silently ignore errors sending logs (just like with pubnub)
+			@apiBinder.logBatch(currentBatch)
+		.catchReturn()
+		.then =>
+			elapsedTime = process.hrtime(startTime)
+			elapsedTimeMs = elapsedTime[0] * 1000 + elapsedTime[1] / 1e6
+			nextDelay = Math.max(@logPublishInterval - elapsedTimeMs, 0)
+			Promise.delay(nextDelay)
+		.then =>
+			if !@shouldStop
+				@publishLoop()
 
-module.exports = class Logger
-	constructor: ({ @eventTracker }) ->
+	_truncateToMaxLength: (message) =>
+		if message.message.length > @maxLineLength
+			message = _.clone(message)
+			message.message = _.truncate(message.message, { length: @maxLineLength, omission: '[...]' })
+		return message
+
+	logDependent: (msg, { uuid }) =>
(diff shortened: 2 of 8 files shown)


pcarranzav:
Still to do: switch the objects sent to the API to snake_case, and adjust rate limit with burst/spike handling.

src/proxyvisor.coffee
```diff
@@ -146,7 +146,7 @@ class ProxyvisorRouter
 			.then ([ device ]) =>
 				return res.status(404).send('Device not found') if !device?
 				return res.status(410).send('Device deleted') if device.markedForDeletion
-				@logger.log(m, { channel: "device-#{device.logs_channel}-logs" })
+				@logger.log(m, { uuid, channel: "device-#{device.logs_channel}-logs" })
```
LucianBuzzo: Why aren't we using `logDependent` here? Ideally we need dep devices to use the API logging.

src/logger.coffee
```diff
@@ -5,114 +5,219 @@ es = require 'event-stream'
...
-module.exports = class Logger
-	constructor: ({ @eventTracker }) ->
+	log: (message) =>
+		return if @offlineMode or !@publishEnabled or @publishQueue.length >= @maxQueueSize
+		if _.isString(message)
+			message = { message: message }
+
+		if message.message.length > 300
+			console.log('Ignored extra-long log line')
+			return
```
LucianBuzzo: We should truncate log lines instead of dropping them completely

LucianBuzzo: Or even better, split the rest of the log message onto a new line

src/proxyvisor.coffee
LucianBuzzo: I've tested this locally, and this needs to be changed to `@logger.logDependent(m, { uuid })` otherwise the log message will be sent using the gateway device's UUID

pcarranzav: Yeah, good catch, I forgot to change it when I introduced the logDependent function. Will fix.

src/logger.coffee
pcarranzav: Good point, for now I'll just truncate it to 300.

pcarranzav: Done

src/proxyvisor.coffee
pcarranzav: Done

src/proxyvisor.coffee
```diff
@@ -146,7 +146,7 @@ class ProxyvisorRouter
 			.then ([ device ]) =>
 				return res.status(404).send('Device not found') if !device?
 				return res.status(410).send('Device deleted') if device.markedForDeletion
-				@logger.log(m, { channel: "device-#{device.logs_channel}-logs" })
+				@logger.logDependent(m, { uuid, channel: "device-#{device.logs_channel}-logs" })
```
LucianBuzzo: The `channel` key is unused in `logDependent`, so this can just be `@logger.logDependent(m, { uuid })`

pcarranzav: Not really @LucianBuzzo - it's unused in the native logger's `logDependent` but used by the pubnub logger's `logDependent`, so we still need it.

LucianBuzzo: okay cool

src/proxyvisor.coffee
```diff
@@ -137,7 +137,7 @@ createProxyvisorRouter = (proxyvisor) ->
 		.then ([ device ]) ->
 			return res.status(404).send('Device not found') if !device?
 			return res.status(410).send('Device deleted') if device.markedForDeletion
-			proxyvisor.logger.log(m, { channel: "device-#{device.logs_channel}-logs" })
+			proxyvisor.logger.logDependent(m, { channel: "device-#{device.logs_channel}-logs" })
```
LucianBuzzo: This call will need to pass the device uuid in the second parameter else the log call will fail for native logging

pcarranzav: Ah, good catch, I lost this in the rebase

src/api-binder.coffee
```diff
@@ -245,6 +250,28 @@ module.exports = class APIBinder
 				body: updatedFields
 			.timeout(conf.apiTimeout)
 
+	_sendLogsRequest: (uuid, data) =>
+		reqBody = _.map data, (msg) ->
+			return _.mapKeys(msg, (v, k) -> _.snakeCase(k))
```
flesler: NIT: No need for `return` in CS no? Even if migrated to JS, you can use the `() => ( ... )` short-hand to avoid it

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+		@intervalHandle = null
+		@publishEnabled = false
+		@offlineMode = false
+		@shouldStop = false
+
+	publishLoop: =>
+		startTime = process.hrtime()
+		Promise.try =>
+			return if @offlineMode or !@publishEnabled or @publishQueue.length is 0
+			currentBatch = _.slice(@publishQueue, 0, @maxLogLinesPerBatch)
```
flesler: NIT: You can do both things in one step with `currentBatch = @publishQueue.splice(0, @maxLogLinesPerBatch)` and you'll be creating less arrays too

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+			@publishQueue = _.slice(@publishQueue, @maxLogLinesPerBatch)
+			# Silently ignore errors sending logs (just like with pubnub)
+			@apiBinder.logBatch(currentBatch)
+		.catchReturn()
+		.then =>
+			elapsedTime = process.hrtime(startTime)
+			elapsedTimeMs = elapsedTime[0] * 1000 + elapsedTime[1] / 1e6
+			nextDelay = Math.min(@logPublishInterval - elapsedTimeMs, 0)
```
flesler: This needs to be `Math.max()` insted

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+			Promise.delay(nextDelay)
+		.then =>
+			if @shouldStop
```
flesler: NIT: Seems like `if !@shouldStop @publishLoop()` would be more straightforward

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+			@publishLoop()
 
-# Pubnub's message size limit is 32KB (unclear on whether it's KB or actually KiB,
-# but we'll be conservative). So we limit a log message to 2 bytes less to account
-# for the [ and ] in the array.
-MAX_LOG_BYTE_SIZE = 30000
-MAX_MESSAGE_INDEX = 9
+	_truncateToMaxLength: (message) =>
+		if message.message.length > @maxLineLength
+			message = _.clone(message)
```
flesler: I'm not too aware of the context but, do you really need to clone it and preserve the original? Can't you just set the truncated message back into the original object?

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+			timestamp: Date.now()
+			message: ''
+		})
+
+		message = @_truncateToMaxLength(message)
+
+		if @publishQueue.length < @maxQueueSize - 1
+			@publishQueue.push(_.pick(message, [ 'message', 'timestamp', 'isSystem', 'isStderr', 'imageId' ]))
+		else
+			@publishQueue.push({
```
flesler: Won't you continue appending notice logs for each log that exceeds the limit?

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+				message: 'Warning! Some logs dropped due to high load'
+				timestamp: Date.now()
+				isSystem: true
```
flesler: We changed the log properties to underscore (in one of the arch calls). It's what the api expects, check the [API tests](https://github.com/resin-io/resin-api/blob/master/test/24_device_log.coffee#L34) when in doubt

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
+				isStderr: true
+			})
+
+	start: (opts) =>
+		console.log('Starting native logger')
+		@apiBinder = opts.apiBinder
+		@publishEnabled = checkTruthy(opts.enable)
+		@offlineMode = checkTruthy(opts.offlineMode)
+		@publishLoop()
+		return null
```
flesler: NIT: Won't `return` be enough?

src/logger.coffee
```diff
@@ -5,114 +5,237 @@ es = require 'event-stream'
...
-						@attached[containerId] = false
+						@attached[stdoutOrStderr][containerId] = false
 					.on 'end', =>
-						@attached[containerId] = false
+						@attached[stdoutOrStderr][containerId] = false
+
+	attach: (docker, containerId, serviceInfo) =>
+		Promise.using @lock(containerId), =>
+			@_attachStream(docker, 'stdout', containerId, serviceInfo)
+			.then =>
```
flesler: Subjective NIT: Can de-indent and keep the promise chain flat

src/logger.coffee
pcarranzav: We can, but I prefer not to - it'd be an unexpected side effect of the function, and this might cause bugs if a caller doesn't expect this in the future

src/logger.coffee
pcarranzav: It will, but: http://bluebirdjs.com/docs/warning-explanations.html#warning-a-promise-was-created-in-a-handler-but-was-not-returned-from-it

src/logger.coffee
pcarranzav: The return in line 50 prevents that

src/logger.coffee
pcarranzav: Yes, but internally we use camel case everywhere. The APIBinder converts to snake in `_sendLogsRequest`

src/api-binder.coffee
pcarranzav: Done

src/logger.coffee
pcarranzav: Nice one, done

src/logger.coffee
pcarranzav: Good catch!

src/logger.coffee
pcarranzav: Done

src/logger.coffee
pcarranzav: Done

PR #547, opened 2018-01-16
Implement an API for proxy and hostname configuration, and centralize…

pcarranzav:
… management of config.json

Change-Type: minor

I had to change bootstrap.coffee to use a centralized interface to modify config.json, so that we can also modify it from host-config.coffee without risking conflicts/races.

Marking as WIP because it's untested.


gosuper/gosuper/main.go
```diff
@@ -26,6 +26,7 @@ func setupApi(router *mux.Router) {
 	apiv1.HandleFunc("/shutdown", ShutdownHandler).Methods("POST")
 	apiv1.HandleFunc("/vpncontrol", VPNControl).Methods("POST")
 	apiv1.HandleFunc("/set-log-to-display", LogToDisplayControl).Methods("POST")
+	apiv1.HandleFunc("/restart-service", RestartServiceHandler).Methods("POST")
```
dfunckt: I think this should be RestartService

pcarranzav: Ah, good catch, fixed!

PR #548, opened 2018-01-20
WIP: First endpoints for multicontainer supervisor API

pcarranzav:
Endpoint definitions are still TBD.

Working on making the deviceState component capable of applying intermediate volatile target states so that complex actions (like starting all services in an app considering dependencies) can be run using that (relevant for purge and app restart).


pcarranzav:
Merging into multicontainer, I've done some testing in dind and the three endpoints worked fine as far as I could tell, even when called in the middle of an update (which is the point of using applyIntermediateTarget for complex actions).

PR #550, opened 2018-02-02
Fix API payload when provisioning a dependent device

LucianBuzzo:


pcarranzav:
Good catch @LucianBuzzo! I think I was originally thinking of using `@resinApiLegacy` to do the provisioning, but this is clearly better.

PR #552, opened 2018-02-05
Fix state endpoint payload

LucianBuzzo:
The data previously sent to the API was in the form of
```
obj = {
  local: apps,
  dependent
};
```

where as according to the API
( https://github.com/resin-io/resin-api/blob/multicontainer2/src/routes/devices.coffee#L964
)
it should be in the form:

```
obj = {
  local: {apps, dependent}
};
```

This also fixes this sentry error https://sentry.io/resinio/api-multi/issues/443300912/events/13511247699/

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -400,7 +400,7 @@ module.exports = class ApplicationManager extends EventEmitter
 						else
 							console.log('Ignoring legacy dependent image', image)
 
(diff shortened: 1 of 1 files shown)


LucianBuzzo:
It appears that it may actually be the state endpoint on the API that needs to be changed, not the supervisor

PR #554, opened 2018-02-16
Small updates for standardization

zwalchuk:
A few small updates to standardize terminology and header with the rest of the docs.

Change-type: patch
Signed-off-by: Zach Walchuk zach@resin.io


pcarranzav:
@zwalchuk could you rebase from master so that the Circle builds can run? After that I can merge it.

PR #557, opened 2018-02-21
IPv6 support

petrosagg:
In order for the supervisor to work correctly in IPv6 environments it needs to be RFC 3484 compliant. This RFC is implemented in the operating system and when a hostname resolution is done the results are returned to the application in the order of highest likelihood of succeeding.

NodeJS doesn't respect that order and re-sorts the records. Furthermore, it will only try the first entry and will not continue attempting the rest on failure.

To fix this we need to always pass the `verbatim: true` flag when resolving that was introduced in nodejs 8.5.0 and will become the default in the next major version. See https://github.com/nodejs/node/pull/14731 and https://github.com/nodejs/node/pull/17793


pcarranzav:
Superseded by #566

PR #558, opened 2018-02-23
Update docker-delta to 2.0.4

dfunckt:
This brings in a fix for an edge case where rsync would exit before we had a chance to register event listeners.

Depends on https://github.com/resin-io/node-docker-delta/pull/23

Change-Type: patch


dfunckt:
Thanks @pcarranzav !

PR #563, opened 2018-03-02
Fix the way commands and entrypoints in string form are parsed

pcarranzav:
Turns out shell-quote's parse function also replaces environment variables, which we don't want in this case. So we escape dollar signs
before calling shell-quote's parse function.

Also shell-quote takes some characters like `>` and globs and returns an object - so we return those objects to string form.

This should still be simpler/better than writing our own shlex.split, I hope...

Change-Type: patch

### DIFF
--- a/src/compose/service.coffee
+++ b/src/compose/service.coffee
@@ -28,17 +28,29 @@ createRestartPolicy = (name) ->
 		name = 'always'
 	return { Name: name, MaximumRetryCount: 0 }
 
+processCommandStr = (s) ->
+	# Escape dollars
+	s.replace(/(\$)/g, '\\$1')
+
+processCommandParsedArrayElement = (arg) ->
+	if _.isObject(arg)
+		if arg.op == 'glob'
+			return arg.pattern
+		return arg.op
+	return arg
+
+ensureCommandIsArray = (s) ->
+	if _.isString(s)
+		s = _.map(parseCommand(processCommandStr(s)), processCommandParsedArrayElement)
+	return s
+
 getCommand = (service, imageInfo) ->
 	cmd = service.command ? imageInfo?.Config?.Cmd ? null
-	if _.isString(cmd)
-		cmd = parseCommand(cmd)
-	return cmd
+	return ensureCommandIsArray(cmd)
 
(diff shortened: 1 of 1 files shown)


pcarranzav:
TODO: Also find a way to avoid shell-quote replacing ops like `>`. Consider PRing an option to shell-quote.

LucianBuzzo:
@pcarranzav What bug/issue does this fix? Are we trying to use the `$` character in shell-quote without it being escaped?

CameronDiver:
@pcarranzav I'm with Lucian here, I'm not 100% on what this PR is doing.

pcarranzav:
@LucianBuzzo @CameronDiver this is about how the service command and entrypoint get parsed by the supervisor. Say the user specifies this:
```
services:
  foo:
    image: alpine
    environment:
      foo: "bar"
    command: sh -c "echo $foo" > somefile.
```

We need to convert this to the array format that docker expects, so we should make command equal `[ "sh", "-c", "echo $foo", ">", "somefile" ]`.
But shell-quote (the package we use to split the string into an array, only one I could find so far) will replace $foo for the value of the environment variable $foo in the supervisor process, which makes no sense and will probably be an empty string. And it will also return `{ op: ">" }` instead of the plain `>`. So the command will equal `[ "sh", "-c", "echo ", { op: ">" }, "somefile" ]` instead of what we wanted.

This PR makes two changes then: it escapes the `$` so that shell-quote doesn't substitute the env vars there, and afterwards replaces the `{op}` objects to get back the original special character (`>` in this case).

src/compose/service.coffee
```diff
@@ -28,17 +28,23 @@ createRestartPolicy = (name) ->
 		name = 'always'
 	return { Name: name, MaximumRetryCount: 0 }
 
+ensureCommandIsArray = (s) ->
+	if _.isString(s)
+		s = _.map parseCommand(s.replace(/(\$)/g, '\\$1')), (arg) ->
```
CameronDiver: I think this would be easier if you moved this out to a utility function like:

pcarranzav: *suspense* :P

PR #574, opened 2018-03-09
Fix bug in require for migrations for legacy preload

CameronDiver:
Change-type: patch
Connects-to: #573 
Closes: #573 
---- Autogenerated Waffleboard Connection: Connects to #573


pcarranzav:
@CameronDiver I think it's Closes and not  Connects-to (supervisor issues are closed when a relevant fix is merged)

PR #586, opened 2018-03-14
Fix using hostname from host when network mode is host, and fix working_dir comparison

pcarranzav:


src/compose/service.coffee
```diff
@@ -459,9 +462,8 @@ module.exports = class Service
 		hostname = container.Config.Hostname
 		# A hostname equal to the first part of the container ID actually
 		# means no hostname was specified
-		if hostname == container.Id.substr(0, 12) or
-			(networkMode == 'host' and hostname == os.hostname())
-				hostname = ''
+		if hostname == container.Id.substr(0, 12)
```
Page-: This could be 
```coffee
if hostname.length is 12 and container.Id.startsWith(hostname)
```

pcarranzav:
Reportedly this didn't work... testing now

PR #587, opened 2018-03-14
Fix the name of the user/provisioning api key so we don't reprovision old devices

Page-:
Change-type: patch


pcarranzav:
Replaced by #588 - closing

PR #592, opened 2018-03-16
Remove the last few parts of Go

pcarranzav:
I recommend reviewing commit-by-commit since the second one just removes all of the Go files


pcarranzav:
Ran some tests on dind, and after fixing a few typos it seems to work fine...

CameronDiver:
Tested on resinOS version 2.12.3+rev1

PR #595, opened 2018-03-19
Trim whitespace from the hostname file on the host before passing it …

### LINKED ISSUE #594: Container hostname with host networking has an additional newline
Need to trim /etc/hostname before passing it to docker.
Connects-to: https://github.com/resin-io/hq/issues/1251

pcarranzav:
…as the hostname for containers with host network

Connects-to: https://github.com/resin-io/hq/issues/1251
Closes #594
Change-Type: patch


src/application-manager.coffee
```diff
@@ -799,7 +799,7 @@ module.exports = class ApplicationManager extends EventEmitter
 				firmware: pathExistsOnHost('/lib/firmware')
 				modules: pathExistsOnHost('/lib/modules')
 			})
-			fs.readFileAsync(path.join(constants.rootMountPoint, '/etc/hostname')).then(String)
+			fs.readFileAsync(path.join(constants.rootMountPoint, '/etc/hostname')).then(String).then(_.trim)
```
Page-: This would probably be better as
```
fs.readFileAsync(path.join(constants.rootMountPoint, '/etc/hostname'), 'utf8').then(_.trim)

Comment (pcarranzav) [reply src/application-manager.coffee]: Good point, done

File: src/application-manager.coffee [hunk 2]
```diff
@@ -799,7 +799,7 @@ module.exports = class ApplicationManager extends EventEmitter
 				firmware: pathExistsOnHost('/lib/firmware')
 				modules: pathExistsOnHost('/lib/modules')
 			})
-			fs.readFileAsync(path.join(constants.rootMountPoint, '/etc/hostname')).then(String)
+			fs.readFileAsync(path.join(constants.rootMountPoint, '/etc/hostname'), 'utf-8').then(_.trim)
```
Page-: I don't know if `utf-8` works or not, I've always used `utf8` and it's what the nodejs docs use everywhere: https://nodejs.org/docs/latest-v6.x/api/fs.html

pcarranzav: Oh, didn't realize that! utf-8 does work, but I changed it to utf8

PR #605, opened 2018-03-26
Handle incorrectly parsed env vars from docker inspect

CameronDiver:
Change-type: patch
Closes: #604


src/lib/conversions.coffee
```diff
@@ -4,5 +4,9 @@ exports.envArrayToObject = (env) ->
 	# env is an array of strings that say 'key=value'
 	toPair = (keyVal) ->
 		m = keyVal.match(/^([^=]+)=(.*)$/)
+		if !m?
+			console.log("WARNING: Could not correctly parse env var #{keyVal}. " +
+				'Please fix this var and recreate the container.')
+			return null
 		return m[1..]
-	_.fromPairs(_.map(env, toPair))
+	_(env).map(toPair).filter(([_, v]) -> v?).fromPairs().value()
```
dfunckt: The following is equivalent: `_(env).map(toPair).filter().fromPairs().value()`.

CameronDiver: It doesn't seem to be:

dfunckt: Ah, yeah you're right -- elements are `[ key, value ]` tuples so of course this won't work.

PR #608, opened 2018-03-29
Respond to reboot and shutdown endpoints with a success object

CameronDiver:
This fixes a bug where `.json` would be called on a timer, which has circular references. Instead just return "OK" as we wouldn't know if the call failed before responding (due to the timer).

Change-type: patch
Closes: #607


src/device-state.coffee
```diff
@@ -369,8 +369,14 @@ module.exports = class DeviceState extends EventEmitter
 				switch step.action
 					when 'reboot'
 						@reboot(force, skipLock)
+						# There isn't really a way that these methods can fail,
+						# and if they do, we wouldn't know about it until after
+						# the response has been sent back to the API. Just return
+						# "OK" for this and the below action
+						Data: 'OK', Error: null
```
dfunckt: This changes the semantics and `@reboot(...)` isn't waited on if it returns a promise (which I'm pretty sure it does).

CameronDiver: A good point, it returns the result which is a timer, but I didn't notice the other stuff going on before that, good catch!

Issue #610, opened 2018-03-30
RESIN_HOST_LOG_TO_DISPLAY can cause reboot loop


CameronDiver:
Setting the LOG_TO_DISPLAY varaible can cause a reboot loop, probably because the unit is missing in production.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_ae6b)

CameronDiver:
The temporary fix for this is to toggle the variable back to `enabled`.

PR #612, opened 2018-04-02
Remove support for broken RESIN_HOST_LOG_TO_DISPLAY variable

CameronDiver:
Channge-type: patch
Closes: #610


src/device-config.coffee
```diff
@@ -259,7 +259,16 @@ module.exports = class DeviceConfig
 		systemd.serviceActiveState(logToDisplayServiceName)
 		.then (activeState) ->
 			return activeState not in [ 'inactive', 'deactivating' ]
-		.catchReturn(UnitNotLoadedError, null)
+		.catch (e) ->
+			# Depending on whether the error is a 'unit not loaded' error, or
+			# otherwise, we need to return the correct value
+			if e[0] is "Unit #{logToDisplayServiceName}.service not loaded."
```
Page-: Why did you stop using the `UnitNotLoadedError` predicate?  It looks like this is the only place using it

CameronDiver: Because it looks like that was exclusively from the go supervisor. Either way, the error that now comes through, `typeof` reports object, and none of the standard error fields (message, stack etc) are present. From what I can tell, this comes from the node dbus library, so we can fix it properly coming up, but this is quite a serious bug, so I'd like to fix it as-is for now.

CameronDiver: Looks like it comes from here: https://github.com/sidorares/dbus-native/blob/7ddc7ebd0b6b988177733a8ac4b2f0aeff70cc73/lib/bus.js#L105 I'd definitely rather not rely on parsing output to detect errors though, I just can't think of a better method just now.

src/lib/errors.coffee
```diff
@@ -3,4 +3,3 @@
 exports.NotFoundError = (err) -> checkInt(err.statusCode) is 404
 exports.ENOENT = (err) -> err.code is 'ENOENT'
 exports.EEXIST = (err) -> err.code is 'EEXIST'
-exports.UnitNotLoadedError = (err) -> err[0]?.endsWith?('not loaded.')
```
pcarranzav: We're still using this in getVPNEnabled

lekkas:
Connects to https://github.com/resin-io/hq/issues/1314

PR #613, opened 2018-04-02
Fix falling back to delta from scratch losing auth info.

Page-:
Change-type: patch

Front logo [Front conversations](https://app.frontapp.com/open/top_ak9v)


pcarranzav:
Connects to https://github.com/resin-io/hq/issues/1259

Issue #603, opened 2018-03-23
shm_size values are not respected by the supervisor


CameronDiver:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_ae6r)

imrehg:
@CameronDiver any progress on this? (just reviewing open tickets)

PR #616, opened 2018-04-04
Fix regex for parsing memory numbers

CameronDiver:
It now allows a trailing `b`, as the docker-compose docs specify.

In addition the regex now specifies a case-insensitive flag, to catch
both upper and lower case memory numbers (the rest of the function
supported these already).

Change-type: patch
Closes: #603


LucianBuzzo:
@CameronDiver Can you post a link to the docker-compose docs that you reference? I'm struggling to find them

LucianBuzzo:
@CameronDiver for my own sanity I've put together a unit test PR into this branch that covers the function changes.

src/compose/service.coffee
```diff
@@ -14,7 +14,7 @@ validRestartPolicies = [ 'no', 'always', 'on-failure', 'unless-stopped' ]
 PORTS_REGEX = /^(?:(?:([a-fA-F\d.:]+):)?([\d]*)(?:-([\d]+))?:)?([\d]+)(?:-([\d]+))?(?:\/(udp|tcp))?$/
 
 parseMemoryNumber = (numAsString, defaultVal) ->
-	m = numAsString?.toString().match(/^([0-9]+)([bkmg]?)$/)
+	m = numAsString?.toString().match(/^([0-9]+)([bkmg])?b?$/i)
```
LucianBuzzo: NIT: This regex allows `bb` but I don't think it actually matters! The function should still behave as if it is given a single `b`

circle.yml
```diff
@@ -32,6 +32,10 @@ defaults: &defaults
             git submodule update --init --recursive
             git clean -fxd base-image
             git submodule foreach --recursive git clean -fxd
+    - run:
+      name: Run test suite
```
LucianBuzzo: I think this needs 4 space indentation

PR #619, opened 2018-04-04
Push version-tagged images from the 6.x branch

pcarranzav:
This will allow us to push critical fixes for devices that are not ready to support the 7.x supervisor line.

Change-Type: patch


pcarranzav:
Closing since it seems we don't need it for now, we can reopen if needed.

PR #620, opened 2018-04-06
WIP: Add /v2/device/target-state endpoints to allow setting a target state…

pcarranzav:
…, and make the API listen on all interfaces, when the device is in local mode

Change-Type: minor

This is an initial -untested- attempt, and we still need to decide on what the interface looks like

Connects-to: https://github.com/resin-io/hq/issues/1285


pcarranzav:
Replaced by #750

Issue #627, opened 2018-04-18
Failing deltas are not handled


afitzek:
The rejection of the promise here: https://github.com/resin-io/resin-supervisor/blob/master/src/lib/docker-utils.coffee#L17 is not called, when the request is aborted here: https://github.com/resin-io/resin-supervisor/blob/master/src/lib/docker-utils.coffee#L27 because the resumeable deltas is not emiting the error event on an abort: https://github.com/resin-io-modules/resumable-request/blob/master/index.coffee#L201

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_b3qb)

PR #628, opened 2018-04-18
Properly propagate delta failure

dfunckt:
The supervisor would never be notified if the delta stream produced an error and hang there forever. This makes sure to abort the download and reject the outer promise correctly.

Change-Type: patch
Closes: #627


dfunckt:
Ah, thanks, I'll amend the commit and edit the description.

PR #630, opened 2018-04-18
Use resinos-in-container for the development dind supervisor

pcarranzav:
Also includes some updates to the README.


Makefile
```diff
@@ -98,66 +108,32 @@ endif
...
-		&& echo "https_proxy=$(rt_https_proxy)" >> config/localenv; \
-	fi; \
-	if [ -n "$(rt_no_proxy)" ]; then \
-		echo "no_proxy=$(rt_no_proxy)" >> config/localenv; \
-	fi \
-	&& docker run -d --name $(CONTAINER_NAME) --privileged ${SUPERVISOR_DIND_MOUNTS} $(IMAGE)
+	&& ./resinos-in-container/resinos-in-container.sh \
+		--detach \
+		--config "$$(pwd)/$(CONFIG_FILENAME)" \
+		--image resin/resinos:2.12.5_rev1-intel-nuc \
```
dfunckt: Should that be configurable via a script argument? Or at least moved to a file-level constant so it's discoverable and easily updated?

src/app.coffee
```diff
@@ -13,5 +13,6 @@ require('log-timestamp')
 
 Supervisor = require './supervisor'
 
+console.log('MOUNTED!!!')
```
dfunckt: Forgotten? :P

README.md
```diff
@@ -4,18 +4,16 @@ Join our online chat at [![Gitter chat](https://badges.gitter.im/resin-io/chat.p
...
 ## Running a supervisor locally
 
 This process will allow you to run a development instance of the supervisor on your local computer. It is not recommended for production scenarios, but allows someone developing on the supervisor to test changes quickly.
-The supervisor is run inside a Docker-in-Docker container that roughly emulates its environment when running on a resinOS device.
+The supervisor is run inside a resinOS instance running in a container, so effectively it's a Docker-in-Docker instance (or more precisely, [balena](https://github.com/resin-os/balena)-in-Docker).
 
 ### Set up `config.json`
 
-Add a `tools/dind/config.json` file from a staging or production device image. It should be configured for an x86 or amd64 device type (e.g. an Intel Nuc) so that you can push apps to it and they run properly on your computer.
+Add a `tools/dind/config.json` file from a staging or production device image. It should be configured for an x86 or amd64 device type (e.g. an Intel NUC) so that you can push apps to it and they run properly on your computer.
```
zwalchuk: Can we clarify the steps here? Something like "Copy the `config.json` file from the `resin-boot` partition of a staging or production device image into the`tools/dind/`  directory. You can also download this file directly from the dashboard. Make sure you choose an x86 or amd64 device type, such as the Intel NUC, ..."

README.md
```diff
@@ -75,42 +73,17 @@ This will mount the ./dist folder into the supervisor container and build the co
 ### Testing with preloaded apps
 To test preloaded apps, add a `tools/dind/apps.json` file according to the preloaded apps spec.
```
zwalchuk: The wording here needs to be a little different if we aren't showing an example of what the file should look like. "To test preloaded apps, run `resin preload`...and copy the `apps.json` file from the `resin-data` partition of the preloaded image to `tools/dind/`."

README.md
```diff
@@ -151,64 +125,22 @@ Build the supervisor with a specific tag, and for a specific architecture, like
...
-
-Once it's running, you can run the test with:
-```bash
-make ARCH=amd64 test-integration
+npm test
 ```
 
-The tests will fail if the supervisor API is down - bear in mind that the supervisor image takes a while to start the actual supervisor program, so you might have to wait a few minutes between running the supervisor and testing it.
-The test expects the supervisor to be already running the application (so that the app is already on the SQLite database), so check the dashboard to see if the app has already downloaded.
+You'll need Node.js 6 installed, and having run `npm install` first.
```
zwalchuk: "You'll need Node.js 6 installed and to have run `npm install` first." 

Is Node.js 6 an exact or minimum requirement?

src/app.coffee
pcarranzav: Oops!

Makefile
pcarranzav: Good point, done

README.md
pcarranzav: There were steps below but I've rearranged this to hopefully make it clearer.

README.md
pcarranzav: Done

README.md
pcarranzav: Exact - I've clarified this a bit more.

PR #632, opened 2018-04-25
Add support for Balena deltas

dfunckt:
Resin’s delta server supports Balena deltas as version 3 deltas. This commit adds support for triggering delta generation for Balena deltas, and applying them locally to the device via a simple pull.

The delta version to use when updating has been abstracted away as an env var that is user-defined. The default value is still instructing use of rsync deltas (v2).

Change-Type: minor


src/config.coffee
```diff
@@ -145,6 +146,7 @@ module.exports = class Config extends EventEmitter
 			deltaApplyTimeout: { source: 'db', mutable: true, default: '' }
 			deltaRetryCount: { source: 'db', mutable: true, default: '30' }
 			deltaRetryInterval: { source: 'db', mutable: true, default: '10000' }
+			deltaVersion: { source: 'db', mutable: true, default: 2 }
```
pcarranzav: If source is 'db', the default should be a string (and then keep in mind that the value will always be a string).
Might actually be good to add a comment with this rule on top of the schema...

src/device-config.coffee
```diff
@@ -50,6 +50,7 @@ module.exports = class DeviceConfig
 			deltaApplyTimeout: { envVarName: 'RESIN_SUPERVISOR_DELTA_APPLY_TIMEOUT', varType: 'int', defaultValue: '' }
 			deltaRetryCount: { envVarName: 'RESIN_SUPERVISOR_DELTA_RETRY_COUNT', varType: 'int', defaultValue: '30' }
 			deltaRetryInterval: { envVarName: 'RESIN_SUPERVISOR_DELTA_RETRY_INTERVAL', varType: 'int', defaultValue: '10000' }
+			deltaVersion: { envVarName: 'RESIN_SUPERVISOR_DELTA_VERSION', varType: 'int', defaultValue: 2 }
```
pcarranzav: Same here: config variables are always strings, so the default should be a string too

src/lib/docker-utils.coffee
```diff
@@ -74,34 +92,48 @@ module.exports = class DockerUtils extends DockerToolbelt
...
 						sendImmediately: true
 				new Promise (resolve, reject) ->
-					request.get("#{deltaEndpoint}/api/v2/delta?src=#{deltaSource}&dest=#{imgDest}", opts)
+					request.get("#{deltaEndpoint}/api/v#{version}/delta?src=#{deltaSource}&dest=#{imgDest}", opts)
 					.on 'response', (res) ->
-						res.resume() # discard response body -- we only care about response headers
 						if res.statusCode in [ 502, 504 ]
 							reject(new Error('Delta server is still processing the delta, will retry'))
-						else if not (300 <= res.statusCode < 400 and res.headers['location']?)
+						else if res.statusCode is 200 and not (300 <= res.statusCode < 400 and res.headers['location']?)
```
pcarranzav: how would this condition happen? Did you mean `res.statusCode isnt 200`?

src/lib/docker-utils.coffee
```diff
@@ -74,34 +92,48 @@ module.exports = class DockerUtils extends DockerToolbelt
...
 							reject(new Error("Got #{res.statusCode} when requesting image from delta server."))
+						else if res.statusCode is 200
+							# v3
+							try
+								name = JSON.parse(res.body).name
+								resolve(applyBalenaDelta(deltaUrl, token, onProgress, log))
+							catch e
+								reject(e)
 						else
```
pcarranzav: Would it be better to check explicitly for the  version rather than depending on the server's response?

src/lib/docker-utils.coffee
```diff
@@ -21,13 +22,23 @@ applyDelta = (imgSrc, deltaUrl, applyTimeout, opts, onProgress) ->
...
+					log("Delta stream emitted error: #{err}")
+					req.abort(err)
 					reject(err)
 
+applyBalenaDelta = (deltaImg, token, onProgress, log) ->
+	log('Applying balena delta...')
+	if token?
+		log('Using registry auth token')
+		auth = { authconfig: registrytoken: token }
+	@dockerProgress.pull(deltaImg, onProgress, auth)
```
pcarranzav: Now that deltas have tags, I wonder if `_removeImageIfNotNeeded` in src/compose/images.coffee will keep working well... I think it should be fine, but we'll have to test that the supervisor cleans up old images well when doing balena delta updates.

src/config.coffee
```diff
@@ -131,6 +132,7 @@ module.exports = class Config extends EventEmitter
 			extendedEnvOptions: { source: 'func' }
 			fetchOptions: { source: 'func' }
 
+			# NOTE: all 'db' values are stored and loaded as *strings*
```
pcarranzav: Thanks for adding this :)

dfunckt:
@pcarranzav, ah, it does, I forgot to mention it. I understand you mean what happens to the source image after a balena delta has been applied, right? I've found there's always one image (per service) at any given time.

pcarranzav:
Cool, yeah, I wondered about that and whether images may be incorrectly cleaned up, mostly because the logic in `compose/images.coffee` to achieve good cleanup is rather complex. But if you tested and found that, then this is good to go imo.

dfunckt:
I didn't test with multiple services sharing the image though I don't think it'd make a difference -- they're differentiated by the tag. I will test though.

pcarranzav:
Yeah, it'd definitely be good to test that - it can make a difference because of how the cleanup works

PR #638, opened 2018-04-30
More typescript conversions

CameronDiver:


src/lib/constants.ts
```diff
@@ -0,0 +1,47 @@
...
+const supervisorNetworkInterface = 'supervisor0';
+
+export = {
+	rootMountPoint,
+	databasePath: checkString(process.env.DATABASE_PATH) || '/data/database.sqlite',
+	dockerSocket: process.env.DOCKER_SOCKET || '/var/run/docker.sock',
+	supervisorImage: checkString(process.env.SUPERVISOR_IMAGE) || 'resin/rpi-supervisor',
+	ledFile: checkString(process.env.LED_FILE) || '/sys/class/leds/led0/brightness',
+	// Only used in development
+	forceSecret: {
```
pcarranzav: This can be removed (it stopped being used with #482)

src/lib/constants.ts
```diff
@@ -0,0 +1,47 @@
...
+	configJsonPathOnHost: checkString(process.env.CONFIG_JSON_PATH),
+	proxyvisorHookReceiver:
+		checkString(process.env.RESIN_PROXYVISOR_HOOK_RECEIVER) || 'http://0.0.0.0:1337',
+	apiEndpointFromEnv: checkString(process.env.API_ENDPOINT),
+	configJsonNonAtomicPath: '/boot/config.json',
+	defaultPubnubSubscribeKey: process.env.DEFAULT_PUBNUB_SUBSCRIBE_KEY,
+	defaultPubnubPublishKey: process.env.DEFAULT_PUBNUB_PUBLISH_KEY,
+	defaultMixpanelToken: process.env.DEFAULT_MIXPANEL_TOKEN,
+	supervisorNetworkInterface: supervisorNetworkInterface,
+	allowedInterfaces: [ 'resin-vpn', 'tun0', 'docker0', 'lo', supervisorNetworkInterface ],
```
pcarranzav: I just realized that this is outdated,we should probably think about what interfaces the supervisor listens on... I think it might only need to be resin-vpn, lo, and the supervisorNetworkInterface. But it might be better left for a separate PR.

src/lib/constants.ts
CameronDiver: Sure, will do it in this PR. In general I'm trying to change as little as possible, but given this is an easy win I can do that.

src/lib/constants.ts
CameronDiver: Yeah I'll save it for another PR, making an issue now.

CameronDiver: https://github.com/resin-io/resin-supervisor/issues/639

src/lib/constants.coffee
```diff
@@ -1,43 +0,0 @@
...
-	allowedInterfaces: ['resin-vpn', 'tun0', 'docker0', 'lo', supervisorNetworkInterface]
-	appsJsonPath: process.env.APPS_JSON_PATH ? '/boot/apps.json'
-	ipAddressUpdateInterval: 30 * 1000
-	imageCleanupErrorIgnoreTimeout: 3600 * 1000
-	maxDeltaDownloads: 3
-	defaultVolumeLabels: {
-		'io.resin.supervised': 'true'
-	}
-
-process.env.DOCKER_HOST ?= "unix://#{module.exports.dockerSocket}"
```
pcarranzav: @CameronDiver this is missing in the ts file and I think it's the reason you're seeing errors

src/lib/validation.ts
```diff
@@ -1,28 +1,24 @@
...
-}
-
-export interface LabelObject {
-	[name: string]: string;
-}
-
 const ENV_VAR_KEY_REGEX = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
 
+type NullableString = string | undefined | null;
```
pcarranzav: Should this also go in types.ts?

test/lib/conversions.spec.coffee
```diff
@@ -0,0 +1,32 @@
...
+conversion = require '../../src/lib/conversions'
+
+describe 'conversions', ->
+
+	describe 'envArrayToObject', ->
+		it 'should convert an env array to an object', ->
+			expect(conversion.envArrayToObject([
+				'key=value'
+				'test1=test2'
+				'k=v'
```
pcarranzav: It would be good to test with a value that has an `=`

src/lib/validation.ts
CameronDiver: I'm not sure, I've tried to stick to only moving types that will be used in another place, whereas this was simply to avoid having to write out the value at every place. If another module finds itself needing it, I'll definitely move it out.

PR #641, opened 2018-05-03
Convert errors and iptables modules to typescript

CameronDiver:
Change-type: patch


src/lib/errors.ts
```diff
@@ -0,0 +1,22 @@
+import { endsWith } from 'lodash';
+
+import { checkInt } from './validation';
+
+export function NotFoundError(err: { statusCode?: string }): boolean {
+	return checkInt(err.statusCode) === 404;
+}
+
+export function ENOENT(err: { code: string }): boolean {
```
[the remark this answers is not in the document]
pcarranzav: Based on https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks

src/lib/errors.ts
```diff
@@ -0,0 +1,22 @@
...
+	return err.code === 'ENOENT';
+}
+
+export function EEXISTS(err: { code: string }): boolean {
+	return err.code === 'EEXISTS';
+}
+
+export function UnitNotLoadedError(err: string[]): boolean {
+	if (err[0] != null) {
```
Page-: Since you've switched to using `_.endsWith` instead of `String.prototype.endsWith` you can just do
```typescript
export function UnitNotLoadedError(err: string[]): boolean {
	return endsWith(err[0], 'not loaded.');
}

File: src/lib/iptables.ts [hunk 3]
```diff
@@ -0,0 +1,38 @@
+import * as Promise from 'bluebird';
+import * as _childProcess from 'child_process';
+
+const childProcess: any = Promise.promisifyAll(_childProcess);
+
+function clearAndAppendIptablesRule(rule: string): Promise<{}> {
+	return childProcess.execAsync(`iptables -D ${rule}`)
+		.catchReturn({})
+		.then(() => {
+			return childProcess.execAsync(`iptables -A ${rule}`);
```
Page-: You could drop the braces + return if you want

src/lib/iptables.ts
```diff
@@ -0,0 +1,38 @@
+import * as Promise from 'bluebird';
+import * as _childProcess from 'child_process';
+
+const childProcess: any = Promise.promisifyAll(_childProcess);
+
+function clearAndAppendIptablesRule(rule: string): Promise<{}> {
+	return childProcess.execAsync(`iptables -D ${rule}`)
+		.catchReturn({})
```
Page-: Can you `catchReturn(null)` instead? or maybe `catchReturn<void>()`

src/lib/iptables.ts
```diff
@@ -0,0 +1,38 @@
...
+		.then(() => {
+			return childProcess.execAsync(`iptables -I ${rule}`);
+		});
+}
+
+export function rejectOnAllInterfacesExcept(
+	allowedInterfaces: string[],
+	port: number,
+): Promise<{}> {
```
Page-: Why `Promise<{}>` over `Promise<void>`?

src/lib/iptables.ts
```diff
@@ -0,0 +1,38 @@
...
+	// We delete each rule and create it again to ensure ordering (all ACCEPTs before the REJECT/DROP).
+	// This is especially important after a supervisor update.
+	return Promise.each(allowedInterfaces, (iface) => {
+		return clearAndInsertIptablesRule(`INPUT -p tcp --dport ${port} -i ${iface} -j ACCEPT`);
+	})
+		.then(() => {
+			return clearAndAppendIptablesRule(`INPUT -p tcp --dport ${port} -j REJECT`)
```
Page-: I think dropping the braces + return would be really helpful here to get rid of this super wonky indentation

src/lib/iptables.ts
```diff
@@ -0,0 +1,30 @@
...
+export function rejectOnAllInterfacesExcept(
+	allowedInterfaces: string[],
+	port: number,
+): Promise<void> {
+	// We delete each rule and create it again to ensure ordering (all ACCEPTs before the REJECT/DROP).
+	// This is especially important after a supervisor update.
+	return Promise.each(allowedInterfaces,
+		(iface) => clearAndInsertIptablesRule(`INPUT -p tcp --dport ${port} -i ${iface} -j ACCEPT`),
+	)
+		.then(() => clearAndAppendIptablesRule(`INPUT -p tcp --dport ${port} -j REJECT`))
```
Page-: This indentation really confuses me, is there any way to keep it indented on the same level it actually executes at?

src/lib/iptables.ts
```diff
@@ -0,0 +1,30 @@
+import * as Promise from 'bluebird';
+import * as _childProcess from 'child_process';
+
+const childProcess: any = Promise.promisifyAll(_childProcess);
```
Page-: It would be much better to do
```typescript
const execAsync = Promise.promisify(childProcess.exec)
```
as that's all you use and it will be able to create much better automatic typings than just using `any`

src/lib/iptables.ts
```diff
@@ -0,0 +1,30 @@
+import * as Promise from 'bluebird';
+import * as _childProcess from 'child_process';
```
Page-: `_` as a prefix has a specific meaning in typescript which is that the variable isn't used at all and that it's intentional, in order to disable the unused vars warnings

src/lib/iptables.ts
CameronDiver: This is the way that I've indented all of my typescript, 
this just looks slightly more weird due to the indented argument above. If this isn't acceptable I think we need to define exactly what to do in these situations, or just use prettier.

PR #651, opened 2018-05-11
Don't generate config fields in offline mode

CameronDiver:
Change-type: patch
Closes: #648


src/config.coffee
```diff
@@ -241,8 +241,9 @@ module.exports = class Config extends EventEmitter
 		deviceRegister.generateUniqueKey()
 
 	generateRequiredFields: =>
-		@getMany([ 'uuid', 'deviceApiKey', 'apiSecret', 'logsChannelSecret' ])
-		.then ({ uuid, deviceApiKey, apiSecret, logsChannelSecret }) =>
+		@getMany([ 'uuid', 'deviceApiKey', 'apiSecret', 'logsChannelSecret', 'offlineMode' ])
+		.then ({ uuid, deviceApiKey, apiSecret, logsChannelSecret, offlineMode }) =>
+			return if offlineMode
```
pcarranzav: @CameronDiver shouldn't we still generate a uuid if it doesn't exist? e.g. for preloaded apps, since they have a RESIN_DEVICE_UUID env var

test/data/testconfig-apibinder-offline.json
```diff
@@ -0,0 +1 @@
+{"applicationName":"supertestrpi3","applicationId":78373,"deviceType":"raspberrypi3","userId":1001,"username":"someone","appUpdatePollInterval":3000,"listenPort":2345,"vpnPort":443,"apiEndpoint":"http://0.0.0.0:3000","vpnEndpoint":"vpn.resin.io","registryEndpoint":"registry2.resin.io","deltaEndpoint":"https://delta.resin.io","pubnubSubscribeKey":"foo","pubnubPublishKey":"bar","mixpanelToken":"baz","apiKey":"boo","version":"2.0.6+rev3.prod","supervisorOfflineMode":true}
```
pcarranzav: You could  also test that the supervisor is in offline mode with a minimal config.json: `{ "deviceType": "raspberrypi3" }` (i.e. there's no apiEndpoint so the supervisor infers it's in offline mode).

src/config.coffee
```diff
@@ -241,10 +241,13 @@ module.exports = class Config extends EventEmitter
...
-				uuid ?= @newUniqueKey()
+		@getMany([ 'uuid', 'deviceApiKey', 'apiSecret', 'logsChannelSecret', 'offlineMode' ])
+		.then ({ uuid, deviceApiKey, apiSecret, logsChannelSecret, offlineMode }) =>
+			if !uuid?
+				uuid = @newUniqueKey()
+			return if offlineMode
+
+			if !deviceApiKey? or !apiSecret? or !logsChannelSecret?
 				deviceApiKey ?= @newUniqueKey()
 				apiSecret ?= @newUniqueKey()
```
pcarranzav: We also need the apiSecret, otherwise the supervisor API will break (and preloaded apps should be able to use the  supervisor API).

test/11-api-binder.spec.coffee
```diff
@@ -135,3 +135,25 @@ describe 'APIBinder', ->
...
+				expect(mode).to.equal(true)
+				# Check that there is no deviceApiKey
+				@config.get('deviceApiKey').then (key) ->
+					expect(key).to.be.undefined
+
+		describe 'Minimal config offline mode', ->
+			before ->
+				initModels.call(this, '/config-apibinder-offline2.json')
+
+			it 'does not generate a key with the minimal config', ->
```
pcarranzav: In this case you can also test that a uuid is still generated?

CameronDiver: I actually meant to add that in and forgot, will add :)

Issue #645, opened 2018-05-09
Supervisor 7.1.18  doesn't update `download_progress` and `status` during download


brentalanmiller:
When running on Intel Edison, ResinOS 2.12.5+rev2, all calls to `GET /v1/device` will always show `"download_progress": null` and `"status":"Idle"` even if there is a download in progress.

For example, if we have a device online that has the update lock in place. Right before we deploy an update of our application we'll see a response like this:
```
[2018-05-08 23:33:39.844] [resin] [trace] Supervisor device response: {"api_port":48484,"ip_address":"10.254.0.87","os_version":"Resin OS 2.12.5+rev2","supervisor_version":"7.1.18","update_pending":false,"update_failed":false,"update_downloaded":false,"commit":"1c7c1f78678673694e3d676007474e809e1fd204","status":"Idle","download_progress":null}
```

When we push the update, we'll see the download progress on the dashboard, but the supervisor won't update `download_progress` and `status`, although `update_pending` seems to get updated.
```
[2018-05-08 23:33:46.904] [resin] [trace] Supervisor device response: {"api_port":48484,"ip_address":"10.254.0.87","os_version":"Resin OS 2.12.5+rev2","supervisor_version":"7.1.18","update_pending":true,"update_failed":false,"update_downloaded":false,"status":"Idle","download_progress":null}
```

Back when we were using Resin OS 2.3.0+rev1 with supervisor 6.1.3, download_progress would show the download percentage and status would show "Downloading":
```
[2018-05-09 00:17:54.580] [resin] [trace] Supervisor device response: {"api_port":48484,"os_version":"Resin OS 2.3.0+rev1","os_variant":"prod","supervisor_version":"6.1.3","download_progress":23,"ip_address":"10.254.0.6","status":"Downloading","commit":"d4daf68510058738be40da18c354cb4cf92cecd8","update_pending":true,"update_failed":false,"update_downloaded":false}
```

CameronDiver:
This will be handled as part of https://github.com/resin-io/resin-supervisor/issues/640, closing this issue to track there.

cywang117:
@basz Could you check with `/v2/state/status`, `/v2/applications/state`, or `/v2/applications/:appId/state` endpoints to see if the result is similar. Additionally, could you describe the conditions under which you queried the download progress and got the results as pasted above?

basz:
Ah... I totally missed there is a v2? I found what I needed and started using that. I'll look into v2 suggestions. thank you

[...]

Issue #640, opened 2018-05-02
V2 state API for the supervisor should be implemented


CameronDiver:
Since the multicontainer upgrades, a lot of the current supervisor endpoints make little sense, or outright don't work. We need to upgrade the endpoints to v2, and provide multicontainer-friendly methods.

Connects to: 
https://github.com/resin-io/resin-supervisor/issues/391
https://github.com/resin-io/resin-supervisor/issues/443
https://github.com/resin-io/resin-supervisor/issues/542
https://github.com/resin-io/resin-supervisor/issues/276

adamtec:
I am very much interested in an updated API. What kind of timeline are we looking at?

I switched to multi-container and I am facing an issue of determining when an update has finished downloading. Previously, on the device, I would write to the update lock file. The device would monitor any ongoing downloads via /v1/device. Once finished a download, I would prompt the user that a download is available, in which case they could apply (where I would then post to /v1/device). 
I don't need necessarily any fine-grained progress percentage nor progress of various service's images just when the complete 'docker-compose pull' has completed.

PR #652, opened 2018-05-11
V2 device state api and fixes

CameronDiver:
- [X] Refactoring of the v1 api into separate modules to aid comprehension
- [X] Refactoring of the v2 api into separate modules
- [X] Add v2 state endpoint, which can be used to query the multicontainer state
- [x] Fix /v1/device returning null when updating

NOTE: The typescript definition files I've added contain the *bare* minimum to get the code to work with the correct types. I plan to recode the entire supervisor in typescript in the very near future, so I'd rather not spend a lot of time producing these definition files which ultimately will not be used. It would be good if you take this into account when reviewing.

Change-type: patch
Connects-to: #640
Connects-to: #645 
---- Autogenerated Waffleboard Connection: Connects to #640

Front logo [Front conversations](https://app.frontapp.com/open/top_c56z)
---- Autogenerated Waffleboard Connection: Connects to #645


src/device-api/v2.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export function createV2Api(router: Router, applications: ApplicationManager) {
+
+	const { _lockingIfNecessary } = applications;
+
+	const handleServiceAction = (
+		req: Request,
+		res: Response,
+		action: any,
+	): Promise<void> => {
```
zvin: NIT: functions in this file would be more readable using async/await.

src/device-api/v2.ts
```diff
@@ -1,10 +1,11 @@
-import { Response, Request, Router } from 'express';
+import * as Promise from 'bluebird';
```
zvin: bluebird should be imported as `Bluebird` to avoid confusion with the regular `Promise` that is used in other files.

Issue #644, opened 2018-05-04
Large port ranges can cause a memory leak


pcarranzav:
Defining a range like `"ports": ["53-65535:53-65535"]` makes the supervisor consume a lot of memory, eventually leading to this message in the logs:

```
May 04 20:54:42 ea1063f healthdog[1062]: FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
```

We should think of better ways to work with port ranges, as they way it currently works in service.coffee, instantiating a new object in portBindings for each port in the range, is not efficient.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_bgk3)

PR #655, opened 2018-05-14
Store port ranges as actual ranges, to reduce ambient memory usage

CameronDiver:
Before this change, port ranges were iterated and stored as an object
per port mapping. Now the port ranges are stored as ranges until they
need to be converted to objects. The need to convert to objects still
exists as this is the format which the docker remote API expects, but
hopefully this should alleviate bugs like #644 by making the memory more
shorter-lived.

Also added more tests.

Change-type: patch
Closes: #644
Signed-of-by: Cameron Diver <cameron@resin.io>


src/compose/service.coffee
```diff
@@ -661,6 +667,10 @@ module.exports = class Service
 		_.isEmpty(_.xor(_.keys(@networks), _.keys(otherService.networks)))
 
 	isSameContainer: (otherService) =>
+		# We need computed fields to be present to compare two services
+		@portMappings ?= @getPortsAndPortBindings()
```
pcarranzav: These two lines shouldn't be necessary if they're generated in the service constructor... and I'd very much prefer it if this function didn't mutate the services (keep it pure)

src/compose/service.coffee
```diff
@@ -273,7 +271,7 @@ module.exports = class Service
 			@extendLabels(opts.imageInfo)
 			@extendAndSanitiseVolumes(opts.imageInfo)
 			@extendAndSanitiseExposedPorts(opts.imageInfo)
-			{ @exposedPorts, @portBindings } = @getPortsAndPortBindings()
+			@portMappings = @getPortsAndPortBindings()
```
pcarranzav: You should also create the portMappings when there is a `@containerId` for the comparison to work (see my comment below on avoiding mutation) - but in that case you'll probably need the inverse function to generatePortBindings, that takes exposedPorts and portBindings (which come from the docker daemon) and produces portMappings

src/compose/ports.ts
```diff
@@ -0,0 +1,326 @@
...
+
+				portList.forEach((port: BindingObject) => {
+					// We need to check that the internal and external ports
+					// are either one above or one below the last ones, if not
+					// we need to start a new range
+					let internal = parseInt(port.internalStrMatch![1]);
+					let external = parseInt(port.hostObj[0].HostPort);
+
+					// If we are not currently tracking a range
+					if (currentRange.internalStart == null) {
```
pcarranzav: Maybe you can avoid all this logic, and just turn the `portBindingList` into a `PortMap[]`, and then call `normalisePortMaps` on that? This would make this function way simpler / shorter, *and* it would mean that the same function is used going both ways, which will make chances of comparisons going wrong smaller.

Unless I'm missing some reason why it wouldn't work?

src/compose/ports.ts
```diff
@@ -0,0 +1,326 @@
...
+export interface PortBindings {
+	[key: string]: Array<{ HostIp: string, HostPort: string }>;
+}
+
+export interface DockerPortOptions {
+	exposedPorts: Map<string, {}>;
+	portBindings: PortBindings;
+}
+
+interface RangeObject {
```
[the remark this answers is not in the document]
CameronDiver: Sure thing, will change.

src/compose/service.coffee
```diff
@@ -537,27 +534,29 @@ module.exports = class Service
...
-					internalEnd ?= internal
-					externalEnd ?= external
-					externalRange = _.map([external..externalEnd], String)
-					internalRange = _.map([internal..internalEnd], String)
-					if externalRange.length == internalRange.length # Ignore invalid port mappings
-						for hostPort, ind in externalRange
-							containerPort = internalRange[ind]
-							exposedPorts["#{containerPort}/#{protocol}"] = {}
-							portBindings["#{containerPort}/#{protocol}"] = [ { HostIp: host, HostPort: hostPort } ]
+				portMaps.push(new PortMap(port))
```
Page-: This would be better as
```coffee
portMaps =
	for port in @ports
		new PortMap(port)
````
or
```coffee
portMaps = _.map @ports, (port) ->
	new PortMap(port)
```
(with the second you can also drop the `if @ports?` as lodash will handle it)

src/compose/ports.ts
```diff
@@ -0,0 +1,206 @@
...
+
+		// Ensure we have the same range
+		if (this.internalEnd - this.internalStart !== this.externalEnd - this.externalStart) {
+			throw new InvalidPortDefinition(
+				`Range for internal and external ports does not match: ${portStr}`,
+			);
+		}
+	}
+
+	private generatePortRange(start: number, end: number): number[] {
```
Page-: It might be worth adding a check that `start <= end` and throwing an error if not

src/compose/ports.ts
```diff
@@ -0,0 +1,206 @@
...
+		if (start === end) {
+			return [ start ];
+		}
+
+		let x = start;
+		let ret = [];
```
Page-: You can replace this with:
```
return _.range(start, end + 1)
```

src/compose/ports.ts
```diff
@@ -0,0 +1,206 @@
...
+
+		// Create a list of portBindings, rather than the map (which we can't
+		// order)
+		const portMaps = _.map(portBindings, (hostObj, internalStr) => {
+
+			const match = internalStr.match(DOCKER_OPTS_PORTS_REGEX);
+			if (match == null) {
+				throw new Error(`Could not parse docker port output: ${internalStr}`);
+			}
+			const internal = parseInt(match![1], 10);
```
Page-: The `!` shouldn't be needed here as we already verified to the compiler that `match` cannot be null

src/compose/ports.ts
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	public toDockerOpts(): DockerPortOptions {
+		const internalRange = this.generatePortRange(this.internalStart, this.internalEnd);
+		const externalRange = this.generatePortRange(this.externalStart, this.externalEnd);
+
+		const exposed: { [key: string]: {} } = {};
+		const portBindings: PortBindings = {};
+
+		_.each(_.zip(internalRange, externalRange), ([internal, external]) => {
```
Page-: I think you can do this with
```typescript
_.zipWith(internalRange, externalRange, (internal, external) => {
```

src/compose/ports.ts
```diff
@@ -0,0 +1,202 @@
...
+				`Range for internal and external ports does not match: ${portStr}`,
+			);
+		}
+	}
+
+	private generatePortRange(start: number, end: number): number[] {
+		if (start > end) {
+			throw new Error('Incorrect port range! The end port cannot be larger than the start port!');
+		}
+		if (start === end) {
```
Page-: You can get rid of this special case - it's already handled by the `_.range(start, end + 1)` below

src/compose/ports.ts
Page-: Done

src/compose/ports.ts
```diff
@@ -0,0 +1,202 @@
...
+
+	public static normalisePortMaps(portMaps: PortMap[]): PortMap[] {
+		// Fold any ranges into each other if possible
+		return _(portMaps)
+			.sortBy((p) => p.protocol)
+			.sortBy((p) => p.host)
+			.sortBy((p) => p.internalStart)
+			.reduce((res: PortMap[], p: PortMap) => {
+				const last = _.last(res);
+				if (last == null) {
```
Page-: You could remove this by using
```
if (last != null &&
	last.internalEnd + 1 === p.internalStart &&
	...
) {
...
} else {
	res.push(p)
}
```
so you just have the one path for `res.push(p)`

src/compose/ports.ts
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	private parsePortString(portStr: string): void {
+		const match = portStr.match(PORTS_REGEX);
+		if (match == null) {
+			throw new InvalidPortDefinition(`Could not parse port definition: ${portStr}`);
+		}
+
+		let [
+			,
```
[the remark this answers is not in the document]
CameronDiver: Why `match.slice(1)`? Seems slightly unnecessary to copy a whole new array - happy to add a comment ofc though.

src/compose/ports.ts
```diff
@@ -0,0 +1,202 @@
...
+			host = '',
```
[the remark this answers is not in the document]
LucianBuzzo: Though thinking about it, it makes the destructuring very awkward

src/compose/ports.ts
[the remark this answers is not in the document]
LucianBuzzo: As long as the intention here is clear, I'm happy. A comment would be fine

src/compose/ports.ts
```diff
@@ -0,0 +1,202 @@
...
+		const externalRange = this.generatePortRange(this.externalStart, this.externalEnd);
+
+		const exposed: { [key: string]: {} } = {};
+		const portBindings: PortBindings = {};
+
+		_.zipWith(internalRange, externalRange, (internal, external) => {
+			exposed[`${internal}/${this.protocol}`] = {};
+
+			portBindings[`${internal}/${this.protocol}`] = [
+				{ HostIp: this.host, HostPort: external!.toString() },
```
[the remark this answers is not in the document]
CameronDiver: Apparently not - I swear it was at some point... Good catch.

Issue #658, opened 2018-05-16
After deconfiguring a device, the supervisor starts up the old application on startup


CameronDiver:
This is because the supervisor's internal state has no concept of *where* an application came from, so to fix this the supervisor should store which environment the application came from, and only start up the application if it matches the current environment.

PR #662, opened 2018-05-17
Don't start applications at startup unless the apiEndpoint fields match the current environment

CameronDiver:
This PR does two things:
1. Migrates the supervisor database to add a source field to the apps table. This will contain the apiEndpoint which provided the application information (or an empty string when this does not apply). 
2. When the supervisor starts up and tries to detect the the app that should be running, it uses the current endpoint to filter the applications from the database, only starting the ones which match

Closes: #658 
---- Autogenerated Waffleboard Connection: Connects to #658


CameronDiver:
@shaunmulligan yeah that should work with this approach, if the apiEndpoint is empty both in the running supervisor and the database state the application code should be started again - but I agree we'll have to make sure that the workflow makes sense and works reliably.

src/application-manager.coffee
```diff
@@ -440,7 +440,9 @@ module.exports = class ApplicationManager extends EventEmitter
 		).get(appId)
 
 	getTargetApp: (appId) =>
-		@db.models('app').where({ appId }).select()
+		@config.get('resinApiEndpoint').then (endpoint) ->
```
pcarranzav: You can do `.then (endpoint = '') ->` and avoid the next line.

src/application-manager.coffee
```diff
@@ -872,7 +876,9 @@ module.exports = class ApplicationManager extends EventEmitter
 			@_targetVolatilePerImageId[imageId] = {}
 
 	getTargetApps: =>
-		Promise.map(@db.models('app').select(), @normaliseAndExtendAppFromDB)
+		@config.get('resinApiEndpoint'). then (source) =>
```
pcarranzav: Same as above, `.then (source = '') ->`.

src/migrations/20180516164000-endpoint-specific-apps.js
```diff
@@ -0,0 +1,32 @@
+const fs = require('fs');
+const configJsonPath = process.env.CONFIG_MOUNT_POINT;
```
pcarranzav: I think this variable doesn't exist? The correct one is BOOT_MOUNT_POINT and then appending `/config.json` I think?

src/device-state.coffee
```diff
@@ -241,17 +241,20 @@ module.exports = class DeviceState extends EventEmitter
...
-				# Apps, deviceConfig, dependent
-				@db.transaction (trx) =>
-					Promise.try =>
-						@config.set({ name: target.local.name }, trx)
-					.then =>
-						@deviceConfig.setTarget(target.local.config, trx)
-					.then =>
-						@applications.setTarget(target.local.apps, target.dependent, trx)
+		Promise.join(
+			@config.get('resinApiEndpoint'),
```
pcarranzav: This is a good opportunity to just replace this with `apiEndpoint` and remove the resinApiEndpoint config field, since it's legacy and not needed anymore.

src/migrations/20180516164000-endpoint-specific-apps.js
CameronDiver: That's what I thought, but the above points to the correct file, and `BOOT_MOUNT_POINT` was undefined in my testing.

src/device-state.coffee
CameronDiver: I'd rather handle that in a seperate PR, just to keep the changes smaller and more focussed.

src/migrations/20180516164000-endpoint-specific-apps.js
pcarranzav: Ah, right, `CONFIG_MOUNT_POINT` is defined in the Dockerfile... yeah, that works fine (but for the record the one I meant was BOOT_MOUNTPOINT, without the second underscore

src/migrations/20180516164000-endpoint-specific-apps.js
```diff
@@ -0,0 +1,32 @@
...
+	return new Promise((resolve, reject) => {
+		if (!configJsonPath) {
+			resolve({});
+		}
+		fs.readFile(configJsonPath, (err, data) => {
+			try {
+				const parsed = JSON.parse(data.toString());
+				resolve(parsed);
+			} catch(e) {
+				resolve({});
```
pcarranzav: I suggest at least adding a console.log warning here, as it would be quite rare not to have a config.json at all (and currently other things would break in that case)

src/migrations/20180516164000-endpoint-specific-apps.js
CameronDiver: Ah ok, which do you think is more correct then? For example I imagine all device types won't  use config.txt (although I know rpi+fin are the only ones which implement this kind of interface so far).

src/migrations/20180516164000-endpoint-specific-apps.js
CameronDiver: Sounds good!

src/migrations/20180516164000-endpoint-specific-apps.js
pcarranzav: Since we're only reading config.json here, this one is fine. If we were writing to it, going through BOOT_MOUNTPOINT would be better to allow an atomic rename.

CameronDiver: Ah ok, makes sense. I'll get this merged then, thanks @pcarranzav

Issue #668, opened 2018-05-30
The supervisor should respect the pinDevice entry in preloaded images


CameronDiver:

pcarranzav:
Connects to https://github.com/resin-io/hq/issues/1254

PR #669, opened 2018-05-30
Pin a device to a commit when preload has a pinDevice field

CameronDiver:
Change-type: minor
Closes: #668
Depends-on: https://github.com/resin-io/resin-api/pull/1056


pcarranzav:
Connects to https://github.com/resin-io/hq/issues/1254

pcarranzav:
(and the supervisor should always act on this config field if it is set, even if the device is already provisioned)

src/api-binder.coffee
```diff
@@ -263,6 +266,32 @@ module.exports = class APIBinder
...
+			@resinApi.get
+				resource: 'release'
+				options:
+					filter:
+						belongs_to__application: app
+						commit: commit
+						status: 'success'
+					select: 'id'
+			.get(0)
+			.get('id')
```
Page-: If no results are returned for whatever reason then this will throw an error along the lines of `cannot get id of undefined`,  and the `if !releaseId?` below implies that you expect that to be a possibility

src/device-state.coffee
```diff
@@ -341,6 +349,11 @@ module.exports = class DeviceState extends EventEmitter
 					@setTarget({
 						local: stateFromFile
 					})
+				.then =>
+					if stateFromFile.pinDevice
+						# multi-app warning!
+						# The following will need to be changed once running multiple applications is possible
+						@pinDevice = { commit: commitToPin, app: appToPin } if commitToPin? and appToPin?
```
Page-: I think it would make sense to add this `if` to the one just above, it feels a little weird as-is

src/api-binder.coffee
CameronDiver: Ah yeah, I was getting confused with lodash's `_.get`, which handles undefined nicely. Will change.

src/api-binder.coffee
```diff
@@ -198,6 +198,21 @@ module.exports = class APIBinder
 				@config.set(configToUpdate)
 			.then =>
 				@eventTracker.track('Device bootstrap success')
+		# Check if we need to pin the device, regardless of if we provisioned
+		.then =>
+			@config.get('pinDevice'). then (pin) ->
```
zvin: this `.then` would look better on the next line

src/config.coffee
```diff
@@ -151,6 +151,8 @@ module.exports = class Config extends EventEmitter
 			lockOverride: { source: 'db', mutable: true, default: 'false' }
 			legacyAppsPresent: { source: 'db', mutable: true, default: 'false' }
 			nativeLogger: { source: 'db', mutable: true, default: 'true' }
+			# A JSON value, which is either null, or { app: number, commit: string }
+			pinDevice: { source: 'db', mutable: true, default: 'false' }
```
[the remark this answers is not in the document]
CameronDiver: ...yeah, missed that, good catch!

src/api-binder.coffee
```diff
@@ -263,6 +277,34 @@ module.exports = class APIBinder
...
+			.then (release) =>
+				releaseId = _.get(release, '[0].id')
+				if !releaseId?
+					throw new Error('Cannot continue pinning preloaded device! No release found!')
+				@resinApi.patch
+					resource: 'device'
+					id: deviceId
+					body:
+						should_be_running__release: releaseId
+			.then ->
```
pcarranzav: Should be fat arrow

src/api-binder.coffee
```diff
@@ -198,6 +198,20 @@ module.exports = class APIBinder
...
 			.then =>
 				@eventTracker.track('Device bootstrap success')
+		# Check if we need to pin the device, regardless of if we provisioned
+		.then =>
+			@config.get('pinDevice')
+			.then(JSON.parse)
+			.tapCatch ->
+				console.log('Warning: Malformed pinDevice value in supervisor database')
+			.catchReturn(null)
+			.then (pinValue) ->
```
pcarranzav: Should be fat arrow

src/api-binder.coffee
CameronDiver: I hate coffeescript... Good catch. I'll test it all again tomorrow before merge.

src/api-binder.coffee
```diff
@@ -198,6 +198,20 @@ module.exports = class APIBinder
 				@config.set(configToUpdate)
 			.then =>
 				@eventTracker.track('Device bootstrap success')
+		# Check if we need to pin the device, regardless of if we provisioned
```
pcarranzav: For this to actually happen regardless of if we provisioned, the check in `provisonDevice` below (line 233) should be `if !conf.provisioned or conf.apiKey? or conf.pinDevice?` (and adding pinDevice to the config.getMany there)

CameronDiver:
@pcarranzav updated and tested - all good.

Issue #671, opened 2018-06-04
Support extlinux configuration for the jetson-TX family devices


CameronDiver:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_c1dn)

imrehg:
@CameronDiver also for the TX1?

CameronDiver:
Good catch :) changed.

PR #672, opened 2018-06-05
Add support for extlinux configuration files and isolcpus config option

CameronDiver:
Firstly this PR abstracts away all of the boot config into a separate module, which behaves differently based on the device type provided. All of the equivalent code is removed from the device-config module. The extlinux support is then built on top of this, with support for the `isolcpus` config option only currently.

Extlinux configuration variables begin `RESIN_HOST_EXTLINUX_`.

It's probably best to review this PR commit by commit.

Connects-to: #671 
---- Autogenerated Waffleboard Connection: Connects to #671


src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+
+export const bootMountPoint = constants.rootMountPoint + constants.bootMountPoint;
+
+export function isRPiDeviceType(deviceType: string): boolean {
+	return _.startsWith(deviceType, 'raspberry') || deviceType === 'fincm3';
+}
+
+export function isExtlinuxDeviceType(deviceType: string): boolean {
+	return _.startsWith(deviceType, 'jetson-tx');
+}
```
dfunckt: I think none of the above need to be exported and could be kept file-private.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+	if (isRPiDeviceType(deviceType)) {
+		return rPiBootConfigPrefix;
+	}	else if (isExtlinuxDeviceType(deviceType)) {
+		return extlinuxBootConfigPrefix;
+	} else {
+		throw new Error(`No boot config var prefix for device type: ${deviceType}`);
+	}
+}
+
+export function bootConfigVarRegex(deviceType: string): RegExp {
```
dfunckt: I think this one can stay private as well.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+	if (!isConfigDeviceType(deviceType)) {
+		return { };
+	}
+	const prefix = bootConfigVarPrefix(deviceType);
+	const regex = bootConfigVarRegex(deviceType);
+
+	let parsedEnv = _.pickBy(env, (_val: string, key: string) => {
+		return _.startsWith(key, prefix);
+	});
+	parsedEnv = _.mapKeys(parsedEnv, (_val, key) => {
```
dfunckt: _.mapKeys takes a callback that receives the key as the first argument, right?

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+			if (_.includes(rpiArrayConfigKeys, key)) {
+				if (!_.startsWith(val, '"')) {
+					return [ val ];
+				} else {
+					return JSON.parse(`[${val}]`);
+				}
+			}
+		}
+		return val;
+	});
```
dfunckt: You can re-write this whole block as:

```
return _(env)
  .pickBy((_val, key) => _.startsWith(key, prefix))
  .mapKeys(key => key.replace(regex, '$2'))
  .mapValues(val => ...)
  .value();
```

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+	const prefix = bootConfigVarPrefix(deviceType);
+	const confWithEnvKeys = _.mapKeys(config, (_val, key) => {
+		return prefix + key;
+	});
+	return _.mapValues(confWithEnvKeys, (val) => {
+		if (_.isArray(val)) {
+			return JSON.stringify(val).replace(/^\[(.*)\]$/, '$1');
+		}
+		return val;
+	});
```
dfunckt: Same here:

```
return _(config)
  .mapKeys(...)
  .mapValues(...)
  .value();
```

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+}
+
+export function filterConfigKeys(
+	deviceType: string,
+	allowedKeys: string[],
+	conf: { [key: string]: any },
+): { [key: string]: any } {
+
+	let isConfigType: boolean = false;
```
dfunckt: The type annotation is redundant, it's inferred by the assigned value.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+	const conf: ConfigOptions = { };
+	const configStatements = confStr.split(/\r?\n/);
+
+	for (const configStr of configStatements) {
+		// Don't show warnings for comments and empty lines
+		const trimmed = _.trimStart(configStr);
+		if (_.startsWith(trimmed, '#') || trimmed === '') {
+			continue;
+		}
+		let keyValue = /^([^#=]+)=(.+)/.exec(configStr);
```
dfunckt: This regex is slightly better I think: `/^([^=]+)=(.*)$/`. Essentially, it matches everything until a `=` as the key and everything until EOL as the value and also allows a value to be empty.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+		if (keyValue != null) {
+			if (!_.includes(rpiArrayConfigKeys, keyValue[1])) {
```
dfunckt: It'll be slightly more readable to assign key and value to local variables before using them: `const [ /* ignore */, key, value ] = keyValue`

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+				conf[keyValue[1]] = keyValue[2];
+			} else {
+				const key = keyValue[1];
+				if (conf[key] == null) {
+					conf[key] = [];
+				}
+				(conf[key] as string[]).push(keyValue[2]);
+			}
+		} else {
```
dfunckt: You can skip the `else` and de-indent the whole block below. Apart from the benefit of not indenting too much, I think it better reflects what you're trying to do -- i.e. try once with a regex and process if there's a match, then re-try with another regex if there was no match.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+
+	return conf;
+}
+
+export function parseExtlinuxBootConfig(confStr: string): ConfigOptions {
+	const parsedBootFile = parseExtlinuxFile(confStr);
+
+	// First find the default label name
+	const defaultLabel = _.find(parsedBootFile.globals, (_v, l) => {
+		if (l === 'DEFAULT') {
```
dfunckt: Maybe compare lower-case (or upper-case)? And you can actually write this as:

```
const defaultLabel = _.find(parsedBootFile.globals, (_v, l) => l.toUpperCase() === 'DEFAULT');
```

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+	const appendEntry = labelEntry.APPEND;
+
+	if (appendEntry == null) {
+		throw new Error('Could not find APPEND directive in default extlinux.conf boot entry');
+	}
+
+	const conf: ConfigOptions = { };
+	const values = appendEntry.split(' ');
+	for(const value of values) {
+		const parts = value.split('=');
```
dfunckt: Maybe raise an error if `parts.length isnt 2`?

src/lib/config-utils.ts
```diff
@@ -0,0 +1,376 @@
...
+	let lastLabel = '';
+
+	for (const line of lines) {
+		const match = line.match(/^\s*(\w+)\s?(.*)$/);
+		if (match == null) {
+			console.log('Warning - Could not read extlinux entry: ${line}');
+			continue;
+		}
+		let directive = match[1].toUpperCase();
+		let value = match[2];
```
dfunckt: It seems you assume `lastLabel` can be `''`, which means *value* must be able to be `''` however from the regex I think this can never be the case. I think you'd have to do:

```
const match = ...(/^\s*(\w+)\s*(.+)?$/)
...
let values = match[2] || '';
```

src/lib/config-utils.ts
CameronDiver: Doesn't seem to: https://lodash.com/docs/4.17.10#mapKeys

src/lib/config-utils.ts
CameronDiver: nice :+1:

src/lib/config-utils.ts
CameronDiver: sure, will change

src/lib/config-utils.ts
CameronDiver: I'm not worried about this too mcuh due to https://github.com/resin-io/resin-supervisor/pull/672/files/89627a33925e218275196bd5fe7b193e5e2737c7#diff-420116c700357465a04fcc85c5baa00eR258

src/lib/config-utils.ts
CameronDiver: Sure, will add.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,198 @@
...
+
+export const rpiArrayConfigKeys = [
+	'dtparam',
+	'dtoverlay',
+	'device_tree_param',
+	'device_tree_overlay',
+];
+
+export const hostConfigConfigVarPrefix = 'RESIN_HOST_';
+export const rPiBootConfigPrefix = hostConfigConfigVarPrefix + 'CONFIG_';
```
telphan: ES6 concat here please.

src/lib/config-utils.ts
```diff
@@ -0,0 +1,198 @@
...
+export const rPiConfigRegex = new RegExp('(' + _.escapeRegExp(rPiBootConfigPrefix) + ')(.+)');
+
+export const bootMountPoint = constants.rootMountPoint + constants.bootMountPoint;
```
telphan: ES6 concat here please

src/lib/config-utils.ts
```diff
@@ -0,0 +1,198 @@
...
+export function getBootConfigPath(deviceType: string): string {
+	if (isRPiDeviceType(deviceType)) {
+		return getRpiBootConfig();
+	} else {
+		throw new Error(`No boot config exists for device type: ${deviceType}`);
+	}
+}
+
+function getRpiBootConfig(): string {
+	return bootMountPoint + '/config.txt';
```
telphan: ES6 concat here too please

src/lib/config-utils.ts
```diff
@@ -0,0 +1,198 @@
...
+	} else {
+		throw new Error(`Could not set boot config for non-boot config device: ${deviceType}`);
+	}
+}
+
+function setRpiBootConfig(target: ConfigOptions): Promise<void> {
+	let confStatements: string[] = [];
+
+	_.each(target, (value, key) => {
+
```
telphan: Extra new line

src/lib/config-utils.ts
```diff
@@ -120,9 +146,15 @@ function getRpiBootConfig(): string {
 	return bootMountPoint + '/config.txt';
 }
 
+function getExtlinuxBootConfig(): string {
+	return bootMountPoint + '/extlinux/extlinux.conf';
```
telphan: ES6 concat here too please

src/lib/config-utils.ts
```diff
@@ -162,9 +194,100 @@ export function parseRpiBootConfig(confStr: string): ConfigOptions {
...
+			console.log('Warning - Could not read extlinux entry: ${line}');
+			continue;
+		}
+		let directive = match[1].toUpperCase();
+		let value = match[2];
+
+		// Special handling for the MENU directive
+		if (directive === 'MENU') {
+			const parts = value.split(' ');
+			directive = 'MENU ' + parts[0];
```
telphan: ES6 concat here too please

src/lib/config-backend.ts
```diff
@@ -0,0 +1,371 @@
...
+			if (key === 'initramfs') {
+				confStatements.push(`${key} ${value}`);
+			} else if(_.isArray(value)) {
+				confStatements = confStatements.concat(_.map(value, (entry) => `${key}=${entry}`));
+			} else {
+				confStatements.push(`${key}=${value}`);
+			}
+		});
+
+		const confStr = confStatements.join('\n') + '\n';
```
telphan: ES6 concat please

src/lib/config-backend.ts
```diff
@@ -0,0 +1,371 @@
...
+				return [ value ];
+			} else {
+				return JSON.parse(`[${value}]`);
+			}
+		}
+		return value;
+	}
+
+	public createConfigVarName(configName: string): string {
+		return RPiConfigBackend.bootConfigVarPrefix + configName;
```
telphan: ES6 concat please

src/device-config.coffee
```diff
@@ -76,11 +45,23 @@ module.exports = class DeviceConfig
 				.tapCatch (err) =>
 					@logger.logConfigChange(logValue, { err })
 			setBootConfig: (step) =>
-				@config.get('deviceType')
-				.then (deviceType) =>
+				Promise.all [
+					@config.get('deviceType')
+					@getConfigBackend()
+				]
+				.then ([ deviceType, configBackend ]) =>
```
dfunckt: Seems `configBackend` is unused. But do you really need to call `getConfigBackend` here? (If yes, Promise.join is typically used in these cases.)

src/lib/config-utils.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+export function bootConfigToEnv(deviceType: string, config: ConfigOptions): EnvVarObject {
+	const prefix = bootConfigVarPrefix(deviceType);
+	const confWithEnvKeys = _.mapKeys(config, (_val, key) => {
+		return prefix + key;
+	});
+	return _.mapValues(confWithEnvKeys, (val) => {
+		if (_.isArray(val)) {
+			return JSON.stringify(val).replace(/^\[(.*)\]$/, '$1');
```
zvin: Using a regexp seems a bit overkill
```typescript
return JSON.stringify(val).slice(1, -1)
```

src/device-config.coffee
```diff
@@ -89,28 +70,30 @@ module.exports = class DeviceConfig
...
-			_.includes(@validKeys, k) or _.startsWith(k, bootConfigEnvVarPrefix)
-
 	getTarget: ({ initial = false } = {}) =>
 		@db.models('deviceConfig').select('targetValues')
-		.then ([ devConfig ]) ->
-			return JSON.parse(devConfig.targetValues)
-		.then (conf) =>
-			conf = @filterConfigKeys(conf)
+		.then ([ devConfig ]) =>
+			return Promise.all [
```
dfunckt: `Promise.join` :)

src/device-config.coffee
```diff
@@ -89,28 +70,30 @@ module.exports = class DeviceConfig
...
 			if initial or !conf.RESIN_SUPERVISOR_VPN_CONTROL?
 				conf.RESIN_SUPERVISOR_VPN_CONTROL = 'true'
 			for own k, { envVarName, defaultValue } of @configKeys
 				conf[envVarName] ?= defaultValue
 			return conf
 
 	getCurrent: =>
-		@config.getMany([ 'deviceType' ].concat(_.keys(@configKeys)))
-		.then (conf) =>
+		Promise.all [
```
dfunckt: `Promise.join`

src/lib/config-utils.ts
```diff
@@ -12,30 +13,51 @@ export interface ConfigOptions {
...
+
 export function isConfigDeviceType(deviceType: string): boolean {
-	return isRPiDeviceType(deviceType);
+	return isRPiDeviceType(deviceType) || isExtlinuxDeviceType(deviceType);
 }
 
 export function bootConfigVarPrefix(deviceType: string): string {
 	if (isRPiDeviceType(deviceType)) {
 		return rPiBootConfigPrefix;
+	}	else if (isExtlinuxDeviceType(deviceType)) {
```
zvin: extra tab / spaces before the `else if`

package.json
```diff
@@ -17,7 +17,9 @@
   },
   "dependencies": {
     "@types/lodash": "^4.14.108",
+    "@types/mz": "0.0.32",
     "mkfifo": "^0.1.5",
+    "mz": "^2.7.0",
```
zvin: NIT: this should have been in the next commit.

src/lib/config-utils.ts
```diff
@@ -187,6 +310,42 @@ function setRpiBootConfig(target: ConfigOptions): Promise<void> {
 	return remountAndWriteAtomic(getRpiBootConfig(), confStatements.join('\n') + '\n');
 }
 
+function setExtlinuxBootConfig(target: ConfigOptions): Promise<void> {
+	// First get a representation of the configuration file, with all resin-supported configuration removed
+	return Promise.resolve(fs.readFile(getExtlinuxBootConfig()))
```
zvin: I don't see the benefit of using `mz.fs` here.
You can use `Bluebird.promisify` for `Bluebird` promises or `util.promisify` for regular promises.
Here you use `mz.fs` methods which return regular promises that you turn into `Bluebird` promises with `Bluebird.resolve`.

src/device-config.coffee
```diff
@@ -126,24 +109,29 @@ module.exports = class DeviceConfig
...
 			return true
 		return false
 
 	getRequiredSteps: (currentState, targetState) =>
 		current = _.clone(currentState.local?.config ? {})
 		target = _.clone(targetState.local?.config ? {})
 		steps = []
-		@config.getMany([ 'deviceType', 'offlineMode' ])
-		.then ({ deviceType, offlineMode }) =>
+		Promise.all [
```
dfunckt: `Promise.join`

src/lib/config-utils.ts
CameronDiver: There's two reasons I used `mz`, one is that I was getting several issues with types when using `promisify`, where the typescript compiler was expecting different parameter counts etc, but more importantly it makes the `stub`bing of these functions in the test suite much easier - that's not to say it can't be done using promisification, I'd prefer this to be hanlded elsewhere, as it's outside of the scope of the supervisor.

src/device-config.coffee
CameronDiver: That's actually a bug, forgot to switch out the `deviceType` parameter for the `configBackend` param (which also means that I can drop the `.all` altogether.

src/device-config.coffee
CameronDiver: I prefer to stick with `Promise.all` when I don't need to worry about performance, mainly because you can keep the same indentation level, and they're equivalent (again ignoring performance).

package.json
CameronDiver: It was actually used in the previous commit, just all of the usages got moved out to the config-backend module after the refactor.

Issue #676, opened 2018-06-11
After deconfiguring and reconfiguring, the supervisor attempts to pull an image with the incorrect key


CameronDiver:
Only happens when the `apiBackend` is the same.

cywang117:
Closing due to age. I've also not seen this behavior when re-provisioning a device in-place to the backend. If you do see this issue again, please reopen with Supervisor & OS versions, reproduction docker-compose, and logs.

PR #678, opened 2018-06-13
Prefer deviceApiKey over apiKey in currentApiKey request

CameronDiver:
Change-type: patch
Closes: #676


pcarranzav:
@CameronDiver is this meant to fix a specific issue? The reason for this to prefer apiKey is that, in general, if there's an apiKey we assume that the deviceApiKey hasn't been registered yet - and when this happens, the first thing we do is to try to register it and delete the apiKey

CameronDiver:
@pcarranzav it's to fix this issue: https://github.com/resin-io/resin-supervisor/issues/676

I was thinking that that should be fine, (testing now), but it should fall back correctly, right?

PR #682, opened 2018-06-18
A lot more typescript conversions

CameronDiver:
Probably best to review commit-by-commit.

Also moved the db conversion to this PR, instead of the v2 api stuff. This needs to be merged, then the v2 api and os unification stuff, because of the refactoring that needs to be done on top of the config module.


src/lib/os-release.ts
```diff
@@ -0,0 +1,47 @@
+import * as Promise from 'bluebird';
+import * as fs from 'fs';
+import * as _ from 'lodash';
+
+// FIXME: Don't use synchronous file reading and change call sites to support a promise
+export function getOSReleaseField(path: string, field: string): string | undefined {
```
dfunckt: Can you please search the code for references to this function and skip exporting it if it's not used? Or if it's easy, change call sites to use the higher-level exported functions below instead?

src/lib/os-release.ts
```diff
@@ -0,0 +1,47 @@
+import * as Promise from 'bluebird';
```
dfunckt: I think we should start importing Bluebird as Bluebird so that we don't override Promise anymore. This will allow an easier migration to async/await.

src/config.ts
```diff
@@ -0,0 +1,361 @@
...
+		resinApiEndpoint: { source: 'func' },
+		provisioned: { source: 'func' },
+		osVersion: { source: 'func' },
+		osVariant: { source: 'func' },
+		provisioningOptions: { source: 'func' },
+		mixpanelHost: { source: 'func' },
+		extendedEnvOptions: { source: 'func' },
+		fetchOptions: { source: 'func' },
+
+		// NOTE: all 'db' values are stored and loaded as *strings*,
```
dfunckt: Might make sense to encode this constraint in the types.

src/lib/os-release.ts
LucianBuzzo: Yep, this will definitely save some headaches in the future

PR #687, opened 2018-06-28
Bind db handle to fn variable

CameronDiver:
During the conversion to typescript, the behaviour of the database
handle changed slightly, meaning storing a reference to the models
function also requires a bind to be applied too.

Change-type: patch


CameronDiver:
My feelings exactly @dfunckt :laughing:

PR #700, opened 2018-07-16
Allow persistent logging to be enabled/disabled via env var

CameronDiver:
The first commit does some refactoring and in effect can be ignored, the feature itself is added in the second commit.


src/device-config.coffee
```diff
@@ -140,29 +144,35 @@ module.exports = class DeviceConfig
 			# If the legacy lock override is used, place it as the new variable
 			if checkTruthy(target['RESIN_OVERRIDE_LOCK'])
 				target['RESIN_SUPERVISOR_OVERRIDE_LOCK'] = target['RESIN_OVERRIDE_LOCK']
-			for own key, { envVarName, varType } of @configKeys
+			for own key, { envVarName, varType, rebootRequired } of @configKeys
 				if !match[varType](current[envVarName], target[envVarName])
 					configChanges[key] = target[envVarName]
 					humanReadableConfigChanges[envVarName] = target[envVarName]
+					@rebootRequired = @rebootRequired || (rebootRequired ? false)
```
pcarranzav: I think it's better to add a `rebootRequired` field to the changeConfig action, and then change `@rebootRequired` from the actionExecutor rather than here (since getRequiredSteps is meant to be a pure -or pure-ish-  function, and this here would break it)

PR #701, opened 2018-07-16
Implement new log backend 

petrosagg:


test/12-logger.spec.coffee
```diff
@@ -1,39 +0,0 @@
-m = require 'mochainon'
-{ expect } = m.chai
-{ spy, useFakeTimers } = m.sinon
-
-Logger = require '../src/logger'
-
-describe 'Logger', ->
```
Page-: Can you add tests for the new logger please

src/logger.coffee
```diff
@@ -209,7 +186,7 @@ module.exports = class Logger
 			if stdoutOrStderr not in [ 'stdout', 'stderr' ]
 				throw new Error("Invalid log selection #{stdoutOrStderr}")
 			if !@attached[stdoutOrStderr][containerId]
-				logsOpts = { follow: true, stdout: stdoutOrStderr == 'stdout', stderr: stdoutOrStderr == 'stderr', timestamps: true }
+				logsOpts = { follow: true, stdout: stdoutOrStderr == 'stdout', stderr: stdoutOrStderr == 'stderr', timestamps: true, since: '0s' }
```
Page-: This means that we'll lose any logs from the application that happen on boot before the supervisor starts will be lost, which at the very least should be flagged up to support agents as I wouldn't be surprised if it comes up

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+
+		# Immediately flush the headers. This gives a chance to the server to
+		# respond with potential errors such as 401 authentication error
+		@_req.flushHeaders()
+
+		# We want a very low writable high watermark to prevent having many
+		# chunks stored in the writable queue of @_gzip and have them in
+		# @_stream instead. This is desirable because once @_gzip.flush() is
+		# called it will do all pending writes with that flush flag. This is
+		# now what we want though. If there are 100 items in the queue we want
```
imrehg: "now" is "not"?

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
 
-	log: (message) =>
-		return if @offlineMode or !@publishEnabled or @publishQueue.length >= @maxQueueSize
-		if _.isString(message)
-			message = { message: message }
+		if _.isString(msg)
+			msg = { message: msg }
 
-		message = _.defaults({}, message, {
+		msg = _.defaults({}, msg, {
```
Page-: You can change this to
```coffee
msg = _.assign({
	timestamp: Date.now()
 	message: ''
 }, msg)
```
to get the same default values/cloning semantics without having to check for each property if it already exists

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+	# latency. When ZLIB_TIMEOUT is 0 the compression ratio is around 5x
+	# whereas when ZLIB_TIMEOUT is infinity the compession ratio is around 10x.
+	_flush: _.throttle(->
+		@_gzip.flush(zlib.Z_SYNC_FLUSH)
+	, ZLIB_TIMEOUT, leading: false)
+
+	_write: (msg) ->
+		@_snooze()
+
+		if @_req is null
```
Page-: Do you specifically want to differentiate between `null` and `undefined`?  If not then this should be `if !@_req?`

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+		# Doing it immediately would potentialy lose logs if it turned out that
+		# the server is unavailalbe because @_req stream would consume our
+		# passthrough buffer
+		@_timeout = setTimeout(=>
+			@_stream.pipe(@_gzip)
+		, RESPONSE_GRACE_PERIOD)
+	, COOLDOWN_PERIOD)
+
+	_teardown: ->
+		if @_req isnt null
```
Page-: Again, do you specifically want to differentiate between `null` and `undefined`?  If not then this should be `if @_req?`

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+			clearTimeout(@_timeout)
+			@_req.removeAllListeners()
+			@_req.on('error', ->)
```
Page-: Can you use `_.noop` for the noop func, it's immediately obvious that it's intentional and can be faster (certainly in node 6 as I've seen a fairly decent speedup in the past from reusing an existing noop function vs creating a new)

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+
+		# We want a very low writable high watermark to prevent having many
+		# chunks stored in the writable queue of @_gzip and have them in
+		# @_stream instead. This is desirable because once @_gzip.flush() is
+		# called it will do all pending writes with that flush flag. This is
+		# now what we want though. If there are 100 items in the queue we want
+		# to write all of them with Z_NO_FLUSH and only afterwards do a
+		# Z_SYNC_FLUSH to maximize compression
+		@_gzip = zlib.createGzip(writableHighWaterMark: 1024)
+		@_gzip.on('error', => @_teardown())
```
Page-: If you change
```coffee
_teardown: ->
```
to
```coffee
_teardown: =>
```
then you can make this
```coffee
@_gzip.on('error', @_teardown)
```
along with all the other cases that do `=> @_teardown()`

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+		@_req.flushHeaders()
+
+		# We want a very low writable high watermark to prevent having many
+		# chunks stored in the writable queue of @_gzip and have them in
+		# @_stream instead. This is desirable because once @_gzip.flush() is
+		# called it will do all pending writes with that flush flag. This is
+		# now what we want though. If there are 100 items in the queue we want
+		# to write all of them with Z_NO_FLUSH and only afterwards do a
+		# Z_SYNC_FLUSH to maximize compression
+		@_gzip = zlib.createGzip(writableHighWaterMark: 1024)
```
Page-: What is the unit of the 1024?  It'd be nice to move it to a named constant at the top of the file I think

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+		# only reason for the server to prematurely respond is to
+		# communicate an error. So teardown the connection immediately
+		@_req.on 'response', (res) =>
+			console.log('LogBackend: server responded with status code:', res.statusCode)
+			@_teardown()
+
+		@_req.on('timeout', => @_teardown())
+		@_req.on('close', => @_teardown())
+		@_req.on 'error', (err) =>
+			console.log('LogBackend: unexpected error:', err)
```
Page-: This should probably be `console.error`

package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "mkfifo": "^0.1.5",
+    "ndjson": "^1.5.0",
```
Page-: I'm guessing this should be in `devDependencies` because it doesn't have a native component (I assume?) and so only needs to be installed whilst doing the webpack bundling, cc @pcarranzav

src/logger.coffee
Page-: The constant was a separate thing; it allows us to see the key config vars and such at the top of the file without having to dig into where it's actually used, otherwise you could just hardcode all the other constants in-place by the same argument.  The reason for wanting to know the units was so that I could see if it was a sane value (and I should have created separate threads for the two in retrospect)

src/logger.coffee
Page-: I'd say it's better to go with `?` because right now I expect there to be cases where it's `null` and cases where it's `undefined`, and for them to have specific behaviour between the two.  ie the same argument you're making about wondering if there's a case where it's `undefined` because there appears to be an explicit choice to differentiate between `null`/`undefined`

src/logger.coffee
Page-: ok

src/logger.coffee
```diff
@@ -1,156 +1,146 @@
...
+			@_writable = @_stream.write(JSON.stringify(msg) + '\n')
+			@_flush()
+		else
+			@_dropCount += 1
 
-	logDependent: (msg, { uuid }) =>
-		msg = @_truncateToMaxLength(msg)
-		@apiBinder.logDependent(uuid, msg)
+	log: (msg) ->
+		if @offlineMode or !@publishEnabled
```
CameronDiver: Shouldn't this be `@opts.offlineMode`? I don't see where you assign `@offlineMode`.

src/logger.coffee
pcarranzav: Isn't `0` the default? Looking at https://docs.docker.com/engine/api/v1.29/#operation/ContainerLogs
(and it should be an integer, not a string?)

Also, (I don't see how this would happen here, but) do we want to lose logs from before the supervisor starts? This used to be an issue that users noted/reported often until it was fixed [here](h.ttps://github.com/resin-io/resin-supervisor/pull/394)

src/config.ts
```diff
@@ -258,10 +250,9 @@ class Config extends EventEmitter {
 						if (offlineMode) {
 							return;
 						}
-						if (deviceApiKey == null || logsChannelSecret == null) {
+						if (deviceApiKey == null) {
 							deviceApiKey = deviceApiKey || this.newUniqueKey();
```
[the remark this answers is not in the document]
petrosagg: This will unconditionally call `this.set` which will write to config.json. We don't want to do superfluous writes, especially to config.json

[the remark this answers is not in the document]
Page-: @petrosagg the issue is the `deviceApiKey || ` part is redundant, because `deviceApiKey` is guaranteed to be falsey here

src/logger.coffee
petrosagg: @pcarranzav for some reason it worked correctly with a recent docker but didn't work with older versions. I changed it to `Math.floor(Date.now() / 1000)`

src/logger.coffee
petrosagg: Fixed

src/logger.coffee
petrosagg: It's evident that the code does not care about `undefined` since it attempts to access fields of the object. You can also wonder if there a case for `@_req` being `false` but I don't think this is something the code should be concerned about.

It's an antipattern to encode conditions that cannot happen. `?` should only be used if something can potentially be `undefined`. This is not the case here. We are abusing it because it's shorter to type. Do a grep for `== null` in the supervisor repo and take the typescript code as further evidence.

src/config.ts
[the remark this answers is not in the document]
petrosagg: oh, right. I'll fix that

src/logger.coffee
```diff
@@ -222,9 +200,9 @@ module.exports = class Logger
 					.on 'data', (logLine) =>
 						space = logLine.indexOf(' ')
 						if space > 0
-							msg = { timestamp: logLine.substr(0, space), message: logLine.substr(space + 1), serviceId, imageId }
+							msg = { timestamp: (new Date(logLine.substr(0, space))).getTime(), message: logLine.substr(space + 1), serviceId, imageId }
```
Page-: Do we know this date parsing is valid both on our newer balena devices and older docker devices for the backports?

src/logger.coffee
pcarranzav: @petrosagg but are you sure we want to bring this behavior back?

This was done in this way (getting all logs) on purpose, because otherwise the first logs are always lost. Getting only the logs after supervisor start only fixes the case of a supervisor restart (which is rare, except on support actions and on supervisor updates) but reintroduces a known issue for all other cases.

Or is there something else I'm missing?

PR #703, opened 2018-07-17
Support changing dns, ntp and country from env vars

CameronDiver:
Added support for the `.dnsServers`, `.ntpServers` and `.country` config.json fields.

Also add some whitespace as the code was starting to get pretty blocky.

Closes: https://github.com/resin-io/resin-supervisor/issues/702


src/config.ts
```diff
@@ -45,6 +45,9 @@ class Config extends EventEmitter {
 		supervisorOfflineMode: { source: 'config.json', default: false },
 		hostname: { source: 'config.json', mutable: true },
 		persistentLogging: { source: 'config.json', default: false, mutable: true },
+		dnsServers: { source: 'config.json', default: '', mutable: true },
```
Page-: A default of an empty string vs undefined?  Does this mean the property will be added but be an empty string?

src/device-config.coffee
```diff
@@ -142,16 +148,20 @@ module.exports = class DeviceConfig
 					checkTruthy(a) == checkTruthy(b)
 				'int': (a, b) ->
 					checkInt(a) == checkInt(b)
+				'string': (a, b) ->
+					a.toString().trim() == b.toString().trim()
```
Page-: If the previous or new value doesn't exist won't this throw?

test/05-device-state.spec.coffee
```diff
@@ -47,6 +47,9 @@ testTarget1 = {
 			'RESIN_SUPERVISOR_POLL_INTERVAL': '60000'
 			'RESIN_SUPERVISOR_VPN_CONTROL': 'true'
 			'RESIN_SUPERVISOR_PERSISTENT_LOGGING': 'false'
+			'RESIN_SUPERVISOR_DNS_SERVERS': ''
```
Page-: Is there a test for removing config vars?

Page-: Also as a note I think it's impossible to set an empty string as the value in the ui

imrehg:
@CameronDiver how's progress on this one? Just checking!

CameronDiver:
@imrehg This needs further discussion, based on where the source of truth should be between the device and API for this. I will add on-device API calls for the time being, until a route forward is decided.

PR #710, opened 2018-07-23
Refactor and improve compose networks support

CameronDiver:
A massive PR because of the typescript changes, but mostly only the networking additions.

Every supported compose field has been extensively tested, by both myself and @hedss


src/compose/network.ts
```diff
@@ -0,0 +1,254 @@
...
+					);
+				}
+
+				// We have a network with the same config and name already created -
+				// we can skip this.
+			})
+			.catch(NotFoundError, () => {
+				return this.docker.createNetwork({
+					Name: this.getDockerName(),
+					Driver: this.config.driver,
```
pcarranzav: I recommend also setting the CheckDuplicate option here (I was planning on doing it in https://github.com/resin-io/resin-supervisor/pull/633/files#diff-d40727b4ebe673427c629249a8bb1db4R108), as it will help avoid some of the weird cases when there's two networks with the same name.

src/compose/utils.ts
```diff
@@ -0,0 +1,438 @@
...
+		// Ensure that the user hasn't added 'supervisor0' to the service's list
+		// of networks
+		delete service.config.networks[constants.supervisorNetworkInterface];
+	}
+
+}
+
+export function serviceUlimitsToDockerUlimits(ulimits: ServiceConfig['ulimits'] | null | undefined)
+	: Array<{ Name: string, Soft: number, Hard: number }> {
+
```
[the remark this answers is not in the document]
CameronDiver: I left this one becase the function was super dense vertically, and I think it's a bit easier to parse with some vertical whitespace there. Also happy to remove if you disagree. Prettier should handle this soon anyway.

src/compose/utils.ts
```diff
@@ -0,0 +1,438 @@
...
+export function serviceRestartToDockerRestartPolicy(restart: string): { Name: string, MaximumRetryCount: number } {
+	return {
+		Name: restart,
+		MaximumRetryCount: 0,
+	};
+}
+
+export function serviceNetworksToDockerNetworks(networks: ServiceConfig['networks'])
+	: Dockerode.ContainerCreateOptions['NetworkingConfig'] {
+
```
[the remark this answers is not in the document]
CameronDiver: Same as https://github.com/resin-io/resin-supervisor/pull/710#discussion_r219437214

test/12-logger.spec.coffee
```diff
@@ -105,3 +105,12 @@ describe 'Logger', ->
...
 			expect(msg).to.deep.equal({ message: 'Hello there!', timestamp: 0, isSystem: true })
+
+	it 'should support non-tty log lines', ->
+		message = '\u0001\u0000\u0000\u0000\u0000\u0000\u0000?2018-09-21T12:37:09.819134000Z this is the message'
+		buffer = Buffer.from(message)
+
+		expect(Logger.extractContainerMessage(buffer)).to.deep.equal({
+			message: 'this is the message',
+			timestamp: 1537533429819
+		})
```
hedss: NL

test/data/docker-states/entrypoint/compose.json
```diff
@@ -9,5 +9,6 @@
 	"releaseId": 597007,
 	"serviceId": 43697,
 	"commit": "ff300a701054ac15281de1f9c0e84b8c",
-	"imageName": "registry2.resin.io/v2/bf9c649a5ac2fe147bbe350875042388@sha256:3a5c17b715b4f8265539c1a006dd1abdd2ff3b758aa23df99f77c792f40c3d43"
+	"imageName": "registry2.resin.io/v2/bf9c649a5ac2fe147bbe350875042388@sha256:3a5c17b715b4f8265539c1a006dd1abdd2ff3b758aa23df99f77c792f40c3d43",
+	"tty": true
 }
```
hedss: NL

test/data/docker-states/simple/compose.json
```diff
@@ -7,5 +7,6 @@
 	"releaseId": 572579,
 	"serviceId": 43697,
 	"commit": "b14730d691467ab0f448a308af6bf839",
-	"imageName": "registry2.resin.io/v2/8ddbe4a22e881f06def0f31400bfb6de@sha256:09b0db9e71cead5f91107fc9254b1af7088444cc6da55afa2da595940f72a34a"
+	"imageName": "registry2.resin.io/v2/8ddbe4a22e881f06def0f31400bfb6de@sha256:09b0db9e71cead5f91107fc9254b1af7088444cc6da55afa2da595940f72a34a",
+	"tty": true
 }
```
hedss: NL

src/application-manager.coffee
```diff
@@ -142,11 +143,25 @@ module.exports = class ApplicationManager extends EventEmitter
 			cleanup: (step) =>
 				@images.cleanup()
 			createNetworkOrVolume: (step) =>
-				model = if step.model is 'volume' then @volumes else @networks
-				model.create(step.target)
+				if step.model is 'network'
+					# TODO: These step targets should be the actual compose objects,
+					# rather than recreating them
+					Network.fromComposeObject({ @docker, @logger },
```
pcarranzav: I see you have a lot of Network.fromComposeObject calls in several places. Wouldn't it be better to just call this when you're getting the current and target state (i.e. in getCurrentForComparison, and in getTarget), and then always work with Network objects? This is what we do with Service (we create Service objects when we get the current and target states, and therefore only call `new Service` or `Service.fromWhatever` in those two specific places.

CameronDiver: Good point, I should have made that a FIXME, not a TODO because I did plan to tacklet it - thanks dude!

PR #721, opened 2018-08-07
Fix: Restart a service when it's memory limit changes

CameronDiver:
Change-type: patch
Closes: #720


CameronDiver:
retest

PR #724, opened 2018-08-13
fix: Apply device name before generating container config

CameronDiver:
Currently the service has the device name applied after the docker
config is generated. This means that is has no effect until the next
restart.

This commit ensures that the device name is applied before the docker
config is generated, meaning that the env var gets applied to the
upcoming container.

Change-type: patch


src/compose/service-manager.coffee
```diff
@@ -101,24 +101,25 @@ module.exports = class ServiceManager extends EventEmitter
...
+				conf = service.toContainerConfig()
+				nets = service.extraNetworksToJoin()
+
 				@logger.logSystemEvent(logTypes.installService, { service })
 				@reportNewStatus(mockContainerId, service, 'Installing')
 
 				@docker.createContainer(conf)
-			.tap (container) =>
-				service.containerId = container.id
+				.tap (container) =>
```
dfunckt: I think you don't need to indent the `tap` here.

CameronDiver: it's because of the use of `nets` here: https://github.com/resin-io/resin-supervisor/pull/724/files#diff-e7a20470a15d19708363a4eb6965b966R119

dfunckt: Ah yeah 👍

CameronDiver:
retest

PR #726, opened 2018-08-15
dev: Add package.json tasks to aid debugging supervisor tests

CameronDiver:
Running mocha on the coffee and ts sources resulted in unreliable
sourcemaps, and breakpoints not working, among other things. To solve
this, this commit adds package.json tasks which will build both the
coffeescript and typescript files with their sourcemaps and places it
into a build directory. It also copies over other dependent files, such
as test data and migrations.

Another task adds a mocha debugging command, which can be connected to
using a node debugger.

Another tasks runs mocha without a debugging interface, but also without
running linters etc first. This enables the tests to be run much faster.

Change-type: minor


package.json
```diff
@@ -12,6 +12,9 @@
     "build": "webpack",
     "lint": "resin-lint --typescript src/ test/",
     "test": "npm run lint && JUNIT_REPORT_PATH=report.xml mocha --exit -r ts-node/register -r coffee-script/register -r register-coffee-coverage test/*.{js,coffee} && npm run coverage",
+    "test:fast": "mocha --exit -r ts-node/register -r coffee-script/register test/*.{js,coffee}",
+    "test:build": "tsc && coffee -m -c -o build . && cp -r test/data build/test/ && cp -r src/migrations build/src && cp package.json build",
```
dfunckt: I think both `tsc` and `coffee ...` will compile everything under the current directory, even node_modules -- is that intended?

CameronDiver: `tsc` won't due to the `tsconfig.json`, but I'll check the `coffee` command

CameronDiver: It doesn't seem to, I guess it adds checks for the `node_modules` directory.

PR #727, opened 2018-08-15
Add more typescript support

CameronDiver:
This PR changes some of the typescript interfaces, ready for https://github.com/resin-io/resin-supervisor/pull/710 to be finished off and put up for real review.

I put these up separately to make the above PR smaller.


typings/docker-toolbelt.d.ts
```diff
@@ -0,0 +1,27 @@
+declare module 'docker-toolbelt' {
```
Page-: Can you put these types directly in docker-toolbelt please since we control that repo and it would be beneficial for everyone that way

src/lib/log-types.ts
```diff
@@ -0,0 +1,171 @@
+export = {
+	stopService: {
```
Page-: I think the export const pattern would make sense here, ie
```typescript
export const stopService = {
```

typings/docker-toolbelt.d.ts
Page-: It's probably still an improvement even then

CameronDiver: I just checked and the types are complete, not sure why I didn't add it, but I'll do that now.

CameronDiver:
@resin-ci retest

PR #729, opened 2018-08-16
ux: Warn on invalid device name when trying to start a service

CameronDiver:
Device names with newlines cause reboot loops, due to newlines not being
supported by docker. This PR will warn when a device name contains a
newline.

Change-type: patch


CameronDiver:
@shaunmulligan the UX here look ok to you?

shaunmulligan:
@CameronDiver seems good to me 👍

src/lib/validation.ts
```diff
@@ -154,6 +154,15 @@ export function isValidLabelsObject(obj: LabelObject): boolean {
 	});
 }
 
+export function isValidDeviceName(name: string): boolean {
+	// currently the only disallowed value in a device name is a newline
+	const newline = name.indexOf('\n') !== -1;
```
dfunckt: This will also consider invalid a name such as `I\\nwant\\nnewlines` even though the new line is escaped and (I expect) shouldn't really cause issues. I don't suggest we should handle this edge-case though -- I see no real use-case -- just raising it.

CameronDiver: It won't match this, because `\\n` won't be converted to the newline character because of the escaping before it:
```
[10:06] ➜  ~ node
> a = "I\\nwant\\nnewlines"
'I\\nwant\\nnewlines'
> a.indexOf('\n')
-1

PR #733, opened 2018-08-27
Further typescipt conversions

CameronDiver:
* Convert logging module
* Convert event-tracker module (with external typings)
* Export specificly typed events from log-types module
* Fix a bug where the enable flag was not respsected in the logging module on instantiation


src/event-tracker.ts
```diff
@@ -0,0 +1,95 @@
...
+		this.logEvent('Event:', event, JSON.stringify(properties));
+		if (this.client == null) {
+			return;
+		}
+
+		properties = this.assignDefaultProperties(properties);
+		this.client.track(event, properties);
+	}
+
+	private logEvent(...args: string[]) {
```
Page-: I think `any` is actually the more accurate type for forwarding to console.log, unless you're intentionally restricting which wouldn't necessarily be a bad thing

CameronDiver: Yeah, I think the onus should be on the caller to ensure that the data will be printed in a useful way, and not just how `console.log` will format it (and there's `toString` for that). If these is any coffeescript code which passes non-strings, it will still work, and any typescript conversions coming up will pick up the type changes and have to conform.

src/logger.ts
```diff
@@ -0,0 +1,428 @@
...
+					.then((stream) => {
+						this.attached[streamType][containerId] = true;
+
+						stream
+						.on('error', (err) => {
+							console.error('Error on container logs', err, err.stack);
+							this.attached[streamType][containerId] = false;
+						})
+						.pipe(es.split())
+						.on('data', (logBuf: Buffer | string) => {
```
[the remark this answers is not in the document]
CameronDiver: Done.

src/logger.ts
```diff
@@ -0,0 +1,428 @@
...
+									imageId,
+								};
+								if (streamType === OutputStream.Stderr) {
+									message.isStdErr = true;
+								}
+								this.log(message);
+							}
+						})
+						.on('error', (err) => {
+							console.error('Error on container logs', err, err.stack);
```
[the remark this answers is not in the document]
CameronDiver: Ah you're right, I was just copying directly from the CS, but yeah we should remove.

typings/json-mask.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module 'json-mask' {
+	function mask(obj: { [key: string]: any }, mask: string): { [key: string]: any };
```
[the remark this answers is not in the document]
CameronDiver: Done.

src/logger.ts
```diff
@@ -0,0 +1,427 @@
...
+		deviceApiKey: string,
+		offlineMode: boolean,
+		enableLogs: boolean,
+	) {
+		this.backend = new ResinLogBackend(apiEndpoint, uuid, deviceApiKey);
+		this.backend.offlineMode = offlineMode;
+		this.backend.publishEnabled = enableLogs;
+	}
+
+	public enable(value: boolean) {
```
pcarranzav: The `value` here is actually expected to be string or boolean (and I think how we call it in deviceState it's always a string), and that's the reason we use checkTruthy on it.

But since you're changing `init` so that the caller is the one that does checkTruthy, you might want to do the same here, and change https://github.com/resin-io/resin-supervisor/blob/master/src/device-state.coffee#L161 to call checkTruthy instead of this function.

CameronDiver: Yeah I was thinking that, what I've been doing is tightening the types for when it's called by typescript, but keeping the function validation present, for when these are called by the existing typescript. It does make sense though that we would switch it out in the coffeescript too, will do that now.

typings/json-mask.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module 'json-mask' {
+
+	type Dictionary<T> = { [key: string]: T };
```
Page-: When I was talking about a dictionary type I was talking about having it shared, ie
```typescript
declare global {
	interface Dictionary<T> {
		[key: string]: T;
	}
}

File: src/logger.ts [hunk 7]
```diff
@@ -0,0 +1,432 @@
...
+			});
+	}
+
+	public attach(
+		docker: Docker,
+		containerId: string,
+		serviceInfo: { serviceId: string, imageId: string },
+	): Bluebird<void> {
+		return Bluebird.using(this.lock(containerId), () => {
+			return this.attachStream(docker, OutputStream.Stdout, containerId, serviceInfo)
```
Page-: Any reason this isn't a `Promise.all([attachStdout, attachStderr])`?  Do we have to attach in order?

src/logger.ts
```diff
@@ -0,0 +1,432 @@
...
+
+
+		// We want a very low writable high watermark to prevent having many
+		// chunks stored in the writable queue of @_gzip and have them in
+		// @_stream instead. This is desirable because once @_gzip.flush() is
+		// called it will do all pending writes with that flush flag. This is
+		// not what we want though. If there are 100 items in the queue we want
+		// to write all of them with Z_NO_FLUSH and only afterwards do a
+		// Z_SYNC_FLUSH to maximize compression
+		this.gzip = zlib.createGzip({ writableHighWaterMark: 1024 } as any);
```
Page-: Is this `any` definitely necessary?

src/logger.ts
```diff
@@ -0,0 +1,432 @@
...
+		if (!_.isObject(message)) {
+			return;
+		}
+
+		message = _.assign({
+			timestamp: Date.now(),
+			message: '',
+		}, message);
+
+		if (!(message.isSystem || message.serviceId != null)) {
```
Page-: I think this would be more readable as
```typescript
if (!message.isSystem && message.serviceId == null)
```

Page-: Also do we want to be silently dropping logs? (Both here and for the !isObject above)

src/logger.ts
```diff
@@ -0,0 +1,432 @@
...
+	headers: Dictionary<string>;
+}
+
+type LogMessage = Dictionary<any>;
+
+abstract class LogBackend {
+	public offlineMode: boolean;
+	public publishEnabled: boolean = true;
+
+	public abstract log(message: Dictionary<any>): void;
```
Page-: I think this should be `message: LogMessage` - I know they're the same but changing it means that when `LogMessage` gets tightened then this automatically does as well

src/logger.ts
CameronDiver: I'm not sure, I was keeping it the same as https://github.com/resin-io/resin-supervisor/pull/733/files#diff-0f68e03b8d0f7c87ec71275f52139ee9L213

src/logger.ts
CameronDiver: Yeah, the typings for zlib don't include `writableHighWaterMark`.

src/logger.ts
CameronDiver: Agreed

CameronDiver: I was just trying to keep it 1:1 with the coffeescript, we can handle that in a separate PR if we want it: https://github.com/resin-io/resin-supervisor/pull/733/files#diff-0f68e03b8d0f7c87ec71275f52139ee9L137

src/logger.ts
CameronDiver: Agreed, good catch

src/logger.ts
Page-: Have you tried using the node 6 typings?  It's possible they'll include it (and we should be using those instead to not accidentally try to use node 10 things).  Aside from that you should be able to extend the interface with
```typescript
declare module 'zlib' {
	export interface ZlibOptions {
		writableHighWaterMark: number;
	}
}

src/logger.ts
Page-: Ok

src/logger.ts
Page-: Ok, doesn't have to be this PR, but cc @pcarranzav do you know the answer to this?

CameronDiver:
@resin-ci retest

PR #737, opened 2018-08-28
fix: Dont bind service class methods in declaration

CameronDiver:
This was causing a bug where the applications were cloned when
restarting all of them. These clones did not carry over the binds, so
when starting the containers back up, two different versions of `this`
were being used.

Change-type: patch
Closes: #736


CameronDiver:
@resin-ci retest

brownjohnf:
@resin-ci retest

PR #738, opened 2018-09-02
Add mixpanel rate limiting, and supervisor version reporting

CameronDiver:
This depends on https://github.com/resin-io/resin-supervisor/pull/733 being merged first.

### DIFF
--- a/src/event-tracker.ts
+++ b/src/event-tracker.ts
@@ -1,6 +1,7 @@
 import * as Bluebird from 'bluebird';
 import mask = require('json-mask');
 import * as _ from 'lodash';
+import * as memoizee from 'memoizee';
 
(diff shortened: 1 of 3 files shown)


src/event-tracker.ts
```diff
@@ -78,9 +82,16 @@ export class EventTracker {
 		}
 
 		properties = this.assignDefaultProperties(properties);
-		this.client.track(event, properties);
+		this.debouncedLogger(event)(properties);
 	}
 
+	private debouncedLogger = _.memoize((event) => {
```
Page-: You can use `memoizee` with `primitive: true` for this - we already have memoizee installed in the supervisor and it's what we normally use for memoization

CameronDiver: Done - merging.

PR #759, opened 2018-10-16
fix: overlay compose fields on image for healthcheck

CameronDiver:
This commit changes a bug where the compose healthcheck would always
overwrite the healthcheck set by the image - even if no compose
healthcheck exists.

Change-type: patch

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -190,8 +190,13 @@ function getNanoseconds(timeStr: string): number {
 
 export function composeHealthcheckToServiceHealthcheck(
 	healthcheck: ComposeHealthcheck | null | undefined,
-): ServiceHealthcheck {
-	if (healthcheck == null || healthcheck.disable) {
+): ServiceHealthcheck | { } {
+
+	if (healthcheck == null) {
+		return  { };
+	}
+
+	if (healthcheck.disable) {
 		return { test: [ 'NONE' ] };
 	}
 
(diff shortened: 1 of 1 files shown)


src/compose/utils.ts
```diff
@@ -230,7 +230,8 @@ export function getHealthcheck(
 		composeHealthcheck,
 	);
 
-	return _.defaults(composeServiceHealthcheck, imageServiceHealthcheck);
+	// Overlay any compose healthcheck fields on the image healthchecks
+	return _.assign(imageServiceHealthcheck, composeServiceHealthcheck);
```
pcarranzav: This will still overwrite the image's healthcheck if no healthcheck is provided, because the compose healthcheck will have `test: [ 'NONE' ]` if it was empty. I think you should remove the `healthcheck == null` check in line 194, and have this line be:

```
return _.assign({ test: [ 'NONE' ] }, imageServiceHealthcheck, composeServiceHealthcheck);
```

src/compose/utils.ts
```diff
@@ -190,8 +190,13 @@ function getNanoseconds(timeStr: string): number {
 
 export function composeHealthcheckToServiceHealthcheck(
 	healthcheck: ComposeHealthcheck | null | undefined,
-): ServiceHealthcheck {
-	if (healthcheck == null || healthcheck.disable) {
+): ServiceHealthcheck | { } {
```
pcarranzav: nit: Should there be no space in the `{ }`?

CameronDiver: Good catch

PR #763, opened 2018-10-17
Replace io.resin labels (and their env vars) with io.balena equivalents

pcarranzav:
But we keep backwards compatibility by normalizing existing io.resin labels
into io.balena ones, and adding both RESIN_ and BALENA_ env vars for these features.

Change-Type: minor

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -300,45 +300,48 @@ export function addFeaturesFromLabels(
 	service: Service,
 	options: DeviceMetadata,
 ): void {
-	if (checkTruthy(service.config.labels['io.resin.features.dbus'])) {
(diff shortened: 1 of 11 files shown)


src/compose/utils.ts
```diff
@@ -442,3 +445,17 @@ export function normalizeNullValues(obj: Dictionary<any>): void {
 		}
 	});
 }
+
+export function normalizeLabels(
+	labels: { [key: string]: string },
+): { [key: string]: string } {
+	const legacyLabels = _.mapKeys(_.pickBy(labels, (_v, k) => _.startsWith(k, 'io.resin')), (_v, k) => {
```
petrosagg: let's switch to `io.resin.` as the predicate

PR #769, opened 2018-10-18
Rename most of the documentation and variable names from resin to balena

pcarranzav:
* References to resin tools are replaced with the balena counterparts (e.g. balena CLI, balenaCloud, etc)
* Variable names with resin renamed to use balena
* Support for BALENA_PROXYVISOR_HOOK_RECEIVER added in the proxyvisor
* References to the resin/scratch delta source are removed, since we've been doing regular pulls instead of delta from scratch for a while now.


pcarranzav:
Depends-on: https://github.com/resin-io/resin-cli/pull/997

(due to the docs referencing balena CLI)

src/lib/docker-utils.coffee
```diff
@@ -76,6 +75,12 @@ module.exports = class DockerUtils extends DockerToolbelt
 			log("Unsupported delta version: #{version}. Falling back to regular pull")
 			return @fetchImageWithProgress(imgDest, fullDeltaOpts, onProgress)
 
+		# Since the supervisor never calls this function without a source anymore,
+		# this should never happen, but we handle it anyways.
+		if !deltaSource?
+			log("Delta from scratch unsupported. Falling back to regular pull")
```
CameronDiver: I would just say something like "Performing regular pull due to lack of delta source" or sth. It might scare people seeing unsupported.

pcarranzav: Good point, will change it

PR #777, opened 2018-10-23
fix: Run checkTruthy on config values before using them

CameronDiver:
Change-type: patch

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -143,7 +143,7 @@ module.exports = class ApplicationManager extends EventEmitter
 			saveImage: (step) =>
 				@images.save(step.image)
 			cleanup: (step) =>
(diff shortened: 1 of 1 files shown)


src/application-manager.coffee
```diff
@@ -905,7 +905,7 @@ module.exports = class ApplicationManager extends EventEmitter
 				@networks.supervisorNetworkReady()
 				@config.get('delta')
 				(cleanupNeeded, availableImages, downloading, supervisorNetworkReady, delta) =>
-					conf = { delta, localMode }
+					conf = _.mapValues({ delta, localMode }, (v) -> checkTruthy(v))
 					if localMode
```
pcarranzav: This should be conf.localMode

CameronDiver: good catch

PR #778, opened 2018-10-23
Two more bugfixes for next release

pcarranzav:
Both only affecting updates from older supervisor / OS versions.

* Symbolic link when /tmp/resin-supervisor exists
* Properly compare images to save the correct image metadata to the db

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -200,8 +200,7 @@ module.exports = class ApplicationManager extends EventEmitter
 			@services.getStatus()
 			@images.getStatus(localMode)
 			@config.get('currentCommit')
-			@db.models('app').select([ 'appId', 'releaseId', 'commit' ])
-			(services, images, currentCommit, targetApps) ->
+			(services, images, currentCommit) ->
 				apps = {}
 				dependent = {}
 				releaseId = null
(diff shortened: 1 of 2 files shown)


src/application-manager.coffee
```diff
@@ -808,10 +807,12 @@ module.exports = class ApplicationManager extends EventEmitter
 			_.map app.services, (service) ->
 				img = _.find(available, { dockerImageId: service.config.image, imageId: service.imageId }) ? _.find(available, { dockerImageId: service.config.image })
 				return _.omit(img, [ 'dockerImageId', 'id' ])
+		allImageDockerIdsForTargetApp = (app) -> _.filter(_.map(app.services, (svc) -> [ svc.imageName, svc.config.image] ), (img) -> !_.isNull(img[1]))
```
Page-: This would be more readable as `_.map(...).filter(...).value()`

richbayliss: +1

Page-: Also `img[1] != null` would be a bit clearer - does null vs undefined matter or was the `!== null` accidental?

Page-: @pcarranzav I think you missed this due to the other comment

pcarranzav: Done

pcarranzav: Done

src/application-manager.coffee
```diff
@@ -823,8 +824,10 @@ module.exports = class ApplicationManager extends EventEmitter
 		imagesToSave = []
 		if !localMode
 			imagesToSave = _.filter targetImages, (targetImage) =>
-				_.some(available, (availableImage) => @images.isSameImage(availableImage, targetImage)) and
-					!_.some availableWithoutIds, (img) -> _.isEqual(img, targetImage)
+				_.some(available, (availableImage) =>
```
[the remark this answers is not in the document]
pcarranzav: Done

PR #780, opened 2018-10-25
test: Use longer delays in logger tests

pcarranzav:
We've seen at least one case where the 10ms delay
wasn't enough, causing the tests to fail where they shouldn't.

(Ideally we should find a better way to detect when the streams
have finished propagating, but this will avoid failures in the meantime).

Change-type: patch


pcarranzav:
(the test that failed was https://circleci.com/gh/resin-io/resin-supervisor/11019 )

PR #787, opened 2018-11-02
V2 service device api improvments

CameronDiver:
This PR adds a couple of fixes and also adds the `serviceName` parameter to v2 service endpoints, to make it a bit more user friendly.

The documentation will be updated in my other pr: https://github.com/balena-io/balena-supervisor/pull/774

### DIFF
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -5,7 +5,11 @@ import { fs } from 'mz';
 
 import { ApplicationManager } from '../application-manager';
 import { Service } from '../compose/service';
-import { appNotFoundMessage, serviceNotFoundMessage } from '../lib/messages';
+import {
+	appNotFoundMessage,
+	serviceNotFoundMessage,
+	v2ServiceEndpointInputErrorMessage,
+} from '../lib/messages';
 import { checkTruthy } from '../lib/validation';
 import { doPurge, doRestart, serviceAction } from './common';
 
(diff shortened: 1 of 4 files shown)


src/device-api/v2.ts
```diff
@@ -42,7 +46,20 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+					// Work if we have a service name or an image id
+					if (imageId == null) {
+						if (serviceName == null) {
+							throw new Error(v2ServiceEndpointInputErrorMessage);
+						}
+					}
+
+					let service: Service | null;
+					if (imageId != null) {
+						service = _.find(app.services, { imageId }) as Service | null;
```
Page-: Is `app.services` not typed?  If it is typed as `Service[]` then I would expect the ` as Service | null` to be redundant

CameronDiver: It is typed, but it looks like the lodash types are extremely permissive here. I tried to get away with not needing this but I was getting weird results such as `number is not assignable to Service | null`.

Page-: If you use
```typescript
let service: Service | undefined;
if (imageId != null) {
	service = _.find(app.services, (service) => service.imageId === imageId);
} else {
	service = _.find(app.services, (service) => service.serviceName === serviceName);
}
```
then it'll work, and `Service | undefined` is the correct typing instead of `Service | null`

PR #790, opened 2018-11-02
Let typescript do the json resolution so we get proper types

Page-:
Change-type: patch


Page-:
Looks like this depends on #788

PR #793, opened 2018-11-05
fix: Also lint typescript declaration files

CameronDiver:
Change-type: patch

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -2,7 +2,7 @@
 	"*.coffee": [
 		"resin-lint"
 	],
-	"*.ts": [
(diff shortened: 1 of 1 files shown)


.lintstagedrc
```diff
@@ -2,7 +2,7 @@
 	"*.coffee": [
 		"resin-lint"
 	],
-	"*.ts": [
+	"*.{d.ts,ts}": [
```
Page-: `*.ts` should match both `abc.ts` and `xyz.d.ts`, so I don't think this change is necessary, could you recheck/point to docs where it's failing?

PR #799, opened 2018-11-05
Update to webpack 4

Page-:
Generated code when unminified is equivalent (the webpack 4 version has slightly better code elim in cases like `if (true) {} else ...`/`if (false) { ... }`), the minified version still needs to be tested though in case changes in the minification are significant

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,7 +3,6 @@ var path = require('path');
 var fs = require('fs');
 var _ = require('lodash');
 var path = require('path');
-var UglifyPlugin = require("uglifyjs-webpack-plugin");
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 
 var externalModules = [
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -66,10 +62,8 @@ module.exports = function (env) {
 			path.resolve(__dirname, 'src/migrations')
 		)
 	]
-	if (env == null || !env.noOptimize) {
```
pcarranzav: This removes the noOptimize option we use in dindctl to debug locally with unminified code. Is there a way to do a similar thing now? Maybe not setting `mode` to "production" if the option is set?

Page-: Yeah, it only defaults to true in production mode, see https://webpack.js.org/configuration/optimization/#optimization-minimize - you can explicitly disable it even in production mode with
```json
  optimization: {
    minimize: false
  }
```
though.  The `mode: 'production'` vs `mode: 'development'` flag triggers a lot more options for debuggability though so it probably makes sense to switch the mode rather than individual flags: https://webpack.js.org/concepts/mode/

Page-: Readded support for `env.noOptimize` @pcarranzav

PR #806, opened 2018-11-13
docs: Clarify that older supervisors use RESIN_ variables

pcarranzav:
We add a note on state endpoint docs to use RESIN_ variables. Since users will likely navigate directly to each endpoint, we add the note on the docs for each endpoint, except for endpoints added in >= 7.21.0 (since that version never made it to production).

We also switch some variables that were still referenced as RESIN_ to now use BALENA_.

Change-type: patch


pcarranzav:
Ah, good catch @hedss, and good idea on using bold. Fixed.

PR #814, opened 2018-11-22
fix: When updating from a legacy supervisor, use updated resource ids…

pcarranzav:
… and image URL from the API

When updating from old supervisors (<7.0.0), we've been so far using a fake id 1 for serviceId, imageId
and releaseId since these were not available in the old supervisor. This causes problems when the supervisor
tries to report these values to the API. Moreover, the app from the legacy supervisor has an image URL
that doesn't include the content hash - this causes the supervisor to believe the image is not really downloaded
and try to fetch it again.

To fix these issues, we add a request to the API when the supervisor starts up and detects that there's a legacy
app that needs to be normalised. We fetch the appropriate release, and use it to populate the resource ids
and the updated image URL.

This should avoid the unnecessary image download, and errors reporting target state after an update.

Change-type: patch

### DIFF
--- a/src/device-state.coffee
+++ b/src/device-state.coffee
@@ -13,7 +13,7 @@ validation = require './lib/validation'
 systemd = require './lib/systemd'
 updateLock = require './lib/update-lock'
 { singleToMulticontainerApp } = require './lib/migration'
-{ ENOENT, EISDIR } = require './lib/errors'
+{ ENOENT, EISDIR, NotFoundError } = require './lib/errors'
 
 DeviceConfig = require './device-config'
 ApplicationManager = require './application-manager'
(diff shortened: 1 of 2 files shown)


src/device-state.coffee
```diff
@@ -143,11 +144,94 @@ module.exports = class DeviceState extends EventEmitter
...
-	normaliseLegacy: =>
+	migrateLegacyApps: (balenaApi) =>
+		console.log('Migrating ids for legacy app...')
+		@db.models('app').select()
+		.then (apps) =>
+			if apps.length == 0
+				console.log('No app to migrate')
+				return
+			app = apps[0]
+			services = JSON.parse(app.services)
```
CameronDiver: Might be worth adding a specific erorr here pointing to an error with the super database, helping users and support agents, e.g.
```
try
	services = JSON.parse(app.services)
catch
	throw new Error('Applications table in database is corrupted')
```
I'm not super bothered about this, but could be useful. Might be good to add it as a `TypedError` too

src/device-state.coffee
```diff
@@ -143,11 +144,94 @@ module.exports = class DeviceState extends EventEmitter
...
+						status: 'success'
+					expand:
+						contains__image: [ 'image' ]
+			)
+			.then (releasesFromAPI) =>
+				if releasesFromAPI.length == 0
+					throw new Error('No compatible releases found in API')
+				release = releasesFromAPI[0]
+				releaseId = release.id
+				image = release.contains__image[0].image[0]
```
CameronDiver: We should check that images are returned from the api here, there always should be, but it's worth checking.
e.g.
```
image = _.get(relase, 'contains__image[0].image[0]')
throw new Error(...) if not image?

PR #815, opened 2018-11-22
compose: Change default grace period to 10s

CameronDiver:
Change-type: patch

Front logo [Front conversations](https://app.frontapp.com/open/top_fjj7)

### DIFF
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -151,7 +151,7 @@ export class Service {
 		delete config.shmSize;
 
 		// time strings
-		let stopGracePeriod = 0;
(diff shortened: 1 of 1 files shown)


src/compose/service.ts
```diff
@@ -493,7 +493,7 @@ export class Service {
 			extraHosts: container.HostConfig.ExtraHosts || [],
 			ulimits,
 			stopSignal: (container.Config as any).StopSignal || '',
```
shaunmulligan: @CameronDiver I think as @imrehg mentioned, we need to also make sure the default stopSignal == SIGTERM otherwise things don't work as expected.

PR #817, opened 2018-11-23
Switch from resin-sync to balena-sync, update knex, sqlite3, lodash, coffee-coverage

pcarranzav:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,6 @@
     "network-checker": "^0.1.1",
     "pinejs-client": "^2.4.0",
     "prettier": "^1.14.3",
-    "register-coffee-coverage": "0.0.1",
     "request": "^2.51.0",
     "resin-lint": "^2.0.1",
     "resin-register-device": "^3.0.0",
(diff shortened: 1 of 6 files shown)


pcarranzav:
This should also close #838

CameronDiver:
What are we waiting on for this to be merged @pcarranzav ?

pcarranzav:
@CameronDiver just needed some time to give it a test run before merging, going for it now

pcarranzav:
I had to add a small fix because a change in knex had broken a necessary hack we have for webpack (check the last commit). Now it works.
@Page- @CameronDiver still approve?

package.json
```diff
@@ -45,6 +45,7 @@
     "body-parser": "^1.12.0",
     "buffer-equal-constant-time": "^1.0.1",
     "chai-events": "0.0.1",
+    "coffee-coverage": "^3.0.0",
```
CameronDiver: We no longer need coffee-coverage actually, because the tests build to javascript and the coverage is run over those files with the maps.

PR #820, opened 2018-11-27
Convert network module to typescript

CameronDiver:


src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+	// - custom docker network bridges (br- + 12 hex characters)
+	return _(os.networkInterfaces())
+		.omitBy((_interfaceFields, interfaceName) => IP_REGEX.test(interfaceName))
+		.map(validInterfaces => {
+			return _(validInterfaces)
+				.pickBy({ family: 'IPv4' })
+				.map('address')
+				.value();
+		})
+		.flatten()
```
Page-: You can combine this with the `.map` above as `.flatMap`

src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+	},
+);
+
+export function enableConnectivityCheck(enable: boolean) {
+	enable = checkTruthy(enable) || true;
+	enableCheck(enable);
+	console.log(`Connectivity check enabled: ${enable}`);
+}
+
+export const connectivityCheckEnabled = Bluebird.method(() => {
```
Page-: If you drop the `{}` you can also drop the `return`

src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+					console.log('Waiting for connectivity...');
+					blink.pattern.start(networkPattern);
+				}
+			},
+		);
+	},
+);
+
+export function enableConnectivityCheck(enable: boolean) {
+	enable = checkTruthy(enable) || true;
```
Page-: This is equivalent to `enable = true` atm, `x ? y` and `x || y` are very different beasts and we generally only use `x ? y` when we specifically need that behavior

src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+let isConnectivityCheckPaused = false;
+let isConnectivityCheckEnabled = true;
+
+function checkHost(
+	opts: networkCheck.ConnectOptions,
+): boolean | PromiseLike<boolean> {
+	if (!isConnectivityCheckEnabled || isConnectivityCheckPaused) {
+		return true;
+	}
+	return networkCheck.checkHost(opts);
```
zvin: NIT: this could be
```typescript
return (
    !isConnectivityCheckEnabled ||
    isConnectivityCheckPaused ||
    networkCheck.checkHost(opts)
)
```

src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+
+		customMonitor(
+			{
+				host: parsedUrl.hostname,
+				port: port || (parsedUrl.protocol === 'https' ? 443 : 80),
+				path: parsedUrl.path || '',
+				interval: 10 * 1000,
+			},
+			connected => {
+				if (onChangeCallback != null) {
```
Page-: Strictly speaking the previously behaviour was `typeof onChangeCallback === 'function'`, I'm not sure if the difference matters here or not

src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+export const startConnectivityCheck = _.once(
+	async (
+		apiEndpoint: string,
+		enable: boolean,
+		onChangeCallback?: networkCheck.MonitorChangeFunction,
+	) => {
+		enableConnectivityCheck(enable);
+		if (!Boolean(apiEndpoint)) {
```
Page-: This was a `!= null` check before, why the `Boolean` now? (especially in comparison to `!apiEndpoint` instead)

zvin: What is the purpose of `Boolean`?
Is there a case where it transforms the truthiness of something?

src/network.ts
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		const parsedUrl = url.parse(apiEndpoint);
+		const port = parseInt(parsedUrl.port!, 10);
+
+		customMonitor(
+			{
+				host: parsedUrl.hostname,
+				port: port || (parsedUrl.protocol === 'https' ? 443 : 80),
+				path: parsedUrl.path || '',
```
Page-: This is new?  What's it for?  Also I'd rather review typescript conversions that are purely conversions and don't introduce wholly new code as it's very hard to spot new changes when the entire file is new

src/network.ts
CameronDiver: This is the way it's compared across all of the other code in the codebase, I'd rather keep it the same.

src/network.ts
CameronDiver: It was required by the typings.

Page-: I'm guessing that's because the types are wrong if it was working before?  Does this mean the network check will now be hitting a different path? (it might be relevant depending on what the path is)

CameronDiver: The path will default to `'/'`, so it will hit the same path, I should probably change the `|| ''` though, I had meant to put `/`.

src/network.ts
```diff
@@ -0,0 +1,156 @@
...
+		Bluebird.resolve(fs.mkdir(constants.vpnStatusPath))
+			.catch(EEXIST, () => {
+				console.log('VPN status path exists.');
+			})
+			.then(() => {
+				fs.watch(constants.vpnStatusPath, vpnStatusInotifyCallback);
+			});
+
+		if (enable) {
+			vpnStatusInotifyCallback();
```
zvin: I'm not sure this is intended: 
the `constants.vpnStatusPath` may not exist yet as we do not await the `mkdir` promise above.

PR #822, opened 2018-11-27
compose: Change default stop signal from '' to SIGTERM

shaunmulligan:
Change-type: patch
Signed-off-by: Shaun Mulligan <shaun@balena.io>


shaunmulligan:
connects to #821

PR #823, opened 2018-11-27
Add container ID api endpoint

CameronDiver:
This is pending https://github.com/balena-io/balena-supervisor/pull/818 and has been rebased on top - only the last commit is part of this PR.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -793,6 +793,41 @@ Response:
 }
 ```
 
+### Container ID (GET /v2/containerId)
+
+Added in supervisor v8.6.0
+
+Use this endpoint to match a service name to a container ID.
+
+From the user container:
+```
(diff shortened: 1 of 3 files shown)


src/device-state.coffee
```diff
@@ -328,61 +328,72 @@ module.exports = class DeviceState extends EventEmitter
...
-					console.log('Preloading complete')
-					if stateFromFile.pinDevice
+		# Ensure that this is actually a file, and not an empty path
+		# It can be an empty path because if the file does not exist
+		# on host, the docker daemon creates an empty directory when
+		# the bind mount is added
+		fs.lstatAsync(appsPath)
+		.then (stats) ->
+			return stats.isFile()
+		.then (found) =>
```
zvin: Checking if a file exists before opening it is not a good practice, it can lead to races (even if in this case it's unlikely).
We should `readFileAsync` directly and catch `ENOENT` and `EISDIR` instead.

Also I can't way for the day we use await / async in the supervisor.

PR #828, opened 2018-12-03
Add application status endpoint

CameronDiver:
Change-type: minor

### DIFF
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -385,6 +385,7 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
 
 	router.get('/v2/state/status', async (_req, res) => {
 		const localMode = await applications.config.get('localMode');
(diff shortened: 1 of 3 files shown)


src/device-api/v2.ts
```diff
@@ -375,4 +375,12 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
 			});
 		}
 	});
+
+	router.get('/v2/state/status', async (_req, res) => {
```
shaunmulligan: @CameronDiver it would be pretty cool if this gave full state with containers, etc? It would be cool on localMode to do something like: `balena device state -w` and then we can see like a simple CLI status of the device, showing, whats going on on the device. I have a need for that often when doing local mode, which i usually have to ssh in and do `balena stat` and `balena ps`

CameronDiver:
@shaunmulligan I'm keen, it would be good to get @curcuz opinion on whether anything else would be helpful though.

PR #835, opened 2018-12-06
WIP: Add the ability to restore volumes from a backup.tgz in the data part…

pcarranzav:
…ition

Change-type: minor


src/api-binder.coffee
```diff
@@ -83,6 +86,21 @@ module.exports = class APIBinder
...
+				return
+			console.log('Migration backup detected')
+			@getTargetState()
+			.then (targetState) =>
+				@deviceState.restoreBackup(targetState)
+		.catch (err) =>
+			console.log('Error restoring migration backup, retrying: ', err)
+			Promise.delay(retryDelay)
+			.then =>
+				@loadBackupFromMigration(retryDelay)
```
CameronDiver: Won't this constantly loop if the backup cannot be loaded?

pcarranzav: @CameronDiver yes... we don't want to continue if we can't load the backup (if there is a backup, that is). I guess we can instead throw the exception and let the supervisor healthcheck kick in? Though probably the effect will be the same, and with this loop we can probably catch transient errors, e.g. a network error when getting the target state

dindctl
```diff
@@ -24,6 +24,7 @@
 # 	--dind-container [name]	container name suffix for the dind host container ( default: "supervisor", which will produce a container named resinos-in-container-supervisor)
 # 	--mount-dist			bind-mount './dist/' (where webpack stores the built js) from local development environment into supervisor container.
 # 	--mount-nm				bind-mount './node_modules/' from local development environment into supervisor container.
+# 	--mount-backup		bind-mount './tools/dind/backup.tgz' to simulate a migration backup.
```
CameronDiver: Weird indentation here.

src/device-state.coffee
```diff
@@ -406,6 +410,37 @@ module.exports = class DeviceState extends EventEmitter
...
 
+	restoreBackup: (targetState) =>
+		@setTarget(targetState)
+		.then =>
+			appId = _.keys(targetState.local.apps)[0]
+			if !appId?
+				throw new Error('No appId in target state')
+			volumes = targetState.local.apps[appId].volumes
+			backupPath = path.join(constants.rootMountPoint, 'mnt/data/backup')
+			rimraf(backupPath) # We clear this path in case it exists from an incomplete run of this function
```
CameronDiver: Given we're already using shell commands further down, and that we're already heavily tied to linux, is it worth bringing in this extra dependency for this task?

src/device-state.coffee
```diff
@@ -406,6 +410,37 @@ module.exports = class DeviceState extends EventEmitter
...
+				Promise.mapSeries dirContents, (volumeName) =>
+					if volumes[volumeName]?
+						console.log("Creating volume #{volumeName} from backup")
+						# If the volume exists (from a previous incomplete run of this restoreBackup), we delete it first
+						@applications.volumes.get({ appId, name: volumeName })
+						.then =>
+							@applications.volumes.remove({ appId, name: volumeName })
+						.catch(NotFoundError, _.noop)
+						.then =>
+							@applications.volumes.createFromPath({ appId, name: volumeName, config: volumes[volumeName] }, path.join(backupPath, volumeName))
```
CameronDiver: I assume this is the part of the code which confused @samothx too? E.g. the volume didn't exist in the target state but it did in the backup - Might be worth printing the volumes which we're ignoring as well, so it's more transparent? This way we could tell the difference between say a corrupted backup and incorrect target state.

dindctl
pcarranzav: Yeah... I now changed all of this to spaces rather than tabs

src/device-state.coffee
pcarranzav: It was already used by some of our dependencies, but true that we don't really needed here... I changed it to another execAsync (see [line 48](https://github.com/balena-io/balena-supervisor/pull/835/files#diff-0011437c96d80cf058a252cafc68dc0bR48))

src/device-state.coffee
pcarranzav: True... now I've changed it so that it errors out if anything in the backup file is unexpected (either it has something that's not a directory, or it has a directory that's not a valid volume in the target state). What do you think?

src/device-state.coffee
```diff
@@ -41,6 +44,9 @@ validateState = Promise.method (state) ->
 	if state.dependent?
 		validateDependentState(state.dependent)
 
+rimraf = (p) ->
+	execAsync("rm -rf \"#{p.replace(/"/g, '\\"')}\"")
```
CameronDiver: Is there a possibility that the backup file can have quotes?

What worries me about this is that it won't cover all escaping necessary for shell commands, but it looks "general". I think we should either be very target (e.g. know exactly that a path will not contain bad chars) or support everything (maybe switch back to rimraf if we cannot guarantee the former? (sorry!))

pcarranzav: I don't know if there's chances for quotes, but I wouldn't like creating a function that can't handle such a simple thing... and it's the kind of thing that could fail in the future if things change. So yeah, let's go back to rimraf :)

Issue #836, opened 2018-12-06
Supervisor fails to report the IP address in dashboard


horia-delicoti:
Tested on: 
balenaos: `v2.28.0+rev1`
supervisor: `8.6.3`

[Flowdock conversation](https://www.flowdock.com/app/rulemotion/resin-devices/threads/OwnbTKNMQnKVlf3CL61sdSrlvVi)

PR #837, opened 2018-12-06
 api: Ensure Supervisor API returns IP addresses

hedss:
The move from pure CoffeeScript to TypeScript has brought a
few changes to the way transpiling happens. Previously, through
serendipity, the way `startIPAddressUpdate` was called worked
because of the binding convention pre-transpiling.

However, with the move to TypeScript, this has altered and
the assumption that a lack of parentheses would call the
method before supplying a callback into the returned function
is incorrect. The method must be specifically called first.

Connects-to: #836 
Change-type: patch
Signed-off-by: Heds Simons <heds@balena.io>

### DIFF
--- a/src/device-state.coffee
+++ b/src/device-state.coffee
@@ -301,7 +301,7 @@ module.exports = class DeviceState extends EventEmitter
 			if changedConfig.connectivityCheckEnabled?
 				network.enableConnectivityCheck(changedConfig.connectivityCheckEnabled)
 		console.log('Starting periodic check for IP addresses')
-		network.startIPAddressUpdate (addresses) =>
+		network.startIPAddressUpdate() (addresses) =>
 			@reportCurrentState(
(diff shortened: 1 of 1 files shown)


src/device-state.coffee
```diff
@@ -300,8 +300,7 @@ module.exports = class DeviceState extends EventEmitter
 		@config.on 'change', (changedConfig) ->
 			if changedConfig.connectivityCheckEnabled?
 				network.enableConnectivityCheck(changedConfig.connectivityCheckEnabled)
-		console.log('Starting periodic check for IP addresses')
-		network.startIPAddressUpdate (addresses) =>
+		network.startIPAddressUpdate() (addresses) =>
```
CameronDiver: Ugh, can't wait for all of the CS to be gone.

hedss: Yup!

PR #844, opened 2018-12-12
Unmanaged + local mode fixes

CameronDiver:
This PR:
* Moves local mode back to being stored in the db, rather than config.json
* Ensures that local mode is always enabled in unmanaged mode
* Adds authentication to unmanaged production mode supervisor API


src/supervisor-api.coffee
```diff
@@ -11,19 +11,19 @@ authenticate = (config) ->
...
-			if queryKey? && bufferEq(new Buffer(queryKey), new Buffer(conf.apiSecret))
-				next()
-			else if headerKey? && bufferEq(new Buffer(headerKey), new Buffer(conf.apiSecret))
-				next()
-			else if checkTruthy(conf.localMode)
-				next()
+			needsAuth = (checkTruthy(conf.unmanaged) and conf.osVariant is 'prod') || not checkTruthy(conf.localMode)
+			if needsAuth
+				key = queryKey ? headerKey
+				if bufferEq(new Buffer(key), new Buffer(conf.apiSecret))
```
zvin: `Buffer.from`

src/config/functions.ts
```diff
@@ -149,5 +149,12 @@ export function createProviderFunctions(
 				]);
 			},
 		},
+		unmanaged: {
```
pcarranzav: I guess this replaces offlineMode then? should we globally replace it rather than keeping both? (I don't think anyone is using the supervisorOfflineMode config.json flag and it's undocumented anyways, so I think we can do the breaking change without problems).

src/supervisor-api.coffee
```diff
@@ -11,19 +11,19 @@ authenticate = (config) ->
...
-		config.getMany([ 'apiSecret', 'localMode' ])
+		config.getMany([ 'apiSecret', 'localMode', 'unmanaged', 'osVariant' ])
 		.then (conf) ->
-			if queryKey? && bufferEq(new Buffer(queryKey), new Buffer(conf.apiSecret))
-				next()
-			else if headerKey? && bufferEq(new Buffer(headerKey), new Buffer(conf.apiSecret))
-				next()
-			else if checkTruthy(conf.localMode)
-				next()
+			needsAuth = (checkTruthy(conf.unmanaged) and conf.osVariant is 'prod') || not checkTruthy(conf.localMode)
```
pcarranzav: So unmanaged devices will always have an unauthenticated supervisor API?

src/device-state.coffee
```diff
@@ -351,7 +351,7 @@ module.exports = class DeviceState extends EventEmitter
 						.then =>
 							@deviceConfig.setTarget(target.local.config, trx)
 						.then =>
-							if localSource
+							if localSource or apiEndpoint == ''
```
Page-: Should this also check for a nullish apiEndpoint?

src/config/functions.ts
```diff
@@ -149,5 +137,12 @@ export function createProviderFunctions(
...
+			get: () => {
+				return config.get('apiEndpoint').then(apiEndpoint => {
```
Page-: This can be
```ts
get: () => config.get('apiEndpoint').then(_.negate)

Comment (CameronDiver) [reply src/config/functions.ts]: The typings aren't good enough yet for that to be allowed, so rather than doing lots of casting, I think the current version would be better.

File: src/device-config.coffee [hunk 6]
```diff
@@ -162,7 +163,7 @@ module.exports = class DeviceConfig
 				return
 
 			# Check if we need to perform special case actions for the VPN
-			if !checkTruthy(offlineMode) &&
+			if !checkTruthy(unmanaged) &&
```
pcarranzav: I think checkTruthy is actually not needed anymore since unmanaged is a boolean

src/local-mode.ts
```diff
@@ -40,9 +40,20 @@ export class LocalModeManager {
 			}
 		});
 
-		const localMode = checkTruthy(
-			(await this.config.get('localMode')) || false,
-		);
+		// On startup, check if we're in unmanaged mode,
+		// as local mode needs to be set
+		let unmanagedLocalMode = false;
+		if (checkTruthy((await this.config.get('unmanaged')) || false)) {
```
pcarranzav: No need for the checkTruthy nor the `|| false` here, right?

src/local-mode.ts
```diff
@@ -40,9 +40,20 @@ export class LocalModeManager {
...
+			console.log('Starting up in unmanaged mode, activating local mode');
+			this.config.set({ localMode: true });
```
pcarranzav: You okay with doing this in the background? Or should this have an `await`?

pcarranzav: Also, now that this is a db value, should this be a string? i.e. `'true'`?

src/migrations/M00002.js
```diff
@@ -0,0 +1,41 @@
...
+			} catch (e) {
+				console.log('Failed to parse config.json! Things may fail unexpectedly!');
+				return resolve(false);
+			}
+		});
+	}).then(localMode => {
+		// We can be sure that this does not already exist in the db because of the previous
+		// migration
+		return knex('config').insert({
+			key: 'explicitLocalMode',
```
pcarranzav: Is this key used anywhere? Or should this just be `'localMode'`?

src/local-mode.ts
CameronDiver: Good catch on the `await`, will change. I'd much rather handle the conversion to string on the way into the database though, as that's an implementation detail, and shouldn't be part of the interface.

src/config/functions.ts
Page-: It's not the typings, I got confused about the _.negate signature - you use it for things like `_.negate(_.isNil)` to wrap a function and negate it's result, so it could be done as `_.negate(_.identity)` but I think that's probably less clear

src/migrations/M00002.js
CameronDiver: oops, missed this...

src/api-binder.coffee
```diff
@@ -102,9 +102,9 @@ module.exports = class APIBinder
 				@loadBackupFromMigration(retryDelay)
 
 	start: =>
-		@config.getMany([ 'apiEndpoint', 'offlineMode', 'bootstrapRetryDelay' ])
-		.then ({ apiEndpoint, offlineMode, bootstrapRetryDelay }) =>
-			if offlineMode
+		@config.getMany([ 'apiEndpoint', 'unmanaged', 'bootstrapRetryDelay' ])
+		.then ({ apiEndpoint, unmanaged, bootstrapRetryDelay }) =>
+			if unmanaged
 				console.log('Offline Mode is set, skipping API binder initialization')
```
pcarranzav: We might wanna change the mentions of "offline mode" in the logs (but will approve regardless)

src/api-binder.coffee
```diff
@@ -338,6 +338,9 @@ module.exports = class APIBinder
 			(currentState, targetConfig, defaultConfig, deviceId) =>
 				currentConfig = currentState.local.config
 				Promise.mapSeries _.toPairs(currentConfig), ([ key, value ]) =>
+					# We want to disable local mode when joining a cloud
+					if key == 'SUPERVISOR_LOCAL_MODE'
+						return @config.set({ localMode: false })
```
pcarranzav: Feels kinda awkward to directly change the config here... If instead you just set `value = 'false'` here, the device state engine will take care of setting the config to false, in the same way as the VPN control var below.

But won't block on this since this will work fine too.

PR #849, opened 2018-12-13
More typescript conversions

CameronDiver:


typings/global.d.ts
```diff
@@ -1,3 +1,7 @@
 interface Dictionary<T> {
 	[key: string]: T;
 }
+
+interface Callback<T> {
+	(err: Error | undefined, res: T): void;
```
Page-:
```suggestion
	(err?: Error, res: T): void;
```

src/lib/update-lock.ts
```diff
@@ -0,0 +1,104 @@
...
+						return Bluebird.mapSeries(
+							lockFilesOnHost(appId, serviceName),
+							tmpLockName => {
+								return Bluebird.try(() => {
+									if (force) {
+										return lockFile.unlockAsync(tmpLockName);
+									}
+								})
+									.then(() => lockFile.lockAsync(tmpLockName))
+									.then(() => (locksTaken[tmpLockName] = true))
```
Page-: Should have `{}` instead - no need to return true

src/lib/update-lock.ts
```diff
@@ -0,0 +1,104 @@
...
+									.catchReturn(ENOENT, null);
```
Page-: `undefined` is preferable imo, as it'll mesh much better with `Promise<void>`

src/lib/update-lock.ts
```diff
@@ -0,0 +1,104 @@
...
+							return dispose(release).throw(
+								new UpdatesLockedError(`Updates are locked: ${err.message}`),
+							);
+						});
+					});
+			})
+			.disposer(dispose);
+	};
+
+	return Bluebird.using(takeTheLock(), () => fn());
```
Page-:
```suggestion
	return Bluebird.using(takeTheLock(), fn);
```
You might need to combine this with making `takeTheLock` return a `PromiseLike<void>`

src/lib/update-lock.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const locker = new Lock();
+const writeLock = Bluebird.promisify(locker.async.writeLock).bind(locker);
+
+export function lock(
+	appId: number | null,
+	{ force = false }: { force: boolean },
+	fn: () => PromiseLike<void>,
+): Bluebird<void> {
+	const dispose = (release: () => void): Bluebird<void> => {
```
Page-: This can be moved outside of the function since it doesn't rely on any of the closure scope, that way it would also never need to be recreated

src/lib/update-lock.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const locksTaken: { [lockName: string]: boolean } = {};
+
+// Try to clean up any existing locks when the program exits
+process.on('exit', () => {
+	const result = [];
+	for (const lockName of _.keys(locksTaken)) {
+		try {
+			result.push(lockFile.unlockSync(lockName));
+			// tslint:disable-next-line
```
Page-: What?

src/lib/request.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// With these settings, the device must be unable to receive a single byte
+// from the network for a continuous period of 20 minutes before we give up.
+// (reqTimeout + retryInterval) * retryCount / 1000ms / 60sec ~> minutes
+const DEFAULT_REQUEST_TIMEOUT = 30000; // ms
+const DEFAULT_REQUEST_RETRY_INTERVAL = 10000; // ms
+const DEFAULT_REQUEST_RETRY_COUNT = 30;
+
+exports.requestOpts = {
```
Page-:
```suggestion
export const requestOpts = {
```

src/lib/migration.ts
```diff
@@ -0,0 +1,64 @@
...
+	const environment: Dictionary<string> = {};
+	for (const key in app.env) {
+		if (!/^RESIN_/.test(key)) {
+			environment[key] = app.env[key];
+		}
+	}
+
+	const { appId } = app;
+	const conf = app.config != null ? app.config : {};
+	const newApp: any = {
```
Page-: Can you make this more specific please (at a minimum `Dictionary<any>`)

src/lib/migration.ts
```diff
@@ -0,0 +1,64 @@
+export const defaultLegacyVolume = () => 'resin-data';
+
+export function singleToMulticontainerApp(app: any) {
```
Page-: Can you make this more specific please (at a minimum `Dictionary<any>`) - if it truly can be `any` then there's a whole bunch of places where the code will blow up

typings/global.d.ts
CameronDiver: This is not possible due to:
```
A required parameter cannot follow an optional parameter.
```

src/lib/update-lock.ts
CameronDiver: We want to ignore the error, but we can't add an empty catch block without the linter complaining. Any better ideas?

src/lib/request.ts
CameronDiver: Ah I missed this one, I've been using decaffeinate you see.

src/lib/migration.ts
CameronDiver: I've correctly typed the output, the input isn't as necessary so as you say I've added `Dictionary<any>`

src/lib/update-lock.ts
CameronDiver: Oh nice, yeah that works.

typings/global.d.ts
CameronDiver: That's very true, nice!

src/lib/update-lock.ts
```diff
@@ -0,0 +1,107 @@
...
+process.on('exit', () => {
+	const result = [];
+	for (const lockName of _.keys(locksTaken)) {
+		try {
+			result.push(lockFile.unlockSync(lockName));
+		} catch (e) {
+			// Ignore unlocking errors
+		}
+	}
+	return result;
```
Page-: Why are you returning a result to `process.on('exit')`?  I'm guessing this is a coffeescript leftover

src/lib/update-lock.ts
```diff
@@ -0,0 +1,107 @@
...
+		path.join(constants.rootMountPoint, lockPath(appId, serviceName), filename),
+	);
+}
+
+const locksTaken: { [lockName: string]: boolean } = {};
+
+// Try to clean up any existing locks when the program exits
+process.on('exit', () => {
+	const result = [];
+	for (const lockName of _.keys(locksTaken)) {
```
Page-: This can be:
```suggestion
	for (const lockName in locksTaken) {
```

CameronDiver: That's what I started with, but our typescript config enforces:
```
[tslint] for (... in ...) statements must be filtered with an if statement [forin]
```

which seems massively out of scope

src/lib/update-lock.ts
CameronDiver: I've just removed `result` altogether, no need for it - I guess a decaffeinate oddity.

Page-: It's to match what coffeescript does with it's automatic returns, but we don't care about returning in this case

src/lib/update-lock.ts
Page-: I presume the idea behind that rule is to do
```ts
if (locksTaken.hasOwnProperty(lockName) {
	...
}
```
to avoid issues with iterating properties on the prototype but it is a little weird to force

CameronDiver: Yeah agreed, and we're also using the keys in a slightly odd way to begin with, the keys either exist with a value of true, or they don't exist, so we might as well use an array - I would rather tackle this as part of a separate PR though.

PR #853, opened 2018-12-19
device-api: Add morgan to log api requests

CameronDiver:
Adds logs which look like:
```
Dec 19 13:37:24 b753d01 resin-supervisor[30304]: SupervisorAPI: GET /v2/applications/state 200 - 69.215 ms
Dec 19 13:37:27 b753d01 resin-supervisor[30304]: SupervisorAPI: GET /v2/applications/state 200 - 60.972 ms
```

Change-type: minor
Closes: #834


src/supervisor-api.coffee
```diff
@@ -29,11 +31,22 @@ authenticate = (config) ->
 		.catch (err) ->
 			res.status(503).send("Unexpected error: #{err}")
 
+expressLogger = morgan (tokens, req, res) -> [
+		'SAPI'
+		tokens.method(req, res)
+		url.parse(tokens.url(req, res)).pathname # Strip the apikey from the request
```
Page-: Should the url already be parsed of part of the `req` object?  If so it'd be good to avoid the work of reparsing it

src/supervisor-api.coffee
```diff
@@ -29,11 +31,22 @@ authenticate = (config) ->
 		.catch (err) ->
 			res.status(503).send("Unexpected error: #{err}")
 
+expressLogger = morgan (tokens, req, res) -> [
+		'SAPI'
```
Page-: I assume `SAPI` here is meant to stand for Supervisor API?  If so could you make it a bit clearer as SAPI can definitely stand for other things

CameronDiver: Sure thing.

src/supervisor-api.coffee
CameronDiver: Done.

PR #857, opened 2018-12-20
Further typescript conversions

CameronDiver:
Mainly host-config and device-config have been converted, but a load of supporting changes have went in too.


src/host-config.ts
```diff
@@ -0,0 +1,174 @@
...
+				hostname,
+			},
+		};
+	});
+}
+
+export async function patch(
+	conf: HostConfig,
+	configModel: Config,
+): Bluebird<void> {
```
Page-: If you change this to `Promise<void>` you can drop the `return Bluebird.resolve()` below

src/host-config.ts
```diff
@@ -0,0 +1,174 @@
...
+				if (match != null) {
+					conf[proxyField] = match[1];
+				}
+			}
+		}
+	} catch (e) {
+		if (!ENOENT(e)) {
+			throw e;
+		}
+		return null;
```
Page-: I'd rather `return` and `| undefined` in the sig

src/host-config.ts
```diff
@@ -0,0 +1,174 @@
...
+const redsocksConfPath = path.join(proxyBasePath, 'redsocks.conf');
+const noProxyPath = path.join(proxyBasePath, 'no_proxy');
+
+interface ProxyConfig {
+	[key: string]: string | string[];
+}
+
+interface HostConfig {
+	network: {
+		proxy?: ProxyConfig | null;
```
Page-: Both null and undefined are allowed?

src/host-config.ts
```diff
@@ -0,0 +1,174 @@
...
+	log_info = on;
+	log = stderr;
+	daemon = off;
+	redirector = iptables;
+}
+
+redsocks {
+	local_ip = 127.0.0.1;
+	local_port = 12345;
+\
```
Page-: I don't think this `\` is necessary?

src/host-config.ts
```diff
@@ -0,0 +1,174 @@
...
+import { ENOENT } from './lib/errors';
+import { writeFileAtomic } from './lib/fs-utils';
+import * as systemd from './lib/systemd';
+
+const mkdirp = Bluebird.promisify(mkdirCb) as (
+	path: string,
+	opts?: any,
+) => Bluebird<mkdirCb.Made>;
+
+const redsocksHeader = `\
```
Page-: I don't think this `\` is necessary?

src/device-config.ts
```diff
@@ -0,0 +1,512 @@
...
+			this.logger.logSystemMessage(
+				`Error setting boot config: ${err}`,
+				{ error: err },
+				'Apply boot config error',
+			);
+			throw err;
+		}
+	}
+
+	private async getVPNEnabled(): Promise<boolean | null> {
```
Page-: I don't see where this can return null?

src/device-config.ts
```diff
@@ -0,0 +1,512 @@
...
+					!DeviceConfig.configTest(
+						varType,
+						current[envVarName],
+						target[envVarName],
+					)
+				) {
+					// Save the change if it is
+					configChanges[key] = target[envVarName];
+					humanReadableConfigChanges[envVarName] = target[envVarName];
+					reboot = reboot || (rebootRequired != null ? rebootRequired : false);
```
Page-: I think `(rebootRequired != null ? rebootRequired : false)` can be simplified to just `rebootRequired` - if `rebootRequired == null` then it'll coerce to false anyway

src/device-config.ts
```diff
@@ -0,0 +1,512 @@
...
+		]);
+
+		const currentConf: Dictionary<string> = {
+			// TODO: Fix this mess of half strings half boolean values everywhere
+			SUPERVISOR_VPN_CONTROL: vpnStatus != null ? vpnStatus.toString() : 'true',
+		};
+
+		for (const key in DeviceConfig.configKeys) {
+			const { envVarName } = DeviceConfig.configKeys[key];
+			const value = conf[key] != null ? (conf[key] as string) : '';
```
Page-: This would be better as:
```suggestion
			currentConf[envVarName] = conf[key] != null ? conf[key].toString() : '';
```
to avoid the need for the cast (which was incorrect anyway if we actually needed the `toString`)

src/device-config.ts
```diff
@@ -0,0 +1,512 @@
...
+	private async getConfigBackend() {
+		if (this.configBackend != null) {
+			return this.configBackend;
+		}
+		const dt = await this.config.get('deviceType');
+		if (!_.isString(dt)) {
+			throw new Error('Could not detect device type');
+		}
+
+		return (this.configBackend = configUtils.getConfigBackend(dt) || null);
```
Page-: Could you separate this to a line for the assignment and a line for the return please

src/device-config.ts
```diff
@@ -0,0 +1,512 @@
...
+		},
+		persistentLogging: {
+			envVarName: 'SUPERVISOR_PERSISTENT_LOGGING',
+			varType: 'bool',
+			defaultValue: 'false',
+			rebootRequired: true,
+		},
+	};
+
+	static validKeys = ['SUPERVISOR_VPN_CONTROL', 'OVERRRIDE_LOCK'].concat(
```
Page-: You could make this:
```suggestion
	static validKeys = ['SUPERVISOR_VPN_CONTROL', 'OVERRRIDE_LOCK', ..._.map(DeviceConfig.configKeys, 'envVarName')]
```

src/application-manager.d.ts
```diff
@@ -51,7 +51,9 @@ export class ApplicationManager extends EventEmitter {
 		opts: Options,
 	): Bluebird<void>;
 
-	public getStatus(): Promise<DeviceApplicationState>;
+	// FIXME: Type this properly as it's some mutant state between
```
Page-: What was wrong with the previous typing?

src/host-config.ts
CameronDiver: Decaffeinate added these, I think you're right - not necessary.

src/host-config.ts
CameronDiver: I'm trying to maintain the external interface for now, no doubt there's somewhere in CS where it's assumed to be a Bluebird promise. Once it's all typescript, I can change the types and get the compiler to find those places for me, but it's a right pain before.

src/device-config.ts
CameronDiver: Ah it used to in the original implementation, but it didn't make any sense and I forgot to take it out of the sig - good catch!

src/device-config.ts
CameronDiver: Nice

src/application-manager.d.ts
CameronDiver: It was correct for `DeviceApplicationState` in the form before this PR, but that interface was wrong. Basically this type is a mix between the state endpoint for a single application and the internal of ApplicationManager. I was going to add the other type to the d.ts but I'm going to be rewriting ApplicaitonManager in typescript very soon, so I'd rather do it then. 

I can add the typing to this PR instead if you prefer?

src/host-config.ts
```diff
@@ -0,0 +1,182 @@
...
+				proxy,
+				hostname,
+			},
+		};
+	});
+}
+
+export function patch(conf: HostConfig, configModel: Config): Bluebird<void> {
+	const promises = [];
+	if (conf != null && conf.network != null) {
```
Page-: Based on the typings `conf` can never be null

src/device-config.ts
```diff
@@ -0,0 +1,515 @@
...
+			return !_.includes(['inactive', 'deactivating'], activeState);
+		} catch (e) {
+			if (UnitNotLoadedError(e)) {
+				return false;
+			}
+			throw e;
+		}
+	}
+
+	private async setVPNEnabled(value: string | boolean) {
```
Page-: Looks like value can also be undefined based on the code

src/device-config.ts
```diff
@@ -0,0 +1,515 @@
...
+			throw new Error(`Corrupted supervisor database! Error: ${e.message}`);
+		}
+		if (initial || conf.SUPERVISOR_VPN_CONTROL == null) {
+			conf.SUPERVISOR_VPN_CONTROL = 'true';
+		}
+		if (unmanaged && conf.SUPERVISOR_LOCAL_MODE == null) {
+			conf.SUPERVISOR_LOCAL_MODE = 'true';
+		}
+
+		_.each(DeviceConfig.configKeys, ({ envVarName, defaultValue }) => {
```
Page-: It looks like you can use
```ts
_.defaults(conf, DeviceConfig.configKeys)
```
for this instead

src/host-config.ts
CameronDiver: This is just what it was in the coffeescript, as I say, when every call site is TS I can be sure to take this out and that the types are stringent enough, but until then I'm really not comfortable taking out the checking, and I'd much prefer the types to as strict as possible.

PR #860, opened 2018-12-31
More typecript conversions

CameronDiver:
* Converted ApiBinder and SupervisorApi modules to typescript
* Changed Config module to export itself in the new es6 style
* Promisify `readLock` and `writeLock` once in update-lock and export them, rather than doing it all over the place


src/supervisor-api.ts
```diff
@@ -0,0 +1,196 @@
...
+
+import Config from './config';
+import { EventTracker } from './event-tracker';
+import blink = require('./lib/blink');
+import * as iptables from './lib/iptables';
+import { checkTruthy } from './lib/validation';
+
+function getKeyFromReq(req: express.Request): string | null {
+	let headerKey: string | null = null;
+	const queryKey = req.query.apikey;
```
Page-: You can add a
```ts
if (queryKey != null) {
	return queryKey
}
```
here to early exit since the `queryKey` takes priority, and then
```ts
if (!maybeHeaderKey) {
	return
}

File: src/lib/update-lock.ts [hunk 2]
```diff
@@ -101,5 +107,9 @@ export function lock(
 			.disposer(dispose);
 	};
 
-	return Bluebird.using(takeTheLock(), fn);
+	const disposer = takeTheLock();
+	if (disposer) {
```
Page-: This changes it so that `fn` is never called in the case of disposer being falsey - is that intentional?

src/api-binder.ts
```diff
@@ -0,0 +1,885 @@
...
+	private async provisionOrRetry(retryDelay: number): Promise<void> {
+		this.eventTracker.track('Device bootstrap');
+		try {
+			await this.provision();
+		} catch (e) {
+			this.eventTracker.track(`Device bootstrap failed, retrying`, {
+				error: e,
+				delay: retryDelay,
+			});
+			Bluebird.delay(retryDelay).then(() => {
```
Page-: Is this meant to be returned?  It was in the previous version

src/api-binder.ts
```diff
@@ -0,0 +1,885 @@
...
+
+		if (pinValue != null) {
+			if (pinValue.app == null || pinValue.commit == null) {
+				console.log(
+					`Malformed pinDEvice fields in supervisor database: ${pinValue}`,
+				);
+				return;
+			}
+			console.log('Attempting to pin device to preloaded release...');
+			this.pinDevice(pinValue);
```
Page-: Is this meant to be returned? It was in the previous version

src/api-binder.ts
```diff
@@ -0,0 +1,885 @@
...
+
+		await this.config.set({ initialConfigReported: apiEndpoint });
+	}
+
+	private async reportInitialConfig(apiEndpoint: string, retryDelay: number) {
+		try {
+			await this.reportInitialEnv(apiEndpoint);
+		} catch (err) {
+			console.error('Error reporting initial configuration, will retry', err);
+			Bluebird.delay(retryDelay).then(() =>
```
Page-: Is this meant to be returned? It was in the previous version

src/supervisor-api.ts
```diff
@@ -0,0 +1,196 @@
...
+			port,
+			allowedInterfaces,
+		);
+
+		// Monitor the switching of local mode, and change which interfaces will
+		// be listened to based on that
+		this.config.on('change', (changedConfig: Dictionary<string>) => {
+			if (changedConfig.localMode != null) {
+				this.applyListeningRules(
+					checkTruthy(changedConfig.localMode || false) || false,
```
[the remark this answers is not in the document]
CameronDiver: Unfortunately they are, until we sort out the typings on config like I mentioned in the previous review. It also stems from the fact that when I converted `checkTruthy` and friends to typescript, I maintained the same interface, which means they return undefined in the case that they cannot parse it and this doesn't play nice with the type system.

src/supervisor-api.ts
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	public async listen(
+		allowedInterfaces: string[],
+		port: number,
+		apiTimeout: number,
+	): Promise<void> {
+		const localMode = (await this.config.get('localMode')) || false;
+		await this.applyListeningRules(
+			checkTruthy(localMode) || false,
```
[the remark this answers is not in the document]
CameronDiver: For the same reasons as https://github.com/balena-io/balena-supervisor/pull/860#discussion_r244699290 the type system thinks it can. This will all be fixed as soon as I cover all the uses with the conversion.

src/supervisor-api.ts
```diff
@@ -0,0 +1,198 @@
...
+	if (queryKey != null) {
+		return queryKey;
+	}
+	const maybeHeaderKey = req.get('Authorization');
+	if (!maybeHeaderKey) {
+		return null;
+	}
+
+	const match = maybeHeaderKey.match(/^ApiKey (\w+)$/);
+	headerKey = match != null ? match[1] : null;
```
Page-: You can change this to
```suggestion
	return match != null ? match[1] : null;
```
and avoid the need for `headerKey` altogether

src/lib/update-lock.ts
```diff
@@ -46,8 +46,14 @@ process.on('exit', () => {
 	}
 });
 
+type LockFn = (key: string | number) => Bluebird<() => void>;
 const locker = new Lock();
-const writeLock = Bluebird.promisify(locker.async.writeLock).bind(locker);
+export const writeLock: LockFn = Bluebird.promisify(
+	locker.async.writeLock,
+).bind(locker);
+export const readLock: LockFn = Bluebird.promisify(locker.async.readLock).bind(
```
Page-:
```suggestion
export const readLock: LockFn = Bluebird.promisify(locker.async.readLock, { context: locker })
```

src/lib/update-lock.ts
```diff
@@ -46,8 +46,14 @@ process.on('exit', () => {
 	}
 });
 
+type LockFn = (key: string | number) => Bluebird<() => void>;
 const locker = new Lock();
-const writeLock = Bluebird.promisify(locker.async.writeLock).bind(locker);
+export const writeLock: LockFn = Bluebird.promisify(
```
Page-:
```suggestion
export const writeLock: LockFn = Bluebird.promisify(locker.async.writeLock, { context: locker })
```

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+				'UUID and apiTimeout should be defined in exchangeKeyAndGetDevice',
+			);
+		}
+
+		// If we have an existing device key we first check if it's
+		// valid, becaise of ot os we can just use that
+		if (opts.deviceApiKey != null) {
+			const device = await this.fetchDevice(
+				uuid,
+				opts.deviceApiKey as string,
```
Page-: Why is this cast necessary?

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+		if (opts == null) {
+			// FIXME: This casting shouldn't be necessary and stems from the
+			// meta-option provioningOptions not returning a ConfigValue
+			opts = ((await this.config.get(
+				'provisioningOptions',
+			)) as any) as KeyExchangeOpts;
+		}
+
+		const uuid = opts.uuid as string;
+		const apiTimeout = opts.apiTimeout as number;
```
Page-: Why is these casts necessary?

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+		}
+
+		try {
+			await this.getAndSetTargetState(false);
+			this.targetStateFetchErrors = 0;
+		} catch (e) {
+			this.targetStateFetchErrors += 1;
+			pollInterval = Math.min(
+				pollInterval,
+				15000 * 2 ** (this.targetStateFetchErrors - 1),
```
Page-: If you put the `++` afterwards you can remove this `-1`

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+				return null;
+			}
+
+			await Bluebird.resolve(this.report()).delay(REPORT_SUCCESS_DELAY);
+			this.reportCurrentState();
+		} catch (e) {
+			this.stateReportErrors += 1;
+			this.eventTracker.track('Device state report failure', { error: e });
+			const delay = Math.min(
+				2 ** this.stateReportErrors * 500,
```
Page-: This will be 1s on the first error not 500ms (due to incrementing first), is that expected/intentional?

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+			_.assign(this.stateForReport.dependent, currentDeviceState.dependent);
+
+			const stateDiff = this.getStateDiff();
+			if (_.size(stateDiff) === 0) {
+				this.reportPending = false;
+				return null;
+			}
+
+			await Bluebird.resolve(this.report()).delay(REPORT_SUCCESS_DELAY);
+			this.reportCurrentState();
```
Page-: Previously this was awaited, I expect it should still be so errors can be caught

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+			this.eventTracker.track('Device state report failure', { error: e });
+			const delay = Math.min(
+				2 ** this.stateReportErrors * 500,
+				MAX_REPORT_RETRY_DELAY,
+			);
+
+			return Bluebird.delay(delay).then(this.reportCurrentState);
+		}
+
+		return null;
```
Page-: Previously this function returned `null` and not the promise, meaning nothing would wait on it, now it returns a `Promise<null>` due to being an async function and any caller will likely wait on the promise, ie it's the equivalent of changing the previous
```cs
	_reportCurrentState: =>
		@reportPending = true
		@deviceState.getStatus()
		.then (currentDeviceState) =>
			_.assign(@stateForReport.local, currentDeviceState.local)
			_.assign(@stateForReport.dependent, currentDeviceState.dependent)
			stateDiff = @_getStateDiff()
			if _.size(stateDiff) is 0
				@reportPending = false
				return
			@_report()
			.delay(REPORT_SUCCESS_DELAY)
			.then =>
				@_reportCurrentState()
		.catch (err) =>
			@stateReportErrors += 1
			@eventTracker.track('Device state report failure', { error: err })
			delay = Math.min((2 ** @stateReportErrors) * 500, MAX_REPORT_RETRY_DELAY)
			Promise.delay(delay)
			.then =>
				@_reportCurrentState()
		return null
```
to
```cs
	_reportCurrentState: =>
		@reportPending = true
		@deviceState.getStatus()
		.then (currentDeviceState) =>
			_.assign(@stateForReport.local, currentDeviceState.local)
			_.assign(@stateForReport.dependent, currentDeviceState.dependent)
			stateDiff = @_getStateDiff()
			if _.size(stateDiff) is 0
				@reportPending = false
				return
			@_report()
			.delay(REPORT_SUCCESS_DELAY)
			.then =>
				@_reportCurrentState()
				return null
		.catch (err) =>
			@stateReportErrors += 1
			@eventTracker.track('Device state report failure', { error: err })
			delay = Math.min((2 ** @stateReportErrors) * 500, MAX_REPORT_RETRY_DELAY)
			Promise.delay(delay)
			.then =>
				@_reportCurrentState()

File: src/api-binder.ts [hunk 17]
```diff
@@ -0,0 +1,888 @@
...
+				apiEndpoint: string;
+			}),
+		).timeout(conf.apiTimeout as number);
+
+		this.stateReportErrors = 0;
+		_.assign(this.lastReportedState.local, stateDiff.local);
+		_.assign(this.lastReportedState.dependent, stateDiff.dependent);
+	}
+
+	private async reportCurrentState(): Promise<null> {
```
Page-: This sig should be just `null` to match the coffeescript version, if you change it then it should throw up typescript errors for most of the comments I've made about the function

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+				.value(),
+			dependent: _(this.stateForReport.dependent)
+				.omitBy((val, key: keyof DeviceApplicationState['dependent']) =>
+					_.isEqual(lastReportedDependent[key], val),
+				)
+				.omit(INTERNAL_STATE_KEYS)
+				.value(),
+		};
+
+		return _.pickBy(diff, _.negate(_.isEmpty));
```
Page-:
```suggestion
		return _.omitBy(diff, _.isEmpty);
```

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+		}
+
+		try {
+			// TODO: Maybe type this response?
+			const res = (await this.balenaApi
+				.get(reqOpts)
+				.timeout(timeout)) as Device[];
+			return res[0];
+		} catch (e) {
+			return Bluebird.resolve(null);
```
Page-: This can be `return null` because we're in an async function so it'll automatically be wrapped

src/api-binder.ts
```diff
@@ -0,0 +1,888 @@
...
+			if (!exists) {
+				return;
+			}
+			console.log('Migration backup detected');
+			const targetState = await this.getTargetState();
+			await this.deviceState.restoreBackup(targetState);
+		} catch (err) {
+			console.log('Error restoring migration backup, retrying: ', err);
+
+			return Bluebird.delay(retryDelay).then(() => {
```
Page-: You can also do
```ts
await Bluebird.delay(retryDelay)
return this.loadBackupFromMigration(retryDelay);

File: src/lib/update-lock.ts (hunk 9 above)
Comment (CameronDiver) [reply src/lib/update-lock.ts]: TIL - thanks!

File: src/api-binder.ts (hunk 11 above)
Comment (CameronDiver) [reply src/api-binder.ts]: Because this is still typed as a `ConfigValue`, which is `string | number | boolean`. This is fixed with the config changes I've been working on and will be PRd as soon as this is merged.

File: src/api-binder.ts (hunk 14 above)
Comment (CameronDiver) [reply src/api-binder.ts]: This is just as it appears in the original coffeescript - I will change though, otherwise the minimum is not actually a minimum.

File: src/api-binder.ts [hunk 21]
```diff
@@ -0,0 +1,890 @@
...
+		_.assign(this.lastReportedState.dependent, stateDiff.dependent);
+	}
+
+	private reportCurrentState(): null {
+		this.reportPending = true;
+		// TODO: Once device-state has been converted to typescript,
+		// remove this any
+		this.deviceState
+			.getStatus()
+			.then((currentDeviceState: any) => {
```
Page-: Isn't there a better typing for this already available?  If not could you change to `Dictionary<any>` at least (assuming it's guaranteed to be an object, the code assumes it is)

src/api-binder.ts
```diff
@@ -0,0 +1,890 @@
...
+
+	private async reportInitialConfig(
+		apiEndpoint: string,
+		retryDelay: number,
+	): Promise<void> {
+		try {
+			await this.reportInitialEnv(apiEndpoint);
+		} catch (err) {
+			console.error('Error reporting initial configuration, will retry', err);
+			return Bluebird.delay(retryDelay).then(() =>
```
Page-: This can be
```suggestion
			await Bluebird.delay(retryDelay);
```
to keep the style more consistent

src/api-binder.ts
```diff
@@ -0,0 +1,890 @@
...
+	private async provisionOrRetry(retryDelay: number): Promise<void> {
+		this.eventTracker.track('Device bootstrap');
+		try {
+			await this.provision();
+		} catch (e) {
+			this.eventTracker.track(`Device bootstrap failed, retrying`, {
+				error: e,
+				delay: retryDelay,
+			});
+			return Bluebird.delay(retryDelay).then(() => {
```
Page-: This can be
```suggestion
			await Bluebird.delay(retryDelay);
```
to keep the style more consistent

src/api-binder.ts
```diff
@@ -0,0 +1,890 @@
...
+			this.targetStateFetchErrors = 0;
+		} catch (e) {
+			pollInterval = Math.min(
+				pollInterval,
+				15000 * 2 ** this.targetStateFetchErrors,
+			);
+			++this.targetStateFetchErrors;
+		}
+
+		return Bluebird.delay(pollInterval).then(this.pollTargetState);
```
Page-: This can be
```suggestion
		await Bluebird.delay(pollInterval);
		return this.pollTargetState()
```
to keep the style more consistent

src/api-binder.ts
```diff
@@ -0,0 +1,890 @@
...
+		this.stateReportErrors = 0;
+		_.assign(this.lastReportedState.local, stateDiff.local);
+		_.assign(this.lastReportedState.dependent, stateDiff.dependent);
+	}
+
+	private reportCurrentState(): null {
+		this.reportPending = true;
+		// TODO: Once device-state has been converted to typescript,
+		// remove this any
+		this.deviceState
```
Page-: If you want you can probably keep this with await/async whilst still keeping the same semantics by using
```ts
(async () => {
	...
})()
return null
```

src/api-binder.ts
CameronDiver: Not yet, but it's next on the list for conversion. Will change to to a `Dictionary<any>` though.

Issue #824, opened 2018-11-28
Listing ports in docker-compose not in ascending order causes container reboot loop


yutter:
If I do not list the ports in the docker-compose file in ascending order the container starts / dies / starts in a continuous loop. This issue started when I upgraded my BalenaOS from 2.13.x to 2.26.x
The issue is the same on my RaspberryPI 3s and UP2s

This works in all BalenaOS versions:
    ports:
      - 80:80
      - 443:443
      - 502:502
      - 4840:4840
      - 8883:8883
      - 10001:10001
      - 30815:30815
      - 30883:30883

This causes the reboot loop in v2.26.x but worked fine in v2.13.x:
    ports:
      - 80:80
      - 443:443
      - 502:502
      - 4840:4840
      - 8883:8883
      - 30815:30815
      - 30883:30883
      - 10001:10001

I have also noticed that when the container gets into this reboot loop it does seem to get out of it and run properly when I am downloading a new container image! No idea why that would happen

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_g9c3)

yutter:
Also just noticed that the BalenaCloud dashboard shows the correct commit IDs for the containers but shows 'factory build' for the device when it is in this loop. After rebuilding my project with ports in ascending order the device commit ID is updated correctly.

FYI I am running two containers in my project and the other container which did not expose any ports ran fine

CameronDiver:
I've managed to reproduce an issue which may be the same or not, I'm not sure. Interestingly this is only exhibited when the code is sent via the builder, and not with the new beta local mode. This implies a version of the composition parser might be at fault, but I'll continue to look into this. I have a load of other stuff to do today, but this is next on my priority list.

> Also just noticed that the BalenaCloud dashboard shows the correct commit IDs for the containers but shows 'factory build' for the device when it is in this loop. After rebuilding my project with ports in ascending order the device commit ID is updated correctly.

This is expected, the supervisor will only report having moved to another release once it has correctly "settled", running the release as it was intended.

agherzan:
I managed to reproduce this on 2.27.0 RaspberryPi 3 after updating the supervisor from 7.16.6 to 8.3.5 with hostOS updates.

janvda:
I have also faced this issue and reordering the ports fixed this issue.

[...]

PR #862, opened 2019-01-02
compose: Normalise target ports for comparison with docker's output

CameronDiver:
Docker always returns ports in ascending order, so if they aren't
specified like that in the compose, a restart loop would occur. This
patch changes the port maps to be stored in ascending order, based on
an alphabetical sort of the internalStart port (not taking into account
the protocol). This is the same as how Docker returns them, so they will
match, regardless of input form.

Change-type: patch
Closes: #824

### DIFF
--- a/src/compose/ports.ts
+++ b/src/compose/ports.ts
@@ -30,7 +30,7 @@ interface PortRange {
 export class PortMap {
 	private ports: PortRange;
 
-	public constructor(portStrOrObj: string | PortRange) {
+	private constructor(portStrOrObj: string | PortRange) {
 		if (_.isString(portStrOrObj)) {
 			this.parsePortString(portStrOrObj);
 		} else {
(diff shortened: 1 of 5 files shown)


src/compose/ports.ts
```diff
@@ -139,6 +139,12 @@ export class PortMap {
 			}, []);
 	}
 
+	public static normaliseComposePorts(portMaps: PortMap[]): PortMap[] {
```
Page-: You could have this accept the raw port info and do the `new PortMap` as part of this function

src/compose/ports.ts
```diff
@@ -139,6 +139,12 @@ export class PortMap {
...
+		return _.sortBy(portMaps, p => {
+			return p.ports.internalStart;
```
Page-: You can do:
```ts
return _.sortBy(portMaps, p => p.ports.internalStart)

src/compose/ports.ts
CameronDiver: I like this idea actually, makes the interface a lot less clunky - will add.

PR #864, opened 2019-01-08
Fix context bind errors by switching to async/await

CameronDiver:
Change-type: patch

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -543,7 +543,8 @@ export class APIBinder {
 			++this.targetStateFetchErrors;
 		}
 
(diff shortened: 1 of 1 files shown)


src/api-binder.ts
```diff
@@ -543,7 +543,8 @@ export class APIBinder {
 			++this.targetStateFetchErrors;
 		}
 
-		return Bluebird.delay(pollInterval).then(this.pollTargetState);
+		await Bluebird.delay(pollInterval);
+		return this.pollTargetState();
```
zvin: you can also `await this.pollTargetState();`

zvin:
And also cast the poll interval to a number at some point

PR #865, opened 2019-01-09
Add types to the config module, and remove unnecessary casts and validations

CameronDiver:
This PR means that we get compile time and runtime validation of all values coming in and out of the config module, and we also get strong types for most usages in the rest of the supervisor.

Also remove mutable config functions as they weren't being used.


src/config/schema-type.ts
```diff
@@ -0,0 +1,254 @@
...
+		type: PermissiveNumber,
+		default: 15 * 60 * 1000,
+	},
+	listenPort: {
+		type: PermissiveNumber,
+		default: 48484,
+	},
+	deltaEndpoint: {
+		type: t.string,
+		default: 'https://delta.resin.io',
```
zvin: balena?

Issue #867, opened 2019-01-09
Supervisor listens on the IPv6 address on all the interfaces, not just the allowed one


imrehg:
This seems because the iptables rules to filter things only thinks about IPv4 addresses. 

https://github.com/balena-io/balena-supervisor/blob/3db260aa5c8f84241728a32831fc2f29ce6e44a3/src/lib/iptables.ts#L32-L44

Can be checked by `curl IP:48484` of the device which will not reply, but in some cases, where the `uuid.local` avahi address is IPv6, it will reply (with unathorized) on `curl uuid.local:48484`

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_gic3)

PR #871, opened 2019-01-21
Apply iptables rules to ipv6

CameronDiver:
Currently unmergable, due to the issue docuemented here: https://github.com/moby/moby/issues/33605

Change-type: patch
Closes: #867

### DIFF
--- a/src/lib/iptables.ts
+++ b/src/lib/iptables.ts
@@ -1,35 +1,42 @@
-import * as Promise from 'bluebird';
+import * as Bluebird from 'bluebird';
 import * as childProcess from 'child_process';
 
 // The following is exported so that we stub it in the tests
-export const execAsync = Promise.promisify(childProcess.exec);
+export const execAsync: (args: string) => Bluebird<string> = Bluebird.promisify(
+	childProcess.exec,
+);
 
-function clearIptablesRule(rule: string): Promise<void> {
-	return execAsync(`iptables -D ${rule}`).return();
+function applyIptablesArgs(args: string): Bluebird<void> {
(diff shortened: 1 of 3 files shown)


CameronDiver:
@shaunmulligan this is waiting on an architecture discussion.

src/lib/iptables.ts
```diff
@@ -1,35 +1,83 @@
...
+		console.log(
+			`Loading ip6tables kernel module for kernel version ${kernelVersion}...`,
+		);
+		const modPath = join(
+			'/mnt/root/lib/modules',
+			kernelVersion,
+			'kernel/net/ipv6/netfilter/ip6_tables.ko',
+		);
+
+		await execAsync(`insmod ${modPath}`);
```
pcarranzav: This will work, as long as the ip6_tables module has no unmet dependencies, but too bad the busybox modprobe doesn't support `modprobe --dirname`... I wonder if it wouldn't be better to just symlink /lib/modules to /mnt/root/lib/modules inside the supervisor container?

src/lib/iptables.ts
```diff
@@ -1,35 +1,83 @@
...
+	} catch (e) {
+		throw new Error(`Unable to load ip6tables kernel module! Error: ${e}`);
+	}
 }
 
-function clearAndAppendIptablesRule(rule: string): Promise<void> {
+function applyIptablesArgs(args: string): Bluebird<void> {
+	return Bluebird.all([
```
pcarranzav: I think Bluebird.join would be preferable here, since it's more performant for a fixed number of promises?

Issue #868, opened 2019-01-14
Supervisor crashes on reboot (unhandled rejection error) following `balena push ip-address`


pdcastro:
There are several different ways to describe and reproduce this issue. I'm choosing one that is plausible in terms of what a platform user could do.

1. Set a device to [local mode](https://www.balena.io/docs/learn/develop/local-mode/).
2. Run `balena push ip-address` with a simple Dockerfile that pulls an image with 3 slashes in its name, which is the format used by the [GCP Container Registry service](https://cloud.google.com/container-registry/):
```
# Dockerfile
FROM eu.gcr.io/repo-name-123456/arm32v7/busybox
```

> Note: such triple-slashed image name triggers a known docker-toolbelt issue (balena-io-modules/docker-toolbelt/issues/32), but this supervisor issue is for the error to be handled in the supervisor repo.

3. This will straight away cause the supervisor to start logging the following errors, but at this point the errors are caught and the supervisor hasn't "crashed":
```
Jan 14 14:09:13 preciouspi resin-supervisor[1204]: [2019-01-14T14:09:13.405Z] Event: Device state report failure {"error":{"message":"Could not parse the image: eu.gcr.io/buoyant-idea-226013/arm32v7/busybox:latest","stack">
Jan 14 14:09:14 preciouspi resin-supervisor[1204]: [2019-01-14T14:09:14.805Z] Event: Device state report failure {"error":{"message":"Could not parse the image: eu.gcr.io/buoyant-idea-226013/arm32v7/busybox:latest","stack">
Jan 14 14:09:15 preciouspi resin-supervisor[1204]: [2019-01-14T14:09:14.999Z] Applying target state
Jan 14 14:09:15 preciouspi resin-supervisor[1204]: [2019-01-14T14:09:15.439Z] Scheduling another update attempt due to failure:  1000 Error: Could not parse the image: eu.gcr.io/buoyant-idea-226013/arm32v7/busybox:latest
Jan 14 14:09:15 preciouspi resin-supervisor[1204]:     at t.<anonymous> (/usr/src/app/dist/app.js:108:30055)
Jan 14 14:09:15 preciouspi resin-supervisor[1204]:     at t.c (/usr/src/app/dist/app.js:9:74529)
Jan 14 14:09:15 preciouspi resin-supervisor[1204]:     at t.getRegistryAndName (/usr/src/app/dist/app.js:304:137154)
[...]

PR #872, opened 2019-01-22
More typescript conversions

CameronDiver:
* Make the config event-emitter strongly typed
* Add `Nullable<T>` type to clean up some code (Could still do with applying it to more places)
* Update typigns for docker-utils, db, and docker-toolbelt
* Convert compose/images module to typescript


src/compose/images.ts
```diff
@@ -0,0 +1,649 @@
...
+			if (this.volatileState[image.imageId] != null) {
+				this.reportChange(image.imageId, {
+					downloadProgress: progress.percentage,
+				});
+			}
+		};
+
+		let success: boolean;
+		try {
+			const imageName = await this.normalise(image.name);
```
pdcastro: 👍 By the way, I don't know if you meant for this PR to fix issue #868. To fix that issue, I believe a change is required to:
* https://github.com/balena-io/balena-supervisor/blob/master/src/lib/docker-utils.coffee#L46
* `@supervisorTagPromise = @normaliseImageName(constants.supervisorImage)`

That line does not handle the possibility of `normaliseImageName` rejecting the promise. It is called by:
* https://github.com/balena-io/balena-supervisor/blob/master/src/application-manager.coffee#L66
* `@docker = new Docker()`

src/compose/images.ts
```diff
@@ -0,0 +1,649 @@
...
+
+			const img = await this.inspectByName(image.name);
+			await this.db
+				.models('image')
+				.update({ dockerImageId: img.Id })
+				.where(image);
+
+			onFinish(true);
+			return null;
+		} catch (e) {
```
pdcastro: If `e` is the error thrown by `await this.normalise(image.name)`, then it seems that the error is ignored? Maybe it's in front of my eyes, but I don't see where variable `e` is referenced...

src/compose/images.ts
CameronDiver: Yeah agreed @pdcastro thanks :) docker-utils was my next conversion module so I planned to pick it up there.

src/compose/images.ts
CameronDiver: Ah this is where the confusion is coming from: https://github.com/balena-io/balena-supervisor/blob/master/src/compose/images.coffee#L59 the error is ignored here and assumed to be because the image does not exist. I'll add a specific check for this.

src/compose/volumes.ts
```diff
@@ -0,0 +1,231 @@
...
+		const [legacyResponse, currentResponse]: [
+			Dockerode.VolumeInfoList,
+			Dockerode.VolumeInfoList
+		] = await Promise.all([
+			this.docker.listVolumes({
+				filters: { label: ['io.resin.supervised'] },
+			}) as Promise<any>,
+			this.docker.listVolumes({
+				filters: { label: ['io.balena.supervised'] },
+			}) as Promise<any>,
```
Page-: Any reason this can't be `Dockerode.VolumeInfoList` here?

src/compose/volumes.ts
```diff
@@ -0,0 +1,231 @@
...
+		// We have to cast the listVolumes call from any[] to any below, until the
+		// relevant PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32383
+		// is merged and released
+		const [legacyResponse, currentResponse]: [
+			Dockerode.VolumeInfoList,
+			Dockerode.VolumeInfoList
+		] = await Promise.all([
+			this.docker.listVolumes({
+				filters: { label: ['io.resin.supervised'] },
+			}) as Promise<any>,
```
Page-: Any reason this can't be `Dockerode.VolumeInfoList` here?

src/compose/images.ts
```diff
@@ -0,0 +1,652 @@
...
+			}
+		}
+	}
+
+	private static hasDigest(name: Nullable<string>): boolean {
+		if (name == null) {
+			return false;
+		}
+		const parts = name.split('@');
+		if (parts[1] == null) {
```
Page-: This could be
```ts
return parts[1] != null

File: src/compose/images.ts [hunk 6]
```diff
@@ -0,0 +1,652 @@
...
+		opts: FetchOptions,
+		onProgress: (evt: FetchProgressEvent) => void,
+	): Bluebird<string> {
+		this.logger.logSystemEvent(LogTypes.downloadImage, { image });
+		return this.docker.fetchImageWithProgress(image.name, opts, onProgress);
+	}
+
+	// TODO: find out if imageId can actually be null
+	private reportChange(imageId: Nullable<number>, status?: Partial<Image>) {
+		if (imageId != null) {
```
Page-: This would be a bit clearer as
```ts
if (imageId == null) {
	return
}

File: src/compose/images.ts [hunk 7]
```diff
@@ -0,0 +1,652 @@
...
+	}
+
+	public static isSameImage(image1: Image, image2: Image): boolean {
+		return (
+			image1.name === image2.name ||
+			Images.hasSameDigest(image1.name, image2.name)
+		);
+	}
+
+	// private async getLocalModeImages() {
```
Page-: Commented code?

src/compose/images.ts
```diff
@@ -0,0 +1,652 @@
...
+		const toCleanup = _(images)
+			.uniq()
+			.filter(
+				image =>
+					this.imageCleanupFailures[image] == null ||
+					Date.now() - this.imageCleanupFailures[image] >
+						constants.imageCleanupErrorIgnoreTimeout,
+			)
+			.value();
+		return toCleanup;
```
Page-: You could return directly rather than using a temp var

src/compose/images.ts
```diff
@@ -0,0 +1,652 @@
...
+								.models('image')
+								.update({ dockerImageId: id })
+								.where(supervisedImage);
+							supervisedImage.dockerImageId = id;
+						}
+					}
+				}
+				return _.filter(
+					supervisedImages,
+					image => !this.isAvailableInDocker(image, dockerImages),
```
Page-: You could use `_.reject` and drop the `!`

src/compose/images.ts
```diff
@@ -0,0 +1,652 @@
...
+
+	public async getAvailable(_localMode: boolean): Promise<Image[]> {
+		const images = await this.withImagesFromDockerAndDB(
+			(dockerImages, supervisedImages) =>
+				_.filter(supervisedImages, image =>
+					this.isAvailableInDocker(image, dockerImages),
+				),
+		);
+
+		// if (localMode) {
```
Page-: Commented code?

src/device-api/v2.ts
```diff
@@ -149,12 +149,13 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
 	);
 
 	// TODO: Support dependent applications when this feature is complete
-	router.get('/v2/applications/state', (_req: Request, res: Response) => {
+	router.get('/v2/applications/state', async (_req: Request, res: Response) => {
 		// It's kinda hacky to access the services and db via the application manager
 		// maybe refactor this code
+		const localMode = await deviceState.config.get('localMode');
```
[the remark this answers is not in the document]
CameronDiver: it's a bug that was caught with the typescript conversion, and seeing as it couldn't be built without fixing this I just added it as part of the same commit.

src/compose/volumes.ts
CameronDiver: It can't be because the dockerode typings currently say that it returns `Promise<any[]>` which is not compatible with `Dockerode.VolumeInfoList`

PR #877, opened 2019-01-24
Typescript conversions

CameronDiver:
There's some places that I'd love to simplify here, but I want at least a 1:1 conversion for now.


src/compose/service-manager.ts
```diff
@@ -0,0 +1,662 @@
...
+		const handoverCompletePaths = service.handoverCompleteFullPathsOnHost();
+
+		const wait = (): Bluebird<void> =>
+			Bluebird.any(
+				handoverCompletePaths.map(file =>
+					fs.stat(file).then(() => fs.unlink(file).catch(_.noop)),
+				),
+			).catch(() => {
+				if (Date.now() < deadline) {
+					return Bluebird.delay(pollInterval).then(wait);
```
Page-: This can be
```ts
await Bluebird.delay(pollInterval)
return wait()

File: src/compose/service-manager.ts [hunk 2]
```diff
@@ -0,0 +1,662 @@
...
+				if (data != null) {
+					const status = data.status;
+					if (status === 'die' || status === 'start') {
+						try {
+							let service: Service | null;
+							try {
+								service = await this.getByDockerContainerId(data.id);
+							} catch (e) {
+								if (NotFoundError(e)) {
+									service = null;
```
Page-: You can just initialise with null and change this to
```ts
if (!NotFoundError(e)) {
	throw e;
}

File: src/compose/service-manager.ts [hunk 3]
```diff
@@ -0,0 +1,662 @@
...
+		let containerId: string | null = null;
+
+		try {
+			const container = await this.create(service);
+			containerId = container.id;
+			this.logger.logSystemEvent(LogTypes.startService, { service });
+
+			this.reportNewStatus(containerId, service, 'Starting');
+
+			await Bluebird.resolve(container.start())
```
Page-: You can drop the `Bluebird.resolve` and just go full await/async here - there's nothing outside of the `.tapCatch` which is relatively easy to keep in await/async

src/compose/service-manager.ts
```diff
@@ -0,0 +1,662 @@
...
+			const nets = service.extraNetworksToJoin();
+
+			this.logger.logSystemEvent(LogTypes.installService, { service });
+			this.reportNewStatus(mockContainerId, service, 'Installing');
+
+			const container = await this.docker.createContainer(conf);
+			service.containerId = container.id;
+
+			await Promise.all(
+				_.map(nets, (endpointConfig, name) =>
```
Page-: I assume this is an object hence the `_.map`?

src/compose/service-manager.ts
```diff
@@ -0,0 +1,662 @@
...
+			);
+		}
+
+		try {
+			await this.docker
+				.getContainer(existingService.containerId)
+				.remove({ v: true });
+		} catch (e) {
+			if (NotFoundError(e)) {
+				return;
```
Page-: Could use `if (!NotFoundError(e))` instead

src/lib/docker-utils.ts
```diff
@@ -0,0 +1,331 @@
...
+						const deltaStream = applyDelta(imgSrc, {
+							log,
+							timeout: applyTimeout,
+						});
+						res
+							.pipe(deltaStream)
+							.on('id', id => resolve(`sha256:${id}`))
+							.on('error', err => {
+								log(`Delta stream emitted error: ${err}`);
+								req.abort();
```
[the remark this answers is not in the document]
CameronDiver: Possibly, I'll have to check, atm it's just 1:1.

src/lib/docker-utils.ts
```diff
@@ -0,0 +1,331 @@
...
+			return '127.0.0.1';
+		}
+
+		const network = await this.getNetwork(networkName).inspect();
+		const config = _.get(network, ['IPAM', 'Config', '0'], null);
+		if (config != null) {
+			if (config.Gateway != null) {
+				return config.Gateway;
+			}
+			if (config.Subnet != null && _.endsWith(config.Subnet, '.0/16')) {
```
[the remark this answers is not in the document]
CameronDiver: Agreed, I've made an issue here: https://github.com/balena-io/balena-supervisor/issues/878

src/compose/service-manager.ts
CameronDiver: It is, coming from https://github.com/balena-io/balena-supervisor/blob/master/src/compose/service.ts#L690

PR #887, opened 2019-02-04
Install newer npm in builds to ensure the use of npm ci

CameronDiver:
Instead of pinning dependencies one-by-one as issues come up, we just install the newer npm and always use npm ci in conjunction with the package-lock.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,10 @@ RUN apt-get update \
 
 COPY package.json package-lock.json /usr/src/app/
 
-RUN JOBS=MAX npm ci --no-optional --unsafe-perm || JOBS=MAX npm install --no-optional --unsafe-perm
+# install the newer NPM so we can use npm ci
+RUN npm install -g npm@6
+
+RUN JOBS=MAX npm ci --no-optional --unsafe-perm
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -49,7 +49,10 @@ RUN apt-get update \
 
 COPY package.json package-lock.json /usr/src/app/
 
-RUN JOBS=MAX npm ci --no-optional --unsafe-perm || JOBS=MAX npm install --no-optional --unsafe-perm
+# install the newer NPM so we can use npm ci
+RUN npm install -g npm@6
```
Page-: You can move this before the `COPY package.json package-lock.json /usr/src/app/` to increase caching

Dockerfile
```diff
@@ -85,10 +88,11 @@ RUN mkdir -p rootfs-overlay && \
 
 COPY package.json package-lock.json /usr/src/app/
 
+# install the newer NPM so we can use npm ci
+RUN npm install -g npm@6
```
Page-: You should be able to move this `RUN npm install -g npm@6` and the `RUN apt-get update ...` above to another build step and share the layers between `node-deps` and `node-build` in order to speed up builds/increase caching

Issue #873, opened 2019-01-22
Exponential backoff doesn't work for image fetch errors


pcarranzav:
In [device-state.coffee](https://github.com/balena-io/balena-supervisor/blob/master/src/device-state.coffee#L573) we have code to backoff before retrying whenever a step towards applying the target state fails.

However, image downloads are [triggered without waiting for the download promise to be fulfilled](https://github.com/balena-io/balena-supervisor/blob/master/src/compose/images.coffee#L94) and therefore run in the background. The device state engine uses the "downloading" status to figure out that it should trigger the download again while this is happening. This means that a download error doesn't cause a call to `applyError`, and therefore the download is retried in the next `applyTarget` iteration, every 1s.

To fix this, triggerFetch could throw an error if not enough time has passed since the last fetch attempt for a specific image - causing the `fetch` action to fail and the `applyError` to be called, backing off the next update attempt.

PR #888, opened 2019-02-04
Backoff on image download error

CameronDiver:
Change-type: patch
Closes: #873

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -4,6 +4,7 @@ import { EventEmitter } from 'events';
 import * as _ from 'lodash';
 import StrictEventEmitter from 'strict-event-emitter-types';
 
+import Config from '../config';
 import Database from '../db';
 import * as constants from '../lib/constants';
 import {
@@ -27,6 +28,7 @@ interface ImageConstructOpts {
 	docker: DockerUtils;
 	logger: Logger;
 	db: Database;
+	config: Config;
 }
 
 interface FetchProgressEvent {
@@ -65,6 +67,8 @@ export class Images extends (EventEmitter as {
 	private logger: Logger;
 	private db: Database;
 
+	public appUpdatePollInterval: number;
+
 	private imageFetchFailures: Dictionary<number> = {};
 	private imageFetchLastFailureTime: Dictionary<
 		ReturnType<typeof process.hrtime>
@@ -92,7 +96,7 @@ export class Images extends (EventEmitter as {
 			// engine, and ensure that we wait a bit lnger
 			const minDelay = Math.min(
 				2 ** this.imageFetchFailures[image.name] * constants.backoffIncrement,
-				constants.maxBackoffTime,
+				this.appUpdatePollInterval,
 			);
 			const timeSinceLastError = process.hrtime(
 				this.imageFetchLastFailureTime[image.name],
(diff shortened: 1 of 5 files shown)


src/lib/constants.ts
```diff
@@ -50,6 +50,10 @@ const constants = {
 	bootBlockDevice: '/dev/mmcblk0p1',
 	hostConfigVarPrefix: 'HOST_',
 	migrationBackupFile: 'backup.tgz',
+	// The amount of time to increase the wait by on each failure
```
Page-: Strictly speaking the backoff is doubled each time starting at this value, ie it'll go 500 -> 1000 -> 2000 -> 4000, rather than 500 -> 1000 -> 1500 -> 2000 which is what this comment implies

CameronDiver: Updated

src/compose/images.ts
```diff
@@ -108,6 +124,13 @@ export class Images extends (EventEmitter as {
 			return null;
 		} catch (e) {
 			if (!NotFoundError(e)) {
+				if (!(e instanceof ImageDownloadBackoffError)) {
+					this.imageFetchLastFailureTime[image.name] = Date.now();
```
pcarranzav: It might be better to use `process.hrtime()` here [like we did in the healthcheck functions](https://github.com/balena-io/balena-supervisor/blob/2c30b55a5c251f16ad1ffeb2b63c21091350f597/src/api-binder.ts#L120), to make this resilient to system clock changes (e.g. if ntp syncing is happening while we're downloading)

src/lib/constants.ts
```diff
@@ -50,6 +50,11 @@ const constants = {
 	bootBlockDevice: '/dev/mmcblk0p1',
 	hostConfigVarPrefix: 'HOST_',
 	migrationBackupFile: 'backup.tgz',
+	// Use this failure multiplied by 2**Number of failures to increase
+	// the backoff on subsequent failures
+	backoffIncrement: 500,
+	// The maximum time to backoff on repeated failure
+	maxBackoffTime: 30000,
```
pcarranzav: While we're at it, we might wanna consider increasing the max backoff time here? In the APIBinder [we backoff all the way up to the polling interval](https://github.com/balena-io/balena-supervisor/blob/master/src/api-binder.ts#L528) (which by default is 10 minutes now), and I wonder if here we should also do the same, to reduce bandwidth usage for devices that are failing to download for whatever reason, and also further reduce pressure on the registry/delta server/API.

src/compose/images.ts
```diff
@@ -74,13 +84,36 @@ export class Images extends (EventEmitter as {
 		this.docker = opts.docker;
 		this.logger = opts.logger;
 		this.db = opts.db;
+		this.config = opts.config;
+	}
+
+	public async init() {
+		this.appUpdatePollInterval = await this.config.get('appUpdatePollInterval');
```
pcarranzav: If the appUpdatePollInterval in config changes, it won't be updated here. You can add a `this.config.on('change',fn)` like in DeviceState (where you update maxPollTime), or have a setter that's called from DeviceState? (if you do the latter, you can just have that setter instead of an init function, and no need to have Config passed on the constructor)

Issue #841, opened 2018-12-11
(HTTP code 409) conflict - conflict: unable to delete (cannot be forced) - image is being used by running container


willswire:
When implementing the **delete-then-download** application update strategy, devices are unable to remove the existing container due to the following error. Confirmed supervisor version running >= v2.5.1.

`10.12.18 14:46:53 (-0500) Killing service 'main'`
`10.12.18 14:46:53 (-0500) Deleting image`
`10.12.18 14:46:53 (-0500) Failed to delete image due to '(HTTP code 409) conflict - conflict: unable to delete (cannot be forced) - image is being used by running container' `

williamwalker88:
It is possible to bypass this issue by first stopping the container, _then_ pushing the update.  The supervisor, under the **delete-then-download** strategy should do this rather than the user having to do so manually.

willswire:
Same error on supervisor ver 9.0.1

```
17.01.19 12:12:52 (-0500) Killing service 'main sha256:08251b0aa8d6c66bd1b240ea4cf172963e81cbd8bac252fb144ba7eaaa0b41a0'
17.01.19 12:12:52 (-0500) Deleting image 'registry2.balena-cloud.com/v2/0f74bea475cd7a34f257da6be491baa8@sha256:d8b60a410da5ab912d725ed19eba5624e71359814dc09f9f83b8b71ba77fc98f'
17.01.19 12:12:52 (-0500) Failed to delete image 'registry2.balena-cloud.com/v2/0f74bea475cd7a34f257da6be491baa8@sha256:d8b60a410da5ab912d725ed19eba5624e71359814dc09f9f83b8b71ba77fc98f' due to '(HTTP code 409) conflict - conflict: unable to delete 08251b0aa8d6 (cannot be forced) - image is being used by running container c3bc0919b429 '
```

CameronDiver:
Hey @willswire thanks for the report. I'll try to reproduce this soon and see what we can do. I imagine it's something like the supervisor not giving the container time to exit, so this would be where I'll start looking.

willswire:
@CameronDiver thanks!  If it helps at all, our current situation is:

- Deploying a single container image (base image: balenalib/amd64-node:jessie) with a total payload of **955.08 MB**
- Device Type: WYSE Zx0 (AMD64 Architecture) with 2GB flash storage

The device will stay in a constant loop, reporting the same error.

CameronDiver:
Thanks for the extra info;

[...]

PR #893, opened 2019-02-05
fix: Rework delete-then-download handling in state engine

CameronDiver:
In the original implementation it was possible that the delete did not
wait for the kill step to be finished, so it would not be deleted.

We seperate this process into two steps, to allow for the container to
have stopped before proceeding.

Change-type: patch
Closes: #841

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -545,8 +545,8 @@ module.exports = class ApplicationManager extends EventEmitter
 				return null
 		'kill-then-download': (current, target) ->
 			return serviceAction('kill', target.serviceId, current, target)
-		'delete-then-download': (current, target, needsDownload) ->
-			return serviceAction('kill', target.serviceId, current, target, removeImage: needsDownload)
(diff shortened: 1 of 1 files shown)


src/application-manager.coffee
```diff
@@ -841,7 +841,10 @@ module.exports = class ApplicationManager extends EventEmitter
 			return @bestDeltaSource(image, available)
 		proxyvisorImages = @proxyvisor.imagesInUse(current, target)
 
-		imagesToRemove = _.filter availableAndUnused, (image) ->
+		potentialDeleteThenDownload = _.filter current.local.apps.services, (svc) ->
+			return svc.config.labels['io.balena.update.strategy'] == 'delete-then-download' and svc.status in ['Stopped']
```
zvin: why not `svc.status == 'Stopped'` ?

CameronDiver: Ah, I was actually testing against `['Stopping', 'Stopped']` then I realised this was wrong, and forgot to change. Good catch!

Issue #895, opened 2019-02-07
Whitespace in front of the env var is causing an issue, getting containers in a restart loop


lekkas:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_h5fv)

lekkas:
The workaround is to delete the leading whitespace characters.

CameronDiver:
Closed by: #901

PR #896, opened 2019-02-07
Don't consider leading whitespace in environment variables on compare

CameronDiver:
Change-type: patch
Closes: #895


pcarranzav:
Weird, I also see what @Page- says. I wonder if it has something to do with how we convert in https://github.com/balena-io/balena-supervisor/blob/master/src/lib/conversions.ts#L26 and/or how we call the docker engine API?

CameronDiver:
Closing until the problem we're facing is better understood.

Issue #897, opened 2019-02-07
Defining adjacent ports in the compose file prevents the container from starting upo


imrehg:
Test case:
```
    ports:
      - '2000:2000'
      - '2001:2001'
```
results in container being continuously killed. Tested on both 9.0.1 and 9.6.2 supervisors
The equivalent
```
    ports:
      - '2000-2001:2000-2001'
```
works on the other hand in both.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_h5r7)

PR #898, opened 2019-02-08
fix: Normalize ports from compose file on instantiation

CameronDiver:
Adjacent ports are always grouped together by docker when reporting the
container state (from an inspect), so adjacent ports defined in the
compose file would not match as they would not have been normalized.

We make sure to always normalize the input port configuration, so that
it will match the docker output (if it should).

We also don't sort in the fromComposePorts function anymore as that is
handled by the normalize function.

Closes: #897
Change-type: patch


CameronDiver:
@resin-ci retest

PR #900, opened 2019-02-11
Make sure to correctly convert config emit events after validation

CameronDiver:
We were validating the input configuration values by coercing them to
the correct type, and then using the initial value to be saved (which
currently is always converted to a string).

We now use the coerced value as the actual value we will store, and more
importantly emit. This means that the config.on('change' ...) calls will
always be properly typed, which before this change was not a guarantee
that we could make.

Change-type: patch

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -174,8 +174,8 @@ export class Config extends (EventEmitter as {
 		};
 
 		return Bluebird.try(() => {
-			// Firstly validate all of the types as they are being set
-			this.validateConfigMap(keyValues);
+			// Firstly validate and coerce all of the types as they are being set
+			keyValues = this.validateConfigMap(keyValues);
 
 			if (trx != null) {
 				return setValuesInTransaction(trx).return();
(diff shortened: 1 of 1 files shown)


src/config/index.ts
```diff
@@ -175,7 +175,7 @@ export class Config extends (EventEmitter as {
 
 		return Bluebird.try(() => {
 			// Firstly validate all of the types as they are being set
-			this.validateConfigMap(keyValues);
+			keyValues = this.validateConfigMap(keyValues) as ConfigMap<T>;
```
Page-: It'd be better to change `validateConfigMap` to specify it returns a `ConfigMap<T>`, that way it keeps the typing as close to the source as possible and less likely to become out of date/wrong

src/config/index.ts
```diff
@@ -282,6 +282,7 @@ export class Config extends (EventEmitter as {
 					}`,
```
Page-: For the `throw new Error` a few lines above it might make sense to use a `TypeError`

PR #901, opened 2019-02-11
Fix whitespace handling in environment variables

CameronDiver:
We stop stripping whitespace from environment variables, and we also allow newlines in the env var body (to match what the docker daemon supports).

This has been tested with newlines and leading/trailing whitespace from a compose file, and leading/trailing whitespace from the dashboard (which does not allow newlines).

### DIFF
--- a/src/lib/conversions.ts
+++ b/src/lib/conversions.ts
@@ -4,7 +4,7 @@ import { EnvVarObject } from './types';
 
 export function envArrayToObject(env: string[]): EnvVarObject {
 	const toPair = (keyVal: string) => {
-		const m = keyVal.match(/^([^=]+)=\s*(.*)\s*$/);
+		const m = keyVal.match(/^([^=]+)=(.*)$/);
 		if (m == null) {
 			console.log(
 				`WARNING: Could not correctly parse env var ${keyVal}. ` +
--- a/src/lib/conversions.ts
+++ b/src/lib/conversions.ts
@@ -4,7 +4,7 @@ import { EnvVarObject } from './types';
 
 export function envArrayToObject(env: string[]): EnvVarObject {
 	const toPair = (keyVal: string) => {
-		const m = keyVal.match(/^([^=]+)=(.*)$/);
+		const m = keyVal.match(/^([^=]+)=([^]*)$/);
 		if (m == null) {
 			console.log(
 				`WARNING: Could not correctly parse env var ${keyVal}. ` +
--- a/test/15-conversions.spec.coffee
+++ b/test/15-conversions.spec.coffee
@@ -43,9 +43,11 @@ describe 'conversions', ->
 			'key2=test\ntest',
 			'key3=test ',
 			'key4= test '
+			'key5=test\r\ntest',
 		])).to.deep.equal({
 			key1: ' test',
 			key2: 'test\ntest',
 			key3: 'test ',
 			key4: ' test ',
+			key5: 'test\r\ntest'
 		})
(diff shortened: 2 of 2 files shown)


src/lib/conversions.ts
```diff
@@ -4,7 +4,7 @@ import { EnvVarObject } from './types';
 
 export function envArrayToObject(env: string[]): EnvVarObject {
 	const toPair = (keyVal: string) => {
-		const m = keyVal.match(/^([^=]+)=\s*(.*)\s*$/);
+		const m = keyVal.match(/^([^=]+)=((?:.|\n)*)$/);
```
Page-: You can use the `s` flag instead:
```suggestion
		const m = keyVal.match(/^([^=]+)=(.*)$/s);
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Advanced_searching_with_flags_2

test/15-conversions.spec.coffee
```diff
@@ -36,14 +36,3 @@ describe 'conversions', ->
 			expect(conversion.envArrayToObject('')).to.deep.equal({})
 			expect(conversion.envArrayToObject([])).to.deep.equal({})
 			expect(conversion.envArrayToObject(1)).to.deep.equal({})
-
-		it 'should ignore leading and trailing whitespace', ->
```
Page-: Can you change this to test that leading/trailing whitespace is handled correctly please

src/lib/conversions.ts
CameronDiver: @Page- we can't use this, as it's not supported by the node version on `i386-nlp`.

Page-: Ah that sucks, in that case it's probably better to use
```suggestion
		const m = keyVal.match(/^([^=]+)=([\s\S]*)$/);
```
as an equivalent that will match everything (your current version will not match `\r`)

test/15-conversions.spec.coffee
```diff
@@ -37,13 +37,15 @@ describe 'conversions', ->
...
-			})
+	it 'should correctly handle whitespace', ->
+		expect(conversion.envArrayToObject([
+			'key1= test',
+			'key2=test\ntest',
+			'key3=test ',
+			'key4= test '
+		])).to.deep.equal({
+			key1: ' test',
+			key2: 'test\ntest',
```
Page-: Can you add a `\r` test as well please

CameronDiver: good catch

src/lib/conversions.ts
```diff
@@ -4,7 +4,7 @@ import { EnvVarObject } from './types';
 
 export function envArrayToObject(env: string[]): EnvVarObject {
 	const toPair = (keyVal: string) => {
-		const m = keyVal.match(/^([^=]+)=\s*(.*)\s*$/);
+		const m = keyVal.match(/^([^=]+)=((?:.|\n|\r)*)$/);
```
Page-: This still doesn't handle all the newline-ish characters, check https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Special_characters_meaning_in_regular_expressions which recommends
```suggestion
		const m = keyVal.match(/^([^=]+)=([^]*)$/);
```
if you don't need old IE (I think my previous suggestion of `[\s\S]` works in old IE as it's what is usually suggested but is maybe less performant and that's why mdn has the other suggestion?)

Issue #843, opened 2018-12-11
Healtheck logic changes


imrehg:
We want to not mark the healthcheck as unhealthy if we recieve any http status back from the api - this means that something about the data is wrong, but the connectivity is not.

[Arch call](https://app.frontapp.com/open/cnv_14sprkr)

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_fz7f)

shaunmulligan:
@CameronDiver did we make the healthcheck improvements for this issue yet?

CameronDiver:
Closed by https://github.com/balena-io/balena-supervisor/pull/902

PR #902, opened 2019-02-12
Don't treat a non-200 status response on patch as report errors

CameronDiver:
Non-200 errors were causing the watchdog to restart the supervisor,
which in some cases could cause a restart loop. Instead we change the
code to only treat communication failures as an error, and report status
code failures directly.

Change-type: patch
Closes: #843

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -3,7 +3,7 @@ import * as bodyParser from 'body-parser';
 import * as express from 'express';
 import * as _ from 'lodash';
 import * as Path from 'path';
-import { PinejsClientRequest } from 'pinejs-client-request';
+import { PinejsClientRequest, StatusError } from 'pinejs-client-request';
 import * as deviceRegister from 'resin-register-device';
 import * as url from 'url';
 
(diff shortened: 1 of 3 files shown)


src/api-binder.ts
```diff
@@ -460,7 +460,18 @@ export class APIBinder {
 
 		await Bluebird.resolve(
 			this.sendReportPatch(stateDiff, { apiEndpoint, uuid }),
-		).timeout(conf.apiTimeout);
+		)
+			.catch(StatusError, (e: StatusError) => {
+				// We don't this to be classed as a report error, as this will cause the
```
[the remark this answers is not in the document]
CameronDiver: heh, good catch :)

src/api-binder.ts
```diff
@@ -460,7 +460,18 @@ export class APIBinder {
 
 		await Bluebird.resolve(
 			this.sendReportPatch(stateDiff, { apiEndpoint, uuid }),
-		).timeout(conf.apiTimeout);
+		)
+			.catch(StatusError, (e: StatusError) => {
```
Page-: This would have made sense to me as a normal `try/catch` as we're in an `await`

CameronDiver: Maybe, but we had to assign the timeout anyway, and the predicate argument of bluebird's catch can make things a bit nicer.

Page-: I know that's why we have the bluebird wrapper, and whilst the sugar is nice it does deviate from what is done everywhere else in the (TS) codebase so I'd go with consistency for `.then`/`.catch` style (ie try/catch instead of .then/.catch)

PR #905, opened 2019-02-13
Better backoff and update lock reporting

CameronDiver:
When we get a new target state, we reset the apply error counter to 0, meaning we'll try immediately to apply the new state.

We also show better logs when updates are locked.

### DIFF
--- a/src/device-state.coffee
+++ b/src/device-state.coffee
@@ -342,6 +342,10 @@ module.exports = class DeviceState extends EventEmitter
 		Promise.using @_inferStepsLock, -> fn()
 
 	setTarget: (target, localSource = false) ->
+		# When we get a new target state, clear any built up apply errors
+		# This means that we can attempt to apply the new state instantly
+		@failedUpdates = 0
+
 		Promise.join(
 			@config.get('apiEndpoint'),
 			validateState(target),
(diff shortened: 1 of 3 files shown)


src/device-state.coffee
```diff
@@ -132,7 +132,8 @@ module.exports = class DeviceState extends EventEmitter
 		@router = createDeviceStateRouter(this)
 		@on 'apply-target-state-end', (err) ->
 			if err?
-				console.log("Apply error #{err}")
+				if not err instanceof UpdatesLockedError
```
Page-: This needs to be
```suggestion
				if not(err instanceof UpdatesLockedError)
```

src/device-state.coffee
```diff
@@ -571,11 +576,15 @@ module.exports = class DeviceState extends EventEmitter
 		@failedUpdates += 1
 		@reportCurrentState(update_failed: true)
 		if @scheduledApply?
-			console.log("Updating failed, but there's another update scheduled immediately: ", err)
+			if not err instanceof UpdatesLockedError
```
Page-: This needs to be
```suggestion
			if not(err instanceof UpdatesLockedError)
```

PR #915, opened 2019-02-22
Increase max payload size in bodyparser to avoid PayloadTooLarge errors

CameronDiver:
Change-type: patch


CameronDiver:
Setting a local state with `balena push` and b-on-b. The env vars pushed the payload size over the default bodyparser limit.

PR #920, opened 2019-02-27
Add migration for delta apply timeout without strict checking

CameronDiver:
Change-type: patch

### DIFF
--- a/src/migrations/M00004.js
+++ b/src/migrations/M00004.js
@@ -1,0 +1,21 @@
+exports.up = function(knex) {
+	return knex('deviceConfig')
+		.select('targetValues')
+		.then(([target]) => {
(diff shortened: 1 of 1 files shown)


src/migrations/M00004.js
```diff
@@ -0,0 +1,21 @@
+exports.up = function(knex) {
+	return knex('deviceConfig')
+		.select('targetValues')
+		.then(([target]) => {
+			const targetValues = target.targetValues;
+			const jsonValues = JSON.parse(targetValues);
+
+			if (jsonValues['SUPERVISOR_DELTA_APPLY_TIMEOUT'] == '') {
```
Page-: Why don't we want strict typing?  What other values are you trying to match against? `0`/`[]`/`false`/?

PR #923, opened 2019-03-07
state-engine: Add rate limited steps to device-config

CameronDiver:
In the case of an airgapped supervisor, with a target state that
requests the vpn be enabled, the supervisor will constantly loop on
trying to set the vpn to on. Unfortunately the vpn requires an internet
connection to be configured, so it will never be turned on.

We add the concept of no-ops to the device-config state change steps,
and don't end the state engine transition while these are present
(similar to how image pulls are implemented).

Change-type: minor

### DIFF
--- a/src/device-state.coffee
+++ b/src/device-state.coffee
@@ -578,8 +578,7 @@ module.exports = class DeviceState extends EventEmitter
 			console.log('Scheduling another update attempt due to failure: ', delay, err)
 			@triggerApplyTarget({ force, delay, initial })
 
-	applyTarget: ({ force = false, initial = false, intermediate = false, skipLock = false } = {}) =>
-		nextDelay = 200
+	applyTarget: ({ force = false, initial = false, intermediate = false, skipLock = false, nextDelay = 200, retryCount = 0 } = {}) =>
 		Promise.try =>
 			if !intermediate
 				@applyBlocker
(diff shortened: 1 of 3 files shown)


src/device-state.coffee
```diff
@@ -594,11 +593,23 @@ module.exports = class DeviceState extends EventEmitter
 					.then ([ currentState, targetState ]) =>
 						@deviceConfig.getRequiredSteps(currentState, targetState)
 						.then (deviceConfigSteps) =>
-							if !_.isEmpty(deviceConfigSteps)
-								return deviceConfigSteps
+							noConfigSteps = _.every(deviceConfigSteps, ({ action }) -> action is 'noop')
+							if !_.isEmpty(deviceConfigSteps) && not noConfigSteps
```
pcarranzav: nit: this could just be `if not noConfigSteps`, since noConfigSteps will be true when deviceConfigSteps is empty

src/device-state.coffee
```diff
@@ -594,11 +593,23 @@ module.exports = class DeviceState extends EventEmitter
...
+								return [true, deviceConfigSteps]
```
pcarranzav: It's not gonna have any real effect, but shouldn't this be `[false, deviceConfigSteps]`?

PR #924, opened 2019-03-08
Only modprobe ip6_tables if it's necessary

pcarranzav:
If the kernel was built with support for ip6tables, there's no need to load the module. This is the case when running
balenaOS in a container in Mac OS, which also can't do modprobes easily.

Change-type: patch


CameronDiver:
@pcarranzav is this waiting for anything? (Let me know if you want me to get this and the other approved PRs merged)

PR #925, opened 2019-03-08
Add libusb support

prodrigestivill:
Add `io.balena.features.usb` label to allow using libusb inside the containers.

From [balena forums](https://forums.balena.io/t/libusb0-py-cannot-open-a-directory/62).

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -338,6 +338,10 @@ export function addFeaturesFromLabels(
 		service.config.volumes.push('/lib/firmware:/lib/firmware');
 	}
 
+	if (checkTruthy(service.config.labels['io.balena.features.usb'])) {
+		service.config.volumes.push('/dev/bus/usb:/dev/bus/usb');
+	}
+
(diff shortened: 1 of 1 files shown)


CameronDiver:
Hey @prodrigestivill thanks for the PR!

I just want to run this past our OS team to ensure everything is good on that front, but I don't expect that should take too long.

In the meantime, would you be ok adding a `change-type: minor` footer at the bottom of your commit body? This enables the automatic versioning system we have to work properly. I can also do this for you if you prefer!

CameronDiver:
Note that it should also be possible to do the same thing using the devices docker-compose field: https://docs.docker.com/compose/compose-file/compose-file-v2/#devices

prodrigestivill:
The problem with docker-compose devices is that the actual usb path to the device is needed to know before hand. So deploying this into several devices would not be viable.

From the forums I can see that apparently this issue is related with balena-os/balena-engine#125.
Not sure if the new approach will include support for libusb to communicate with any device.

For now we will use the hack [mentioned](https://forums.balena.io/t/libusb0-py-cannot-open-a-directory/62/17) in the forums.

PR #929, opened 2019-03-11
Fix changing local mode environment variables from compose file

CameronDiver:


test/14-application-manager.spec.coffee
```diff
@@ -262,7 +262,7 @@ describe 'ApplicationManager', ->
 					image: @applications.imageForService(target.local.apps[0].services[0])
 					serviceId: 23
 					appId: 1234
-				}])
+				}, { action: 'noop', appId: 1234 }])
```
pcarranzav: I think this should use `expect(steps).to.eventually.have.deep.members` as the test case below, so that it works if the ordering changes

PR #930, opened 2019-03-12
Pin lodash types to avoid compile error

CameronDiver:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@types/express": "^4.11.1",
     "@types/knex": "^0.14.14",
     "@types/lockfile": "^1.0.0",
-    "@types/lodash": "^4.14.119",
+    "@types/lodash": "4.14.122",
     "@types/memoizee": "^0.4.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "@types/express": "^4.11.1",
     "@types/knex": "^0.14.14",
     "@types/lockfile": "^1.0.0",
-    "@types/lodash": "^4.14.119",
+    "@types/lodash": "4.14.119",
```
Page-: 4.14.122 should be fine to use, it's 4.14.123 that probably broke it due to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33228

Page-: (I've experienced the issue elsewhere that suddenly `_.isObject` type casts everything to `{}` which you then can't do anything with)

CameronDiver: I just took it back to the last known working version (which luckily was only 4 patch versions). I'll give this a try, but the PR looks unrelated.

Issue #918, opened 2019-02-25
Docker-compose network service names not resolving correctly when `networks:` configuration is used


mikesimos:
<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_hi43)

CameronDiver:
![](https://github.trello.services/images/mini-trello-icon.png) [Use service names as domain names e.g. curl http://api/test](https://trello.com/c/1GO4dlHb/55-use-service-names-as-domain-names-eg-curl-http-api-test)

gelbal:
Thread in forums where users are waiting for the release of this fix:
https://forums.balena.io/t/docker-compose-network-service-names-not-resolving-correctly/5313/11

Please ping the thread when we release.

OdysLam:
can we get a confirmation on whether thus update will be included in the BalenaOs64 as in [here](https://github.com/balena-os/balena-raspberrypi/issues/181)

CameronDiver:
@OdysLam Yes this fix will be included in all device types, in balenaOS version 2.32.0

OdysLam:
> @OdysLam Yes this fix will be included in all device types, in balenaOS version 2.32.0

Hey @CameronDiver ,

Any ETA on that version? I am asking as I am using Balena embedded in the solution for my Thesis and I would love to program vaguely when I will be back at looking to it.

Thanks for the swift response.

Cheers!

CameronDiver:
It's currently going through internal testing, and things look good. I imagine it will be available by the end of this week or early next week, but that's not completely certain. 

If testing doesn't bring up any regressions we should be fine, but obviously, if it does those will need to be addressed first.

OdysLam:
Amazing news, this timeline is more than accurate that I need. Thanks!

On a side note, it got my attention that `hostname` is also bugged, as although I included it in my `docker-compose.yaml` the service was unreachable using the host-name but reachable using the service name.

The new version will also fix this particular issue?

Thanks!

CameronDiver:
Closed by https://github.com/balena-io/balena-supervisor/pull/933

CameronDiver:
@OdysLam Sorry I missed your message. Is your container using host networking?

If so, it could be this issue: https://github.com/balena-io/balena-supervisor/issues/994
There's a workaround posted here: https://forums.balena.io/t/hostname-issues/5813

PR #933, opened 2019-03-13
Fix network alias creation and comparison

CameronDiver:
Before this change, service name resolution would only occur in the
default network. This was because we were not explicitly adding aliases
of the service names to the aliases fields.

We also fix the comparison, which would do funny things based on
container IDs, which was correct but unnecessary.

Also fix a couple of lint errors.

Front logo [Front conversations](https://app.frontapp.com/open/top_i3ln)

### DIFF
--- a/src/app.coffee
+++ b/src/app.coffee
@@ -9,8 +9,6 @@ do ->
 			return lookup(name, { verbatim: true }, opts)
 		return lookup(name, Object.assign({ verbatim: true }, opts), cb)
 
-require('log-timestamp')
-
 Supervisor = require './supervisor'
 
 supervisor = new Supervisor()
--- a/src/compose/service-manager.ts
+++ b/src/compose/service-manager.ts
@@ -20,6 +20,7 @@ import {
 import * as LogTypes from '../lib/log-types';
 import { checkInt, isValidDeviceName } from '../lib/validation';
 import { Service } from './service';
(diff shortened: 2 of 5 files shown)


pcarranzav:
I think this closes #918

jo-tham:
Hey guys, thanks very much for this change. Lots happening here - so thanks in advance for you time. 

Any idea on the time frame for this to be merged/released? 

Anything users can do to prioritise/contribute?

pcarranzav:
Hi @jo-tham - this still needs some work, and we're currently working on fixing a few other issues so that we can do a balenaOS release. We'll get to this one after that, so I *think* it's likely this'll get merged next week (unless anything unexpected comes up) and it would then come out with the next OS release after that.

pcarranzav:
@CameronDiver I think for this to work, you'll need to add a call to `ComposeUtils.serviceNetworksToDockerNetworks` in `service.extraNetworksToJoin` or when it's called inside ServiceManager's `create` ( https://github.com/balena-io/balena-supervisor/blob/2a105723b4061e4c065c2f546f4a6c5cb93732c7/src/compose/service-manager.ts#L260 ) so that the `aliases` field is converted to `Aliases` before sending it to docker.

CameronDiver:
Great catch @pcarranzav ! I've made that change and rebased.

PR #934, opened 2019-03-14
Add a random jitter to target state polls, and a config var to ignore…

pcarranzav:
… update notifications and not poll immediately after startup

This commit does two related things:

* We make the poll interval a random time between 0.5 and 1.5 times the configured interval.
* We introduce the BALENA_SUPERVISOR_INSTANT_UPDATE_TRIGGER configuration variable, that defaults to true. If this variable is set
to false, then calls to /v1/update are ignored, and on startup the supervisor waits for a poll interval before getting the target state.

This will help especially on cases where there's a large number of devices on a single network. By disabling instant updates and setting a large
poll interval, we can now achieve a sitation where not all devices apply an update at the same time, which can help avoid
overwhelming the network.

Change-type: minor


src/api-binder.ts
```diff
@@ -912,7 +920,15 @@ export class APIBinder {
 		router.post('/v1/update', (req, res) => {
 			apiBinder.eventTracker.track('Update notification');
 			if (apiBinder.readyForUpdates) {
-				apiBinder.getAndSetTargetState(req.body.force, true).catch(_.noop);
+				this.config.get('instantUpdates').then(instantUpdates => {
+					if (instantUpdates) {
+						apiBinder.getAndSetTargetState(req.body.force, true).catch(_.noop);
+					} else {
+						console.log(
+							'Ignoring update notification because instant updates are disabled',
```
CameronDiver: Perhaps we should return a different http header, for example `405 - method not allowed` ?

pcarranzav: @CameronDiver maybe a different code would make sense, but I think that can cause issues since the balena API won't be expecting that?

pcarranzav: I guess we could use a 202 Accepted?

> The request has been accepted for processing, but the processing has not been completed. The request might or might not eventually be acted upon, as it might be disallowed when processing actually takes place.

From https://www.restapitutorial.com/httpstatuscodes.html

Cause I wouldn't want to consider this case an error.

CameronDiver: Yeah I think that makes sense, I would just like a method of detecting that this isn't going to work. If the API doesn't support it then that needs to be changed IMO.

pcarranzav:
@CameronDiver made it so that we use 202 if we don't apply immediately (also if APIBinder isn't ready), and added a 503 for any unexpected errors.

PR #945, opened 2019-03-21
Fixes for the provisioning and pinning process

pcarranzav:
* Fix awaiting for provisioning to complete before continuing the APIBinder initialization
* Fix running the device pinning code when we're retrying but the device is already provisioned

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -875,7 +875,7 @@ export class APIBinder {
 				delay: retryDelay,
 			});
 			await Bluebird.delay(retryDelay);
-			this.provisionOrRetry(retryDelay);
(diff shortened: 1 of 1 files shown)


src/api-binder.ts
```diff
@@ -875,7 +875,7 @@ export class APIBinder {
 				delay: retryDelay,
 			});
 			await Bluebird.delay(retryDelay);
-			this.provisionOrRetry(retryDelay);
+			return this.provisionOrRetry(retryDelay);
```
Page-: I think this should probably be `await` as we don't actually want to return a value and just want to wait on the promise chain (but either version will work for our purposes)

PR #949, opened 2019-03-27
When a device config variables requested value is not valid, fallback to the default

CameronDiver:
We also show that this is being ignored in the dashboard logs, to avoid confusion. This change also includes the ability to not track an event, taken from https://github.com/balena-io/balena-supervisor/pull/905/commits/6846229a0d77bc349f18cfc4bf7a3aec3318485c but that is not ready for merge yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,18 +2,18 @@ ARG ARCH=amd64
 
 # The node version here should match the version of the runtime image which is
 # specified in the base-image subdirectory in the project
-FROM resin/rpi-node:6.13.1-slim as rpi-node-base
-FROM resin/armv7hf-node:6.13.1-slim as armv7hf-node-base
-FROM resin/aarch64-node:6.13.1-slim as aarch64-node-base
+FROM balenalib/rpi-node:6.16.0 as rpi-node-base
+FROM balenalib/armv7hf-node:6.16.0 as armv7hf-node-base
+FROM balenalib/aarch64-node:6.16.0 as aarch64-node-base
 RUN [ "cross-build-start" ]
 RUN sed -i '/security.debian.org jessie/d' /etc/apt/sources.list
 RUN [ "cross-build-end" ]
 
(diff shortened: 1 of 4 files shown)


src/device-config.ts
```diff
@@ -395,10 +397,39 @@ export class DeviceConfig {
...
+							false,
+						);
+						if (
+							!DeviceConfig.configTest(
+								varType,
+								current[envVarName],
+								defaultValue,
+							)
+						) {
+							// Set it to the default value if it is different to the current
```
pcarranzav: If you move the warning log here, you can make it be logged only once, otherwise it'll be logged on every call to getRequiredSteps (because current and target will never be the same, but current and default will)

CameronDiver: Ah yeah, very good point - only report it when we initially fall back to the default :+1:

Dockerfile
```diff
@@ -2,18 +2,18 @@ ARG ARCH=amd64
 
 # The node version here should match the version of the runtime image which is
 # specified in the base-image subdirectory in the project
-FROM resin/rpi-node:6.13.1-slim as rpi-node-base
-FROM resin/armv7hf-node:6.13.1-slim as armv7hf-node-base
-FROM resin/aarch64-node:6.13.1-slim as aarch64-node-base
+FROM balenalib/rpi-node:6.16.0 as rpi-node-base
+FROM balenalib/armv7hf-node:6.16.0 as armv7hf-node-base
```
pcarranzav: I think this will be fine, but this *could* cause issues given the build and runtime will be using slightly different nodejs versions. I'd strongly recommend either sticking to 6.13.1 here (and creating balenalib/ARCH-node builds) or moving balena-supervisor-base to 6.16.0.

CameronDiver: Yeah I think this is fine with just a minor version, but it's worth moving the base too. The only problem there is the base is on v8, and that causes big issues with deltas, so we can't do that. Getting the other base images generated is probably the easiest.

PR #954, opened 2019-03-29
CHANGELOG.md: Add entry for 9.11.4 that was skipped by versionbot

pcarranzav:


pcarranzav:
Versionbot checks will fail because there's no change-type, and we'll have to merge using admin rights.

PR #955, opened 2019-04-01
Better supervisor and container logging 

CameronDiver:
We remove the `Container already running` log, which was a massive source of confusion for users, replacing it with a `Supervisor starting` message instead.

We then refactor the code which requests logs from the docker daemon, into a purpose built class, keeping track of the streams required for this.

Commit:
> Container logging is now handled by a class which attaches and emits
information from the container. We add these to the directory
logging-backends/, and rename it to logging/.

We then add a database table to keep track of which logs have been sent, to ensure we send all logs of a container's lifecycle.
Commit:
> We add a database table, which holds information about the last
timestamp of a log successfully reported to a backend (local or remote).
We then use this value to calculate from which point in time to start
reporting logs from the container. If this is the first time we've seen
a container, we get all logs, and for every log reported we save the
timestamp. If it is not the first time we've seen a container, we
request all logs since the last reported time, ensuring no interruption
of service.


src/logging/container.ts
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	public async attach(lastSentTimestamp: number) {
+		const logOpts = {
+			follow: true,
+			timestamps: true,
+			since: Math.floor(lastSentTimestamp / 1000),
+		};
+		const stdoutLogOpts = _.merge({ stdout: true, stderr: false }, logOpts);
+		const stderrLogOpts = _.merge({ stderr: true, stdout: false }, logOpts);
```
zvin:
```typescript
const stdoutLogOpts = { stdout: true, stderr: false, ...logOpts};
const stderrLogOpts = { stderr: true, stdout: false, ...logOpts};
```

src/logging/container.ts
```diff
@@ -0,0 +1,98 @@
...
+						);
+					})
+					.pipe(es.split())
+					.on('data', (logBuf: Buffer | string) => {
+						if (_.isString(logBuf)) {
+							logBuf = Buffer.from(logBuf);
+						}
+						const logMsg = ContainerLogs.extractMessage(logBuf);
+						if (logMsg != null) {
+							this.emit('log', _.merge({ isStdout }, logMsg));
```
zvin:
```typescript
this.emit('log', { isStdout, ...logMsg });
```

src/logging/container.ts
```diff
@@ -0,0 +1,98 @@
...
+						);
+					})
+					.on('end', () => this.emit('closed'));
+			},
+		);
+	}
+
+	private static extractMessage(
+		msgBuf: Buffer,
+	): { message: string; timestamp: number } | null {
```
Page-: I'd usually go with `| undefined` and `return`

src/logging/container.ts
```diff
@@ -0,0 +1,98 @@
...
+		// Non-tty message format from:
+		// https://docs.docker.com/engine/api/v1.30/#operation/ContainerAttach
+		if (msgBuf[0] in [0, 1, 2] && _.every(msgBuf.slice(1, 7), c => c === 0)) {
```
Page-: This is typescript, `in` doesn't mean the same thing as coffeescript so this needs to be
```suggestion
		if (_.includes([0, 1, 2], msgBuf[0]) && _.every(msgBuf.slice(1, 7), c => c === 0)) {
```

src/logging/container.ts
Page-: That's equivalent to `_.assign` not to `_.merge` @zvin, that said I expect @CameronDiver actually wants the assign behaviour and not merge behaviour here

src/logging/container.ts
Page-: Same here, I expect the assign behaviour is what is wanted but it's worth checking as this isn't a 1-1 with `_.merge`

src/logger.ts
```diff
@@ -139,20 +133,47 @@ export class Logger {
...
+
+		return Bluebird.using(this.lock(containerId), async () => {
+			const logs = new ContainerLogs(containerId, docker);
+			this.containerLogs[containerId] = logs;
+			logs.on('error', err => {
+				console.error(`Container log retrieval error: ${err}`);
+				delete this.containerLogs[containerId];
+			});
+			logs.on('log', async logMessage => {
+				this.log(_.merge({}, serviceInfo, logMessage));
```
Page-: Do you actually want merge behaviour here or are you expecting assign behaviour?

src/logging/container.ts
CameronDiver: I always forget about that syntax, but yeah this is what I want.

src/logging/container.ts
CameronDiver: Good catch! This is actually code from a while ago, so that bug has been there for ages!

src/application-manager.coffee
```diff
@@ -75,6 +75,16 @@ module.exports = class ApplicationManager extends EventEmitter
 		@_targetVolatilePerImageId = {}
 		@_containerStarted = {}
 
+		# Rather than relying on removing out of data database entries when we're no
+		# longer using them, set a task that runs periodically to clear out the database
+		# This has the advantage that if for some reason a container is removed while the
+		# supervisor is down, we won't have zombie entries in the db
+		setInterval =>
+			@docker.listContainers(all: true).then (containers) =>
+				@logger.clearOutOfDateDBLogs(_.map(containers, 'Id'))
```
pcarranzav: So this runs only after the supervisor has been up for 24 hours... would it be good to also run it on startup? so that we it doesn't get out of hand if the supervisor is rebooted within less than a day every time

PR #958, opened 2019-04-05
avahi: Remove default service definitions

hedss:
Removes default 'example' service definitions that
are included by Avahi 0.7+. These conflict with
our balenaOS advertised services, causing potential
issues.

Connects-to: #957
Change-type: patch
Signed-off-by: Heds Simons <heds@balena.io>

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -2,7 +2,9 @@
 
 set -o errexit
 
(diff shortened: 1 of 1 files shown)


nazrhom:
@resin-ci retest

PR #966, opened 2019-04-16
fix: Run db log cleanup on init - after migrations

CameronDiver:
Before this change the first time the cleanup code runs would be before
the migrations have had a chance to execute. This change makes it so
that the cleanup code always runs once the migrations have finished.

Change-type: patch

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -75,19 +75,6 @@ module.exports = class ApplicationManager extends EventEmitter
 		@_targetVolatilePerImageId = {}
 		@_containerStarted = {}
 
-		# Rather than relying on removing out of data database entries when we're no
-		# longer using them, set a task that runs periodically to clear out the database
-		# This has the advantage that if for some reason a container is removed while the
-		# supervisor is down, we won't have zombie entries in the db
-		setInterval =>
-			@docker.listContainers(all: true).then (containers) =>
-				@logger.clearOutOfDateDBLogs(_.map(containers, 'Id'))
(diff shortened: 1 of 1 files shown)


src/application-manager.coffee
```diff
@@ -200,6 +187,19 @@ module.exports = class ApplicationManager extends EventEmitter
 		.then (interval) =>
 			@images.appUpdatePollInterval = interval
 			@images.cleanupDatabase()
+		.then =>
+			# Rather than relying on removing out of data database entries when we're no
+			# longer using them, set a task that runs periodically to clear out the database
+			# This has the advantage that if for some reason a container is removed while the
+			# supervisor is down, we won't have zombie entries in the db
+			setInterval =>
```
Page-: This would be nicer as
```coffee
fn = =>
	@docker.listContainers(all: true).then (containers) =>
		@logger.clearOutOfDateDBLogs(_.map(containers, 'Id'))
setInterval(fn, 1000 * 60 * 60 * 24)
fn()
```

src/application-manager.coffee
```diff
@@ -200,6 +187,19 @@ module.exports = class ApplicationManager extends EventEmitter
 		.then (interval) =>
 			@images.appUpdatePollInterval = interval
 			@images.cleanupDatabase()
+		.then =>
+			# Rather than relying on removing out of data database entries when we're no
```
Page-:
```suggestion
			# Rather than relying on removing out of date database entries when we're no
```

PR #981, opened 2019-05-10
Cache delta auth token request results for 10 minutes

CameronDiver:
Change-type: patch

### DIFF
--- a/src/lib/docker-utils.ts
+++ b/src/lib/docker-utils.ts
@@ -1,6 +1,7 @@
 import { DockerProgress, ProgressCallback } from 'docker-progress';
 import * as Dockerode from 'dockerode';
 import * as _ from 'lodash';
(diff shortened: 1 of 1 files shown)


src/lib/docker-utils.ts
```diff
@@ -27,6 +28,18 @@ interface RsyncApplyOptions {
...
+interface ImageNameParts {
+	registry: string;
+	imageName: string;
+	tagName: string;
+	digest: string;
+}
+
+// How long do we keep a delta token before invalidating it
+// (10 mins)
+const DELTA_TOKEN_TIMEOUT = 60 * 10 * 1000;
```
Page-: I'd prefer
```suggestion
const DELTA_TOKEN_TIMEOUT = 10 * 60 * 1000;
```
to go M/S/ms instead of `S/M/ms`

Issue #987, opened 2019-05-28
TimeoutError: Knex: Timeout acquiring a connection 


CameronDiver:
This error stops all state application and logging, as this relies on the database. Seen on-and-off with several devices.

This [forum thread](https://forums.balena.io/t/when-logging-a-lot-the-supervisor-seems-to-have-issues/11824) seems to point to it coming from logging too quickly, this makes sense, as the supervisor will record the timestamp of the latest reported message, to ensure nothing is lost. My first thought it to apply some kind of debounce to this, to stop unnecessary thrashing of the db. We often overwrite the same values too, which means this is a good optomisation for SD cards etc.

PR #988, opened 2019-05-28
Add in-memory cache around container logs saving, to reduce db load

CameronDiver:
Changes are collected together and exist in memory, for querying and
saving. Once every 10 mins, every changed timestamp is flushed to the
database.

Change-type: patch
Closes: #987

### DIFF
--- a/src/logging/monitor.ts
+++ b/src/logging/monitor.ts
@@ -1,0 +1,82 @@
+import Database from '../db';
+
+// Flush every 10 mins
+const DB_FLUSH_INTERVAL = 10 * 60 * 1000;
+
+/**
+ * This class provides a wrapper around the database for
+ * saving the last timestamp of a container
+ */
+export class LogMonitor {
+	private timestamps: { [containerId: string]: number } = {};
+	private writeRequired: { [containerId: string]: boolean } = {};
+
+	public constructor(private db: Database) {
+		setInterval(() => this.flushDb(), DB_FLUSH_INTERVAL);
+	}
+
+	public updateContainerSentTimestamp(
+		containerId: string,
+		timestamp: number,
+	): void {
+		this.timestamps[containerId] = timestamp;
+		this.writeRequired[containerId] = true;
+	}
+
+	public async getContainerSentTimestamp(containerId: string): Promise<number> {
+		// If this is the first time we are requesting the
+		// timestamp for this container, request it from the db
+		if (this.timestamps[containerId] == null) {
+			// Set the timestamp to 0 before interacting with the
+			// db to avoid multiple db actions at once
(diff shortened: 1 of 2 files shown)


src/logging/monitor.ts
```diff
@@ -0,0 +1,84 @@
...
+		return this.timestamps[containerId] || 0;
+	}
+
+	private async flushDb() {
+		console.log('Attempting container log timestamp flush...');
+		const containerIds = Object.getOwnPropertyNames(this.timestamps);
+		try {
+			for (const containerId of containerIds) {
+				// Avoid writing to the db if we don't need to
+				if (this.writeRequired[containerId]) {
```
zvin: There's a missing `!` I think.

src/logging/monitor.ts
```diff
@@ -0,0 +1,84 @@
...
+
+				this.writeRequired[containerId] = false;
+				await this.db
+					.models('containerLogs')
+					.where({ containerId })
+					.update({ lastSentTimestamp: this.timestamps[containerId] });
+			}
+		} catch (e) {
+			console.error(
+				`There was an error storing the container log timestamps: ${e}\n${
```
Page-: I think `${e}` will already include the stack trace, so the explicit e.stack will result in double logging the trace

src/logging/monitor.ts
```diff
@@ -0,0 +1,84 @@
...
+
+	public updateContainerSentTimestamp(
+		containerId: string,
+		timestamp: number,
+	): void {
+		this.timestamps[containerId] = timestamp;
+		this.writeRequired[containerId] = true;
+	}
+
+	public async getContainerSentTimestamp(containerId: string): Promise<number> {
```
Page-: This function isn't safe to be called multiple times in parallel with the same `containerId` when this.timestamps[containerId] is null btw

CameronDiver: Whilst this should be completely fine in the way that it's used, I've added an `this.timestamps[containerId] = 0` before ever `await`ing on the database calls. This *could* make some logs be sent more than once, but we're talking the time it takes to write to the database, so one or two absolute max (assuming nothing else is going on).

src/logging/monitor.ts
```diff
@@ -0,0 +1,82 @@
...
+		const containerIds = Object.getOwnPropertyNames(this.timestamps);
+		try {
+			for (const containerId of containerIds) {
+				// Avoid writing to the db if we don't need to
+				if (!this.writeRequired[containerId]) {
+					continue;
+				}
+
+				this.writeRequired[containerId] = false;
+				await this.db
```
zvin: If the update fails, `this.writeRequired[containerId]` will be `false` anyway.

PR #989, opened 2019-05-29
Update balena-supervisor-base to v1.4.7

CameronDiver:
Change-type: patch


zrzka:
@hedss go away!

PR #992, opened 2019-05-29
Update and improve debug and development tools

CameronDiver:
Add a debounce to the livepush invocations, execute on start and also
add a wait on the supervisor CMD line for those rare occassions where
the supervisor enters a restart loop.

Change-type: patch


CameronDiver:
@balena-ci retest

PR #995, opened 2019-06-06
Standardise all supervisor output logging

CameronDiver:
Also use the supervisor's own container logging monitoring code when
running livepush on the supervisor container.

Change-type: minor

### DIFF
--- a/src/lib/supervisor-console.ts
+++ b/src/lib/supervisor-console.ts
@@ -1,0 +1,90 @@
+import * as _ from 'lodash';
+import { TransformableInfo } from 'logform';
+import * as winston from 'winston';
+
+const levels = {
+	error: 0,
+	warn: 1,
+	success: 2,
+	event: 3,
+	api: 4,
+	info: 5,
+	debug: 6,
+};
+
+type logLevel = keyof typeof levels;
+
+const colors: { [key in logLevel]: string | string[] } = {
+	error: 'red',
+	warn: 'yellow',
+	success: 'green',
+	event: 'cyan',
+	info: 'blue',
+	debug: 'magenta',
+	api: ['black', 'bgWhite'],
+};
+
+const maxLevelLength = _(levels)
+	.map((_v, k) => k.length)
+	.max();
+
+const uncolorize = winston.format.uncolorize();
+
+const formatter = winston.format.printf(args => {
+	const { level, message } = args;
+	const { level: strippedLevel } = uncolorize.transform(args, {
+		level: true,
+		message: true,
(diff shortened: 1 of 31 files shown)


src/lib/supervisor-console.ts
```diff
@@ -0,0 +1,102 @@
...
+	// is what happens internally in winston)
+}) as unknown) as { [key in logLevel]: (message: string) => void };
+
+winston.addColors(colors);
+
+const messageFormatter = (printFn: (message: string) => void) => {
+	return (...parts: any[]) => {
+		let message = '';
+		let start = '';
+		for (const part of parts) {
```
Page-: I think the gist of this function is
```ts
parts
	.map((p) => {
		if (p instanceof Error) {
			return p.stack
		}
		return p
	})
	.join(' ')
	.replace('\n','\n  ')
	.split('\n')
	.forEach((p) => {
		printFn(p)
	})
```
right?  Does `printFn` not handle multi-line messages very well?

src/lib/supervisor-console.ts
```diff
@@ -0,0 +1,102 @@
...
+			switch (typeof part) {
+				case 'string':
+					message += `${start}${part}`;
```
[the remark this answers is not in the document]
hedss: Or if you want to be explicit, just:
```
case 'string':
default:
    ....
```

src/lib/supervisor-console.ts
CameronDiver: It breaks the formatting a little bit, but it's not the end of the world. This way just makes sure it's consistent. Will change to more functional, the idea was that I'd add more specific handlers for data types, but I didn't get there yet (For example a consistent way of showing docker output etc).

src/lib/supervisor-console.ts
```diff
@@ -0,0 +1,102 @@
+import * as _ from 'lodash';
+import stripAnsi from 'strip-ansi';
+import * as winston from 'winston';
+
+const levels = {
```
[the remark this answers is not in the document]
CameronDiver: Only that winston requires them in this format.

src/lib/supervisor-console.ts
Page-: You can still add more handlers in the
```ts
	.map((p) => {
		if (p instanceof Error) {
			return p.stack
		}
		return p
	})
```
btw

CameronDiver: Yeah I know, I was mainly talking about the `switch`, which i will end up adding back in when/if we get more specific handlers.

PR #999, opened 2019-06-10
Re-parallelise current image retrieval code

CameronDiver:
Change-type: patch

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -219,15 +219,15 @@ export class Images extends (EventEmitter as new () => ImageEventEmitter) {
 	private async withImagesFromDockerAndDB<T>(
 		cb: (dockerImages: NormalisedDockerImage[], composeImages: Image[]) => T,
 	) {
-		const images = await this.docker.listImages({ digests: true });
-		const newImages = await Bluebird.map(images, async image => {
-			const newImage = _.clone(image) as NormalisedDockerImage;
-			newImage.NormalisedRepoTags = await this.getNormalisedTags(image);
-			return newImage;
-		});
-
-		const dbImages = await this.db.models('images').select();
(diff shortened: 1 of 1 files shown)


src/compose/images.ts
```diff
@@ -219,15 +219,18 @@ export class Images extends (EventEmitter as new () => ImageEventEmitter) {
...
-		const images = await this.docker.listImages({ digests: true });
-		const newImages = await Bluebird.map(images, async image => {
-			const newImage = _.clone(image) as NormalisedDockerImage;
-			newImage.NormalisedRepoTags = await this.getNormalisedTags(image);
-			return newImage;
-		});
-
-		const dbImages = await this.db.models('images').select();
-		return cb(newImages, dbImages);
+		return Bluebird.join(
```
Page-: If you want this to be async/await and parallel that can be done via
```ts
const [images, newImages] = await Promise.all([
	this.docker.listImages({ digests: true }),
	Bluebird.map(images, async image => {
		const newImage = _.clone(image) as NormalisedDockerImage;
		newImage.NormalisedRepoTags = await this.getNormalisedTags(image);
		return newImage;
	})
])
```
or the equivalent with assigning to vars and then awaiting later

src/compose/images.ts
```diff
@@ -219,15 +219,18 @@ export class Images extends (EventEmitter as new () => ImageEventEmitter) {
...
+			Bluebird.map(
+				await this.docker.listImages({ digests: true }),
```
pcarranzav: Is the `await` needed here?

CameronDiver: Ah it's not actually, I forgot you could pass a promise which resolves to an array as the first argument.

PR #1005, opened 2019-06-18
Make sync-debug much more reliable with logs

CameronDiver:
Change-type: patch

### DIFF
--- a/sync-debug.js
+++ b/sync-debug.js
@@ -20,58 +20,55 @@ const dockerode = require('dockerode');
 const chokidar = require('chokidar');
 const _ = require('lodash');
 
-require('ts-node/register');
-const { ContainerLogs } = require('./src/logging/container');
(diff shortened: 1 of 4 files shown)


sync-debug.js
```diff
@@ -115,7 +113,10 @@ const performLivepush = _.debounce(async (livepush, containerId, docker) => {
 		console.log('SYNC: executing:', command);
 	});
 	livepush.on('commandOutput', ({ output }) => {
-		console.log(`\t${output.data.toString()}`);
+		const message = output.toString().trim();
```
zvin: There's no `toString` method on `CommandOutput`

zvin: Actually there is but it will output `'[object Object]'`

CameronDiver: Fixed.

PR #1006, opened 2019-06-19
Allow writing test files in typescript

CameronDiver:
Change-type: patch


CameronDiver:
I don't really see the advantage of a build system really, short of keeping package.json clean. There's just several different ways of building and running the supervisor based on environment.

Issue #959, opened 2019-04-05
The supervisor should update it's ip address on the dashboard in local mode


CameronDiver:
Because this is useful for doing a push to device, the supervisor should update the ip address on the dashboard even in local mode, where currently the ip address stays at the value that it was when local mode was turned on.

Requested here: https://forums.balena.io/t/the-dashboard-doesnt-update-device-ips/5842/6

roman-mazur:
@CameronDiver I'm a bit confused here.

IP address reporting is initiated with `startIPAddressUpdate`, and it does get called in local mode for me. I will test the behaviour with a real IP address change, but it may be not a supervisor issue.

roman-mazur:
Ok, I found it. The IP address change is detected by the supervisor, but the API call to update the state is not made in local mode.

PR #1007, opened 2019-06-19
Report device state in local mode

roman-mazur:
In local mode, we now update device status on the backend,
but omit applications info in our updates.

Closes: #959
Change-type: minor

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -38,7 +38,7 @@ const INTERNAL_STATE_KEYS = [
 	'update_failed',
 ];
 
-interface APIBinderConstructOpts {
+export interface APIBinderConstructOpts {
 	config: Config;
 	// FIXME: Remove this
 	db: Database;
@@ -450,12 +450,22 @@ export class APIBinder {
 
 	private async sendReportPatch(
 		stateDiff: DeviceApplicationState,
-		conf: { apiEndpoint: string; uuid: string },
+		conf: { apiEndpoint: string; uuid: string; localMode: boolean },
 	) {
 		if (this.cachedBalenaApi == null) {
 			throw new InternalInconsistencyError(
 				'Attempt to send report patch without an API client',
 			);
+		}
+
+		let body = stateDiff;
+		if (conf.localMode) {
+			body = this.stripDeviceStateInLocalMode(stateDiff);
+			// In local mode, check if it still makes sense to send any updates after data strip.
+			if (_.isEmpty(body.local)) {
+				// Nothing to send.
+				return;
+			}
 		}
 
 		const endpoint = url.resolve(
@@ -467,12 +477,24 @@ export class APIBinder {
 			{
 				method: 'PATCH',
 				url: endpoint,
-				body: stateDiff,
+				body,
 			},
 			this.cachedBalenaApi.passthrough,
 		);
 
 		await this.cachedBalenaApi._request(requestParams);
+	}
+
+	// Returns an object that contains only status fields relevant for the local mode.
+	// It basically removes information about applications state.
+	public stripDeviceStateInLocalMode(
+		state: DeviceApplicationState,
+	): DeviceApplicationState {
+		return {
+			local: _.cloneDeep(
+				_.omit(state.local, 'apps', 'is_on__commit', 'logs_channel'),
+			),
+		};
 	}
 
 	private async report() {
@@ -484,17 +506,12 @@ export class APIBinder {
 			'localMode',
 		]);
 
-		if (conf.localMode) {
-			return;
-		}
-
 		const stateDiff = this.getStateDiff();
 		if (_.size(stateDiff) === 0) {
 			return 0;
 		}
 
-		const apiEndpoint = conf.apiEndpoint;
-		const uuid = conf.uuid;
+		const { apiEndpoint, uuid, localMode } = conf;
 		if (uuid == null || apiEndpoint == null) {
 			throw new InternalInconsistencyError(
 				'No uuid or apiEndpoint provided to ApiBinder.report',
@@ -503,7 +520,7 @@ export class APIBinder {
 
 		try {
 			await Bluebird.resolve(
-				this.sendReportPatch(stateDiff, { apiEndpoint, uuid }),
+				this.sendReportPatch(stateDiff, { apiEndpoint, uuid, localMode }),
 			).timeout(conf.apiTimeout);
 
 			this.stateReportErrors = 0;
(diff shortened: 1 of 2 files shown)


src/api-binder.ts
```diff
@@ -450,14 +450,24 @@ export class APIBinder {
...
 	) {
 		if (this.cachedBalenaApi == null) {
 			throw new InternalInconsistencyError(
 				'Attempt to send report patch without an API client',
 			);
 		}
 
+		let body = stateDiff;
+		if (conf.localMode) {
+			body = this.stripDeviceStateInLocalMode(stateDiff);
```
roman-mazur: I was also exploring a possibility to do this `strip` on a collected state rather than on the diff, so that we can make diff on potentially smaller data object. But since there is a lot of logic that both `dependent` and `local` fields have to be present on state objects, everything becomes slightly more complicated. So, I ended up with this approach.

test/api-binder.ts
```diff
@@ -0,0 +1,44 @@
...
+				supervisor_version: '9.16.3',
+				provisioning_progress: null,
+				provisioning_state: '',
+				status: 'Idle',
+				logs_channel: null,
+				apps: {},
+				is_on__commit: 'whatever',
+			},
+			dependent: { apps: {} },
+		} as DeviceApplicationState;
```
roman-mazur: @CameronDiver @Page- This is an example of state object collected on my machine.
I'm a bit worried by the fact it does not really match `DeviceApplicationState` interface. Although it all works, `DeviceApplicationState` has `config`, which, to my understanding, is data dictated by the backend. So, I'm not sure it's the best idea to use this interface for data originating from both backend and device.

test/api-binder.ts
```diff
@@ -0,0 +1,44 @@
...
+		it('should strip applications data', () => {
+			const result = apiBinder.stripDeviceStateInLocalMode(
+				sampleState,
+			) as Dictionary<any>;
+			expect(result).to.not.have.property('dependent');
+
+			const local = result['local'];
+			expect(local).to.not.have.property('apps');
+			expect(local).to.not.have.property('is_on__commit');
+			expect(local).to.not.have.property('logs_channel');
```
roman-mazur: @CameronDiver @Page- 
Please take a look at this list of stripped fields.

`is_on__commit` is filled with `"localrelease"` in local mode, but I decided to remove it from the patch at all.

`logs_channel` - not sure how it's used, but felt like it's not relevant in local mode :)
That said, it's not even present on the list used by the backend:
https://github.com/balena-io/open-balena-api/blob/master/src/routes/devices.ts#L750

roman-mazur:
Rebased on master.

src/api-binder.ts
```diff
@@ -450,14 +450,24 @@ export class APIBinder {
...
+			// In local mode, check if it still makes sense to send any updates after data strip.
+			if (_.size(body.local) === 0) {
```
Page-:
```suggestion
			if (_.isEmpty(body.local)) {
```

src/api-binder.ts
```diff
@@ -467,14 +477,27 @@ export class APIBinder {
...
 
 		await this.cachedBalenaApi._request(requestParams);
 	}
 
+	// Returns an object that contains only status fields relevant for the local mode.
+	// It basically removes information about applications state.
+	public stripDeviceStateInLocalMode(
+		state: DeviceApplicationState,
+	): DeviceApplicationState {
+		const result = _.cloneDeep(_.pick(state, 'local'));
```
Page-: I think this might well be better as
```suggestion
		const result = {
			local: _.cloneDeep(_.omit(state.local, 'apps', 'is_on__commit', 'logs_channel'));
		}
```
because it keeps the same behaviour without needing to do the `cloneDeep` on properties that are just going to be removed immediately after

test/api-binder.ts
```diff
@@ -0,0 +1,44 @@
+import { expect } from 'chai';
+import { APIBinder, APIBinderConstructOpts } from '../src/api-binder';
+import { DeviceApplicationState } from '../src/types/state';
+
+describe.only('APIBinder', () => {
+	let apiBinder: APIBinder;
+
+	before(() => {
+		apiBinder = new APIBinder({} as APIBinderConstructOpts);
```
Page-: Does this cast actually work? I would expect it to fail if this is actually being type checked

src/api-binder.ts
```diff
@@ -500,9 +519,10 @@ export class APIBinder {
 				'No uuid or apiEndpoint provided to ApiBinder.report',
 			);
 		}
+		const localMode = conf.localMode;
```
Page-: This can be
```suggestion
		const { localMode } = conf;
```

test/api-binder.ts
roman-mazur: It does, the test passes.

roman-mazur:
@balena-ci retest

roman-mazur:
@Page- thanks for the suggestions, I believe I addressed all of them

test/api-binder.ts
roman-mazur: @CameronDiver 

>is_on__commit and logs_channel can be removed

So this implementation and test ensure that these fields are not present in device status payload. Could you please clarify if you are confirming it's correct or asking to changing this?

test/api-binder.ts
roman-mazur: Any thoughts here?

CameronDiver: This is fine, the `config` field will only be present when the supervisor wants to let the api that that state has changed.

test/api-binder.ts
CameronDiver: I can confirm that both of these values (`is_on__commit` and `logs_channel`) can and should be omitted when reporting state whilst in local mode.

test/api-binder.ts
roman-mazur: @CameronDiver sorry, I'm not deeply experienced with TypeScript details :), would you elaborate how this can be written in a different way?

Also, I have to disagree with your :)
>This shouldn't work

I does work: the code does get compiled, as you can see from build results.

If understand this correctly, TypeScript allows casts to interfaces as long as the dictionary is a subset of interface properties. 

For instance, from a [different commit](https://github.com/balena-io/balena-supervisor/commit/75863ccf79cf45fad95157eef18b037b12c4cdbf):
```
{Id: 'image-1'} as Docker.ImageInfo
```
This dictionary keys form a subset of `ImageInfo` interface properties, and the cast works.

In this case, we have an empty set, which is a subset of any set ;)

CameronDiver: Ugh, this wan't something I was aware of: https://github.com/Microsoft/TypeScript/issues/13788

Pretty awful. Anyway fair enough, it's just tests.

roman-mazur:
@balena-ci retest

roman-mazur:
@Page- @CameronDiver rebased on master, PTAL

PR #1008, opened 2019-06-19
Don't drop failed api patch data

CameronDiver:
Change-type: patch

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -501,24 +501,27 @@ export class APIBinder {
 			);
 		}
 
-		await Bluebird.resolve(
-			this.sendReportPatch(stateDiff, { apiEndpoint, uuid }),
-		)
-			.catch(StatusError, (e: StatusError) => {
+		try {
+			await Bluebird.resolve(
+				this.sendReportPatch(stateDiff, { apiEndpoint, uuid }),
+			).timeout(conf.apiTimeout);
+
+			this.stateReportErrors = 0;
+			_.assign(this.lastReportedState.local, stateDiff.local);
+			_.assign(this.lastReportedState.dependent, stateDiff.dependent);
+		} catch (e) {
+			if (e instanceof StatusError) {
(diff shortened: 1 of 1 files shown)


src/api-binder.ts
```diff
@@ -501,24 +501,27 @@ export class APIBinder {
...
+					e,
 				);
-			})
-			.timeout(conf.apiTimeout);
-
-		this.stateReportErrors = 0;
-		_.assign(this.lastReportedState.local, stateDiff.local);
-		_.assign(this.lastReportedState.dependent, stateDiff.dependent);
+			} else {
+				throw e;
```
roman-mazur: To be honest, I don't understand what we are trying to achieve here.

CameronDiver: These lines are the problem: https://github.com/balena-io/balena-supervisor/pull/1008/files#diff-721adcdf9dbd4a5b5abcf5055bc115d5L519

If something bad happens whilst trying to report the state to the api, we'll silently drop all of the data that we were going to send. With this change, the data that we were going to send will be sent on the next successful try.

roman-mazur: Ok, I see now. Because we don't assign data to the last reported state in case of an error, the diff will contain unreported data next time.

But I'm concerned about this specific throw. If it's not a status code error, most likely it's connection problem. Why do we want to throw in such case?

CameronDiver: Because this is handled further up, and will apply a backoff on the connection attempts, so we don't needlessly try to report the state. 

This may not be the best thing to do, but I'm preserving the behavior of the current code.

roman-mazur: Got it, thanks.

Another question.
We use `stateReportErrors` in our healthceck. I find this a bit weird: it is not really connected to supervisor health. At the same time, it's a bit weird we don't apply back-off logic in case of backend errors.

Shouldn't we remove stateReportErrors from the healtch check and use back-off logic for both backend and connection problems (so that we actually do best effort to report the state).

Issue #993, opened 2019-05-29
On leaving local mode the supervisor should clean up dangling images


CameronDiver:
This would be equivalent to doing
```
balena rmi $(balena images -f "dangling=true" -q)
```

Given that these images are dangling, it shouldn't interact with any normal usage of the device.

roman-mazur:
@CameronDiver Does it make sense to also take care of unused networks and volumes?

We could use `/prune` APIs of the engine (if our version is based on something later than 1.25).
https://docs.docker.com/engine/api/v1.25/#operation/ImagePrune

Similar APIs exist for volumes and networks, and, to my understanding, are used by `docker system prune` command:
https://docs.docker.com/engine/reference/commandline/system_prune/

CameronDiver:
We would need to test that they remove only what we want them to remove, but in general, yeah we could do that. I'm pretty sure that our balena-engine is based on > 1.25.

roman-mazur:
Checked on BoB: API version is 1.35

roman-mazur:
Reproduced this with [hello rust project](https://github.com/balena-io-projects/balena-rust-hello-world) (which has a multistage build).

In case of this project, `balena push <ip>` created `local_image_main` image (for the main service to run) and 2 dangling images (which should map to `builder` and `dependencies` images from the multistage build described in the project Dockerfile).

```
REPOSITORY                          TAG                 IMAGE ID            CREATED             SIZE
<none>                              <none>              6b104f6621d6        11 minutes ago      1.26GB
local_image_main                    latest              dd97d909a94f        11 minutes ago      145MB
<none>                              <none>              f16281f6026c        11 minutes ago      1.25GB
```

The dangling images don't have `io.resin.local.image` label, so will not be removed when local mode is disabled.

CameronDiver:
This also holds true for images that are not built on device and instead are provided using the docker-compose `image:` directive I believe.

roman-mazur:
Yes, confirmed with the following docker-compose file:
```
version: '2'
services:
  main:
    build: .
    links:
      - "redis:redis-db"
    volumes:
      - test-volume:/var/lib/test-data
  redis:
    image: "redis:alpine"
  main2:
    build: .
    networks: 
      - test-network
networks:
  test-network:
volumes:
  test-volume:
```

Networks and volumes seem to be deleted normally as a result of services deletion when local mode is disabled.

[...]

PR #1012, opened 2019-06-21
Clean up local mode engine objects using snapshots

roman-mazur:
Snapshot is collected to compare with engine state when local mode is turned off.
All the objects (containers, images, volumes, and networks) that didn't exist before local mode was enabled will be removed once local mode is deactivated.

Closes: #993 

Change-type: minor


roman-mazur:
I'm testing this on a device and researching the best way to lock remote state application while this cleanup happens.

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
+	// Query the engine to get currently running containers and installed images.
+	public async collectEngineSnapshot(): Promise<EngineSnapshotRecord> {
+		const containers = this.docker
+			.listContainers()
+			.then(resp => _.map(resp, 'Id'));
+		const images = this.docker.listImages().then(resp => _.map(resp, 'Id'));
+		const volumes = this.docker
+			.listVolumes()
+			.then(resp => _.map(resp.Volumes, 'Name'));
+		const networks = this.docker.listNetworks().then(resp => _.map(resp, 'Id'));
```
CameronDiver: Can you change these names slightly to make it more clear that their intended to be promises, I initially thought you'd dropped the `await` accidentally.

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
+				),
+		);
+	}
+
+	// Store engine snapshot data in the local database.
+	public async storeEngineSnapshot(
+		record: EngineSnapshotRecord,
+	): Promise<void> {
+		const timestamp = record.timestamp.toISOString();
+		log.info(
```
CameronDiver: This should probably be `debug`. It's currently not that well-defined though...

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
+		);
+		return this.db
+			.models('engineSnapshot')
+			.delete()
+			.then(() => {
+				return this.db.models('engineSnapshot').insert({
+					snapshot: JSON.stringify(record.snapshot),
+					timestamp,
+				});
+			});
```
CameronDiver: Please use `await` here (or drop the `async` qualifier in the function definition, but I'd prefer you use `await`).

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
+			.orderBy('rowid', 'DESC')
+			.first()
+			.then(r => {
+				if (!r) {
+					return null;
+				}
+				return new EngineSnapshotRecord(
+					EngineSnapshot.fromJSON(r.snapshot),
+					new Date(Date.parse(r.timestamp)),
+				);
```
CameronDiver: `await`.

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
 			});
-			await Bluebird.map(images, imageId => {
-				log.debug('Removing local mode image: ', imageId);
-				return this.docker.getImage(imageId).remove({ force: true });
+	}
+
+	private async removeLocalModeArtifacts(
+		objects: EngineSnapshot,
+	): Promise<void> {
+		log.debug(`Going to delete following objects: ${objects}`);
```
CameronDiver: This is probably going to need to be formatted a bit nicer, can you paste what it currently prints please?

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
+				// Remove any local mode state added to the database.
+				return this.db
+					.models('app')
+					.del()
+					.where({ source: 'local' })
+					.catch(e =>
+						log.error('Cannot delete local app entries in the database', e),
+					);
 			});
+	}
```
CameronDiver: Please use `async/await`.

src/local-mode.ts
```diff
@@ -56,42 +90,172 @@ export class LocalModeManager {
...
+						currentRecord.snapshot.diff(previousRecord.snapshot),
+					);
+				});
+			})
+			.catch(e => {
+				log.error(
+					`Problems managing engine state on local mode switch. Local mode: ${local}.`,
+					e,
+				);
+			});
```
CameronDiver: Please use `async/await`.

test/local-mode.ts
```diff
@@ -0,0 +1,160 @@
...
+import * as tmp from 'tmp-promise';
+import Config from '../src/config';
+import DB from '../src/db';
+import LocalModeManager, {
+	EngineSnapshot,
+	EngineSnapshotRecord,
+} from '../src/local-mode';
+import Logger from '../src/logger';
+
+describe.only('LocalModeManager', () => {
```
CameronDiver: Don't forget to remove this `.only.`

src/local-mode.ts
roman-mazur: I used debug for dumping actual object IDs (because it's more verbose). But I'm fine with debug here too.
Anyway, we don't eliminate debug messages from the actual container output.

src/local-mode.ts
roman-mazur:
```
Going to delete the following objects: containers [container-1,container-2], images [image-1,image-2], volumes [volume-1,volume-2], networks [network-1,network-2]
```

EngineSnapshot has toString method. But for some reason it's not committed in this draft... I will push soon.

src/local-mode.ts
roman-mazur: So, in this method, I kept `.catch()` constructions. Here's why.

I wanted to catch every individual object deletion failure, so that we do attempt all objects deletion instead of stopping on the first failure.
Adding `try/catch` in every close there would be too ugly to my taste.

The last db operation catch could be done with `try/catch`, but I kept the promise-like approach for consistency with the rest of the code in this method.

src/local-mode.ts
roman-mazur: See https://github.com/balena-io/balena-supervisor/pull/1012/files#diff-0f5fedf56e0f6f78a6c52a828857e5e4R40

src/local-mode.ts
```diff
@@ -56,51 +99,133 @@ export class LocalModeManager {
...
+			.then(resp => _.map(resp, 'Id'));
+		const imagesPromise = this.docker
+			.listImages()
+			.then(resp => _.map(resp, 'Id'));
+		const volumesPromise = this.docker
+			.listVolumes()
+			.then(resp => _.map(resp.Volumes, 'Name'));
+		const networksPromise = this.docker
+			.listNetworks()
+			.then(resp => _.map(resp, 'Id'));
```
roman-mazur: Ah, and the only use of `then`. :)

My goal was to make sure that there is no limitation in how queries to objects are made (we don't need to make them serially), hence I moved the `await` call down to the `.all()`.

roman-mazur:
@balena-ci retest

roman-mazur:
@balena-ci retest

rebased on master

roman-mazur:
@CameronDiver the last commit ensure the correct order of local mode cleanup and DeviceState/ApplicationsManager actions.

src/local-mode.ts
```diff
@@ -56,51 +102,150 @@ export class LocalModeManager {
 
 		if (!localMode) {
 			// Remove any leftovers if necessary
-			await this.removeLocalModeArtifacts();
+			await this.handleLocalModeStateChange(false);
```
roman-mazur: Found that this piece does not play very nice atm.

@CameronDiver would you give me more information on the use cases it handles?
Meanwhile, I'm going to make sure we update the snapshot after each new target state application, so that supervisor restart works correctly.

src/local-mode.ts
```diff
@@ -56,51 +102,150 @@ export class LocalModeManager {
...
+			);
+	}
+
+	// Handle local mode state change.
+	// Input parameter is a target (new) state.
+	public async handleLocalModeStateChange(local: boolean) {
+		let switchResolve: (() => void) | null = null;
+		this.switchInProgress = new Promise<void>(resolve => {
+			switchResolve = resolve;
+		});
```
CameronDiver: This is a weird pattern. Why not just wrap it all inside a new promise?

src/local-mode.ts
```diff
@@ -56,51 +102,150 @@ export class LocalModeManager {
...
+
+		// Remove any local mode state added to the database.
+		await this.db
+			.models('app')
+			.del()
+			.where({ source: 'local' })
+			.catch(e =>
+				log.error('Cannot delete local app entries in the database', e),
+			);
```
CameronDiver: Is this new behavior?

src/local-mode.ts
CameronDiver: I'd still prefer something like
```
await Bluebird.all([
  _.map(await this.docker.listContainers(), 'Id'),
...
```

but good idea on running them in parallel.

src/local-mode.ts
```diff
@@ -56,51 +102,150 @@ export class LocalModeManager {
...
+		});
+		try {
+			const currentRecord = await this.collectEngineSnapshot();
+			if (local) {
+				return await this.storeEngineSnapshot(currentRecord);
+			}
+
+			const previousRecord = await this.retrieveLatestSnapshot();
+			if (!previousRecord) {
+				log.warn('Previous engine snapshot was not stored. Skipping clanup.');
```
CameronDiver:
```suggestion
				log.warn('Previous engine snapshot was not stored. Skipping cleanup.');
```

src/local-mode.ts
```diff
@@ -56,51 +102,150 @@ export class LocalModeManager {
...
+			return null;
+		}
+		return new EngineSnapshotRecord(
+			EngineSnapshot.fromJSON(r.snapshot),
+			new Date(Date.parse(r.timestamp)),
+		);
+	}
+
+	private async removeLocalModeArtifacts(objects: EngineSnapshot) {
+		log.debug(`Going to delete the following objects: ${objects}`);
```
CameronDiver: I do worry about how this will look with something like bob for example, that has a bunch of services and volumes.

src/local-mode.ts
roman-mazur: No, it existed before - just carrying over.
https://github.com/balena-io/balena-supervisor/blob/master/src/local-mode.ts#L77

src/local-mode.ts
roman-mazur: Are you worried about logs pollution?

Can replace this with counters in logs. Specific object deletion behaviour is covered with tests, so counters here should be fine.

CameronDiver: By counters do you mean `removing n images, n networks` etc?

src/local-mode.ts
CameronDiver: Fair enough, no worries then, I had forgotten about that.

src/local-mode.ts
CameronDiver: And also unwrapping them specifically from the promise:
```typescript
const [ containers, images, volumes, networks ] = await Bluebird.all([
```

src/local-mode.ts
roman-mazur: Yep, exactly.

src/local-mode.ts
roman-mazur: Yeah, I agree. Let me fix it.

src/local-mode.ts
CameronDiver: SGTM

src/local-mode.ts
roman-mazur: Addressed in the new commit - we remove the stored snapshot after successful cleanup.

src/local-mode.ts
roman-mazur: Done.

src/local-mode.ts
roman-mazur: But in construction like
```
await Bluebird.all([
  _.map(await this.docker.listContainers(), 'Id'),
  _.map(await this.docker.listImages(), 'Id'),
...
```
won't it pause until containers are collected, and only then make a request to get images?

Unwrapping is done - thanks for the hint.

roman-mazur:
@balena-ci retest

src/local-mode.ts
```diff
@@ -1,12 +1,59 @@
 import * as Bluebird from 'bluebird';
 import * as Docker from 'dockerode';
 import * as _ from 'lodash';
-
```
CameronDiver: I generally leave these newlines as a separator for internal and external dependencies, at a glance. It's not a hard and fast rule, but it does lay out the imports better when there's quite a few.

src/local-mode.ts
```diff
@@ -1,12 +1,59 @@
...
+// A snapshot is taken before entering local mode in order to perform cleanup when we exit local mode.
+export class EngineSnapshot {
+	constructor(
+		public readonly containers: string[],
+		public readonly images: string[],
+		public readonly volumes: string[],
+		public readonly networks: string[],
+	) {}
+
+	public static fromJSON(json: string): EngineSnapshot {
```
CameronDiver: Some cursory validation here would be good I think, fs corruption is something we have to deal with constantly, so any logs detailing that something read from the database is wrong definitely helps when debugging.

src/local-mode.ts
```diff
@@ -1,12 +1,59 @@
...
+			`${this.containers.length} containers, ` +
+			`${this.images.length} images, ` +
+			`${this.volumes.length} volumes, ` +
+			`${this.networks.length} networks`
+		);
+	}
+}
+
+// Record in a database that stores EngineSnapshot.
+export class EngineSnapshotRecord {
```
CameronDiver: I like this :+1: We should start to define all our models in this way for the typescript conversion.

src/local-mode.ts
CameronDiver: Ah yeah you're right, my bad.

src/local-mode.ts
CameronDiver: Fair enough, the code is clear, so I'm happy.

src/local-mode.ts
```diff
@@ -56,51 +102,159 @@ export class LocalModeManager {
...
-			await this.db
-				.models('app')
-				.del()
-				.where({ source: 'local' });
-		} catch (e) {
-			log.error('There was an error clearing local mode artifacts: ', e);
-		}
+	public startLocalModeChangeHandling(local: boolean) {
+		this.switchInProgress = new Promise(async resolve => {
+			try {
```
CameronDiver: This `try...finally` makes me a little uneasy, if the `handleLocalModeStateChange(..)` function ever changes to throw an error (I know currently it can't) we'd be ignoring it and printing nothing. I think given that we know it cannot fail, I'd prefer this to be removed, and in the case where the `handleLocalModeStateChange` function changes to ever throw an error, we add this back in with an explicit `catch`.

I'm happy to hear if you have any other thoughts on it though.

src/local-mode.ts
```diff
@@ -56,51 +102,159 @@ export class LocalModeManager {
...
+	}
+
+	// Store engine snapshot data in the local database.
+	public async storeEngineSnapshot(record: EngineSnapshotRecord) {
+		const timestamp = record.timestamp.toISOString();
+		log.debug(
+			`Storing engine snapshot in the database. Timestamp: ${timestamp}`,
+		);
+		await this.cleanEngineSnapshots();
+		return this.db.models('engineSnapshot').insert({
```
CameronDiver:
```suggestion
		await this.db.models('engineSnapshot').insert({
```

src/local-mode.ts
```diff
@@ -56,51 +102,159 @@ export class LocalModeManager {
...
+				e,
+			);
+		} finally {
+			log.debug('Handling of local mode switch is completed');
+		}
+	}
+
+	// Returns a promise to await local mode switch completion started previously.
+	public async switchCompletion() {
+		if (!this.switchInProgress) {
```
CameronDiver: Generally the pattern we use here is `if (this.switchInProgress == null)`. Two reasons for this, one it catches both `null` and `undefined` (as does your method), but the second is that it's very clear what we're *actually* checking for.

src/local-mode.ts
roman-mazur: It's changed because I auto-organize imports with my editor. Will check if it can allow configuration like this.

src/local-mode.ts
roman-mazur: Good point. Let me check how error here works, and how it's logged...

src/local-mode.ts
roman-mazur: `startLocalModeChangeHandling` is responsible for managing `switchInProgress` variable, and I was trying to make it as independent from the rest of the logic as possible.
Its responsibility is to construct a promise which gets resolved whenever local mode switch handler completes regardless how it completes to avoid any possibility of a deadlock. That is why, I put a `finally` block inhere.

In fact, it can be written "simpler", I think:
```
this.switchInProgress = this.handleLocalModeStateChange(local)
    .finally(() => { this.switchInProgress = null; });
```
But it does not compile with our current setup :) TS compiler says there is no `finally` property in `Promise`.

Does it make sense? It's not really about handling errors, but about separating the task of unblocking other operations whenever local mode switch completes.

roman-mazur: Side note: this method should actually be private.
But I need to use it in tests... Do you know a good way how I could limit the method visibility to tests only?

Should I look into namespaces?

src/local-mode.ts
roman-mazur: Changed to `== null`.

src/local-mode.ts
CameronDiver: I don't think there's a good solution to this, it's either make it public, or do `(instance as any).method` in the tests, losing types.

CameronDiver: I'm pretty sure the method you mention here https://github.com/balena-io/balena-supervisor/pull/1012#discussion_r297566134 will propagate errors so I'd be fine with this. To do that, you'd have to wrap it in a `Bluebird.resolve`.

roman-mazur: Done: https://github.com/balena-io/balena-supervisor/pull/1012/files#diff-0f5fedf56e0f6f78a6c52a828857e5e4R110

roman-mazur: So do we need to use Bluebird because of older nodejs versions?

CameronDiver: Actually, I forgot that `.finally` isn't a bluebird extension, and should be present in the native. I'd check if we need to upgrade our `@types/node`.

src/local-mode.ts
roman-mazur: There is no such option :(
Do you know if we can configure `prettify` to do this?

src/local-mode.ts
roman-mazur: There is no such option :(
Do you know if we can configure prettify to do this?

CameronDiver: I don't believe so. I'd prefer our imports didn't reorganise with every commit in a different file though.

roman-mazur: Yep, completely agree - I removed this change.

src/local-mode.ts
roman-mazur: Ok, resolving this thread in such case.

src/local-mode.ts
roman-mazur: In 5de1410e13fa0b3905000f6ab58201d5536a672b, I added code to ensure we log bad input when parsing errors happen. It basically yields to skipping local mode leave changes - see tests in that commit.

Is it good enough for you?

CameronDiver: That's great, works for me, thanks @roman-mazur

roman-mazur:
@balena-ci retest
rebased on master

roman-mazur:
@balena-ci retest

Issue #1003, opened 2019-06-17
The supervisor should be able to stream journald logs from it's api


CameronDiver:
We should provide an api endpoint to stream journald logs, which would be useful for diagnostics, among other things.

PR #1013, opened 2019-06-21
Add an endpoint and module for reading journald logs

CameronDiver:
Change-type: minor
Closes: #1003

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1149,6 +1149,7 @@ Response:
 	]
 }
 ```
+
 ### V2 Utilities
 
 #### Cleanup volumes with no references
@@ -1179,3 +1180,37 @@ Unsuccessful response:
 }
 ```
 
+#### Journald logs
+
+Added in supervisor version v10.1.0
+
+Retrieve a stream to the journald logs on device. This is
+equivalent to running `journalctl -o export`. Options
+supported are:
+
+##### all: boolean
+Show all fields in full, equivalent to `journalctl --all`.
+
+##### follow: boolean
+Continuously stream logs as they are generated, equivalent
+to `journalctl --follow`.
+
+##### count: integer
+Show the most recent `count` events, equivalent to
+`journalctl --line=<count>`.
+
+##### unit
+Show journal logs from `unit` only, equivalent to
+`journalctl --unit=<unit>`.
+
+Fields should be provided via POST body in JSON format.
+
+From an application container (with systemd installed):
+```
+$ curl -X POST --data '{"follow":true,"all":true}' "$BALENA_SUPERVISOR_ADDRESS/v2/journal-logs?apikey=$BALENA_SUPERVISOR_API_KEY" | systemd-journal-remote - -o log.journal
+```
+
+The `log.journal` file can then be viewed with
+```
+journalctl --file log.journal -f
+```
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -13,8 +13,11 @@ import {
 import { doPurge, doRestart, serviceAction } from './common';
 
 import Volume from '../compose/volume';
+import { spawnJournalctl } from '../lib/journald';
+
 import log from '../lib/supervisor-console';
 import supervisorVersion = require('../lib/supervisor-version');
+import { checkInt, checkTruthy } from '../lib/validation';
 
 export function createV2Api(router: Router, applications: ApplicationManager) {
 	const { _lockingIfNecessary, deviceState } = applications;
@@ -528,6 +531,34 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
 			});
 		} catch (e) {
 			res.status(503).json({
+				status: 'failed',
+				message: messageFromError(e),
+			});
+		}
+	});
+
+	router.post('/v2/journal-logs', (req, res) => {
+		try {
+			const all = checkTruthy(req.body.all) || false;
+			const follow = checkTruthy(req.body.follow) || false;
+			const count = checkInt(req.body.count, { positive: true }) || undefined;
+			const unit = req.body.unit;
+
+			const journald = spawnJournalctl({ all, follow, count, unit });
+			res.status(200);
+			journald.stdout.pipe(res);
+			res.on('close', () => {
+				journald.kill('SIGKILL');
+			});
+			journald.on('exit', () => {
+				res.end();
+			});
+		} catch (e) {
+			log.error('There was an error reading the journalctl process', e);
+			if (res.headersSent) {
+				return;
+			}
+			res.json({
 				status: 'failed',
 				message: messageFromError(e),
 			});
--- a/src/lib/journald.ts
+++ b/src/lib/journald.ts
@@ -1,0 +1,41 @@
+import { ChildProcess, spawn } from 'child_process';
+
+import constants = require('./constants');
+import log from './supervisor-console';
+
+export function spawnJournalctl(opts: {
+	all: boolean;
+	follow: boolean;
+	count?: number;
+	unit?: string;
+}): ChildProcess {
+	const args = [
+		// The directory we want to run the chroot from
+		constants.rootMountPoint,
+		'journalctl',
+		'-o',
+		'export',
+	];
+	if (opts.all) {
+		args.push('-a');
+	}
+	if (opts.follow) {
+		args.push('--follow');
+	}
+	if (opts.unit != null) {
+		args.push('-u');
+		args.push(opts.unit);
+	}
+	if (opts.count != null) {
+		args.push('-n');
+		args.push(opts.count.toString());
+	}
+
+	log.debug('Spawning journald with: chroot ', args.join(' '));
+
+	const journald = spawn('chroot', args, {
+		stdio: 'pipe',
+	});
+
+	return journald;
+}


docs/API.md
```diff
@@ -1152,3 +1152,87 @@ Response:
...
+Added in supervisor version v9.18.0
+
+Get access to the journald logs from the device. The logs
+are parsed to extract certain information, and the original 
+log is included too. The original log format can be found
+[here](https://www.freedesktop.org/software/systemd/man/journalctl.html#%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20json%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20).
+The fields added by the supervisor are:
+```json
+{
+	message: "A string representation of the message",
```
zvin: Keys need to be surrounded by double quotes.
Or just change `json` to `typescript`.

docs/API.md
```diff
@@ -1152,3 +1152,87 @@ Response:
...
+  "_SYSTEMD_SLICE": "system-dropbear.slice",
+  "_SYSTEMD_INVOCATION_ID": "dba13abf52a543c6b66a665f7d37130d",
+  "_STREAM_ID": "a069e1e503f5415380d7adfa84ad1569",
+  "_PID": "18907"
+}
+{
+  "message": "This is a second journald log",
+  "priority": 6,
+  "timestamp": "2019-06-21T10:24:01.215Z",
+	"name": "dropbear@1-fe80::2f68:98bf:10bd:a43:22222-fe80::2021:f1a5:53bc:8f46:47974.service",
```
zvin: wrong indentation

docs/API.md
```diff
@@ -1152,3 +1152,87 @@ Response:
...
+	<snip>
+}
+{
+  "message": "This is a third journald log",
+  "priority": 6,
+  "timestamp": "2019-06-21T10:24:02.391Z",
+	"name": "dropbear@1-fe80::2f68:98bf:10bd:a43:22222-fe80::2021:f1a5:53bc:8f46:47974.service",
```
zvin: wrong indentation

src/lib/journald.ts
```diff
@@ -0,0 +1,138 @@
...
+export function getJournald(opts: {
+	follow: boolean;
+	count?: number;
+	unit?: string;
+}): Journald {
+	const args = [
+		CHROOT_DIRECTORY,
+		JOURNALD_COMMAND,
+		JOURNALD_FORMAT_FLAG,
+		JOURNALD_FORMAT,
```
zvin: NIT:
```typescript
const args = ['/mnt/root', 'journalctl', '--output', 'json']
```
would require less effort to read. The constants are only used in one place.

src/lib/journald.ts
```diff
@@ -0,0 +1,138 @@
...
+	];
+	if (opts.follow) {
+		args.push(JOURNALD_FOLLOW_FLAG);
```
zvin: same NIT:
```typescript
args.push('--follow');
```

src/lib/journald.ts
```diff
@@ -0,0 +1,138 @@
...
+	if (opts.unit != null) {
+		args.push(JOURNALD_SERVICE_NAME_FLAG);
+		args.push(opts.unit);
+	}
+	if (opts.count != null) {
+		args.push(JOURNALD_COUNT_FLAG);
+		args.push(opts.count.toString());
+	}
+
+	log.debug(`Spawning journald with: ${CHROOT_COMMAND}`, args.join(' '));
```
zvin:
```typescript
log.debug('Spawning journald with:', CHROOT_COMMAND, args.join(' '));
```

docs/API.md
CameronDiver: Strange, not sure why that happened.

docs/API.md
CameronDiver: I actually changed that, must've forgot to push the changed version...

src/lib/journald.ts
CameronDiver: Yeah fair point, I had thought I'd need to do something a bit more complex, but as it turns out I didn't.

src/lib/journald.ts
```diff
@@ -0,0 +1,138 @@
...
+		[key: string]: unknown;
+	};
+	stderrLine: string;
+	disconnect: { code: number; signal: string };
+	error: Error;
+}
+
+export type JournaldEventEmitter = StrictEventEmitter<EventEmitter, Events>;
+
+export class Journald extends (EventEmitter as new () => JournaldEventEmitter) {
```
roman-mazur: Should not it also be "Closable", so that when HTTP connection is closed, we can close the pipe on our side?

docs/API.md
```diff
@@ -1152,3 +1152,87 @@ Response:
...
+
+Logs are streamed back in [NDJSON](http://ndjson.org/)
+format, and if follow is specified, this connection is never
+closed by the supervisor.
+
+Supported POST options are as follows:
+```typescript
+{
+	follow: true,
+	unit: string,
```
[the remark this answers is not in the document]
CameronDiver: Normally I'd agree, but that implies that `name` and `unit` will always be the same, unfortunately due to the output format, this isn't the case.

docs/API.md
```diff
@@ -1152,3 +1152,87 @@ Response:
...
+  "_HOSTNAME": "ab3d181",
+  "_TRANSPORT": "stdout",
+  "MESSAGE": "This is a journald log",
+  "_COMM": "cat",
+  "_SYSTEMD_CGROUP": "/system.slice/system-dropbear.slice/dropbear@1-fe80::2f68:98bf:10bd:a43:22222-fe80::2021:f1a5:53bc:8f46:47974.service",
+  "_SYSTEMD_UNIT": "dropbear@1-fe80::2f68:98bf:10bd:a43:22222-fe80::2021:f1a5:53bc:8f46:47974.service",
+  "_SYSTEMD_SLICE": "system-dropbear.slice",
+  "_SYSTEMD_INVOCATION_ID": "dba13abf52a543c6b66a665f7d37130d",
+  "_STREAM_ID": "a069e1e503f5415380d7adfa84ad1569",
+  "_PID": "18907"
```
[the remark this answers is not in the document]
CameronDiver: I'd much rather only store the journald logs and not lose any information than store our best guesses. The fields we add are only helpers, and anybody wanting to do some serious debugging could be quite annoyed that we're dropping important information.

src/device-api/v2.ts
```diff
@@ -470,4 +473,43 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+					status: 'failed',
+					message: `Journald stderr: ${line}`,
+				});
+			});
+			res.on('close', () => {
+				journald.end();
+			});
+			journald.on('disconnect', () => res.end());
+		} catch (e) {
+			if (!statusSent) {
```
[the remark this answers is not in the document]
CameronDiver: Good point.

src/device-api/v2.ts
```diff
@@ -470,4 +473,43 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+				res.status(500).json({
+					status: 'failed',
+					message: messageFromError(e),
```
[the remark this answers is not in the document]
CameronDiver: Normally I'd agree, and our backend services follow this, but the supervisor API is very closed down as it is, and the wide range of tasks it performs means that it should return all the information available.

src/device-api/v2.ts
```diff
@@ -470,4 +473,43 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+				if (!statusSent) {
+					res.status(500);
+				}
+				res.json({
+					status: 'failed',
+					message: `Journald stderr: ${line}`,
+				});
+			});
+			res.on('close', () => {
+				journald.end();
```
roman-mazur: So this closes journald when response stream is closed.

But when `follow` is set to `true`, we should probably also react on `req` events.
https://nodejs.org/api/http.html#http_event_abort

Still relatively new to nodejs, so may be mistaking. :)

docs/API.md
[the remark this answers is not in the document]
roman-mazur: I'm just trying to imagine how we'd use this in the dashboard UI, for instance.
What fields should we rely on, so that we can keep the contract not broken?

Maybe it's worth specifying what a client should usually use if he wants to implement logs display.

src/device-api/v2.ts
CameronDiver: `res` is closed when the client closes the connection too, what other events are you interested in catching?

roman-mazur: I'm good in such case :)

docs/API.md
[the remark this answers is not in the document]
CameronDiver: The only thing we can say for certain is that `message` `priority` and `timestamp` are present. Somebody serious about making a good display for this should probably look at the source of `journalctl` to see what they do to ensure everything is printed correctly.

docs/API.md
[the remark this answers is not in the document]
roman-mazur: Ok, makes sense.

src/lib/journald.ts
```diff
@@ -0,0 +1,134 @@
...
+		'/mnt/root',
+		'journalctl',
+		'-o',
+		'json',
+	];
+	if (opts.all) {
+		args.push('-a');
+	}
+	if (opts.follow) {
+		args.push('--follow');
```
roman-mazur: 👍 
much easier to read

src/device-api/v2.ts
```diff
@@ -509,4 +512,28 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+			const all = checkTruthy(req.body.all) || false;
+			const follow = checkTruthy(req.body.follow) || false;
+			const count = checkInt(req.body.count, { positive: true }) || undefined;
+			const unit = req.body.unit;
+
+			const journald = getJournaldStream({ all, follow, count, unit });
+			res.status(200);
+			journald.pipe(res);
+			res.on('close', () => {
+				journald.destroy();
```
roman-mazur: This is a `destroy()` call on the child process stream. Does the process gets killed too?

src/device-api/v2.ts
```diff
@@ -509,4 +512,28 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+			});
+		} catch (e) {
+			if (res.headersSent) {
+				return;
+			}
+			res.json({
+				status: 'failed',
+				message: messageFromError(e),
```
roman-mazur: Should we also log the error?

src/device-api/v2.ts
CameronDiver: It does in this case, but thinking about it, relying on that shouldn't really be done. Will change the journald module to also return the spawn handle, so it can be closed explicitly.

PR #1015, opened 2019-06-21
Dont setup a logstream until we're provisioned

CameronDiver:
This PR includes a bunch of typescript changes as well, as if we'd done that in the first place this bug wouldn't exist.


src/api-binder.ts
```diff
@@ -79,7 +75,7 @@ export class APIBinder {
 	};
 	private eventTracker: EventTracker;
 
-	private balenaApi: PinejsClientRequest | null = null;
+	public balenaApi: PinejsClientRequest | null = null;
```
roman-mazur: can it be readonly if it's public?
I think ideally, it can be injected (through constructor) from outside...

src/app.ts
```diff
@@ -0,0 +1,21 @@
+// This was originally wrapped in a do block in
+// coffeescript, and it's not clear now why that was the
+// case, so I'm going to maintain that behaviour
```
roman-mazur: I also don't see any reason. Suggest dropping the closure.

src/logger.ts
```diff
@@ -65,6 +69,44 @@ export class Logger {
 
 		this.backend.unmanaged = unmanaged;
 		this.backend.publishEnabled = enableLogs;
+
+		// Only setup a config listener if we have to
+		if (!this.balenaBackend.initialised) {
```
roman-mazur: I don't understand this logic fully yet, but should this flag be swapped in the result of the actions below?
I don't see any code like that.

roman-mazur: Ah, found it in `assignFields`...

src/app.ts
CameronDiver: agreed, after a conversation with @Page-

src/api-binder.ts
CameronDiver: I don't believe that this can be marked `readonly`, as we don't initialize it in the constructor.

src/supervisor.ts
```diff
@@ -0,0 +1,118 @@
...
+	'unmanaged',
+	'deviceApiKey',
+	'mixpanelToken',
+	'mixpanelHost',
+	'loggingEnabled',
+	'localMode',
+	'legacyAppsPresent',
+];
+
+export class Supervisor {
```
pdcastro: Thumbs up for more Coffeescript to Typescript conversion 👍

src/logger.ts
roman-mazur: @CameronDiver Want to give some thoughts on the matter, but I'm fully ok with merging this for now.

I may lack some extra background, but based on my current experience/impressions.

`balenaBackend` seems to be initialized asynchronously.

If we want the users of balena backend to deal with that, my suggestion would to work with the backend as a promise, so that for the user it would look like.

```
const client = await this.balenaBackend();
client.callSomeApi();
```

I don't really like this approach, but the main point is that it's a bit weird that user of `balenaBackend` has to deal with its initialisation. There may be many users - we get a question who then is responsible for full initialization. Ultimately, it results in a situation when every user has to care about this.

A more prefereable approach would be to hide asynchronous initalization in the backend class.
API calls are asynchronous anyway, so if we hide asynchronous initalization behind that API call promise, the user will not feel the difference. Like
```
// user code:
const res = await this.balenaBackend.callSomeApi();

// BalenaBackend class:
public async callSomeApi() {
  await config();
  return await doCallSomeApi();
}
```

To summarize: state has to be encapsulated and hidden in balenaBackend. Users of balenaBackend, should not care about its `initalized` state. Hope it makes sense.

CameronDiver:
@roman-mazur I did think about this, but it would require the `balenaBackend` module getting a handle to the config, rather than the primitives it's currently provided with.

I fully agree that perhaps the caller should not worry about this, but given that the caller *is* the one that has to initialise when it becomes possible, I'm ok with it.

I'm not sure I fully understand your promise based interface, could you explain it a little more please? If we're always waiting on the backend being available, whole threads of runtime are going to pause whilst we perform provisioning etc, or have I missed your point?

PR #1017, opened 2019-06-24
Ensure that every architecture uses npm ci

CameronDiver:
We do this by using the standalone installer pinned at v6.9.0. We use the standalone installer because npm itself fails to upgrade on the i386-nlp build (npm v3). The standalone installer is installed in /tmp, and then to avoid mysterious CI errors, we use the original npm to run the tests.

Change-type: patch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 ARG ARCH=amd64
+ARG NPM_VERSION=6.9.0
 
 # The node version here should match the version of the runtime image which is
 # specified in the base-image subdirectory in the project
@@ -33,6 +34,8 @@ FROM balenalib/amd64-node:6-build as i386-nlp-node-build
 
 FROM $ARCH-node-build as node-build
 
+ARG NPM_VERSION
+
 WORKDIR /usr/src/app
 
 RUN apt-get update \
@@ -43,12 +46,20 @@ RUN apt-get update \
 		make \
 		python \
 		rsync \
-		wget \
+		curl \
 	&& rm -rf /var/lib/apt/lists/
 
 COPY package.json package-lock.json /usr/src/app/
 
-RUN JOBS=MAX npm ci --no-optional --unsafe-perm || JOBS=MAX npm install --no-optional --unsafe-perm
+# We first ensure that every architecture has an npm version
+# which can do an npm ci, then we perform the ci using this
+# temporary version
+RUN curl -LOJ https://www.npmjs.com/install.sh && \
+	# This is required to avoid a bug in uid-number
+	# https://github.com/npm/uid-number/issues/7
+	npm config set unsafe-perm true && \
+	npm_install="#{NPM_VERSION}" npm_config_prefix=/tmp sh ./install.sh && \
+	JOBS=MAX /tmp/bin/npm ci --no-optional --unsafe-perm
 
 COPY webpack.config.js fix-jsonstream.js hardcode-migrations.js tsconfig.json tsconfig.release.json /usr/src/app/
 COPY src /usr/src/app/src
@@ -63,6 +74,7 @@ RUN npm test \
 # Build nodejs dependencies
 FROM $ARCH-node-base as node-deps
 ARG ARCH
+ARG NPM_VERSION
 
 RUN [ "cross-build-start" ]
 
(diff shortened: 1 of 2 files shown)


tools/update-npm.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "Ensuring npm version is >= ${MINIMUM_VERSION}"
+
+# Sort the versions using `sort`s version sort, and take the
+# first line
+NEWEST_VERSION="$(echo -e "${CURRENT_VERSION}\n${MINIMUM_VERSION}" | sort -V -r | head -n1)"
+
+# If the head is the same as the current, we are already >=
+# the minimum version
+if [ "${CURRENT_VERSION}" = "${NEWEST_VERSION}" ]; then
+	echo 'Npm already supports npm ci, not upgrading';
```
roman-mazur: I wonder if we can just run `npm ci --help` and see if it succeeds, so we don't need to care about versions management, but just if feature is present.

imrehg: That's a very interesting thought, feature-driven testing is more robust indeed.

tools/update-npm.sh
```diff
@@ -0,0 +1,23 @@
...
+else
+	echo 'Upgrading npm version to support npm ci';
+	npm install -g npm@${MINIMUM_VERSION}
```
roman-mazur: So, to my understansing, we can have different base images with different node/npm versions.

Why not updating base images to make sure they do have a minimal npm version we need.
So, in i386-nlp image, we know that npm version is low, so we add a command there to upgrade it.

tools/update-npm.sh
wrboyce: @roman-mazur great idea

```bash-session
# nvm use 6; npm ci --help &>/dev/null || echo "need to update npm"
Now using node v6.17.1 (npm v3.10.10)
need to update npm
# nvm use 10; npm ci --help &>/dev/null || echo "need to update npm"
Now using node v10.16.0 (npm v6.9.0)
```

roman-mazur: Yeah, but I also wonder if we should update specific base images instead.

CameronDiver: Brilliant idea. As for the updating base images, I've always planned to create specific ones, but never gotten round to it. As it is, I'd like to have this working first, and move onto that.

tools/update-npm.sh
CameronDiver: https://github.com/balena-io/balena-supervisor/pull/1017#discussion_r296667485

tools/update-npm.sh
roman-mazur: Fair enough, let's proceed in this way.

Dockerfile
```diff
@@ -43,12 +43,20 @@ RUN apt-get update \
...
 
-RUN JOBS=MAX npm ci --no-optional --unsafe-perm || JOBS=MAX npm install --no-optional --unsafe-perm
+# We first ensure that every architecture has an npm version
+# which can do an npm ci, then we perform the ci using this
+# temporary version
+RUN curl -LOJ https://www.npmjs.com/install.sh && \
+	# This is required to avoid a bug in uid-number
+	# https://github.com/npm/uid-number/issues/7
+	npm config set unsafe-perm true && \
+	npm_install="6.9.0" npm_config_prefix=/tmp sh ./install.sh && \
```
roman-mazur: Can we somehow share variables in multistage build?
I would extract the version to such a variable if yes, so that there is a single place in the file where it has to be modified if we need to bump it.

CameronDiver: Done.

Dockerfile
```diff
@@ -17,7 +18,7 @@ FROM balenalib/i386-node:10-run as i386-node-base
 RUN echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-start && chmod +x /usr/bin/cross-build-start \
 	&& echo '#!/bin/sh\nexit 0' > /usr/bin/cross-build-end && chmod +x /usr/bin/cross-build-end
 
-FROM balenalib/i386-nlp-node:6-run as i386-nlp-node-base
+FROM balenalib/i386-nlp-node:6-jessie as i386-nlp-node-base
```
roman-mazur: So it is `jessie-run`, is it?

PR #1021, opened 2019-06-26
Apply targetState immediately after update request

### LINKED ISSUE #980: Device doesn't instantly update when POST request  with { “force”: true } on /v1/update is sent.
When a device with updatelocks applied, sends a with POST request to /v1/update with:
```
{
“force”: true
}
```
It takes some minutes to apply, but is should be near instant.

mwohlert:
Cancel delayed promise if exists and schedule a new one without delay, when /v1/update is called

Change-type: patch

Fixes #980


CameronDiver:
I've asked @pcarranzav who originally wrote this code to have a cursory glance, just to ensure there's nothing that I'm missing, but I'll merge after he gives the go ahead.

mwohlert:
I was not quite sure about reusing applyInProgress for saving the promise and resetting it to null. I would really like to know if that is good practice here or if I should rename that variable.

mwohlert:
I also saw, that it would not run the coffeescript mocha tests, when I ran npm run test. It failed, because of the config.json not beeing in the right spot, but I didn't know where to put it.
Seems to be the same error in your CI: 
`  LocalModeManager
[debug]   Test db: /tmp/tmp-168tZklQVelZeBM
Unable to locate config.json! Things may fail unexpectedly!
Unable to locate config.json! Things may fail unexpectedly!
Unable to locate config.json! Things may fail unexpectedly!
    engine snapshots collection
      local mode switch
[error]   Problems managing engine state on local mode switch. Local mode: false. Error: Cannot parse snapshot data {"snapshot":"bad json","timestamp":"2019-06-26T22:04:18.560Z"}.Original message: [Unexpected token b in JSON at position 0].
[error]         at LocalModeManager.<anonymous> (/usr/src/app/build/src/local-mode.js:9:19726)
[error]       at step (/usr/src/app/build/src/local-mode.js:9:7039)
[error]       at Object.next (/usr/src/app/build/src/local-mode.js:9:2703)
[error]       at fulfilled (/usr/src/app/build/src/local-mode.js:9:653)
[debug]   Handling of local mode switch is completed
        ✓ skips cleanup in case of data corruption


  1 passing (1s)
`

Are you aware of this? The build process should fail if there is an error with the tests imho.

roman-mazur:
Hey @mwohlert 
Speaking about that specific `LocalModeManager` test, it exercises supervisor behavior in some unexpected conditions. The error is logged, but the test expects the error and verifies it was handled.

Message
```
Unable to locate config.json! Things may fail unexpectedly!
```
must be generated by one of our db migration scripts and definitely needs a review, but it also should not affect the test execution.

The problem why CS tests are not run, because I missed to remove `.only` on that particular LocalMode manager test before merging to master.
Will fix in a minute.

roman-mazur:
See #1022

CameronDiver:
> I was not quite sure about reusing applyInProgress for saving the promise and resetting it to null. I would really like to know if that is good practice here or if I should rename that variable.

IMO it's fine, but perhaps adding a comment above the initial declaration that documents what type it should be and what it's used for might help.

src/device-state.coffee
```diff
@@ -675,24 +675,25 @@ module.exports = class DeviceState extends EventEmitter
...
 
-	triggerApplyTarget: ({ force = false, delay = 0, initial = false } = {}) =>
-		if @applyInProgress
-			if !@scheduledApply?
+	triggerApplyTarget: ({ force = false, delay = 0, initial = false, isFromApi = false } = {}) =>
+		if @applyInProgress?
+			if !@scheduledApply? || isFromApi
 				@scheduledApply = { force, delay }
+				if isFromApi
+					@applyInProgress.cancel()
```
pcarranzav: I think we need to call `Promise.config({ cancellation: true })` somewhere for this to actually work. I'd recommend putting it in the app's entrypoint, app.coffee.

test/05-device-state.spec.coffee
```diff
@@ -290,6 +290,13 @@ describe 'deviceState', ->
 			done()
 		, 5
 
+	it 'cancels current promise applying the target state', ->
+		@deviceState.applyInProgress = Promise.finally => @applyInProgress = null
```
pcarranzav: `Promise.finally` is actually undefined. I think you could instead do:

```
@deviceState.applyInProgress = (new Promise(->)).finally => @deviceState.applyInProgress = null
```

src/device-state.coffee
mwohlert: Yes I just read through the bluebird promise documentation for cancellation. Sorry for the oversight.

test/05-device-state.spec.coffee
mwohlert: Thank you. You're of course 100% correct.

mwohlert:
@pcarranzav Thanks for your helpful comments.
I agree with you about the cancellation of the promise and implemented your suggested solution to this. Thanks for catching that.

test/05-device-state.spec.coffee
mwohlert: I changed it slighly to make it testable due to the asynchronous nature of promises. Feel free to comment on my approach. Not sure if this is the ideal way

src/app.coffee
```diff
@@ -1,3 +1,5 @@
+Promise.config({ cancellation: true })
```
Page-: You need to import bluebird before you try to enable cancellation on it

src/device-state.coffee
```diff
@@ -675,24 +676,28 @@ module.exports = class DeviceState extends EventEmitter
...
-	triggerApplyTarget: ({ force = false, delay = 0, initial = false } = {}) =>
-		if @applyInProgress
-			if !@scheduledApply?
+	triggerApplyTarget: ({ force = false, delay = 0, initial = false, isFromApi = false } = {}) =>
+		if @applyInProgress?
+			if !@scheduledApply? || (isFromApi && @applyDelayInProgress)
 				@scheduledApply = { force, delay }
+				if isFromApi
+					# Cancel promise delay if call came from api to prevent waiting due to backoff
+					@applyInProgress.cancel()
```
Page-: So this is a little of a weird one but I think you can avoid the need for cancellation by just doing something like:
```coffee
@applyInProgress = new Promise (resolve) =>
	setTimeout(resolve, delay)
	@cancelDelay = resolve
```
and then calling `@cancelDelay()` here

mwohlert: Interesting approach, but I like it. It's more transparent than just cancelling the delayed promise. I'll implement that now.

src/app.coffee
mwohlert: I'll remove this since I'm going to implement your suggestion.

src/device-state.coffee
mwohlert: I changed @cancelDelay to reject since it would otherwise execute the .then chain, which executes the applyTargetState with the old force and initial parameters.
Please have a look.

src/device-state.coffee
```diff
@@ -675,24 +676,30 @@ module.exports = class DeviceState extends EventEmitter
...
+					@cancelDelay()
 			else
 				# If a delay has been set it's because we need to hold off before applying again,
 				# so we need to respect the maximum delay that has been passed
 				@scheduledApply.delay = Math.max(delay, @scheduledApply.delay)
 				@scheduledApply.force or= force
 			return
-		@applyInProgress = true
-		Promise.delay(delay)
+		@applyInProgress = new Promise (resolve, reject) =>
```
pcarranzav: Now that there's no cancellation, I think there's no need to assign the Promise to `@applyInProgress` anymore, so this could just be:

```
@applyInProgress = true
new Promise (resolve, reject) =>
 	setTimeout(resolve, delay)
	@cancelDelay = reject
.then =>
	@cancelDelay = null
	...
.finally =>
	@applyInProgress = false
	...
```

(But this doesn't hurt anyways so just a suggestion)

mwohlert: Oh yeah you are right. We don't need the promise reference anymore. I've reverted to using booleans for applyInProgress again. Thanks for the input.

src/device-state.coffee
```diff
@@ -675,19 +676,25 @@ module.exports = class DeviceState extends EventEmitter
...
+					@cancelDelay()
 			else
 				# If a delay has been set it's because we need to hold off before applying again,
 				# so we need to respect the maximum delay that has been passed
 				@scheduledApply.delay = Math.max(delay, @scheduledApply.delay)
 				@scheduledApply.force or= force
 			return
 		@applyInProgress = true
-		Promise.delay(delay)
+		new Promise (resolve, reject) =>
```
pcarranzav: I just realized we need to add a `.catch(->)` for this promise, otherwise there'll be an unhandled rejection when the promise is cancelled. There's a few ways to do this that I can think of, I would suggest something like:

```
@applyInProgress = true
@applyCancelled = false
new Promise (resolve, reject) =>
	setTimeout(resolve, delay)
	@cancelDelay = reject
.catch =>
	@applyCancelled = true
.then =>
	@cancelDelay = null
	if @applyCancelled
		log.info('Skipping applyTarget because of a cancellation')
		return
	log.info('Applying target state')
	@applyTarget({ force, initial })
.finally =>
...
```

Sorry I didn't spot this before...

mwohlert: No worries, but you're right.
Why not place the catch below the then and save the @applyCanceled?

```
@applyInProgress = true
new Promise (resolve, reject) =>
	setTimeout(resolve, delay)
	@cancelDelay = reject
.then =>
	@cancelDelay = null
	log.info('Applying target state')
	@applyTarget({ force, initial })
.catch =>
        @cancelDelay = null
        log.info('Skipping applyTarget because of a cancellation')
.finally =>
...
```
This would come at the cost of setting cancelDelay null in two places in the code, but I think it might be better than introducing another variable.

mwohlert: hmm but then on the other side my suggestion would also catch exceptions or errors from the then block lead to a log message that is not descriptive of what happend.

pcarranzav: Yup, that’s exactly what I was going to say :) after the then block we don’t know for sure where the error comes from. Alternatively you could use typed errors but it gets more complicated imo.

mwohlert: I implemented your suggestion. Since we know if the promise is cancelled I simplified the test while doing that.

CameronDiver:
@mwohlert I'm happy to merge this once you've rebased on master - let me know I'll merge and release. Thanks again!

mwohlert:
@CameronDiver done :)

CameronDiver:
@balena-ci retest

PR #1022, opened 2019-06-27
Fix tests run removing .only

roman-mazur:
The issue was introduced with one of the recent changes (#1012) related to local mode cleanup.

Change-type: patch


roman-mazur:
@CameronDiver my apology here :)
it was a different `.only` from what you pointed me to in that PR

I wonder if we could add it as a PR check

CameronDiver:
Yeah I think a git commit hook or something simlar could go a long way to finding and preventing these.

roman-mazur:
See #1025.

PR #1023, opened 2019-06-28
fix: When reporting initial config, avoid creating a "VPN enabled" un…

### LINKED ISSUE #1024: Supervisor can create an unnecessary SUPERVISOR_VPN_CONTROL = true device config var

pcarranzav:
…necessary override

Since we were comparing the VPN's value before adding the explicit "true", there were cases
were the VPN is off, and therefore "value" didn't match the default, so the supervisor would
create a device specific SUPERVISOR_VPN_CONTROL = true, which is unnecessary and causes issues if users don't expect this and move the device to an app that has VPN disabled. The correct behavior
is to compare "varValue" and only create a device config var if this value differs from the default.

(This was the behavior before the TS conversion in 01ed7bb103b4df8fb0679cf858220db42d4a0b92 )

Closes #1024 

Change-type: patch

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -708,7 +708,7 @@ export class APIBinder {
 				varValue = 'true';
 			}
 
-			if (targetConfig[key] == null && value !== defaultConfig[key]) {
+			if (targetConfig[key] == null && varValue !== defaultConfig[key]) {
 				const envVar = {
 					value: varValue,
 					device: deviceId,


src/api-binder.ts
```diff
@@ -708,7 +708,7 @@ export class APIBinder {
 				varValue = 'true';
 			}
 
-			if (targetConfig[key] == null && value !== defaultConfig[key]) {
+			if (targetConfig[key] == null && varValue !== defaultConfig[key]) {
```
roman-mazur: @pdcastro would you clarify for me if these `SUPERVISOR_LOCAL_MODE` and `SUPERVISOR_VPN_CONTROL` variables can be practically set (changed) on the device side (not only on the backend side). If yes, how?

pdcastro: Did you mean to ask this question to @pcarranzav ? :-)
Browsing the supervisor API reference, I have found this [Set a target state (POST /v2/local/target-state)](https://www.balena.io/docs/reference/supervisor/supervisor-api/#set-a-target-state-post-v2localtarget-state) action, which I believe can be used locally on the device side, and thus may count as a way of "practically setting (changing) the variables on the device side". And these other docs may also be of interest:
* https://www.balena.io/docs/reference/supervisor/bandwidth-reduction/
* https://www.balena.io/docs/learn/manage/serv-vars/

roman-mazur: Yes, sorry @pdcastro, the question addressed to @pcarranzav 

And thanks for the references.
That said, the API you point to is used in local mode context only, and it does not affect the backend state in any way.

From this PR, it looks like the API used by supervisor to report its state actually updates the variables on the backend. I wonder if we actually have use cases, when source of truth for those variables has to be the supervisor app, and not the backend.

CameronDiver: @roman-mazur all environment and configuration vars should have the api as the source of truth. You can temporarily change them whilst in local mode (but the cli doesn't support this), but when reconnecting to a backend, it will mirror the state on the backend.

pcarranzav: @roman-mazur this report is only done on the supervisor's first run. As @CameronDiver says, the source of truth is the API, *but* on first boot, the supervisor "initializes" these variables to reflect what is currently on the device, as there are a few cases where these can differ from the default, e.g. if the user has modified config.txt locally before flashing the SD card. After this initial report, the source of truth is the API and any local changes will be overridden by the supervisor following what the API says.

(And also as Cameron says, local mode is an exception where we temporarily allow local changes)

pcarranzav:
@balena-ci retest

nazrhom:
@balena-ci retest

src/api-binder.ts
roman-mazur: Thanks for the answers, guys.

PR #1028, opened 2019-07-02
Use instantiations of compose objects throughout application-manager

CameronDiver:
To do this, we first split the volumes module into a volume-manager and volume manager, the same as the network model.

We then make sure to initialise these objects at the start of the state comparison, and these values are what are passed throughout and operated on.

This is work for both getting more of the codebase in typescript, but more importantly it's the basis for https://github.com/balena-io/balena-supervisor/issues/1026 as it fixed a bug where appIds will not reference the actual application that volumes and networks come from.


src/compose/volume.ts
```diff
@@ -0,0 +1,143 @@
...
+				error: e,
+			});
+		}
+	}
+
+	public static generateDockerName(appId: number, name: string) {
+		return `${appId}_${name}`;
+	}
+
+	private static deconstructDockerName(
```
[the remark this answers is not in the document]
roman-mazur: Getting the network would need to be changes to a "list networks with label filter" call in such case.

src/compose/volume.ts
```diff
@@ -0,0 +1,143 @@
...
+		} catch (e) {
+			this.logger.logSystemEvent(LogTypes.removeVolumeError, {
+				volume: { name: this.name, appId: this.appId },
+				error: e,
+			});
+		}
+	}
+
+	public static generateDockerName(appId: number, name: string) {
+		return `${appId}_${name}`;
```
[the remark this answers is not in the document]
CameronDiver: Yeah we should, but changing anything like this is a huge job, because it requires migrating of the volumes, and I'd much rather have a supervisor which doesn't delete volumes first.

src/compose/volume.ts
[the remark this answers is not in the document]
CameronDiver: Me too, but the same problem applies as here: https://github.com/balena-io/balena-supervisor/pull/1028#discussion_r299444489, I do plan to move to this kind of architecture, but I want to make sure that we're not going to delete any user data.

src/compose/volume.ts
```diff
@@ -0,0 +1,143 @@
...
+			isEqual(this.config.labels, volume.config.labels)
+		);
+	}
+
+	public async create(): Promise<void> {
+		this.logger.logSystemEvent(LogTypes.createVolume, {
+			volume: { name: this.name },
+		});
+		// Check that we're not trying to recreate a volume that
+		// already exists
```
zvin: Is this a TODO? It seems like it's already checked in `VolumeManager.create(volume: Volume)`.

CameronDiver: Ah it's meant to be in that other module, forgot to move it.

src/application-manager.coffee
```diff
@@ -398,57 +387,19 @@ module.exports = class ApplicationManager extends EventEmitter
...
-				target: {
-					name
-					appId
-					config: target[name]
-				}
-			})
-		toBeUpdated = _.filter _.intersection(targetNames, currentNames), (name) =>
+			outputPairs.push({ current: null, target: target[name] })
+		toBeUpdated = _.filter _.intersection(targetNames, currentNames), (name) ->
 			# While we're in this in-between state of a network-manager, but not
```
pcarranzav: This comment is no longer needed :)

src/application-manager.coffee
```diff
@@ -732,26 +699,33 @@ module.exports = class ApplicationManager extends EventEmitter
...
-				volumes = _.mapValues volumes, (volumeConfig) ->
+				volumes = _.mapValues volumes, (volumeConfig, volumeName) =>
 					volumeConfig ?= {}
 					volumeConfig.labels ?= {}
-					return volumeConfig
+					@createTargetVolume(volumeName, app.appId, volumeConfig)
+
+				networks = JSON.parse(app.networks)
+				networks = _.mapValues networks, (networkConfig, networkName) =>
+					networkConfig ? {}
```
pcarranzav: I think you meant `networkConfig ?=  {}`? In which case you could define the default above as `networks = _.mapValues networks, (networkConfig = {}, networkName) =>`

src/application-manager.coffee
```diff
@@ -151,24 +153,11 @@ module.exports = class ApplicationManager extends EventEmitter
...
-				if step.model is 'network'
-					Network.fromComposeObject({ @docker, @logger },
-						step.current.name,
-						step.appId,
-						step.current.config
-					).remove()
-				else
-					@volumes.remove(step.current)
+			removeNetworkOrVolume: (step) ->
+				step.current.remove()
```
pcarranzav: imo it's a bit odd that models are created through their Manager but removed directly - I'd rather have everything go to the Manager for consistency, even if it doesn't add much at this point.

src/compose/network-manager.ts
```diff
@@ -40,17 +42,45 @@ export class NetworkManager {
...
-				docker: this.docker,
-			},
-			network.name,
-			network.appId,
+	public async get(network: {
+		name: string;
+		appId: number;
+	}): Bluebird<Network> {
+		const dockerNet = await this.docker
+			.getNetwork(`${network.appId}_${network.name}`)
```
pcarranzav: Might be nice to have a `Network.generateDockerName(appId, name)` like you did for Volume

src/compose/volume.ts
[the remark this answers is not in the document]
pcarranzav: Note that having the names defined like this is also useful to easily reference them when creating service containers - otherwise we'd also have to find the correct name for an app's volume when including it in each container. And I made it so that we also get these properties (appId, volume name) from the name so that there's no duplication.

src/application-manager.coffee
roman-mazur: +1 here

src/application-manager.coffee
roman-mazur: How can `networkConfig` be null or undefined?

src/compose/network-manager.ts
roman-mazur: +1 here

Though I would rather have it like
```
this.docker.getNetwork(network.dockerName())
```

src/compose/volume.ts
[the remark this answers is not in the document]
roman-mazur: It's 100% useful to have such names, but I don't think supervisor logic should rely on name format: we might want to change it. 
To me, it makes sense to have friendly names helping with support/triaging/debugging, but not rely on decoding metadata from them. It would be easier on the long term to rely only on labels for structured metadata.

src/application-manager.coffee
CameronDiver: It's like my brain just doesn't see outdated comments :P

src/application-manager.coffee
CameronDiver: We can't be certain of any of the values that come in from docker-compose files, as users can do anything like to add them to the api. I'm going to add runtime validation with io-ts but for now, we use checks like this.

src/application-manager.coffee
CameronDiver: Fair point, will change.

src/compose/network-manager.ts
CameronDiver: @roman-mazur it's a lot more flexible to have a static function when there's no need to create an object for the method's task.

package-lock.json
```diff
@@ -1222,7 +1222,7 @@
         "fetch-ponyfill": "^4.0.0",
         "fetch-readablestream": "^0.2.0",
         "lodash": "^4.6.1",
-        "node-web-streams": "github:resin-io-modules/node-web-streams#emit-errors",
+        "node-web-streams": "github:resin-io-modules/node-web-streams#46f98300b69090bde3f6b4983877ccfe283a892c",
```
roman-mazur: Why do we get changes like this?

CameronDiver: Different npm versions I believe.

CameronDiver:
@balena-ci retest

Issue #1026, opened 2019-07-01
The supervisor should not delete volumes from an application without an explicit action


CameronDiver:
If a volume is not referenced after an update, it will be deleted. This can cause an accidental push to remove a volume (and all of the data) from a device. 

Instead of implicitly removing volumes, we should provide a supervisor endpoint which will reap all volumes which are no longer referenced.

PR #1032, opened 2019-07-05
Don't delete volumes implicitly

CameronDiver:
We only delete volumes from the device when moving between applications, where the action is explicit. We also provide an endpoint to remove orphaned volumes. 

A couple of bugfixes have made it into this PR too.

Closes: #1026

### DIFF
--- a/src/compose/volume-manager.ts
+++ b/src/compose/volume-manager.ts
@@ -1,7 +1,5 @@
 import * as Docker from 'dockerode';
-import filter = require('lodash/filter');
-import get = require('lodash/get');
-import unionBy = require('lodash/unionBy');
+import * as _ from 'lodash';
 import * as Path from 'path';
 
 import constants = require('../lib/constants');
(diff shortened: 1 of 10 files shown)


src/compose/volume-manager.ts
```diff
@@ -131,6 +129,36 @@ export class VolumeManager {
...
+			.uniq()
+			.value();
+		const volumeNames = _.map(dockerVolumes.Volumes, 'Name');
+
+		const volumesToRemove = _.difference(volumeNames, containerVolumes);
+		const removalPromises: Array<Promise<void>> = [];
+		for (const volume of volumesToRemove) {
+			removalPromises.push(this.docker.getVolume(volume).remove());
+		}
+		await Promise.all(removalPromises);
```
zvin: NIT:
```typescript
await Promise.all(volumesToRemove.map(v => this.docker.getVolume(v).remove()))
```

CameronDiver:
@roman-mazur I just added the tests to the same file as I wanted to use some of the infrastructure there (and I think all of the tests will need to be recreated when we change the applications model). 

The dashboard will have to use version comparisons to detect this, I'm going to chat with @shaunmulligan about the best way to roll this out.

PR #1035, opened 2019-07-11
Don't attempt to cleanup any target state referenced volumes

CameronDiver:
The code before this change could potentially remove a volume which
should not be removed if a container was deleted before the call that
references said volume.

To avoid this, we additionally filter the list of volumes to cleanup by
any that are referenced in the target state. This means that cleanup
will never remove it, as long as it's still supposed to be there,
regardless of if a container references it or not.

Change-type: patch

### DIFF
--- a/src/application-manager.d.ts
+++ b/src/application-manager.d.ts
@@ -11,11 +11,14 @@ import ServiceManager from './compose/service-manager';
 import DB from './db';
 
 import { APIBinder } from './api-binder';
-import { Service } from './compose/service';
 import Config from './config';
 
 import NetworkManager from './compose/network-manager';
 import VolumeManager from './compose/volume-manager';
+
+import Network from './compose/network';
+import Service from './compose/service';
+import Volume from './compose/volume';
 
 declare interface Options {
 	force?: boolean;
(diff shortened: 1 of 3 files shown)


src/application-manager.d.ts
```diff
@@ -65,6 +68,14 @@ export class ApplicationManager extends EventEmitter {
...
 	// the state endpoint and the ApplicationManager internals
 	public getStatus(): Promise<Dictionay<any>>;
+	// The return type is incompleted
+	public getTargetApps(): Promise<
+		Dictionary<{
+			services: Dictionary<Service>;
+			volumes: Dictionary<Volume>;
+			networks: Dictionary<Network>;
+		}>
+	>;
```
CameronDiver: I'm currently in the process of rewriting this module in typescript, so it would be counter productive to define this fully and the only typescript user is this module.

src/device-api/v2.ts
```diff
@@ -488,7 +489,16 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
 		try {
-			await applications.volumes.removeOrphanedVolumes();
+			const targetState = await applications.getTargetApps();
+			const referencedVolumes: string[] = [];
+			_.each(targetState, app => {
+				_.each(app.volumes, vol => {
+					referencedVolumes.push(
+						Volume.generateDockerName(vol.appId, vol.name),
+					);
+				});
```
roman-mazur: nit: this might work too
```
referencedVolumens = _.flatMap(targetState, (app) => app.volumes)
  .map((vol) => Volume.generateDockerName(vol.appId, vol.name));
```

CameronDiver: It's a bit strange, the initial map gets a list of dictionaries, and because the names can be the same across applications, I wouldn't like to have lodash do the merging (if it even can).

Issue #851, opened 2018-12-19
Support the service: <name> option for network_mode


CameronDiver:
Users would like to assign a network mode for a service to another service.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_g5bf)

edkeeble:
Hi @CameronDiver,

We are very interested in this feature. I would be happy to take a crack at it, if you can point me in the right direction, as I haven't worked on supervisor before and have just taken a quick initial look. I see where these values are being filtered out in service.ts, but haven't yet figured out where they are used following that.

Thanks!

CameronDiver:
Hey @edkeeble thanks for the interest!

I'll try to outline the steps necessary for this to be implemented;

Firstly, this causes an implicit dependency between the services, as a service `a` having a `network_mode` of `service: b`, would require that `b` is running, so that we can convert the service name to the container id, which would be required by docker (https://docs.docker.com/engine/api/v1.30/#operation/ContainerCreate). To do this we would have to add a `depends_on` entry, in the service configuration.

The next issues arises when comparing the current and target state of the container with a `service:` network mode. We'd have to convert the service name to a container id, to ensure that if the depended on service gets recreated, we will recreate the service with a service with a `service:` network mode. This could also create a deadlock if done improperly, so it's something to watch out for.

There may be other things which come up because of it too, so all in all it's not a small (or simple) change. 

What is the reason that you would like to use this feature? Would it be possible to use a private network instead?

I'll be away for a week or two, but can try to prioritise this feature when I get back if you prefer.

edkeeble:
Thanks for the explanation, Cameron.

In our case, we're supporting a legacy application which needs to route traffic from/to our cloud services to/from specific containers on our edge device through an openvpn container (streaming video from the device, for instance). We actually have a workaround setup, using iptables on each of the relevant containers to handle the routing, but for the sake of long-term maintenance, we'd prefer to use the more typical network_mode approach.

I don't mind taking a look at it, as I'm sure I'll learn a lot about supervisor in the process, but given the level of complexity and possibility of side effects, if you would prefer I leave it to you, then we can certainly wait until you get back and have a chance to look into it.

Thanks again.

CameronDiver:
[...]

PR #1036, opened 2019-07-15
Support network_modes of service:<servicename>

CameronDiver:
That way that this is performed is by first adding a depends_on entry
for the target service if it appears in a network mode. Then when we
generate the docker container for this service, we use the containerId
of the target container and replace the network_mode with
`container:<id>`.

When comparing state, we check that the containerId still points to the
contianerId of the target container, and in this way we ensure that
when a network mode target container changes, we change the dependent
container too.

Change-type: minor
Closes: #851


src/compose/service-manager.ts
```diff
@@ -93,9 +93,13 @@ export class ServiceManager extends (EventEmitter as new () => ServiceManagerEve
 	}
 
 	public async get(service: Service) {
```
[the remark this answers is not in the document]
CameronDiver: It's an artefact of the coffeescript days, when we didn't pass around instantiations of the objects in the state engine. It can be removed but it requires some refactors which I haven't gotten around to yet.

src/compose/service.ts
```diff
@@ -565,6 +575,16 @@ export class Service {
...
 
+		const match = this.config.networkMode.match(SERVICE_NETWORK_MODE_REGEX);
+		if (match != null) {
+			const containerId = opts.containerIds[match[1]];
+			if (!containerId) {
+				throw new Error(
+					`No container for network_mode: 'service: ${match[1]}'`,
+				);
+			}
+			this.config.networkMode = `container:${containerId}`;
```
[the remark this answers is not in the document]
CameronDiver: It should really be a private variable tbh, but I'm waiting until everything is in typescript before making large changes like this.

src/compose/service.ts
```diff
@@ -702,24 +725,70 @@ export class Service {
...
+			[this.config.networkMode, service.config.networkMode],
+			[service.config.networkMode, this.config.networkMode],
+		]) {
+			const aMatch = a.match(SERVICE_NETWORK_MODE_REGEX);
+			const bMatch = b.match(SERVICE_NETWORK_MODE_REGEX);
+			if (aMatch != null) {
+				if (bMatch != null) {
+					// They're both service: <name> network modes
+					if (aMatch[1] !== bMatch[1]) {
+						sameNetworkMode = sameNetworkMode || false;
```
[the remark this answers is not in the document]
CameronDiver: What I was going for here is that once `sameNetworkMode` goes to `true`, it never changes again (because we check both 'orders' of the network mode. I can see how this code could be confusing though, I'll look into making it a little more obvious.

src/compose/service.ts
```diff
@@ -702,24 +725,76 @@ export class Service {
...
+		// Check the network mode separetely, as if it is a
+		// service: network mode, the container id needs to be
+		// checked against the running containers
+
+		// When this function is called, it's with the current
+		// state as a parameter and the target as the instance.
+		// We shouldn't rely on that because it's not enforced
+		// anywhere. For that reason we need to consider both
+		// network_modes in the correct way
+		let sameNetworkMode = true;
```
pcarranzav: Shouldn't this start as `false`?

src/compose/service.ts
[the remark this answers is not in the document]
roman-mazur: I would suggest adding a unit test that exercises different variants and includes both equal and not-equal cases (right now, you only have a test for `isEqualConfig.to.be.true`). I still don't feel this implementation does what it should.

roman-mazur:
I agree with @pcarranzav that moving conversion to `container:id` to earlier time should make stuff more well organized. 

But do we need to care about containers recreation in such case?
I we can assume that containers we care about may be created by supervisor only, things should be simple enough to apply the conversion from `service:` to `container:` early, when target state is received.

src/compose/service.ts
CameronDiver: Woah, it most certainly should be, I'm not sure how this slipped through, good catch!

src/compose/service.ts
```diff
@@ -190,8 +180,27 @@ export class Service {
...
+				}
+				service.dependsOn.push(match[1]);
+				serviceNetworkMode = true;
+			} else if (CONTAINER_NETWORK_MODE_REGEX.test(config.networkMode)) {
+				log.warn(
+					'A network_mode referencing a container is not supported. Ignoring.',
+				);
+				delete config.networkMode;
+			}
+		} else if (!config.networkMode) {
```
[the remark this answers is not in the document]
CameronDiver: Nope, good catch.

src/application-manager.coffee
```diff
@@ -860,35 +860,45 @@ module.exports = class ApplicationManager extends EventEmitter
...
+			if diff.length > 0
+				oldApps = diff
+
+		_(current.local.apps)
+			.keys()
+			.concat(_.keys(target.local.apps))
+			.uniq()
+			.each (id) =>
+				intId = checkInt(id)
+				containerIdsByAppId[intId] = @services.getContainerIdMap(intId)
```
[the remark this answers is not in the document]
CameronDiver: Agreed, changed.

src/device-state.coffee
```diff
@@ -605,13 +605,13 @@ module.exports = class DeviceState extends EventEmitter
...
-						@getCurrentForComparison()
-						@getTarget({ initial, intermediate })
-					])
-					.then ([ currentState, targetState ]) =>
+				Promise.all([
+					@getCurrentForComparison()
+					@getTarget({ initial, intermediate })
+				])
+				.then ([ currentState, targetState ]) =>
+					@applications.getExtraStateForComparison(currentState, targetState)
```
pcarranzav: I'm afraid this change of ordering would re-introduce the race condition fixed in https://github.com/balena-io/balena-supervisor/commit/42737cb9e96b889d0fdfd48926778a38566d927b

Not sure what'd be the best way to fix it though - I think you can get around it by calling `@images.getAvailable` before getting the current state, and removing it from `getExtraStateForComparison`. But the "correct" fix probably involves changing ApplicationManager's `normaliseAndExtendAppFromDB` to not do calls to `@images` directly and instead take the available images as an argument.

src/compose/service.ts
```diff
@@ -702,24 +726,76 @@ export class Service {
...
+			[service.config.networkMode, this.config.networkMode],
+		]) {
+			const aMatch = a.match(SERVICE_NETWORK_MODE_REGEX);
+			const bMatch = b.match(SERVICE_NETWORK_MODE_REGEX);
+			if (aMatch != null) {
+				if (bMatch != null) {
+					// They're both service: <name> network modes
+					if (aMatch[1] !== bMatch[1]) {
+						continue;
+					}
```
pcarranzav: I think this is missing an `else { sameNetworkMode = true; }`?

(I don't think we'll ever have a case where both have `service:` network modes, but for consistency this should be true if they match, shouldn't it?

src/device-state.coffee
```diff
@@ -244,7 +244,7 @@ module.exports = class DeviceState extends EventEmitter
 			@applications.services.killAllLegacy()
 		.then =>
 			log.debug('Migrating legacy app volumes')
-			@applications.getTargetApps()
+			@applications.getTargetApps({ availableImages: [] })
```
pcarranzav: nit: It might be nicer to make `availableImages` default to `[]` inside the function?

test/14-application-manager.spec.coffee
```diff
@@ -191,14 +192,15 @@ describe 'ApplicationManager', ->
 	after ->
 		@applications.images.inspectByName.restore()
 		@applications.docker.getNetworkGateway.restore()
+		@applications.docker.listContainers.restore()
 		Service.extendEnvVars.restore()
 
 	it 'infers a start step when all that changes is a running state', ->
 		Promise.join(
 			@normaliseCurrent(currentState[0])
 			@normaliseTarget(targetState[0], availableImages[0])
```
pcarranzav: Note that this above calls `normaliseAndExtendAppFromDB`, which now needs to get an `extraState` argument (this is causing some tests to fail)

src/compose/service.ts
```diff
@@ -702,24 +726,76 @@ export class Service {
...
+
+		// When this function is called, it's with the current
+		// state as a parameter and the target as the instance.
+		// We shouldn't rely on that because it's not enforced
+		// anywhere. For that reason we need to consider both
+		// network_modes in the correct way
+		let sameNetworkMode = false;
+		for (const [a, b] of [
+			[this.config.networkMode, service.config.networkMode],
+			[service.config.networkMode, this.config.networkMode],
```
roman-mazur: I keep thinking about this in bg. The comparison logic here is complex enough to deserve a separate `NetworkMode` class.

General idea:
```typescript

class NetworkMode {

  private constructor(
    public readonly type: 'host'|'bridge'|'none'|'service_ref'|'container_ref',
    public readonly serviceReference: string|undefined,
    public readonly containerReference: string|undefined
  )

  public static from(composeLiteral: string): NetworkMode {
    // Parse the literal and
    return new NetworkMode(type, serviceRef, containerRef);
  }

  public equals(another: NetworkMode): boolean {
    // Extremely easy to do now - just check if all 3 fields match.
    return this.type === another.type 
        && this.containerRef === another.containerRef 
        && this.serviceRef === another.containerRef;
  }

}

```

And then here we'd do
```typescript
  const sameNetworkMode = this.config.networkMode.equals(service.config.networkMode);
```

CameronDiver: The code you posted doesn't do the same thing that is necessary and this PR does. That being said the overall idea is a good one, but I'd like to handle it another time and get this out.

src/compose/service.ts
```diff
@@ -702,24 +725,76 @@ export class Service {
...
+				} else {
+					const containerMatch = b.match(CONTAINER_NETWORK_MODE_REGEX);
+					if (containerMatch == null) {
+						continue;
+					} else {
+						// We never want to switch sameNetworkMode back
+						// to false when it's true
+						if (!sameNetworkMode) {
+							sameNetworkMode =
+								currentContainerIds[aMatch[1]] === containerMatch[1];
```
roman-mazur: how about 
```
if (currentContainerIds[aMatch[1]] === containerMatch[1]) {
  sameNetworkMode = true;
  break;
}
```

src/compose/service.ts
```diff
@@ -702,24 +725,76 @@ export class Service {
...
+						}
+					}
+				}
+			} else {
+				// Only consider the hostname when it's not a
+				// service: entry
+				if (!sameNetworkMode) {
+					sameNetworkMode =
+						a === b && this.config.hostname === service.config.hostname;
```
roman-mazur:
```
if (a === b && this.config.hostname === service.config.hostname) {
  sameNetworkMode = true;
  break;
}
```

src/compose/service.ts
```diff
@@ -702,24 +725,76 @@ export class Service {
...
+			[this.config.networkMode, service.config.networkMode],
+			[service.config.networkMode, this.config.networkMode],
+		]) {
+			const aMatch = a.match(SERVICE_NETWORK_MODE_REGEX);
+			const bMatch = b.match(SERVICE_NETWORK_MODE_REGEX);
+			if (aMatch != null) {
+				if (bMatch != null) {
+					// They're both service: <name> network modes
+					if (aMatch[1] !== bMatch[1]) {
+						continue;
```
roman-mazur: With `break`s I suggested, this if construction may be simplified
```typescript
if (aMatch !== null) {
  if (bMatch === null) {
    const containerMatch = b.match(CONTAINER_NETWORK_MODE_REGEX);
    if (containerMatch !== null && currentContainerIds[aMatch[1]] === containerMatch[1]) {
      sameNetworkMode = true;
      break;
    }
  }
} else if (a === b && this.config.hostname === service.config.hostname) {
  sameNetworkMode = true;
  break;
}
```

src/device-state.coffee
roman-mazur: @CameronDiver thoughts here?

src/compose/service.ts
```diff
@@ -33,7 +35,7 @@ export class Service {
 	public imageName: string | null;
 	public containerId: string | null;
 
-	public dependsOn: string | null;
+	public dependsOn: string[] | null;
```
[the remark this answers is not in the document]
CameronDiver: This was a bug - the types are only at compile time though and the usage allowed arrays, it just was not defined as such.

src/device-state.coffee
CameronDiver: This is fixed already (see Pablo's approval).

src/compose/service.ts
CameronDiver: I'm not sure this is any more readable really, I'm not sure what you're trying to optimise for?

src/compose/service.ts
```diff
@@ -23,6 +23,8 @@ import { sanitiseComposeConfig } from './sanitise';
 
 import log from '../lib/supervisor-console';
 
+const SERVICE_NETWORK_MODE_REGEX = /service:\s*(.+)/;
+const CONTAINER_NETWORK_MODE_REGEX = /container:\s*(.+)/;
 export class Service {
```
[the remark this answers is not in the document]
CameronDiver: Done.

src/compose/service.ts
roman-mazur: It's much more readable to me because it contains only branches that we care about to say that configs are equal. There are basically only 2 branches in this code which maps to the actual definition of equality: either (networkMode refers to a service and its container ID matches the config) or (both networkMode and hostname are identical).

Other branches make the code bigger but don't help with deriving this definition from the code.

src/device-state.coffee
roman-mazur: Got it, thanks.

src/compose/service.ts
CameronDiver: Fair enough, although this is pretty subjective. I'll take a look and see if it can be made clearer.

CameronDiver: I implemented something similar.

src/compose/service.ts
```diff
@@ -702,24 +726,73 @@ export class Service {
...
+					}
+				} else {
+					// They're both service entries, we shouldn't get here
+					// but it's technically an equal configuration
+					if (a === b) {
+						sameNetworkMode = true;
+						break;
+					}
+				}
+				// We consider the hostname when it's not a service: entry
```
roman-mazur: nit: indent

blackjid:
🎉

blackjid:
@CameronDiver when could I expect the supervisor 10.1.0 (the one that has this changes) to be available in BalenaOS?? or is any other way to using without waiting for BalenaOS.

CameronDiver:
@blackjid sorry about the delay, I received the alert when on time off and it somehow slipped through the net! 

The release is currently going through testing so I imagine a balenaOS release will be available soon. Which device type are you using? I can try and chase up a more specific date.

blackjid:
Thanks, don't worry.. :)

I'm using raspberry pi's 3b and 3b+

Juan Ignacio Donoso

CameronDiver:
These devices should be ready within the next week or two, I can't really give a more precise estimate right now unfortunately.

PR #1038, opened 2019-07-17
Prevent supervisor from deleting itself

roman-mazur:
With recent changes to local mode behavior, supervisor removes all engine objects
that were not present before entering local mode.
With this change, supervisor also detects its own resources and does not delete them
if it was updated during local mode operations.

Related to https://github.com/balena-os/meta-balena/pull/1590 and https://github.com/balena-os/balena-engine/issues/173

Change-type: patch

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -9,6 +9,7 @@ const constants = {
 	rootMountPoint,
 	databasePath:
 		checkString(process.env.DATABASE_PATH) || '/data/database.sqlite',
+	containerId: checkString(process.env.SUPERVISOR_CONTAINER_ID) || undefined,
 	dockerSocket: process.env.DOCKER_SOCKET || '/var/run/docker.sock',
 	supervisorImage:
 		checkString(process.env.SUPERVISOR_IMAGE) || 'resin/rpi-supervisor',
--- a/src/local-mode.ts
+++ b/src/local-mode.ts
@@ -155,8 +155,8 @@ export class LocalModeManager {
 		return new EngineSnapshot(
 			[inspectInfo.Id],
 			[inspectInfo.Image],
-			inspectInfo.Mounts.filter(m => m.Name !== undefined).map(m => m.Name!),
-			_.values(inspectInfo.NetworkSettings.Networks).map(ns => ns.NetworkID),
+			inspectInfo.Mounts.filter(m => m.Name != null).map(m => m.Name!),
+			_.map(inspectInfo.NetworkSettings.Networks, n => n.NetworkID),
 		);
 	}
 
--- a/src/local-mode.ts
+++ b/src/local-mode.ts
@@ -5,6 +5,7 @@ import * as _ from 'lodash';
 import Config from './config';
 import Database from './db';
 import * as constants from './lib/constants';
+import { SupervisorContainerNotFoundError } from './lib/errors';
 import log from './lib/supervisor-console';
 import { Logger } from './logger';
 
@@ -176,7 +177,7 @@ export class LocalModeManager {
 				);
 				return this.collectContainerResources(fallback);
 			}
-			throw e;
+			throw new SupervisorContainerNotFoundError(e);
 		}
 	}
 
--- a/src/local-mode.ts
+++ b/src/local-mode.ts
@@ -4,6 +4,7 @@ import * as _ from 'lodash';
 
 import Config from './config';
 import Database from './db';
+import * as constants from './lib/constants';
 import log from './lib/supervisor-console';
 import { Logger } from './logger';
 
@@ -56,6 +57,9 @@ export class EngineSnapshotRecord {
 	) {}
 }
 
+/** Container name used to inspect own resources when container ID cannot be resolved. */
+const SUPERVISOR_CONTAINER_NAME_FALLBACK = 'resin_supervisor';
+
 /**
  * This class handles any special cases necessary for switching
  * modes in localMode.
@@ -70,6 +74,7 @@ export class LocalModeManager {
 		public docker: Docker,
 		public logger: Logger,
 		public db: Database,
+		private containerId: string | undefined = constants.containerId,
 	) {}
 
 	// Indicates that switch from or to the local mode is not complete.
@@ -140,6 +145,39 @@ export class LocalModeManager {
 			new EngineSnapshot(containers, images, volumes, networks),
 			new Date(),
 		);
+	}
+
+	private async collectContainerResources(
+		nameOrId: string,
+	): Promise<EngineSnapshot> {
+		const inspectInfo = await this.docker.getContainer(nameOrId).inspect();
+		return new EngineSnapshot(
+			[inspectInfo.Id],
+			[inspectInfo.Image],
+			inspectInfo.Mounts.filter(m => m.Name !== undefined).map(m => m.Name!),
+			_.values(inspectInfo.NetworkSettings.Networks).map(ns => ns.NetworkID),
+		);
+	}
+
+	// Determine what engine objects are linked to our own container.
+	private async collectOwnResources(): Promise<EngineSnapshot> {
+		try {
+			return this.collectContainerResources(
+				this.containerId || SUPERVISOR_CONTAINER_NAME_FALLBACK,
+			);
+		} catch (e) {
+			if (this.containerId !== undefined) {
+				// Inspect operation fails (container ID is out of sync?).
+				const fallback = SUPERVISOR_CONTAINER_NAME_FALLBACK;
+				log.warn(
+					'Supervisor container resources cannot be obtained by container ID. ' +
+						`Using '${fallback}' name instead.`,
+					e.message,
+				);
+				return this.collectContainerResources(fallback);
+			}
+			throw e;
+		}
 	}
 
 	private async cleanEngineSnapshots() {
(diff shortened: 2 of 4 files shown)


roman-mazur:
Keeping this as draft until the arch call.

roman-mazur:
@balena-ci retest

roman-mazur:
Promoting it to "Ready for review".
Exposing container ID without bind mount is tracked in balena-os/balena-engine#173

roman-mazur:
Rebased.
@balena-ci retest please

roman-mazur:
Got a test failure
https://ci.balena-dev.com/builds/730541
```
1) deviceState
       allows triggering applying the target state:
     Uncaught AssertionError: expected applyTarget to have been called with arguments { force: true, initial: false }
      at Timeout._onTimeout (build/test/05-device-state.spec.js:358:55)
```

roman-mazur:
@balena-ci retest once more please

roman-mazur:
@CameronDiver we seem to have a flaky test. It succeeded on re-run:
https://ci.balena-dev.com/builds/730719

Have you seen this failure before?

roman-mazur:
Rebased again.
@balena-ci retest please

roman-mazur:
@balena-ci retest please

src/lib/constants.ts
```diff
@@ -9,6 +9,7 @@ const constants = {
 	rootMountPoint,
 	databasePath:
 		checkString(process.env.DATABASE_PATH) || '/data/database.sqlite',
+	containerId: checkString(process.env.SUPERVISOR_CONTAINER_ID) || undefined,
```
roman-mazur: @CameronDiver @robertgzr this is the env variable we'll use to set the CID with `--env-cid` parameter in `balena run`.

I didn't use BALENA_ prefix, since we don't use it for other variables.

roman-mazur:
@balena-ci retest

src/lib/constants.ts
CameronDiver: As long as this is definable in the engine, that's fine by me.

src/local-mode.ts
```diff
@@ -142,6 +147,39 @@ export class LocalModeManager {
...
+	private async collectOwnResources(): Promise<EngineSnapshot> {
+		try {
+			return this.collectContainerResources(
+				this.containerId || SUPERVISOR_CONTAINER_NAME_FALLBACK,
+			);
+		} catch (e) {
+			if (this.containerId !== undefined) {
+				// Inspect operation fails (container ID is out of sync?).
+				const fallback = SUPERVISOR_CONTAINER_NAME_FALLBACK;
+				log.error(
```
CameronDiver:
```suggestion
				log.warn(
```

src/local-mode.ts
```diff
@@ -142,6 +147,39 @@ export class LocalModeManager {
...
+					'Supervisor container resources cannot be obtained by container ID. ' +
+						`Using '${fallback}' name instead.`,
+					e.message,
+				);
+				return this.collectContainerResources(fallback);
+			}
+			throw e;
```
CameronDiver: Can you make this a typed error please? Often errors can come from seemingly nowhere, so wrapping this would help know it's origin.

test/local-mode.ts
```diff
@@ -1,6 +1,7 @@
 import * as assert from 'assert';
```
CameronDiver: This fine should be named `local-mode.spec.ts`

test/local-mode.ts
```diff
@@ -1,6 +1,7 @@
...
 import { expect } from 'chai';
 import * as Docker from 'dockerode';
+import * as fs from 'fs';
```
CameronDiver: This is unused, and for future reference, we tend to use `import { fs } from 'mz';` to get the free promisified functions.

src/local-mode.ts
roman-mazur: The problem here is that I don't think dockerode/docker-modem throws a typed error.
https://github.com/apocas/docker-modem/blob/master/lib/modem.js#L257

roman-mazur: Or do you mean wrapping it with our own type?

CameronDiver: Wrapping it with our type, for example
```typescript
import * as TypedError from 'typed-error';
export class SupervisorContainerNotFoundError extends TypedError {}

throw new SupervisorContainerNotFoundError(e)
```

roman-mazur: normal typescript import does not work (lint fails), so I used an approach from other files (like `src/lib/errors.ts`):
```
import TypedError = require('typed-error');
```
I believe the package needs an upgrade, but this should be done in a separate PR.

roman-mazur:
@balena-ci retest please

src/local-mode.ts
CameronDiver: Ah yeah I forgot about that - I was waiting for 100% typescript before doing that.

src/local-mode.ts
```diff
@@ -56,6 +58,12 @@ export class EngineSnapshotRecord {
 	) {}
 }
 
+/** Container name used to inspect own resources when container ID cannot be resolved. */
+const SUPERVISOR_CONTAINER_NAME_FALLBACK = 'resin_supervisor';
+
+/** An error thrown if our own container cannot be inspected. */
+class SupervisorContainerNotFoundError extends TypedError {}
```
CameronDiver: Please add this to `./errors.ts`

roman-mazur: Done.

roman-mazur:
@balena-ci retest please

src/local-mode.ts
```diff
@@ -142,6 +148,39 @@ export class LocalModeManager {
...
 	}
 
+	private async collectContainerResources(
+		nameOrId: string,
+	): Promise<EngineSnapshot> {
+		const inspectInfo = await this.docker.getContainer(nameOrId).inspect();
+		return new EngineSnapshot(
+			[inspectInfo.Id],
+			[inspectInfo.Image],
+			inspectInfo.Mounts.filter(m => m.Name !== undefined).map(m => m.Name!),
```
CameronDiver: I think this should be `== null`. The docker daemon can sometimes return `null` for various fields, or omit the field altogether. The best thing to do with using the docker api is to always use `== null` to cover both instances.

src/local-mode.ts
```diff
@@ -142,6 +148,39 @@ export class LocalModeManager {
...
+			_.values(inspectInfo.NetworkSettings.Networks).map(ns => ns.NetworkID),
```
CameronDiver:
```suggestion
			_.mapValues(inspectInfo.NetworkSettings.Networks, 'NetworkID'),
```

src/local-mode.ts
roman-mazur: Good to know, thanks.

src/local-mode.ts
roman-mazur: It's not type safe! :)
Do you insist?

src/local-mode.ts
roman-mazur: Changed.

src/local-mode.ts
CameronDiver: Hm I thought it was type safe? Otherwise I'd prefer `mapValues` with a lambda

src/local-mode.ts
```diff
@@ -70,6 +75,7 @@ export class LocalModeManager {
 		public docker: Docker,
 		public logger: Logger,
 		public db: Database,
+		private containerId: string | undefined = constants.containerId,
```
Page-: Just to check, the `| undefined` here is to allow assigning `constants.containerId` right?  Because if it's expected to be able to explicitly pass `undefined` as the value then that will fail due to the default

src/local-mode.ts
Page-: I'd go for
```suggestion
			_.mapValues(inspectInfo.NetworkSettings.Networks, ns => ns.NetworkID),
```
to get the best of both

roman-mazur: In fact, `mapValues` transforms object to an object - it does not return an array
https://lodash.com/docs/4.17.14#mapValues

src/local-mode.ts
roman-mazur: It's supposed to be used either as
```
new LocalModeManager(config, docker, logger, db)
```
from the main code where the manager is created - in such case the ID will be taken from constants (using the environment variable)
or as
```
new LocalModeManager(config, docker, logger, db, 'someSpecificValue')
```
which is used in tests to inject a specific container ID

roman-mazur:
>Commit message Filter both null and ommited values should be omitted instead

Has already been fixed. Thanks.

@balena-ci retest please

src/local-mode.ts
Page-: Right, I was just looking at the type safe bit, you can use straight `_.map` in that case to have an array returned

roman-mazur: Changed.

roman-mazur:
@balena-ci retest please

roman-mazur:
@CameronDiver I agree it's a sensitive word, so I renamed the test and adjusted commit messages.
Branch will be deleted after the merge, so we should be good.

roman-mazur:
@balena-ci retest

roman-mazur:
Rebased again
@balena-ci retest please

Issue #1039, opened 2019-07-17
Supervisor spews "Error creating volume" after update


imrehg:
Raspberry Pi3 2.12.7+rev2 / supervisor 7.4.3 updated to 2.38.3(?) test build that has supervisor 9.18.8 in it, these errors keep being posted in the logs:

```
17.07.19 12:40:50 (+0100) Error creating volume 'resin-data' due to 'Trying to create volume with name: resin-data, but a volume with that name and a different configuration already exists'
```

* Updating a Pi1, 2.32.0 (supervisor v9.14.0) to the same version doesn't result in a problem
* 

Also note, when looking at the whole logs, it seems like after the update the `resin-data` paritition is attempted to be removed:
```
17.07.19 14:35:30 (+0100) Supervisor starting
17.07.19 14:33:46 (+0100)  main  Idling...
17.07.19 14:35:43 (+0100) Killing service 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:35:54 (+0100) Service exited 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:35:55 (+0100) Killed service 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:36:11 (+0100) Removing volume 'resin-data'
17.07.19 14:36:11 (+0100) Installing service 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:36:13 (+0100) Installed service 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:36:13 (+0100) Starting service 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:36:29 (+0100) Started service 'main sha256:b0d484522897a0ae241381acdd21585a69d993151a8c25267c7358d5f6ec3edd'
17.07.19 14:36:27 (+0100)  main  Idling...
17.07.19 14:36:37 (+0100) Error creating volume 'resin-data' due to 'Trying to create volume with name: resin-data, but a volume with that name and a different configuration already exists'
[...]

PR #1043, opened 2019-07-18
fix: Don't consider volume supervised labels when comparing

CameronDiver:
Change-type: patch
Closes: #1039

### DIFF
--- a/src/compose/volume.ts
+++ b/src/compose/volume.ts
@@ -1,6 +1,7 @@
 import * as Docker from 'dockerode';
 import assign = require('lodash/assign');
 import isEqual = require('lodash/isEqual');
+import omitBy = require('lodash/omitBy');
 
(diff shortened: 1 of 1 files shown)


src/compose/volume.ts
```diff
@@ -148,6 +152,14 @@ export class Volume {
 			name: match[2],
 		};
 	}
+
+	private static omitSupervisorLabels(labels: LabelObject): LabelObject {
+		// TODO: Export these to a constant
```
imrehg: What does this todo item mean?

CameronDiver: We shouldn't really be hardcoding any strings if we can help it, other than to add them to the constants module. That being said, there's a bunch of them all over the codebase so I'd rather handle them at once.

PR #1045, opened 2019-07-19
Improve build speed

Page-:


Page-:
Ignore the very last commit, I've still got it there for the moment just to see if there's any more low hanging fruit but will remove it before a merge

roman-mazur:
Thanks @Page- I started looking at the docker files myself today after waiting for 1+ h to get check results.

roman-mazur:
@Page- so I'm watching builds of another PR that takes 1h+ to complete the checks.

It looks like the majority of time is spent on building architecture-specific node modules (the `node-deps` part of our multistage build).
If I read our Dockerfile and scripts correctly, this image is never cached (it's not in the tree of the target image, so it's never pushed/pulled to/from the registry), though its cache should have a good hit rate since the only "source" is `package.json`, and it's changed rarely.

Do you think tagging and pushing it would make sense?

PR #1057, opened 2019-08-09
modification to unpacking volume backup

samothx:
Due to changes in migrator volume backups cannot be unpacked any more with 
tar -xzf backup.tgz -C #{backupPath} .
Leaving away the '.' works for old and new backups


samothx:
@CameronDiver sorry, I have not run into this before. What do I do about the Versionist failure ?

CameronDiver:
@samothx don't worry, I'll take over this branch :+1:

PR #1067, opened 2019-08-19
Only consider certain array fields without order

CameronDiver:
Various fields returned from the docker daemon don't retain order (for
example the volumes field). We now only select certain array values to
compare taking order into account.

Change-type: patch

### DIFF
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -46,8 +46,6 @@ export class Service {
 		'devices',
 		'capAdd',
 		'capDrop',
-		'dns',
-		'dnsSearch',
 		'dnsOpt',
 		'tmpfs',
 		'extraHosts',
@@ -56,6 +54,13 @@ export class Service {
 		'groupAdd',
 		'securityOpt',
 	];
+	public static orderedConfigArrayFields: ServiceConfigArrayField[] = [
+		'dns',
+		'dnsSearch',
+	];
+	public static allConfigArrayFields: ServiceConfigArrayField[] = Service.configArrayFields.concat(
+		Service.orderedConfigArrayFields,
+	);
 
 	// A list of fields to ignore when comparing container configuration
 	private static omitFields = [
@@ -69,7 +74,7 @@ export class Service {
 		// These fields are special case, due to network_mode:service:<service>
 		'networkMode',
 		'hostname',
-	].concat(Service.configArrayFields);
+	].concat(Service.allConfigArrayFields);
 
 	private constructor() {}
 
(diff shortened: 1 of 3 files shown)


src/compose/service.ts
```diff
@@ -684,17 +689,31 @@ export class Service {
...
 				}
 				return eq;
-			});
+			}) &&
+			_.every(
+				Service.orderedConfigArrayFields,
+				(field: keyof ServiceConfig) => {
+					const eq = _.isEqual(this.config[field], service.config[field]);
+					if (!eq) {
+						differentArrayFields.push(field);
```
zvin: This will never be reached if the previous `_.every(...)` return false, making `differentArrayFields` potentially incorrect (missing some ordered fields).

src/compose/service.ts
```diff
@@ -684,17 +689,31 @@ export class Service {
...
+					_.sortBy(service.config[field] as unknown[]),
+				);
 				if (!eq) {
 					differentArrayFields.push(field);
 				}
 				return eq;
-			});
+			}) &&
+			_.every(
+				Service.orderedConfigArrayFields,
```
roman-mazur: Consider
```typescript
type extractFn = (input: unkown[]) => unknown[];
const comparator = (fieldsArray: string[], extractor: extractFn) => 
  _.every(fieldsArray, (field: keyof ServiceConfig) => {
    const eq = _.isEqual(extractor(this.config[field]), extractor(service.config[field]));
    if (!eq) {
      differentArrayFields.push(field);
    }
    return eq;
  });

sameConfig = sameConfig && comparator(Service. orderedConfigArrayFields, (x) => x) 
  && comparator(Service.configArrayFields, (array) => _.sortBy(array));
```

src/compose/service.ts
```diff
@@ -56,6 +54,13 @@ export class Service {
 		'groupAdd',
 		'securityOpt',
 	];
+	public static orderedConfigArrayFields: ServiceConfigArrayField[] = [
+		'dns',
+		'dnsSearch',
```
roman-mazur: What happens if `devices` contains an overlap in container paths? For instance,
```
devices:
  - "/dev/device-1:/dev/device-1"
  - "/dev/device-2:/dev/device-1"
```

Does it result in an error or does the order matter?

src/compose/utils.ts
```diff
@@ -499,3 +499,60 @@ export function normalizeLabels(labels: {
...
+	orderedFields: Array<keyof T>,
+): { equal: true };
+export function compareArrayFields<T extends Dictionary<unknown>>(
+	obj1: T,
+	obj2: T,
+	nonOrderedFields: Array<keyof T>,
+	orderedFields: Array<keyof T>,
+): { equal: boolean; difference?: string[] } {
+	let equal = true;
+	const difference: string[] = [];
```
zvin: NIT: this is also an `Array<keyof T>`

src/compose/service.ts
CameronDiver: We only need to worry about order when the order has some meaning, and the docker daemon won't change it when reporting the current state. So the above shouldn't be an issue.

src/compose/utils.ts
CameronDiver: Ah good point, we can avoid a cast there too.

src/compose/service.ts
roman-mazur: My question was basically if order matters for `devices:`. And a quick test shows it does.
```
romans-mbp:resin-ui roman$ docker run --device=/dev/zero:/dev/test1 --device=/dev/urandom:/dev/test1  -it --rm busybox
/ # cat /dev/test1
^C
```

Here, `cat` does not produce any output, so it means `/dev/zero` got mapped to `/dev/test1` in the container. So order of `--device` matters, and container has to be recreated if elements there are swapped...

So, should we add `devices` to this array too?

CameronDiver: Ah sorry I misunderstood your original point - this is actually a different instance of the order being important, as something is being overwritten as it were.  This actually bears more thinking about, as I believe the volumes could also suffer from this problem, if you mount something on the same mountpoint. This is even more confusing as the volumes returned from docker can definitely be in a different order to being referenced. 

Given that exists as a problem now, I'd rather think about this as a whole with all of the possible ways that this could fail, and merge this now to at least support `dns` and `dnsSearch` properly, as this will not have the same problem.

PR #1069, opened 2019-08-21
Correctly generate network or volume creation steps

CameronDiver:
Change-type: patch


CameronDiver:
@roman-mazur yep, one of the step names wasn't changed, causing a bug when creating.

PR #1078, opened 2019-09-02
entry.sh: conform to shellcheck syntax

xginn8:
Change-type: patch


xginn8:
https://github.com/koalaman/shellcheck/wiki/SC2236

Issue #1080, opened 2019-09-09
Cancelling delayed update promises throws a TypeError: `TypeError: this.cancelDelay is not a function`


xginn8:
Reproducible by triggering many updates sequentially or parallelized, such as this simple SDK request:

```
await Promise.map(_.range(0, 100), i => sdk.models.device.envVar.set('79da5e2e1a92b26f566cfef14534021b', `MY_ENVVAR${i}`, String(i)))
```

Introduced here: https://github.com/balena-io/balena-supervisor/commit/7a42b6719a25275363b458b4e2fac07298b73b59#diff-0011437c96d80cf058a252cafc68dc0bR686

PR #1082, opened 2019-09-09
state update: set cancelDelay to true rather than calling the nonexis…

xginn8:
…tent fn

Closes: #1080
Change-type: patch

### DIFF
--- a/src/device-state.coffee
+++ b/src/device-state.coffee
@@ -683,7 +683,7 @@ module.exports = class DeviceState extends EventEmitter
 				@scheduledApply = { force, delay }
 				if isFromApi
 					# Cancel promise delay if call came from api to prevent waiting due to backoff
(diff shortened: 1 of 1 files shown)


src/device-state.coffee
```diff
@@ -694,7 +694,8 @@ module.exports = class DeviceState extends EventEmitter
 		@applyInProgress = true
 		new Promise (resolve, reject) =>
 			setTimeout(resolve, delay)
-			@cancelDelay = reject
+			if @cancelDelay
+				@cancelDelay = reject()
```
CameronDiver: This is a little odd, where we don't know whether the class var is a function or boolean, perhaps this whole workflow should be revisited.

xginn8:
sorry for the noise @CameronDiver, meant to abandon this

PR #1089, opened 2019-09-12
Convert some coffeescript tests to typescript

CameronDiver:
Change-type: patch


test/lib/prepare.ts
```diff
@@ -0,0 +1,56 @@
...
+		fs.writeFileSync(
+			'./test/data/config-apibinder-offline.json',
+			fs.readFileSync('./test/data/testconfig-apibinder-offline.json'),
+		);
+		return fs.writeFileSync(
+			'./test/data/config-apibinder-offline2.json',
+			fs.readFileSync('./test/data/testconfig-apibinder-offline2.json'),
+		);
+	} catch (e) {
+		/* ignore /*/
```
roman-mazur: should we remove these `catch`es? If preparation fails, we cannot be sure in test results. So why not to fail tests?

I would suggest removing them before merging. Otherwise, looks good.

test/lib/prepare.ts
```diff
@@ -0,0 +1,56 @@
...
+
+		fs.writeFileSync(
+			'./test/data/config-apibinder-offline.json',
+			fs.readFileSync('./test/data/testconfig-apibinder-offline.json'),
+		);
+		return fs.writeFileSync(
+			'./test/data/config-apibinder-offline2.json',
+			fs.readFileSync('./test/data/testconfig-apibinder-offline2.json'),
+		);
+	} catch (e) {
```
Page-:
```suggestion
	} catch {
```

PR #1090, opened 2019-09-12
Add support for container contracts

CameronDiver:
It's worth reviewing these commits one-by-one, as they tackle a bunch of slightly related stuff.

In terms of the container contracts, we look for a contract field on the service object, and before committing this to the supervisor database, we will reject the target state if any container contract is not fulfilled. The reason we do it here is that if we overwrite a current state with a release which cannot run, we won't be able to roll back.


src/application-manager.coffee
```diff
@@ -693,17 +716,32 @@ module.exports = class ApplicationManager extends EventEmitter
 					Promise.map appsForDB, (app) =>
 						@db.upsertModel('app', app, { appId: app.appId }, trx)
 				.then (appsForDB) ->
-					trx('app').where({ source }).whereNotIn('appId', _.map(appsForDB, 'appId')).del()
+					trx('app').where({ source }).whereNotIn('appId',
+						# Use apps here, rather than filteredApps, to
+						# avoid removing a release whose subsequent
```
hedss: `whose subsequent` what? Or `which subsequently`? Bit confused. (And 'which' not 'whose').

src/application-manager.coffee
```diff
@@ -693,17 +716,32 @@ module.exports = class ApplicationManager extends EventEmitter
...
 			.then =>
 				@proxyvisor.setTargetInTransaction(dependent, trx)
 
-		Promise.try =>
+		contractViolators = {}
+		Promise.props(contractsFulfilled).then (fulfilledContracts) ->
+			filteredApps = _.cloneDeep(apps)
+			_.each fulfilledContracts, ({ valid, unmetServices }, appId) ->
+				if not valid
+					console.log(apps[appId])
```
hedss: Debug left in, or should be logged?

src/device-state.coffee
```diff
@@ -17,7 +17,13 @@ validation = require './lib/validation'
 systemd = require './lib/systemd'
 updateLock = require './lib/update-lock'
 { singleToMulticontainerApp } = require './lib/migration'
-{ ENOENT, EISDIR, NotFoundError, UpdatesLockedError } = require './lib/errors'
+{
+	ContractViolationError,
```
hedss: `ContractViolationError` is now imported but doesn't seem to get used?

test/24-contracts.ts
```diff
@@ -0,0 +1,212 @@
...
+import supervisorVersion = require('../src/lib/supervisor-version');
+
+describe('Container contracts', () => {
+	// Because the supervisor version will change whenever the
+	// package.json will, we generate values which are above
+	// and below the current value, and use these to reason
+	// about the contract engine results
+	const supervisorVersionGreater = `${semver.major(supervisorVersion)! +
+		1}.0.0`;
+	const supervisorVersionLesser = `${semver.major(supervisorVersion)! - 1}.0.0`;
```
hedss: Good job we're not on version 0.9.0... ;-)

test/24-contracts.ts
```diff
@@ -0,0 +1,212 @@
...
+
+	before(async () => {
+		// We ensure that the versions we're using for testing
+		// are the same as the time of implementation, otherwise
+		// these tests could fail or succeed when they shouldn't
+		expect(await osRelease.getOSSemver(constants.hostOSVersionPath)).to.equal(
+			'2.0.6',
```
hedss: I think I'm confused here by this version number.
Edit: Ah, no I'm not, I've seen the end of this commit list now.

src/lib/os-release.ts
```diff
@@ -1,49 +1,61 @@
...
+): Promise<string | undefined> {
+	try {
+		const data = await getOSReleaseData(path);
+		const value = data[field];
+		if (value == null) {
+			log.warn(
+				`Field ${field} is not available in OS information file: ${path}`,
+			);
+		}
+		return data[field];
```
zvin: `return value`

src/lib/request.ts
```diff
@@ -53,9 +28,55 @@ type PromisifiedRequest = typeof requestLib & {
...
+		},
+	};
+
+	const resumableOpts = {
+		timeout: DEFAULT_REQUEST_TIMEOUT,
+		maxRetries: DEFAULT_REQUEST_RETRY_COUNT,
+		retryInterval: DEFAULT_REQUEST_RETRY_INTERVAL,
+	};
+
+	const requestHandle = requestLib.defaults(exports.requestOpts);
```
zvin: I don't see `requestOpts` being exported in this file.

src/api-binder.ts
```diff
@@ -580,9 +587,25 @@ export class APIBinder {
...
+					{},
+					'targetStateRejection',
+					false,
+				);
 			})
+			.tapCatch(
+				(e: unknown) =>
+					!(
+						e instanceof ContractValidationError ||
+						e instanceof ContractViolationError
```
zvin: If the above `.tapCatch` was a `.catch`, there would be no need for filtering on `e`'s class here.

test/data/etc/os-release
```diff
@@ -1,2 +1,3 @@
-PRETTY_NAME="Resin OS 2.0.6 (fake)"
+PRETTY_NAME="balenaOS 2.0.6 (fake)"
```
roman-mazur: 👍

src/api-binder.ts
```diff
@@ -408,9 +415,9 @@ export class APIBinder {
 				);
 			}
 			return {
-				id: id.value,
-				name: name.value,
-				value: value.value,
+				id: id.right,
+				name: name.right,
+				value: value.right,
```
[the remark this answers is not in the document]
CameronDiver: I'm extremely happy with it tbh, if you look at the config module, the typings it provides are incredible.

src/proxyvisor.coffee
```diff
@@ -595,10 +595,12 @@ module.exports = class Proxyvisor
 					"#{constants.proxyvisorHookReceiver}/v1/devices/"
 
 	sendUpdate: (device, timeout, endpoint) =>
-		request.putAsync "#{endpoint}#{device.uuid}", {
-			json: true
-			body: device.target
-		}
+		Bluebird.resolve(request.getRequestInstance())
```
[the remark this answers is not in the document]
CameronDiver: It's wrapping it in a Bluebird promise (rather than native) to maintain compat with other modules.

src/lib/contracts.ts
```diff
@@ -0,0 +1,122 @@
...
+		[osContract, supervisorContract].concat(
+			containers.map(c => new Contract(c)),
+		),
+	);
+
+	const solution = blueprint.reproduce(universe);
+
+	if (solution.length > 1) {
+		throw new InternalInconsistencyError(
+			'More than one solution available for container contracts when only one is expected!',
```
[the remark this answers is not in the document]
CameronDiver: How could we get multiple OS versions?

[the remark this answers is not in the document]
roman-mazur: We can't :)
I'm just theorizing to confirm I correctly understand how this snippet works.

src/proxyvisor.coffee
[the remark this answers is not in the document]
roman-mazur: Got it, thanks.

src/lib/contracts.ts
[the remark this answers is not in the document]
CameronDiver: Oh I sorry I misunderstood your question. Because of the `1+` in the container membership of the blueprint, we can rely on this. I figured it was worth checking against because I accidentally made a typo in development and this occurred. I'd rather see this in development, rather than debug why suddenly all contracts are failing :P

src/api-binder.ts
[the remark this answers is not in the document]
roman-mazur: yeah, I agree on typings, but using Either for errors handling looks weird to me.
I even understand it from Haskell, and it indeed makes sense to keep the successful (right) value in `right` :)
But `result` type in Rust, for instance, is much more expressive for errors handling.
It's close to the debates why some languages don't have Tuple or Pair types in std SDKs.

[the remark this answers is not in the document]
CameronDiver: I'm not sure I follow, using `io-ts` (and by extension `fp-ts`) means we don't get a choice, as this is the return type of the module `decode` function. Is this what you're saying?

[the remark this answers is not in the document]
roman-mazur: It's not a call for action but thoughts on the code expressiveness in a written form.
We are not throwing `io-ts` out in foreseeable future, so we keep using Either too. However, I think we should keep its usage strictly in the files that perform validation and forbid it to cross the boundary of a module it's used in.

src/lib/request.ts
CameronDiver: good catch!

src/api-binder.ts
CameronDiver: The thing is we still want these values to be raised up to the caller (device state) which will cause things to back off. I couldn't think of a nice way to do it really.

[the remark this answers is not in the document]
CameronDiver: Good point I'll just handle it in one :+1:

src/lib/contracts.ts
```diff
@@ -0,0 +1,122 @@
...
+				const found = _.find(children, child => {
+					return _.isEqual((child as any).raw, contract);
+				});
+				if (found == null) {
+					return serviceName;
+				}
+				return;
+			})
+			.filter(n => n != null)
+			.value() as string[];
```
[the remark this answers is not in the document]
CameronDiver: Not really, we need to compare a certain field of the return array items against the value of the inputs, but return the keys themselves.

src/lib/os-release.ts
```diff
@@ -1,49 +1,60 @@
...
+// Retrieve the data for the OS once only per path
+const getOSReleaseData = _.memoize(
+	async (path: string): Promise<Dictionary<string>> => {
+		const releaseItems: Dictionary<string> = {};
+		try {
+			const releaseData = await fs.readFile(path, 'utf-8');
+			const lines = releaseData.split('\n');
+
+			for (const line of lines) {
+				const [key, ...values] = line.split('=');
```
roman-mazur:
```suggestion
				const [key, quatedValue] = line.split('=', 2);
```

This would remove the necessity to join values 2 lines below:
```
const value = _.trim(quatedValue.replace(...))
```

src/lib/request.ts
```diff
@@ -53,9 +28,55 @@ type PromisifiedRequest = typeof requestLib & {
 	) => Bluebird<any>;
 };
 
-const requestHandle = requestLib.defaults(exports.requestOpts);
+const getRequestInstances = once(async () => {
+	// Generate the user agents with out versions
```
roman-mazur:
```suggestion
	// Generate the user agents with our versions
```
otherwise, it does not make much sense...

PR #1092, opened 2019-09-16
Correctly handle partial IPAM configurations for networks

CameronDiver:
Currently, the driver can be missing from the IPAM configuration if it's not provided in the docker-compose. This would cause the supervisor to think the network needed to be recreated, and would stop all associated services. This patch fixes the parsing of incomplete configuration to make sure all fields are set.

Change-type: patch


roman-mazur:
Looks good.

PR #1095, opened 2019-09-23
Unify API errors processing

roman-mazur:
With this change, we define a custom error handler as express middleware
which renders 503 error with JSON response that includes status and message
fields.

The handler also logs the error, so the stack can be inspected in supervisor
logs. It's also a point where we can report the error to analytics services.

This removes a bunch of error handlers written in every request handler
function. Behaviour should remain unchanged except the fact that
`/healthy` endpoint now returns 503 in case of failure instead of 500.

Implemented while investigating problems with `GET /target-state` endpoint.

Change-type: patch

### DIFF
--- a/src/device-api/v1.coffee
+++ b/src/device-api/v1.coffee
@@ -9,7 +9,7 @@ exports.createV1Api = (router, applications) ->
 
 	{ eventTracker } = applications
 
-	router.post '/v1/restart', (req, res) ->
+	router.post '/v1/restart', (req, res, next) ->
 		appId = checkInt(req.body.appId)
 		force = checkTruthy(req.body.force)
 		eventTracker.track('Restart container (v1)', { appId })
@@ -18,10 +18,9 @@ exports.createV1Api = (router, applications) ->
 		doRestart(applications, appId, force)
 		.then ->
 			res.status(200).send('OK')
-		.catch (err) ->
-			res.status(503).send(err?.message or err or 'Unknown error')
+		.catch(next)
 
-	v1StopOrStart = (req, res, action) ->
+	v1StopOrStart = (req, res, next, action) ->
 		appId = checkInt(req.params.appId)
 		force = checkTruthy(req.body.force)
 		if !appId?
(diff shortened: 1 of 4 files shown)


src/device-api/v1.coffee
```diff
@@ -47,16 +46,15 @@ exports.createV1Api = (router, applications) ->
 					return service
 			.then (service) ->
 				res.status(200).json({ containerId: service.containerId })
-		.catch (err) ->
-			res.status(503).send(err?.message or err or 'Unknown error')
+		.catch(next)
 
-	router.post '/v1/apps/:appId/stop', (req, res) ->
-		v1StopOrStart(req, res, 'stop')
+	router.post '/v1/apps/:appId/stop', (req, res, next) ->
```
Page-: I think these would be better if you used a partial like
```suggestion
	router.post('/v1/apps/:appId/stop', createV1StopOrStart('start'))
```
or
```suggestion
	router.post('/v1/apps/:appId/stop', _.partial(v1StopOrStart, _, _, _, 'start'))
```

src/device-api/v2.ts
```diff
@@ -85,286 +74,254 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
+				.catch(next);
 		},
 	);
 
 	router.post(
 		'/v2/applications/:appId/restart-service',
-		(req: Request, res: Response) => {
-			return handleServiceAction(req, res, 'restart');
+		(req: Request, res: Response, next: NextFunction) => {
+			return handleServiceAction(req, res, next, 'restart');
```
Page-: Same for these, it'd be nice to use a partial to reduce duplication

roman-mazur: Thanks, good point.

roman-mazur: Addressed. PTAL again.

src/device-api/v1.coffee
```diff
@@ -47,16 +46,14 @@ exports.createV1Api = (router, applications) ->
...
-			res.status(503).send(err?.message or err or 'Unknown error')
+		.catch(next)
 
-	router.post '/v1/apps/:appId/stop', (req, res) ->
-		v1StopOrStart(req, res, 'stop')
+	createV1StopOrStartHandler = (action) -> _.partial(v1StopOrStart, _, _, _, action)
 
-	router.post '/v1/apps/:appId/start', (req, res) ->
-		v1StopOrStart(req, res, 'start')
+	router.post '/v1/apps/:appId/stop', createV1StopOrStartHandler('stop')
```
CameronDiver: Please add parens for single line functions calls in coffeescript

roman-mazur: Done. But why doesn't our linter require this?

roman-mazur:
@balena-ci retest, please

roman-mazur:
aarch64-debug build hangs on making sqlite... restarted

CameronDiver:
@roman-mazur it's not hanging, it just takes a bunch of time due to qemu.

roman-mazur:
aarch64-debug took 47:36 - my personal record for CI builds :)

src/device-api/v1.coffee
Page-: You can either enforce parens in all cases or none @roman-mazur, what we decided on for coffeescript was the more nuanced "always for functions without a callback/single line callback, optional/never for functions with a multiline callback" - overall that makes the best use of significant whitespace in coffeescript whilst avoid a lot of the confusion around things like `a b c, d, e, f (g h), (i j)` where things can get crazy with various interactions of parens and whitespace and it's hard to parse without really thinking about it (that example compiles to `a(b(c, d, e, f(g(h), i(j))))`)
Either way we're steadily moving to typescript which removes the coffeescript linting capabilities as an issue :)

PR #1099, opened 2019-09-26
Add a cache around the database application target state

CameronDiver:
This should stop OOM errors when the application containers are logging excessively in local mode.

Change-type: minor

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -17,7 +17,7 @@ updateLock = require './lib/update-lock'
 { ContractViolationError, ContractValidationError, NotFoundError } = require './lib/errors'
 { pathExistsOnHost } = require './lib/fs-utils'
 
-{ ApplicationTargetStateWrapper } = require './target-state'
+{ TargetStateAccessor } = require './target-state'
 
 { ServiceManager } = require './compose/service-manager'
 { Service } = require './compose/service'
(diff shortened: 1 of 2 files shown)


src/target-state.ts
```diff
@@ -0,0 +1,80 @@
...
+ * receiving of target state. Because the target state can
+ * only be set from a single place, but several workflows
+ * rely on getting the target state at one point or another,
+ * we cache the values using this class. Accessing the
+ * database is inherently expensive, and for example the
+ * local log backend accesses the target state for every log
+ * line. This can very quickly cause serious memory problems
+ * and database connection timeouts.
+ */
+export class ApplicationTargetStateWrapper {
```
roman-mazur: It's a small enough and isolated component. Can be covered with tests.

roman-mazur: On the naming, my suggestion is "TargetStateAccessor".

src/application-manager.coffee
```diff
@@ -79,6 +81,8 @@ module.exports = class ApplicationManager extends EventEmitter
 		@_targetVolatilePerImageId = {}
 		@_containerStarted = {}
 
+		@targetStateWrapper = new TargetStateAccessor(this, @config, @db)
```
roman-mazur: May be worth renaming this property too.

roman-mazur:
@CameronDiver it's a bit weird, but the merged commit here contains only a rename of the cache implementation. https://github.com/balena-io/balena-supervisor/commit/12f68634d8d1afefebbe95b229fb3003dab8bdde

Did we merge the actual implementation as a part of some other PR?

CameronDiver:
Wow, I have no idea what happened here. Something obviously got lost in the rebase but quite frankly I'm at a loss now. Great catch @roman-mazur

roman-mazur:
It got merged as a part of #1090 (7239b93f4a04851e4adbf1478af95c4a337aaadd).

PR #1106, opened 2019-10-02
event-tracker: Use std mixpanel types

roman-mazur:
Custom type definitions are removed for mixpanel module since they are embedded
into their npm package.

This change also adds an extra check for events tracking logic in a hope we can
investigate some weird events we see in the analytics system.

Change-type: patch

### DIFF
--- a/src/event-tracker.ts
+++ b/src/event-tracker.ts
@@ -2,8 +2,7 @@ import * as Bluebird from 'bluebird';
 import mask = require('json-mask');
 import * as _ from 'lodash';
 import * as memoizee from 'memoizee';
-
-import Mixpanel = require('mixpanel');
+import * as mixpanel from 'mixpanel';
 
 import { ConfigType } from './config';
 import log from './lib/supervisor-console';
@@ -35,7 +34,7 @@ const mixpanelMask = [
 
 export class EventTracker {
 	private defaultProperties: EventTrackProperties | null;
-	private client: any;
+	private client: mixpanel.Mixpanel | null;
 
 	public constructor() {
 		this.client = null;
@@ -54,10 +53,10 @@ export class EventTracker {
 				uuid,
 				supervisorVersion,
 			};
-			if (unmanaged || mixpanelHost == null) {
+			if (unmanaged || mixpanelHost == null || mixpanelToken == null) {
 				return;
 			}
-			this.client = Mixpanel.init(mixpanelToken, {
+			this.client = mixpanel.init(mixpanelToken, {
 				host: mixpanelHost.host,
 				path: mixpanelHost.path,
 			});
@@ -94,7 +93,9 @@ export class EventTracker {
 			// Call this function at maximum once every minute
 			return _.throttle(
 				properties => {
-					this.client.track(event, properties);
+					if (this.client != null) {
+						this.client.track(event, properties);
+					}
 				},
 				eventDebounceTime,
 				{ leading: true },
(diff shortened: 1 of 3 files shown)


src/event-tracker.ts
```diff
@@ -94,7 +93,20 @@ export class EventTracker {
...
-					this.client.track(event, properties);
+					if (this.client != null) {
+						if (!properties.uuid) {
+							// In some weird cases we see events that seem to be submitted by the supervisor
+							// but not containing any default properties.
+							// This code is supposed to clarify if it actually happens because of the supervisor.
+							const availableData = _.merge({}, this.defaultProperties, {
+								details: 'uuid is missing on event properties',
+							});
+							this.client.track(`Problematic ${event}`, availableData);
```
roman-mazur: @gelbal this way we'll at least have a prove it's coming from supervisor if it's reproduced in newer versions

@CameronDiver it's related to the fact we see "Supervisor start" events without uuid reported. It may be coming from some old supervisor versions. In our current version, I could find any real race that would lead to firing an event without default properties being set. But just to make sure, I'm adding this. What do you think?

gelbal: Sounds good Roman. Talking in general, for such debugging cases, is it possible to send events only to Sentry and not to Mixpanel/Amplitude?

If possible, I'd prefer not to pollute the Mixpanel/Amplitude prod schema for debug reasons. Then of course you could argue that the schema is already "polluted" with an unexpected event : )

roman-mazur: The problem here is that Sentry is not integrated into the supervisor. Hence, we can only write to the logs or send via mixpanel client.

src/event-tracker.ts
```diff
@@ -57,7 +56,7 @@ export class EventTracker {
 			if (unmanaged || mixpanelHost == null) {
 				return;
 			}
-			this.client = Mixpanel.init(mixpanelToken, {
+			this.client = mixpanel.init(mixpanelToken as string, {
```
CameronDiver: This cast shouldn't be required, the `ConfigType<'mixpanelToken'>` should resolve to a string.

src/event-tracker.ts
CameronDiver: I've not tested this at all, but there strictly is some time when the supervisor starts up and it doesn't have a UUID, e.g. before provision. Is this not the event you're investigating?

In general, I'm for these kinds of changes, but I'd like to understand a little more about this stray event you're referencing. Furthermore, I'm wondering if the supervisor start event is even useful, but maybe that is a bigger conversation.

roman-mazur: An unprovisioned device is a good point and must explain this. Thanks!

src/event-tracker.ts
roman-mazur: it's `string | undefined`

src/event-tracker.ts
roman-mazur: Removed this code.

roman-mazur:
@balena-ci retest please, rebased on master

src/event-tracker.ts
CameronDiver: I'd prefer we check for thay here then https://github.com/balena-io/balena-supervisor/pull/1106/files#diff-aaf08893d3e6fd02df6a69b14f239847R56

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest please again

roman-mazur:
@balena-ci retest please

roman-mazur:
Waiting for github events to get back to normal so the checks are completed.

roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest

PR #1127, opened 2019-10-29
Support optional containers based on their contract

CameronDiver:
Change-type: minor

### DIFF
--- a/src/lib/contracts.ts
+++ b/src/lib/contracts.ts
@@ -22,6 +22,7 @@ export async function containerContractsFulfilled(
 	valid: boolean;
 	unmetServices: string[];
 	fulfilledServices: string[];
+	unmetAndOptional: string[];
 }> {
 	const containers = _(serviceContracts)
 		.map('contract')
@@ -76,6 +77,7 @@ export async function containerContractsFulfilled(
 			valid: false,
 			unmetServices: _.keys(serviceContracts),
 			fulfilledServices: [],
+			unmetAndOptional: [],
 		};
 	}
 
@@ -90,6 +92,7 @@ export async function containerContractsFulfilled(
 			valid: true,
 			unmetServices: [],
 			fulfilledServices: _.keys(serviceContracts),
+			unmetAndOptional: [],
 		};
 	} else {
 		// If we got here, it means that at least one of the
@@ -112,15 +115,18 @@ export async function containerContractsFulfilled(
 			},
 		);
 
-		const valid = !_.some(
+		const [unmetAndRequired, unmetAndOptional] = _.partition(
 			unfulfilledServices,
-			svcName => !serviceContracts[svcName].optional,
+			serviceName => {
+				return !serviceContracts[serviceName].optional;
+			},
 		);
 
 		return {
-			valid,
+			valid: unmetAndRequired.length === 0,
 			unmetServices: unfulfilledServices,
 			fulfilledServices,
+			unmetAndOptional,
 		};
 	}
 }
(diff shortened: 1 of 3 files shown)


src/lib/contracts.ts
```diff
@@ -75,23 +89,45 @@ export async function containerContractsFulfilled(
 	});
 
 	if (children.length === containers.length) {
-		return { valid: true, unmetServices: [] };
+		console.log(`Got no problems with this contract setup`);
```
richbayliss: debug output?

src/lib/contracts.ts
```diff
@@ -75,23 +89,45 @@ export async function containerContractsFulfilled(
...
+					null
+				);
+			},
+		);
+
+		const nonOptional = _.filter(
+			unfulfilledServices,
+			svcName => !serviceContracts[svcName].optional,
+		);
+		const valid = nonOptional.length === 0;
```
Page-:
```suggestion
		const valid = _.some(
			unfulfilledServices,
			svcName => !serviceContracts[svcName].optional,
		);
```

src/lib/contracts.ts
CameronDiver: haha yes, good catch!

src/lib/contracts.ts
```diff
@@ -75,23 +89,45 @@ export async function containerContractsFulfilled(
...
+					return true;
 				}
-				return;
-			})
-			.filter(n => n != null)
-			.value() as string[];
-		return { valid: false, unmetServices };
+				// Did we find the contract in the generated state?
+				return (
+					_.find(children, child => _.isEqual((child as any).raw, contract)) !=
```
Page-:
```suggestion
					_.some(children, child => _.isEqual((child as any).raw, contract))
```

src/lib/contracts.ts
```diff
@@ -53,7 +61,9 @@ export async function containerContractsFulfilled(
 
 	universe.addChildren(
 		[osContract, supervisorContract].concat(
```
Page-:
```suggestion
		[osContract, supervisorContract, ...containers.map(c => new Contract(c))]
```

src/lib/contracts.ts
```diff
@@ -13,13 +13,21 @@ import supervisorVersion = require('./supervisor-version');
...
-	const containers = _.values(serviceContracts);
+): Promise<{
+	valid: boolean;
+	unmetServices: string[];
+	fulfilledServices: string[];
+}> {
+	const containers = _(serviceContracts)
+		.values()
+		.map('contract')
+		.filter(c => c != null)
```
Page-:
```suggestion
		.compact()
```
(and with this I think you can drop some `!`s below)

CameronDiver: nice! (the `!` dropping)

Page-: The other way to do it (eg with native filter) is with a type guard, ie
```ts
arr.filter((x): x is NonNullable<typeof x> => x != null)

File: src/lib/contracts.ts [hunk 6]
```diff
@@ -13,13 +13,21 @@ import supervisorVersion = require('./supervisor-version');
...
 	serviceContracts: ServiceContracts,
-): Promise<{ valid: boolean; unmetServices: string[] }> {
-	const containers = _.values(serviceContracts);
+): Promise<{
+	valid: boolean;
+	unmetServices: string[];
+	fulfilledServices: string[];
+}> {
+	const containers = _(serviceContracts)
+		.values()
```
Page-: There's no need for this `.values()` since lodash can map on an object directly

Issue #1074, opened 2019-08-28
The supervisor should expose whether it's running on a dev or prod device


CameronDiver:

iamhosseindhv:
+1.

(I never do +1 as a comment but also couldn't find any other way to bring attention to this issue.)

klutchell:
We should note that prod and dev devices have been combined into the same image in the latest meta-balena releases (https://github.com/balena-os/meta-balena/pull/2227).

I don't think we've updated the documentation yet, but this OS change will impact how we implement this feature.

PR #1130, opened 2019-11-03
env: expose osVariant as environment variable

xginn8:
Connects-to: #1074
Change-type: minor


src/lib/os-release.ts
```diff
@@ -51,7 +51,7 @@ export async function getOSVersion(path: string): Promise<string | undefined> {
 	return getOSReleaseField(path, 'PRETTY_NAME');
 }
 
-export function getOSVariant(path: string): Promise<string | undefined> {
+export async function getOSVariant(path: string): Promise<string | undefined> {
```
CameronDiver: No need for async here, as we're not awaiting the promise.

PR #1132, opened 2019-11-06
Extract loadTargetFromFile into it's own module

CameronDiver:
This is the start of some work on refactoring the device-state module, starting with removing some non-related code into other places, and then refactoring the class itself.

Also define some more types for interacting with target states, and upgrade typescript.


src/application-manager.d.ts
```diff
@@ -6,7 +6,7 @@ import { EventTracker } from './event-tracker';
 import { Logger } from './logger';
 import { DeviceApplicationState } from './types/state';
 
-import Images from './compose/images';
+import Images, { Images } from './compose/images';
```
zvin: Oo

CameronDiver: uhh, supposed to be `Image` (this needs refactored too). It's weird that this even builds though, much less passes the tests...

src/device-state/preload.ts
```diff
@@ -0,0 +1,114 @@
...
+): Promise<void> {
+	log.info('Attempting to load any preloaded applications');
+	if (!appsPath) {
+		appsPath = constants.appsJsonPath;
+	}
+
+	try {
+		const content = await fs.readFile(appsPath, 'utf8');
+
+		// It's either a tgarget state
```
zvin: tgarget -> target

Also:

There are two types of people in this world:
1. Those who can extrapolate from incomplete data.

src/device-state/preload.ts
```diff
@@ -0,0 +1,115 @@
...
+		const localState = { local: { name: '', ...preloadState } };
+
+		await deviceState.setTarget(localState);
+
+		log.success('Preloading complete');
+		if (stateFromFile.pinDevice) {
+			// Multi-app warning!
+			// The following will need to be changed once running
+			// multiple applications is possible
+			log.debug('Device will be pinned');
```
zvin: NIT: the debug message could be inside the if below

src/lib/migration.ts
```diff
@@ -67,3 +69,18 @@ export function singleToMulticontainerApp(app: Dictionary<any>): Application {
 	};
 	return newApp;
 }
+
+export async function convertLegacyAppsJson(
```
zvin: This function does not need to be `async`.

CameronDiver:
Tested on an rpi3 and works great.

PR #1133, opened 2019-11-06
Extract normaliseLegacy as normalise to migration module

telphan:
Continuing work to refactor device-state module. 

Change-type: patch
Signed-off-by: Theodor Gherzan <theodor@balena.io>


src/lib/migration.ts
```diff
@@ -84,3 +94,167 @@ export async function convertLegacyAppsJson(
 	const apps = _.keyBy(_.map(appsArray, singleToMulticontainerApp), 'appId');
 	return { apps, config: deviceConfig };
 }
+
+export async function normalise(
```
CameronDiver: Let's call this `normalizeLegacyDb`

PR #1135, opened 2019-11-07
Hide init method in configJson

roman-mazur:
ConfigJson users now don't need to worry about initialization.
It's done lazily on the first public method call.

Implementation is also rewritten to follow async/await pattern.

Change-type: patch


roman-mazur:
@CameronDiver this is more like a suggestion...

As you know, I'm investigating that case when event tracker can submit events without device UUID. I didn't find how it's possible with recent versions of the supervisor since `config.init` is called early on startup and is awaited to complete.
I also identified that those events usually come from devices that belong to applications that have older resinOS versions.

While looking at config implementation, I got a few suggestions that I'm now submitting as separate PRs.

roman-mazur:
@balena-ci retest please

Issue #1109, opened 2019-10-04
Does not seem to be working for purging the data volume


heri16:
volume "resin-data" not cleared even during doPurge with supervisor 10.3.8

https://github.com/balena-io/balena-supervisor/blob/19582a1551cac8fa58d14d536c0fabc4e157ec72/src/device-api/common.coffee#L47

roman-mazur:
Related to #1083.

CameronDiver:
@heri16 could you clarify what you're seeing? Specifically, a way to reproduce this would be helpful.

heri16:
When deploying this repo
https://github.com/zoobc/zoobc-labnet

The services are restarted and recreated, but resin-data is not removed or purged.

heri16:
@CameronDiver 

Here is my docker-compose file: https://github.com/zoobc/zoobc-labnet/blob/master/docker-compose.yml

Here is my logs when purge is triggered:
```bash
[Logs]    [10/15/2019, 4:39:19 PM] [n2] INFO[0891] ZOOBC Shutdown                               
[Logs]    [10/15/2019, 4:39:19 PM] [main] INFO[0891] ZOOBC Shutdown                               
[Logs]    [10/15/2019, 4:39:20 PM] Service exited 'main sha256:5ca8d5d37ae5e6c5bae0b88cd87cff34da6f78352408851ea45ba6e9a9c1483f'
[Logs]    [10/15/2019, 4:39:20 PM] Killed service 'main sha256:5ca8d5d37ae5e6c5bae0b88cd87cff34da6f78352408851ea45ba6e9a9c1483f'
[Logs]    [10/15/2019, 4:39:21 PM] Service exited 'grcpweb sha256:c6a1cbcb06c243bd30008d090abd50be98e8b879296936bd596c3626b95e44f6'
[Logs]    [10/15/2019, 4:39:21 PM] Service exited 'n2 sha256:5ca8d5d37ae5e6c5bae0b88cd87cff34da6f78352408851ea45ba6e9a9c1483f'
[Logs]    [10/15/2019, 4:39:21 PM] Killed service 'grcpweb sha256:c6a1cbcb06c243bd30008d090abd50be98e8b879296936bd596c3626b95e44f6'
[Logs]    [10/15/2019, 4:39:21 PM] Killed service 'n2 sha256:5ca8d5d37ae5e6c5bae0b88cd87cff34da6f78352408851ea45ba6e9a9c1483f'
[...]

PR #1137, opened 2019-11-10
Check if appId is a number before purging

mwohlert:
Check appId param before calling doPurge to make sure it'a a number, because otherwise the filter result won't be as expected.


mwohlert:
Fixes #1109

CameronDiver:
@balena-ci retest

PR #1141, opened 2019-11-13
Compare Linux4Tegra versions in contracts

CameronDiver:
We now allow matching on a l4t version if it exists, and don't allow OS matching.

The upgrade of prettier also caused a bunch of files to change their formattiing, so the only commit worth looking at is https://github.com/balena-io/balena-supervisor/pull/1141/commits/e2785e9d371fe7aaf42db854a041ab1bf59ddd8e


src/lib/os-release.ts
```diff
@@ -58,3 +58,23 @@ export function getOSVariant(path: string): Promise<string | undefined> {
 export function getOSSemver(path: string): Promise<string | undefined> {
 	return getOSReleaseField(path, 'VERSION');
 }
+
+const L4T_REGEX = /^.*-l4t-r(\d+).(\d+).*$/;
```
telphan: I do not see any need to capture two groups here. You could just write this regex as:
`/^.*-l4t-r(\d+\.\d+).*$/`
This can simplify the regex operation and simplify the code at line 75.
Also make sure to escape `.` to express the version correctly.

src/lib/contracts.ts
```diff
@@ -145,14 +181,76 @@ const contractObjectValidator = t.type({
...
+			const svc = app.services[svcId];
+
+			if (svc.contract) {
+				try {
+					await validateContract(svc.contract);
+
+					serviceContracts[svc.serviceName] = {
+						contract: svc.contract,
+						optional:
+							checkTruthy(svc.labels?.['io.balena.features.optional']) || false,
```
telphan: So impressive :heart_eyes:

src/lib/os-release.ts
```diff
@@ -58,3 +58,23 @@ export function getOSVariant(path: string): Promise<string | undefined> {
...
+export async function getL4tVersion(): Promise<string | undefined> {
+	// We call `uname -r` on the host, and look for l4t
+	try {
+		const [stdout] = (await child_process.exec('uname -r')).map(b =>
```
telphan: NIT The map over the second element of the array (i.e. stderr) is dead code.

PR #1148, opened 2019-11-21
Fix wrong table name in 9.15.x branch

roman-mazur:
Change-type: patch
Signed-off-by: Roman Mazur <roman@balena.io>

PR #1149, opened 2019-11-21
Add regression test for app manager init

roman-mazur:
If we had it earlier, it would have prevented #1148.
When init is called, withImagesFromDockerAndDB in images.ts is called as well.

Change-type: patch


roman-mazur:
@balena-ci retest please

roman-mazur:
@balena-ci retest please

PR #1152, opened 2019-11-28
Handle cancelDelay not being a function

CameronDiver:
Change-type: patch
Closes: #1080

### DIFF
--- a/src/device-state.coffee
+++ b/src/device-state.coffee
@@ -469,8 +469,10 @@ module.exports = class DeviceState extends EventEmitter
 			if !@scheduledApply? || (isFromApi && @cancelDelay)
 				@scheduledApply = { force, delay }
 				if isFromApi
-					# Cancel promise delay if call came from api to prevent waiting due to backoff
-					@cancelDelay()
+					# Cancel promise delay if call came from api to
+					# prevent waiting due to backoff (and if we've
+					# previously setup a delay)
+					@cancelDelay?()
 			else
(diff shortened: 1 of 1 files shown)


src/device-state.coffee
```diff
@@ -469,8 +469,11 @@ module.exports = class DeviceState extends EventEmitter
 			if !@scheduledApply? || (isFromApi && @cancelDelay)
 				@scheduledApply = { force, delay }
 				if isFromApi
-					# Cancel promise delay if call came from api to prevent waiting due to backoff
-					@cancelDelay()
+					# Cancel promise delay if call came from api to
+					# prevent waiting due to backoff (and if we've
+					# previously setup a delay)
+					if _.isFunction(@cancelDelay)
+						@cancelDelay()
```
Page-: @cancelDelay?()

CameronDiver: I'd rather make sure it was a function explicitly to be honest.

Page-: @cancelDelay does explicitly check it's a function and only calls it if it is, you can check the generated code to verify. If you're thinking it cutie be something other than null or a function though that should definitely be looked into as it's probably a much deeper bug

CameronDiver: TIL about `?()` being specifically a function check. Yeah it's certainly not ideal, but I've made big inroads into converting this file to typescript (and removing the mess that this `cancelDelay` setup is)

PR #1154, opened 2019-12-02
volumes: Allow the use of the 'driver' setting in volumes

richbayliss:
A compose file can now contain a volume which uses a different driver
from the default one; local.

Change-type: patch

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -666,11 +666,12 @@ module.exports = class ApplicationManager extends EventEmitter
 
 				Promise.map(JSON.parse(app.services), (service) => @createTargetService(service, configOpts))
 				.then (services) =>
-					# If a named volume is defined in a service, we add it app-wide so that we can track it and purge it
+					# If a named volume is defined in a service but NOT in the volumes of the compose file, we add it app-wide so that we can track it and purge it
+					# !! DEPRECATED, WILL BE REMOVED IN NEXT MAJOR RELEASE !!
 					for s in services
 						serviceNamedVolumes = s.getNamedVolumes()
 						for name in serviceNamedVolumes
-							volumes[name] = @createTargetVolume(name, app.appId, { labels: {} })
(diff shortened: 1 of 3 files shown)


src/compose/volume.ts
```diff
@@ -17,34 +17,28 @@ export interface VolumeConstructOpts {
...
 	private logger: Logger;
 	private docker: Docker;
 
 	private constructor(
-		name: string,
-		appId: number,
-		config: VolumeConfig,
+		public name: string,
+		public appId: number,
+		public config: VolumeConfig,
```
richbayliss: Simplified the fact these are passed in, and are publicly exposed on the instance 👍

src/application-manager.coffee
```diff
@@ -666,11 +666,12 @@ module.exports = class ApplicationManager extends EventEmitter
 
 				Promise.map(JSON.parse(app.services), (service) => @createTargetService(service, configOpts))
 				.then (services) =>
-					# If a named volume is defined in a service, we add it app-wide so that we can track it and purge it
+					# If a named volume is defined in a service but NOT in the volumes of the compose file, we add it app-wide so that we can track it and purge it
+					# !! TO BE DEPRECATED IN NEXT MAJOR RELEASE !!
```
CameronDiver: super super NIT: `Deprecated, will be removed in next major release`

src/compose/volume.ts
```diff
@@ -54,6 +48,7 @@ export class Volume {
 		// Convert the docker inspect to the config
 		const config: VolumeConfig = {
 			labels: inspect.Labels || {},
+			driver: inspect.Driver,
```
CameronDiver: I can't fully remember, but `local` is *always* returned by the daemon when nothing is passed, right?

richbayliss: every volume has a driver; local is just the default one, so I would expect Dockerode to return local here. It's not an optional value in the types...

CameronDiver: Yeah I figured, just wanted to make sure. The tests will catch this anyway.

PR #1163, opened 2019-12-12
Always do a standard docker download on first pull

CameronDiver:
Change-type: patch

### DIFF
--- a/src/application-manager.coffee
+++ b/src/application-manager.coffee
@@ -906,11 +906,28 @@ module.exports = class ApplicationManager extends EventEmitter
 							if _.every(nextSteps, { action: 'noop' })
 								volumePromises.push(@removeAllVolumesForApp(checkInt(appId)))
 			newDownloads = _.filter(nextSteps, (s) -> s.action == 'fetch').length
+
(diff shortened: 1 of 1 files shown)


src/application-manager.coffee
```diff
@@ -909,11 +909,18 @@ module.exports = class ApplicationManager extends EventEmitter
...
-				downloadsToBlock = downloading.length + newDownloads - constants.maxDeltaDownloads
-				while downloadsToBlock > 0
-					_.pull(nextSteps, _.find(nextSteps, action: 'fetch'))
-					downloadsToBlock -= 1
+
+			if !ignoreImages and delta and newDownloads > 0 and availableImages > 0
+				# Check that this is not the first pull for an
+				# application, as we want to download all images then
+				availableFromThisApp = _.filter availableImages, (img) ->
+					img.appId = appId
```
dfunckt: You're doing an assignment here, when you really meant:

```suggestion
					img.appId is appId
```

CameronDiver: woah, good catch

PR #1168, opened 2019-12-18
Report the correct error when we cannot remove a network

CameronDiver:
Change-type: patch

### DIFF
--- a/src/compose/network.ts
+++ b/src/compose/network.ts
@@ -193,7 +193,7 @@ export class Network {
 				.getNetwork(Network.generateDockerName(this.appId, this.name))
 				.remove(),
 		).tapCatch(error => {
-			this.logger.logSystemEvent(logTypes.createNetworkError, {
+			this.logger.logSystemEvent(logTypes.removeNetworkError, {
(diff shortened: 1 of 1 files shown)


src/compose/network.ts
```diff
@@ -193,7 +193,7 @@ export class Network {
 				.getNetwork(Network.generateDockerName(this.appId, this.name))
 				.remove(),
 		).tapCatch(error => {
-			this.logger.logSystemEvent(logTypes.createNetworkError, {
+			this.logger.logSystemEvent(logTypes.removeNetworkError, {
 				network: { name: this.name, appId: this.appId },
```
richbayliss: is this right?

CameronDiver: How do you mean?

richbayliss: Ok - reason for the false flag; the PR title is "... cannot remove a volume" and this is all about networks...

CameronDiver: Ah yeah, accidentally put volume instead of network in the commit message - changed now. Nice catch!

Issue #1171, opened 2019-12-31
Take into account mixed device-type fleets when applying default configuration parameters


chrisys:
There exists a reproducible scenario where adding a Pi 4 to a Pi 3 app with existing Pi 3 devices causes that Pi 4 device to fail to boot more than once.

I was able to reproduce this in my Pi 3, Pi-hole application, which already had the `RESIN_HOST_CONFIG_dtoverlay` device configuration variable set. I added a Pi 4 device, it boots the first time and is added to the application correctly, but as soon as the supervisor applies the configuration from the application, it appears as though the `vc4-fkms-v3d` dtoverlay is removed from the default device `config.txt` file, the device then attempts to reboot and can't because this critical overlay has been removed.

I also attempted to reproduce this in a clean application with no parameters set, but in that scenario the `RESIN_HOST_CONFIG_dtoverlay` is correctly set to `vc4-fkms-v3d` for the Pi 4 device even when it's added alongside an existing Pi 3 device.

[Related discussion](https://www.flowdock.com/app/rulemotion/r-resinos/threads/JSfQzO8ZT12EOfUg90QtC55lbPO)

PR #1172, opened 2020-01-08
Ensure that we always have the pi4 overlay assigned

CameronDiver:
Change-type: patch
Closes: #1171

### DIFF
--- a/src/device-config.ts
+++ b/src/device-config.ts
@@ -337,18 +337,14 @@ export class DeviceConfig {
 		target: Dictionary<string>,
 		deviceType: string,
 	): boolean {
-		let targetBootConfig = configUtils.envToBootConfig(configBackend, target);
+		const targetBootConfig = configUtils.envToBootConfig(configBackend, target);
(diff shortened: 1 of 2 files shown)


src/device-config.ts
```diff
@@ -613,17 +610,34 @@ export class DeviceConfig {
...
+				this.ensureDtoverlay(conf, 'vc4-fkms-v3d');
+				break;
+		}
+
+		return conf;
+	}
+
+	// Modifies conf
+	private static ensureDtoverlay(conf: ConfigOptions, field: string) {
 		if (conf.dtoverlay != null) {
```
zvin:
```typescript
if (conf.dtoverlay == null) {
  conf.dtoverlay = [];
} else if (_.isString(conf.dtoverlay)) {
  conf.dtoverlay = [conf.dtoverlay]
}
if (!_.includes(conf.dtoverlay, field)) {
  conf.dtoverlay.push(field);
}
```

CameronDiver: Much nicer! :+1:

PR #1173, opened 2020-01-10
Add repo.yml

nazrhom:
Declare ESR version next

Change-type: none
Signed-off-by: Giovanni Garufi <giovanni@balena.io>

### DIFF
--- a/repo.yml
+++ b/repo.yml
@@ -1,0 +1,4 @@
+type: node
+esr:
+  line: next
+  version: 10.3


repo.yml
```diff
@@ -0,0 +1,4 @@
+type: node
```
roman-mazur: Why node and not docker?

nazrhom:
@roman-mazur you are right, closing in favour of #1184. Will backport the same to the ESR branches. I thought we were missing some changes in versionist for this, as the supervisor has a package.json that maintains the node version, but I forgot about having added support for this in versionist several months ago ^^

PR #1175, opened 2020-01-10
Keep existing locks when purging

### LINKED ISSUE #1164: Deleting wifi credentials during purge breaks device if update lock is set
When deleting wifi credentials in combination with a lockfile and the staged releases a device could potentially not start the wifi-connect because the applying state changes. "Force" from the purge also overwrites the update and the device does not recover the containers.

mwohlert:
When purging the device keep the preexisting locks, when using force in place to make sure, that the locks persist through the purge.
This prevents a scheduled update from happening after the purge.

Fixes #1164


src/application-manager.coffee
```diff
@@ -945,14 +945,14 @@ module.exports = class ApplicationManager extends EventEmitter
 				.then =>
 					delete @_containerStarted[service.containerId]
 
-	_lockingIfNecessary: (appId, { force = false, skipLock = false } = {}, fn) =>
+	_lockingIfNecessary: (appId, { force = false, skipLock = false, keepLocks = false } = {}, fn) =>
```
CameronDiver: I wonder if there's any time that we don't want this behaviour? It seems to be that when using force, at the end of the task being forced, we should be in the same situation regarding locks.

mwohlert: Yes. I agree on that. I also thought about a different approach, where we keep track of the locks and have a predefined timeout, modifiable by env, that will wait a certain amount of time for the locks to be reinstated and fail the update-lock lock method, to kind of have a virtual lock, that is not really there, but update-lock is aware of that. 
That way we won't be stuck in case the locks were supposed to be removed, since a service maybe forgot to remove it.

CameronDiver: I like this idea, and I've just thought of a problem with the current implementation of this PR.

If a service tries to take the update lock before doing something critical that cannot be interrupted, there's a good chance the service will sit and wait for the lock to be taken, but given the supervisor will have already taken it, we're going to be in a deadlock.

mwohlert: Yes this could be a problem with the current implementation.

Are you fine with going ahead with my suggestion in my last comment? So that we have a timeout for locks to be reinstated.

CameronDiver: I'd need to understand the suggestion a little better before committing I think. So we have a timeout, in which the services cannot take the lock, but the update flow cannot also take the lock - with the intent being that any started services will have time to take the lock themselves?

Would we be treating locks from the update flow in a different way to locks taken by services? I don't understand what would stop the application update flow taking the lock before the services, regardless of this timeout?

I appreciate you taking the time to investigate and implement these solutions, it feels like we're super close - I'd just like to make sure that the level of complexity we add isn't detrimental to the codebase.

mwohlert: I agree with making sure to not increase the level of complexity too much, don't get me wrong here.

I would not treat those virtual locks different. Maybe writing up a little sample code would make things more clear here:
```
...
const existingLocks: { [lockName: string]: Date } = {};
...
function dispose(release: () => void): Bluebird<void> {
	return Bluebird.map(_.keys(locksTaken), lockName => {
		delete locksTaken[lockName];
		return lockFile.unlockAsync(lockName);
	})
		.finally(release)
		.return();
}
...
    if (force) {
        if (lockFile.checkSync(tmpLockName) && keepLocks) {
            existingLocks[tmpLockName] = Date.now();
        }
        return lockFile.unlockAsync(tmpLockName);
    } else {
        return Bluebird.map(_.keys(existingLocks), lockName => {
            if (existingLocks[lockName] + lockTimeout < Date.now()) {
                delete existingLocks[lockName];
            } else {
                throw new Error("Lock was set previously and is enforced by reaquire timeout")
            }
        })
    }
})
    .then(() => lockFile.lockAsync(tmpLockName))
...``

It would just keep a timestamp of when the lock was unlocked by the supervisor if force was used and check if the Date.now() + timeout is bigger than the current Date resulting in some kind of virtual lock, which is not really there, but the supervisor keeps track, that this lockfile existed before.

Hope the codeexample helps to convey my idea.

CameronDiver: Generally I like the idea, and with some comments in the code to explain the various properties I think it should be good - thanks!

Apologies for the delay in getting back to you, I've had some unexpected time off recently.

mwohlert: I'll do this tomorrow. Sorry I was short on time the last couple of weeks.

mwohlert:
@CameronDiver can you take a look again?

CameronDiver:
@mwohlert did you test this code? For me I was unable to update due to locks being taken, even when the lock override was set. Something is very much going wrong with this change unfortunately. I'm testing on an intel-nuc fwiw.

I've recently update the documentation on developing the supervisor in the README: https://github.com/balena-io/balena-supervisor#developing-the-supervisor so if you give that a try, we should get this PR moving forward much quicker.

mwohlert:
@CameronDiver Thanks for the feedback. I think since the PR is open for so long and a lot of files changed to js or ts, I'll redo this PR properly :).

Therefore I'm closing this for now and creating a new branch.

CameronDiver:
Thanks @mwohlert - I often find it easier to do this myself too.

PR #1177, opened 2020-01-16
Increase the minimum backoff delay and use poll inteval as the max

CameronDiver:
This change will start the minimum backoff from 15s (up from 500ms) and
will use the appUpdatePollInterval configuration variable as the max.

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -30,8 +30,8 @@ import log from './lib/supervisor-console';
 import DeviceState = require('./device-state');
 import Logger from './logger';
 
-const REPORT_SUCCESS_DELAY = 1000;
-const MAX_REPORT_RETRY_DELAY = 60000;
+// The exponential backoff starts at 15s
+const MINIMUM_BACKOFF_DELAY = 15000;
 
 const INTERNAL_STATE_KEYS = [
 	'update_pending',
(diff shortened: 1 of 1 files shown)


src/api-binder.ts
```diff
@@ -549,18 +549,20 @@ export class APIBinder {
 				}
 
 				await this.report();
-				await Bluebird.delay(REPORT_SUCCESS_DELAY);
-				await this.reportCurrentState();
+				this.reportCurrentState();
```
roman-mazur: I cannot get what this recursive call does... Should not it be called in the error case only? It would cause the second attempt to report the state which most likely will not happen because of the zero diff. But what is the point? We call `report` on a state change anyway?

CameronDiver:
@balena-ci retest

src/api-binder.ts
CameronDiver: I'm not sure actually, this is still left over from the typescript rewrite. I'm going to make some heavy architectural changes to this module soon so I'll look at it properly then.

Page-: It's in case a new change has come in whilst we were checking the previous one, and if there's nothing to report then we just exit early. I expect there could be nicer/clearer ways to do it though

PR #1179, opened 2020-01-16
Update types and fix issue with treating string as a number

Page-:
Change-type: patch


Page-:
@balena-ci retest

Page-:
@balena-ci retest

Issue #1183, opened 2020-01-27
Change changelog format


nazrhom:
With the current changelog format, knowing the version and date the verison was released is required to create a link that has an anchor to a specific changelog version. Aligning to a format like the one in https://github.com/balena-os/meta-balena/blob/development/CHANGELOG.md would make creating such references easier

nazrhom:
Connects-to: https://github.com/balena-os/balena-os/issues/380

PR #1184, opened 2020-01-28
Set docker type and update CHANGELOG.md to the correct format

nazrhom:
Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>
Connects-to: #1183


nazrhom:
@roman-mazur @CameronDiver it is a simple regex :P 
It doesn't make too much sense to support this on the VB level, as changelogs are meant to be free form and fully templatable. We could support this on balena-versionist directly, which will know about all the templates that we use, but are only 2 for now and the transformation between them is trivial

Issue #1176, opened 2020-01-16
Allow the GETting of target-state when not in local mode


CameronDiver:
There is a `GET /v2/target-state` api endpoint, which can report the target state when not in local mode with no problems, but it will return an error if used when not in local mode.

There's no reason to keep this endpoint from working when not in local mode.

PR #1186, opened 2020-01-29
Allow getting the target state when not in local mode

CameronDiver:
Change-type: patch
Closes: #1176 


roman-mazur:
IIRC, our docs say `/local` is used for local mode. We should either change the semantics of this path or treat it as an exception in the docs.

CameronDiver:
Good catch @roman-mazur - I've updated the docs

PR #1187, opened 2020-01-29
Added support for loading SSDT via ConfigFS for up-board

### LINKED ISSUE #1113: The supervisor should be able to load ACPI SSDTs for the Up Board(s)
The way to load ACPI tables is the following:

1 - `modprobe acpi_configfs`
2 - `mount -t configfs none [some location to be defined??]`
3 - `cd [some location to be defined??]/acpi/table`
4 - `mkdir my_ssdt`
5 - `cat /boot/acpi-tables/[desired SSDT] > my_ssdt/aml`

and in `/boot/acpi-tables/` there will be 4 tables: `spidev1.0.aml`, `spidev1.1.aml`, `spidev2.0.aml`, `spidev2.1.aml`

and they are currently to be used like so:

`spidev1.0.aml`, `spidev1.1.aml` on `up-core-plus` and `up-squared`
`spidev2.0.aml`, `spidev2.1.aml` on `up-core`

The above steps are not persistent between reboots so the supervisor would need to do them at each reboot if needed.

richbayliss:
Resolves #1113

### DIFF
--- a/src/config/backend.ts
+++ b/src/config/backend.ts
@@ -1,10 +1,12 @@
 import * as _ from 'lodash';
 import { child_process, fs } from 'mz';
+import * as path from 'path';
 
 import * as constants from '../lib/constants';
 import { writeFileAtomic } from '../lib/fs-utils';
 
 import log from '../lib/supervisor-console';
+import Logger from '../logger';
 
 export interface ConfigOptions {
 	[key: string]: string | string[];
@@ -32,7 +34,13 @@ async function remountAndWriteAtomic(
 	await writeFileAtomic(file, data);
 }
 
+export interface BackendOptions {
+	logger?: Logger;
+}
+
 export abstract class DeviceConfigBackend {
+	protected options: BackendOptions = {};
+
 	// Does this config backend support the given device type?
 	public abstract matches(deviceType: string): boolean;
 
@@ -64,6 +72,12 @@ export abstract class DeviceConfigBackend {
 
 	// Return the env var name for this config option
 	public abstract createConfigVarName(configName: string): string;
+
+	// Allow a chosen config backend to be initialised
+	public async initialise(opts: BackendOptions): Promise<DeviceConfigBackend> {
+		this.options = { ...this.options, ...opts };
+		return this;
+	}
 }
 
 export class RPiConfigBackend extends DeviceConfigBackend {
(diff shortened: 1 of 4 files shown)


src/config/utils.ts
```diff
@@ -2,21 +2,33 @@ import * as _ from 'lodash';
...
+		const backend = await getConfigBackend(deviceType);
+		if (backend) {
+			backend.initialise(opts);
+			return backend;
+		}
+	},
+);
 
-export function getConfigBackend(
+export async function getConfigBackend(
```
richbayliss: I think this might be able to be dropped now...

richbayliss:
@balena-ci re-test

richbayliss:
@balena-ci retest

src/device-config.ts
```diff
@@ -223,8 +223,10 @@ export class DeviceConfig {
 			return this.configBackend;
 		}
 		const dt = await this.config.get('deviceType');
-
-		this.configBackend = configUtils.getConfigBackend(dt) || null;
+		this.configBackend =
+			(await configUtils.initialiseConfigBackend(dt, {
+				logger: this.logger,
+			})) || null;
```
Page-: I'd prefer
```suggestion
			})) ?? null;
```
to clarify that it's just the `undefined` response you want to convert to null

src/config/backend.ts
```diff
@@ -424,3 +439,223 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+
+	public processConfigVarValue(name: string, value: string): string | string[] {
+		switch (this.stripPrefix(name)) {
+			case 'ssdt':
+				// value could be a single value, so just add to an array and return...
+				if (!_.startsWith(value, '"')) {
+					return [value];
+				} else {
+					// or, it could be parsable as the content of a JSON array; "value" | "value1","value2"
+					return JSON.parse(`[${value}]`) as string[];
```
Page-: Why do we allow literally any json? If all you want here is to convert a string into a string[] then using `.split` is probably a better path to take

src/config/backend.ts
```diff
@@ -424,3 +439,223 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+
+	public processConfigVarName(name: string): string {
+		return name;
+	}
+
+	public processConfigVarValue(name: string, value: string): string | string[] {
+		switch (this.stripPrefix(name)) {
+			case 'ssdt':
+				// value could be a single value, so just add to an array and return...
+				if (!_.startsWith(value, '"')) {
```
Page-: This could also be
```suggestion
				if (!value.startsWith('"')) {
```
and just use the native function, assuming the typing is correct

src/config/backend.ts
```diff
@@ -424,3 +439,223 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+		}
+	}
+
+	public isBootConfigVar(name: string): boolean {
+		switch (this.stripPrefix(name)) {
+			case 'ssdt':
+				return true;
+			default:
+				return false;
+		}
```
Page-: This could be
```suggestion
		return this.stripPrefix(name) === 'ssdt'
```

src/config/backend.ts
```diff
@@ -424,3 +439,223 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+		await this.writeConfigJSON(config);
+	}
+
+	public isSupportedConfig(name: string): boolean {
+		switch (this.stripPrefix(name)) {
+			case 'ssdt':
+				return true;
+			default:
+				return false;
+		}
```
Page-: This could also be
```suggestion
		return this.stripPrefix(name) === 'ssdt'
```
or if you really want to leave it potentially extendable then
```suggestion
		return ['ssdt'].includes(this.stripPrefix(name))
```
(I don't think it's necessary as it's an easy change to make when adding more anyway)

src/config/backend.ts
```diff
@@ -424,3 +439,223 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+	public async setBootConfig(opts: ConfigOptions): Promise<void> {
+		// read the config file...
+		const config = await this.readConfigJSON();
+
+		// see if the target state defined some SSDTs...
+		const ssdtKey = `${this.ConfigVarNamePrefix}ssdt`;
+		if (opts[ssdtKey]) {
+			// it did, so update the config with theses...
+			const ssdt = opts[ssdtKey];
+			const ssdts = _.isArray(ssdt) ? ssdt : [ssdt];
```
Page-:
```suggestion
			const ssdts = _.castArray(ssdt);
```

src/config/backend.ts
richbayliss: This was taken from other implementations in the RPi backend; so perhaps we should adopt a standard approach and use it all over 👍

src/config/backend.ts
richbayliss: TIL

src/config/backend.ts
richbayliss: agreed

Page-: Btw @richbayliss the bit I felt was unnecessary was the `[].includes` part, I think moving away from the `switch` makes sense

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+		// read the config file...
+		const config = await this.readConfigJSON();
+
+		// see if the target state defined some SSDTs...
+		const ssdtKey = `${this.ConfigVarNamePrefix}ssdt`;
+		if (opts[ssdtKey]) {
+			// it did, so update the config with theses...
+			const ssdt = opts[ssdtKey];
+			const ssdts = _.castArray(ssdt);
+			config['ssdt'] = ssdts;
```
Page-: I think trimming this down to
```suggestion
			config['ssdt'] = _.castArray(opts[ssdtKey]);
```
makes it clearer in this case

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+	}
+
+	private async writeConfigJSON(config: ConfigfsConfig): Promise<void> {
+		await remountAndWriteAtomic(this.ConfigFilePath, JSON.stringify(config));
+	}
+
+	private async loadConfiguredSsdt(config: ConfigfsConfig): Promise<void> {
+		if (config['ssdt'] !== undefined && _.isArray(config['ssdt'])) {
+			log.info('Loading configured SSDTs');
+			await Bluebird.mapSeries(config['ssdt'], aml => this.loadAML(aml));
```
Page-: Since this is done in serial anyway I think it'd be a bit nicer as
```suggestion
			for (const aml of config['ssdt']) {
				await this.loadAML(aml);
			}
```
Reasoning being is it'll work better with zero-cost async traces (which require `await` and going via the bluebird async removes that) and it also avoids the need to convert/adapt between native and bluebird promises

src/config/backend.ts
```diff
@@ -64,6 +73,12 @@ export abstract class DeviceConfigBackend {
 
 	// Return the env var name for this config option
 	public abstract createConfigVarName(configName: string): string;
+
+	// Allow a chosen config backend to be initialised
+	public initialise(opts: BackendOptions): Promise<DeviceConfigBackend> {
```
Page-: Why does this need to return a promise?

src/config/backend.ts
```diff
@@ -64,6 +73,12 @@ export abstract class DeviceConfigBackend {
...
+		this.options = { ...this.options, ...opts };
+		return Promise.resolve(this);
```
Page-: It'd be nicer to just make the function `async` and then this can be `return this` and it'll be automatically wrapped, also any exceptions will also get automatically wrapped

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+			log.info('Loading configured SSDTs');
+			await Bluebird.mapSeries(config['ssdt'], aml => this.loadAML(aml));
+		}
+	}
+
+	public async initialise(
+		opts: BackendOptions,
+	): Promise<ConfigfsConfigBackend> {
+		try {
+			super.initialise(opts);
```
Page-: This works because whilst `super.initialise` returns a promise it actually does all it's work synchronously, if we need to keep it as a promise returning function though then we should make sure to `await` this

richbayliss: yep - I spotted this before I read this comment, good spot 👍

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+		try {
+			const loadedTables = await this.listLoadedAcpiTables();
+
+			if (loadedTables.indexOf(aml) < 0) {
+				await fs.mkdir(amlDstPath);
+			}
+
+			log.info(`Loading AML ${aml}`);
+			await child_process.exec(
+				`cat ${amlSrcPath} > ${path.join(amlDstPath, 'aml')}`,
```
CameronDiver: Why this over a `cp` out of interest? Or even better a
```
fs.createReadStream(amlSrcPath).pipe(fs.createWriteStream(amlDstPath))
```
?

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+			);
+
+			const [oemId, oemTableId, oemRevision] = await Bluebird.all([
```
CameronDiver: I'd just use a Promise.all here and you can drop the Bluebird import

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+			return {};
+		}
+	}
+
+	private async writeConfigJSON(config: ConfigfsConfig): Promise<void> {
+		await remountAndWriteAtomic(this.ConfigFilePath, JSON.stringify(config));
+	}
+
+	private async loadConfiguredSsdt(config: ConfigfsConfig): Promise<void> {
+		if (config['ssdt'] !== undefined && _.isArray(config['ssdt'])) {
```
CameronDiver:
```suggestion
		if (_.isArray(config['ssdt'])) {
```

src/config/backend.ts
```diff
@@ -64,6 +73,12 @@ export abstract class DeviceConfigBackend {
 
 	// Return the env var name for this config option
 	public abstract createConfigVarName(configName: string): string;
+
+	// Allow a chosen config backend to be initialised
+	public async initialise(opts: BackendOptions): Promise<DeviceConfigBackend> {
+		this.options = { ...this.options, ...opts };
+		return this;
```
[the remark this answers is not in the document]
richbayliss: @CameronDiver constructors cannot be async, and the initialisation of a backend my require some async operations (read/write files, run system commands etc) so the base case has this method which allows the passing of some options, and is invoked async before returning it to the consumer of the backend.

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+/**
+ * A backend to handle ConfigFS host configuration for ACPI SSDT loading
+ *
+ * Supports:
+ * 	- {BALENA|RESIN}_HOST_CONFIGFS_ssdt = value | "value" | "value1","value2"
+ */
+export class ConfigfsConfigBackend extends DeviceConfigBackend {
+	private readonly SystemAmlFiles = path.join(
+		constants.rootMountPoint,
+		'/boot/acpi-tables',
```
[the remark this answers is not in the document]
richbayliss: ... so I always tend to adopt this style, as I would consider it more confusing without the leading `/` 😁 

I am of course happy to adopt a standard approach, so if we agree that they shouldn't contain the slash then I'll do that and we should make sure all other instances are aligned.

src/config/backend.ts
richbayliss: @CameronDiver I originally tried `cp` and it didn't work... the instructions on how to load the SSDTs manually specifically details using `cat` so I went with that.

src/config/backend.ts
richbayliss: 👍

src/config/backend.ts
```diff
@@ -424,3 +439,218 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
...
+		constants.rootMountPoint,
+		'/sys/kernel/config',
+	);
+	private readonly ConfigVarNamePrefix = `${constants.hostConfigVarPrefix}CONFIGFS_`;
+
+	// supported backend for the following device types...
+	public static readonly SupportedDeviceTypes = ['up-board'];
+
+	private stripPrefix(name: string): string {
+		return name.substr(this.ConfigVarNamePrefix.length);
```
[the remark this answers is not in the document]
richbayliss: I wouldn't want to throw an error, but the usage of this method is either returned or directly compared... so I wouldn't expect any negative fallout from keeping it like this.

```suggestion
		if (!name.startsWith(this.ConfigVarNamePrefix)) {
			return name;
		}
		return name.substr(this.ConfigVarNamePrefix.length);
```

This would make sure the value is only stripped if it's desired though, so I think this is a good option.

src/config/backend.ts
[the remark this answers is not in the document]
CameronDiver: Everywhere else that this is done doesn't use a leading `/` so preferably yeah that'd be better.

src/config/backend.ts
CameronDiver: Super strange. Can you leave a comment in the code with this info please? Could totally see somebody looking to "optimize" this in future

richbayliss: Yes, will do 👍 that's a good point

PR #1197, opened 2020-02-07
Make balena-sync optional

CameronDiver:
Drivelist won't build with node12, and it won't build at all with osx.
We make balena-sync optional so that everything but sync.js will work
fine on these platforms.

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "@types/sinon": "^7.5.1",
     "@types/sinon-chai": "^3.2.3",
     "@types/tmp": "^0.1.0",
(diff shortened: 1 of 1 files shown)


richbayliss:
@CameronDiver it still will not build on macos due to the `dbus-native` dependency, which if you make optional will allow it to build. Without both being optional, the PR doesn't satisfy your description, so not sure how you want to proceed... 😕

CameronDiver:
Closing in favour of updating dependencies separately and solving this in a better way.

PR #1204, opened 2020-02-11
Add a noinit option to sync-debug

CameronDiver:
The noinit option will avoid doing a complete sync to the supervisor
on startup. This is useful when there has only been code changes, and no
modules etc have been added. It can be helpful when running on a network
with no external internet access, as everything needed is already inside
the container.

To aid this we also move to yargs, rather than try to do all of the
parsing ourselves.

Change-type: patch

### DIFF
--- a/sync-debug.js
+++ b/sync-debug.js
@@ -1,18 +1,32 @@
 #!/usr/bin/env node
 
-if (!process.argv[2] || ['help', '-h', '--help'].includes(process.argv[2])) {
-	console.log(`
-Sync changes in the javascript code to a running local mode supervisor on a device on the local network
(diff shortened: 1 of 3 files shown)


sync-debug.js
```diff
@@ -1,18 +1,33 @@
...
+		'Sync changes in code to a running debug mode supervisor on a local device',
+		yargs =>
+			yargs.positional('device IP', {
+				type: 'string',
+				describe: 'The address of a local device',
+			}),
+	)
+	.usage(helpText)
+	.version(false)
+	// .boolean('noinit')
```
zvin: leftover commented code

PR #1205, opened 2020-02-12
Clean up labels code, and add sysfs label

CameronDiver:
This PR adds the label `io.balena.features.sysfs`, which when applied will mount the host `/sys` into the containers `/sys` for monitoring purposes.

I've also tried to make the code a little cleaner and more declarative and it should make adding these features marginally easier in the future.

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -371,6 +371,8 @@ export function addFeaturesFromLabels(
 			);
 		},
 		'io.balena.features.sysfs': () => service.config.volumes.push('/sys:/sys'),
+		'io.balena.features.procfs': () =>
+			service.config.volumes.push('/proc:/proc'),
 	};
 
 	_.each(features, (fn, label) => {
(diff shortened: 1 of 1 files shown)


src/compose/utils.ts
```diff
@@ -320,64 +320,74 @@ export function addFeaturesFromLabels(
...
+					`http://127.0.0.1:${options.listenPort}`,
+				);
+			} else {
+				setEnvVariables('SUPERVISOR_HOST', options.supervisorApiHost);
+				setEnvVariables(
+					'SUPERVISOR_ADDRESS',
+					`http://${options.supervisorApiHost}:${options.listenPort}`,
+				);
+				service.config.networks[constants.supervisorNetworkInterface] = {};
+			}
```
zvin: NIT:
```typescript
let host: string;
if (service.config.networkMode === 'host') {
	host = '127.0.0.1';
} else {
	host = options.supervisorApiHost;
	service.config.networks[constants.supervisorNetworkInterface] = {};
}
setEnvVariables('SUPERVISOR_HOST', host);
setEnvVariables('SUPERVISOR_ADDRESS', `http://${host}:${options.listenPort}`);
```

PR #1216, opened 2020-03-13
Update dependencies

Page-:
This reduces the size of the generated bundle from 3.85 to 3.75 MB due to deduped packages, I believe it also reduces the time to build as there's less code to process (but it's fairly variable and running enough times to reduce the error margin would take forever)


package.json
```diff
@@ -62,9 +62,9 @@
     "@types/sinon-chai": "^3.2.3",
     "@types/tmp": "^0.1.0",
     "balena-sync": "^10.0.0",
-    "blinking": "~0.0.2",
-    "bluebird": "^3.5.3",
-    "body-parser": "^1.12.0",
+    "blinking": "~0.0.3",
```
Page-: This is not strictly updated, just setting the package.json version to match the package-lock.json

package.json
```diff
@@ -80,32 +80,31 @@
     "docker-toolbelt": "^3.3.7",
     "duration-js": "^4.0.0",
     "event-stream": "3.3.4",
-    "express": "^4.0.0",
+    "express": "^4.17.1",
     "fork-ts-checker-webpack-plugin": "^0.5.2",
-    "fp-ts": "^2.0.0",
-    "husky": "^4.2.1",
+    "fp-ts": "^2.0.5",
```
Page-: This is not actually updated, just setting the package.json version to match the package-lock.json

package.json
```diff
@@ -80,32 +80,31 @@
...
+    "husky": "^4.2.3",
     "io-ts": "^2.0.1",
     "io-ts-reporters": "^1.0.0",
     "istanbul": "^0.4.5",
     "json-mask": "^0.3.8",
-    "knex": "~0.15.2",
-    "lint-staged": "^10.0.7",
+    "knex": "^0.15.2",
```
Page-: The range is equivalent for both but npm decided it wanted ^ over ~ so I let it

package.json
```diff
@@ -80,32 +80,31 @@
...
+    "lint-staged": "^10.0.8",
     "livepush": "^2.0.2",
-    "lockfile": "^1.0.1",
-    "lodash": "^4.17.13",
+    "lockfile": "^1.0.4",
```
Page-: This is not actually updated, just setting the package.json version to match the package-lock.json

package.json
```diff
@@ -80,32 +80,31 @@
...
+    "lodash": "^4.17.15",
     "log-timestamp": "^0.1.2",
-    "memoizee": "^0.4.1",
+    "memoizee": "^0.4.14",
```
Page-: This is not actually updated, just setting the package.json version to match the package-lock.json

package.json
```diff
@@ -80,32 +80,31 @@
...
     "mixpanel": "^0.10.1",
     "mkdirp": "^0.5.1",
     "mocha": "^5.2.0",
     "morgan": "^1.9.1",
     "mz": "^2.7.0",
     "network-checker": "^0.1.1",
     "pinejs-client-request": "^5.2.0",
-    "prettier": "1.19.1",
```
Page-: Removed this as it's only used via resin-lint now

package.json
```diff
@@ -119,8 +118,8 @@
     "ts-loader": "^5.4.5",
     "ts-node": "^8.6.2",
     "typed-error": "^2.0.0",
-    "typescript": "^3.7.5",
-    "webpack": "^4.25.0",
+    "typescript": "^3.8.3",
+    "webpack": "^4.28.0",
```
Page-: This is not actually updated, just setting the package.json version to match the package-lock.json

src/application-manager.coffee
```diff
@@ -202,9 +202,7 @@ module.exports = class ApplicationManager extends EventEmitter
 					else
 						log.debug('Ignoring legacy dependent image', image)
 
-				obj = { local: apps, dependent }
-				obj.commit = currentCommit
-				return obj
+				return { local: apps, dependent, commit: currentCommit }
```
Page-: I rewrote like this to make the return type a little clearer to see

src/application-manager.d.ts
```diff
@@ -97,7 +97,11 @@ class ApplicationManager extends EventEmitter {
 		transaction: Knex.Transaction,
 	): Promise<void>;
 
-	public getStatus(): Promise<DeviceStatus>;
+	public getStatus(): Promise<{
+		local: DeviceStatus.local.apps;
```
Page-: Updated typescript caught that we were using the types expecting them to look like this and it turns out that we were using them correctly but typing incorrectly, so I've now fixed the typing

webpack.config.js
```diff
@@ -7,6 +7,7 @@ const CopyWebpackPlugin = require('copy-webpack-plugin');
 var ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 
 var externalModules = [
+	'async_hooks',
```
Page-: This is to make webpack on node 6 happy with the bluebird update, bluebird only loads `async_hooks` when it's supported so it's fine at runtime but was causing webpack issues when trying to resolve the module for node 6 (as it doesn't exist there)

PR #1220, opened 2020-03-23
Update dependencies

CameronDiver:
We also pin livepush in place while
https://github.com/balena-io-modules/livepush/issues/67
is fixed.

Change-type: patch


package.json
```diff
@@ -55,74 +55,75 @@
...
     "mocha": "^5.2.0",
-    "morgan": "^1.9.1",
+    "morgan": "^1.10.0",
     "mz": "^2.7.0",
     "network-checker": "^0.1.1",
     "pinejs-client-request": "^5.2.0",
     "pretty-ms": "^5.1.0",
     "request": "^2.88.2",
-    "resin-lint": "^3.2.4",
+    "resin-lint": "^3.3.1",
```
Page-: Ideally also switch to balena-lint:
```suggestion
    "@balena/lint": "^4.0.1",
```

package.json
```diff
@@ -55,74 +55,75 @@
...
-    "io-ts": "^2.0.1",
+    "io-ts": "^2.1.2",
     "io-ts-reporters": "^1.0.0",
     "istanbul": "^0.4.5",
-    "json-mask": "^0.3.8",
+    "json-mask": "^0.3.9",
     "knex": "^0.15.2",
     "lint-staged": "^10.0.8",
-    "livepush": "^2.0.2",
+    "livepush": "2.1.0",
```
Page-: Why is this now a fixed version? Can we not trust the livepush maintainer? ;)

CameronDiver: As it turns out, no we can't trust the maintainer :p 
https://github.com/balena-io-modules/livepush/issues/67

Page-: And in a patch version no less! :O

PR #1221, opened 2020-03-23
Update to balena-lint and enable javascript linting

Page-:
Change-type: patch


.lintstagedrc
```diff
@@ -2,7 +2,7 @@
 	"*.coffee": [
 		"resin-lint"
 	],
-	"*.ts": [
+	"*.ts$|!(*webpack.config).js$": [
```
Page-: We cannot lint the webpack config as then it won't be usable on node 6..

webpack.config.js
```diff
@@ -134,13 +133,13 @@ module.exports = function(env) {
...
 				{
-					from: './src/migrations',
+					from: './build/migrations',
 					to: 'migrations',
 				},
 			]),
 			new webpack.ContextReplacementPlugin(
 				/\.\/migrations/,
-				path.resolve(__dirname, 'src/migrations')
+				path.resolve(__dirname, 'build/migrations')
```
Page-: @CameronDiver would you be able to test the migrations work in a minified build please - this change is probably right but the migrations are a bit of black magic so I'm not certain

PR #1222, opened 2020-03-23
Convert device-api/v1 from coffeescript to javascript

Page-:


src/lib/validation.ts
```diff
@@ -61,7 +61,7 @@ export function checkString(s: unknown): string | void {
  * Given a value which can be a string, boolean or number, return a boolean
  * which represents if the input was truthy
  */
-export function checkTruthy(v: unknown): boolean | void {
+export function checkTruthy(v: unknown): boolean | undefined {
```
Page-: Specifying this as `undefined` means we can use `??` like `checkTruthy(x) ?? false`

src/device-api/v1.js
```diff
@@ -0,0 +1,126 @@
...
+import * as constants from '../lib/constants';
+import { checkInt, checkTruthy } from '../lib/validation';
+import { doRestart, doPurge, serviceAction } from './common';
+
+export const createV1Api = function(router, applications) {
+	const { eventTracker } = applications;
+
+	router.post('/v1/restart', function(req, res, next) {
+		const appId = checkInt(req.body.appId);
+		const force = checkTruthy(req.body.force) ?? false;
```
Page-: I had to add the `?? false` here to pass the type checks, previously `force` could be `undefined` - is `false` the right default/

src/device-api/v1.js
```diff
@@ -0,0 +1,126 @@
...
+					appToSend.commit = status.commit;
+				}
+				return res.json(appToSend);
+			},
+		).catch(next);
+	});
+
+	router.post('/v1/purge', function(req, res, next) {
+		const appId = checkInt(req.body.appId);
+		const force = checkTruthy(req.body.force) ?? false;
```
Page-: I had to add the `?? false` here to pass the type checks, previously `force` could be `undefined` - is `false` the right default/

src/device-api/v1.js
```diff
@@ -0,0 +1,126 @@
...
+			return res.status(400).send('Missing app id');
+		}
+		return doRestart(applications, appId, force)
+			.then(() => res.status(200).send('OK'))
+			.catch(next);
+	});
+
+	const v1StopOrStart = function(req, res, next, action) {
+		const appId = checkInt(req.params.appId);
+		const force = checkTruthy(req.body.force) ?? false;
```
Page-: This `?? false` wasn't technically required by typescript but it seemed the right thing to do based on the two cases it did catch

src/device-api/v1.js
CameronDiver: It is yeah.

src/device-api/v1.js
CameronDiver: Yeah, the `checkTruthy` function is a mess, but I'd like the entire codebase to be in typescript before changing fundamentals like that.

PR #1225, opened 2020-03-24
Add type checking for javascript files

Page-:
Depends on #1224

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -347,7 +347,7 @@ export class Images extends (EventEmitter as new () => ImageEventEmitter) {
 	}
 
 	private async getImagesForCleanup(): Promise<string[]> {
-		const images = [];
+		const images: string[] = [];
 
 		const [
 			supervisorImageInfo,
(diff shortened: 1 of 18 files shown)


tsconfig.js.json
```diff
@@ -0,0 +1,8 @@
+{
+	"extends": "./tsconfig.release.json",
+	"compilerOptions": {
+		"noImplicitAny": false,
```
Page-: Without this the js files have a lot of implicit any errors which makes sense and isn't something we're interested in

src/migrations/20180516164000-endpoint-specific-apps.js
```diff
@@ -2,12 +2,12 @@ const fs = require('fs');
...
 
 exports.up = function(knex, Promise) {
-	return new Promise((resolve, reject) => {
+	return new Promise((resolve) => {
 		if (!configJsonPath) {
 			console.log(
 				'Unable to locate config.json! Things may fail unexpectedly!',
 			);
-			resolve({});
+			return resolve({});
```
Page-: I don't know what would have happened but previously this fell through the `fs.readFile` with an empty path (typescript caught this)

src/compose/images.ts
```diff
@@ -347,7 +347,7 @@ export class Images extends (EventEmitter as new () => ImageEventEmitter) {
 	}
 
 	private async getImagesForCleanup(): Promise<string[]> {
-		const images = [];
+		const images: string[] = [];
```
Page-: All these changes to typescript files for vars that don't have an initial type are because they fail under `noImplicitAny: false`

PR #1227, opened 2020-03-26
Convert proxyvisor to javascript

Page-:
Change-type: patch


src/proxyvisor.js
```diff
@@ -0,0 +1,1013 @@
+import * as Promise from 'bluebird';
+import * as _ from 'lodash';
+import * as express from 'express';
+import * as fs from 'fs';
+import * as request from './lib/request';
```
Page-: This was previously `{ request }` which was broken

src/proxyvisor.js
```diff
@@ -0,0 +1,1014 @@
...
+					const targetState = formatTargetAsState(device);
+					return this.getHookEndpoint(device.appId).then(endpoint => {
+						if (device.markedForDeletion) {
+							return this.sendDeleteHook(device, apiTimeout, endpoint);
+						} else if (
+							device.targetCommit != null &&
+							!_.isEqual(targetState, currentState) &&
+							!_.isEqual(targetState, this.acknowledgedState[device.uuid])
+						) {
+							return this.sendUpdate(device, apiTimeout, endpoint);
```
Page-: This used to be `@sendUpdate(device, targetState, apiTimeout, endpoint)` but there is no `targetState` accepted so I figured this was the correct fix

src/proxyvisor.js
```diff
@@ -0,0 +1,1012 @@
...
+							const { uuid } = device;
+							// Only consider one app per dependent device for now
+							const appId = _(device.apps)
+								.keys()
+								.head();
+							if (appId == null) {
+								throw new Error(
+									'Could not find an app for the dependent device',
+								);
+							}
```
Page-: I had to add this to make typescript happy, I assume it never happens but at least if it does we now get a slightly better error message

src/proxyvisor.js
```diff
@@ -0,0 +1,1012 @@
...
+		/** @type {_.Dictionary<any>} */
+		const fieldsToUpdateOnAPI = _.pick(
+			fieldsToUpdateOnDB,
+			'status',
+			'is_online',
+			'releaseId',
+		);
+		if (fieldsToUpdateOnDB.commit != null) {
+			fieldsToUpdateOnAPI.is_on__commit = fieldsToUpdateOnDB.commit;
+		}
```
Page-: I simplified this from
```coffee
		fieldsToUpdateOnAPI = _.pick(fieldsToUpdateOnDB, 'status', 'is_online', 'commit', 'releaseId')
		if fieldsToUpdateOnAPI.commit?
			fieldsToUpdateOnAPI.is_on__commit = fieldsToUpdateOnAPI.commit
			delete fieldsToUpdateOnAPI.commit
```
whilst trying to get the types to be happy but ended up having to use `/** @type {_.Dictionary<any>} */` - I've left the simplified version though as I think it's an improvement

src/proxyvisor.js
```diff
@@ -0,0 +1,1012 @@
...
+import * as url from 'url';
+
+import { log } from './lib/supervisor-console';
+
+/** @type {(command: string, options: childProcess.ExecOptions) => Promise<string>} */
+const execAsync = Promise.promisify(childProcess.exec);
+const lstatAsync = Promise.promisify(fs.lstat);
+const unlinkAsync = Promise.promisify(fs.unlink);
+const readdirAsync = Promise.promisify(fs.readdir);
+const mkdirpAsync = Promise.promisify(mkdirp);
```
CameronDiver: Can you change these to mz please?

src/proxyvisor.js
CameronDiver: We use `Dictionary<T>` everywhere already, is that accepted as a type, even though it's defined in typings/global.d.ts?

src/proxyvisor.js
```diff
@@ -0,0 +1,1012 @@
...
+			.catch(function(err) {
+				log.error(`Error on ${req.method} ${url.parse(req.url).pathname}`, err);
+				return res.status(503).send(err?.message || err || 'Unknown error');
+			}),
+	);
+
+	return router;
+};
+
+export default class Proxyvisor {
```
CameronDiver: I don't think this is the same? You'll need to update some imports, right?

src/application-manager.coffee
```diff
@@ -28,7 +28,7 @@ updateLock = require './lib/update-lock'
 { Volume } = require './compose/volume'
 compositionSteps = require './compose/composition-steps'
 
-Proxyvisor = require './proxyvisor'
+{ Proxyvisor } = require './proxyvisor'
```
Page-: There's no `export =` in javascript modules so it was necessary to make this change

package.json
```diff
@@ -18,7 +18,7 @@
     "coverage": "istanbul report text && istanbul report html",
     "test:fast": "TEST=1 mocha --opts test/fast-mocha.opts",
     "test:debug": "npm run test:build && TEST=1 mocha --inspect-brk",
-    "prettify": "balena-lint -e ts -e js --typescript --fix src/ test/ typings/ webpack.config.js",
+    "prettify": "balena-lint -e ts -e js --typescript --fix src/ test/ typings/",
```
Page-: This removes webpack.config.js from prettifying because it needs to be able to run directly on node 6 and prettifying it breaks that

src/proxyvisor.js
Page-: Pushed as a separate commit

src/proxyvisor.js
Page-: Done

src/proxyvisor.js
```diff
@@ -0,0 +1,1013 @@
...
+import * as path from 'path';
+import * as mkdirp from 'mkdirp';
+import * as bodyParser from 'body-parser';
+import * as childProcess from 'child_process';
+import * as url from 'url';
+
+import { log } from './lib/supervisor-console';
+
+/** @type {(command: string, options: childProcess.ExecOptions) => Promise<string>} */
+const execAsync = Promise.promisify(childProcess.exec);
```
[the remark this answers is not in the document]
Page-: Done

PR #1228, opened 2020-03-26
Convert src/device-api/common.coffee to javascript

Page-:
Change-type: patch


src/device-api/common.js
```diff
@@ -0,0 +1,88 @@
...
+			return deviceState
+				.pausingApply(() =>
+					deviceState
+						.applyIntermediateTarget(currentState, { skipLock: true })
+						.then(() =>
+							// Now that we're not running anything, explicitly
+							// remove the volumes, we must do this here, as the
+							// application-manager will not remove any volumes
+							// which are part of an active application
+							Bluebird.each(volumes.getAllByAppId(appId), vol => vol.remove()),
```
CameronDiver: It looks like you're missing a return here?

CameronDiver: And in fact I would prefer to do this all in parallel.

Page-: It's an implicit return function, just the comment makes that hard to spot - I'll make it an explicit return function instead. 
For parallel vs serial I'd rather leave it as-is to keep the same behavior during the conversion and make that change in a separate PR since it could bring up edge cases bugs

Page-: I changed this to an explicit return

CameronDiver: I'm happy enough leaving it as-is for now.

PR #1229, opened 2020-03-26
Application manager improvements

Page-:


src/application-manager.coffee
```diff
@@ -68,11 +68,11 @@ createApplicationManagerRouter = (applications) ->
 
 	return router
 
-module.exports = class ApplicationManager extends EventEmitter
+exports.ApplicationManager = class ApplicationManager extends EventEmitter
```
Page-: Named export makes life much easier in a javascript/typescript world

src/application-manager.coffee
```diff
@@ -68,11 +68,11 @@ createApplicationManagerRouter = (applications) ->
...
 	constructor: ({ @logger, @config, @db, @eventTracker, @deviceState }) ->
 		@docker = new Docker()
 		@images = new Images({ @docker, @logger, @db, @config })
-		@services = new ServiceManager({ @docker, @logger, @images, @config })
+		@services = new ServiceManager({ @docker, @logger, @config })
```
Page-: There were a load of examples like this where we pass unexpected/unused parameters that I removed

src/compose/images.ts
```diff
@@ -465,7 +465,10 @@ export class Images extends (EventEmitter as new () => ImageEventEmitter) {
 		}
 	}
 
-	public static isSameImage(image1: Image, image2: Image): boolean {
+	public static isSameImage(
+		image1: Pick<Image, 'name'>,
+		image2: Pick<Image, 'name'>,
+	): boolean {
```
Page-: These types are more accurate to what the function actually uses and what application-manager might pass in

src/device-api/common.d.ts
```diff
@@ -25,6 +25,6 @@ declare function serviceAction(
 	action: string,
 	serviceId: number,
 	current: Service,
-	target: Service,
-	options: any,
+	target?: Service,
+	options?: any,
```
Page-: These aren't guaranteed to be passed in by the application-manager

src/lib/docker-utils.ts
```diff
@@ -45,7 +45,7 @@ const DELTA_TOKEN_TIMEOUT = 10 * 60 * 1000;
 export class DockerUtils extends DockerToolbelt {
 	public dockerProgress: DockerProgress;
 
-	public constructor(opts: Dockerode.DockerOptions) {
+	public constructor(opts?: Dockerode.DockerOptions) {
```
Page-: This is actually optional and the application-manager doesn't pass any options in

PR #1234, opened 2020-03-30
Don't mangle names when minimising with webpack

CameronDiver:
Change-type: patch


package.json
```diff
@@ -114,6 +114,7 @@
     "sinon-chai": "^3.5.0",
     "strict-event-emitter-types": "^2.0.0",
     "terser": "^3.17.0",
```
Page-: `terser-webpack-plugin` looks to use `terser` 4.x - should this be updated/removed?

CameronDiver: good catch

Issue #1232, opened 2020-03-30
Add BALENA_DEVICE_ARCH and BALENA_MACHINE_NAME env vars in containers for runtime use


rahul-thakoor:
`BALENA_DEVICE_ARCH` and `BALENA_MACHINE_NAME` can be useful at runtime to customize scripts/programs for particular device family or architecture

PR #1235, opened 2020-03-30
Add BALENA_DEVICE_ARCH environment variable for containers

CameronDiver:
Closes: #1232
Change-type: minor


src/config/functions.ts
```diff
@@ -1,12 +1,14 @@
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
+import { fs } from 'mz';
```
zvin: We could use `fs.promises` as we're on node >= 10

src/config/functions.ts
```diff
@@ -32,6 +34,22 @@ export const fnSchema = {
...
+			const data = await fs.readFile(
+				`${constants.rootMountPoint}/resin-boot/device-type.json`,
+				'utf8',
+			);
+			const deviceInfo = JSON.parse(data);
+
+			return deviceInfo.arch;
+		} catch (e) {
+			log.error(`Unable to get architecture: ${e}`);
+			return 'unavailable';
```
zvin: Why not `unknown` as in `schema-type.ts` ?

src/device-api/v2.ts
```diff
@@ -311,23 +310,25 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
-		});
+			return res.status(200).json({
+				status: 'success',
+				info: {
+					arch: deviceArch,
+					deviceType,
+				},
+			});
+		} catch (e) {
+			res.status(400).json({
```
zvin: `Bad Request` doesn't seem appropriate here.

test/04-service.spec.ts
```diff
@@ -1,4 +1,5 @@
 import * as _ from 'lodash';
+import * as mz from 'mz';
```
zvin: This is not used.

src/config/functions.ts
Page-: We still have a node 6 device :(

src/config/functions.ts
CameronDiver: I was originally thinking it was for different failure modes, but it's not actually possible with the current situation for `unavailable` to be returned. I guess it's fine.

PR #1240, opened 2020-03-31
Convert some tests to javascript

Page-:
Change-type: patch


src/compose/network.ts
```diff
@@ -94,7 +94,9 @@ export class Network {
 	public static fromComposeObject(
 		name: string,
 		appId: number,
-		network: Partial<ComposeNetworkConfig>,
+		network: Partial<Omit<ComposeNetworkConfig, 'ipam'>> & {
+			ipam?: Partial<ComposeNetworkConfig['ipam']>;
+		},
```
Page-: This is more accurate to what tests pass in and the code also handles the `ipam` properties being undefined/null

src/compose/ports.ts
```diff
@@ -30,7 +30,7 @@ interface PortRange {
 export class PortMap {
 	private ports: PortRange;
 
-	private constructor(portStrOrObj: string | PortRange) {
+	public constructor(portStrOrObj: string | PortRange) {
```
Page-: This constructor is used a lot by the tests and the easiest way to make them work was to make it public

test/16-ports.spec.js
```diff
@@ -0,0 +1,378 @@
...
+	});
+
+	describe('toDockerOpts', function() {
+		it('should correctly generate docker options', () =>
+			expect(new PortMap('80').toDockerOpts()).to.deep.equal({
+				exposedPorts: {
+					'80/tcp': {},
+				},
+				portBindings: {
+					'80/tcp': [{ HostIp: '', HostPort: '80' }],
```
Page-: I had to convert all of these to strings to match the types - I'm assuming that's correct but just want to double check

test/17-config-utils.spec.js
```diff
@@ -0,0 +1,142 @@
...
+TIMEOUT 30
+
+MENU TITLE Boot Options
+LABEL primary
+MENU LABEL primary Image
+LINUX /Image
+APPEND \${cbootargs} \${resin_kernel_root} ro rootwait isolcpus=3\
+`;
+
+				let readFileStub = stub(fs, 'readFile').resolves(text);
```
Page-: I've changed to assigning the stub in a bunch of places as it helps a lot with the typings when we then call sinon specific methods

test/18-startup.ts
```diff
@@ -0,0 +1,16 @@
+import { Supervisor } from '../src/supervisor';
+import { expect } from './lib/chai-config';
+
+describe('Startup', () => {
+	it('should startup correctly', function() {
+		const supervisor = new Supervisor();
+		expect(supervisor.init()).to.not.throw;
+		// Cast as any to access private properties
+		const anySupervisor = supervisor as any;
```
Page-: I'm not a massive fan of this cast but it's probably preferable to making all the properties public

test/18-startup.ts
```diff
@@ -0,0 +1,16 @@
+import { Supervisor } from '../src/supervisor';
```
Page-: There's 2 files with `18-`, not sure if that's a problem but it surprised me slightly

src/compose/ports.ts
CameronDiver: There's a good reason that this is private, and I don't want anybody to think that this is the way to use that throughout the codebase. I'd 100% prefer a `@ts-ignore` or cast to any.

src/compose/ports.ts
```diff
@@ -18,7 +18,7 @@ export interface DockerPortOptions {
 	portBindings: PortBindings;
 }
 
-interface PortRange {
+export interface PortRange {
```
Page-, line 21: I had to expose this to be able to force cast with it in the tests

src/compose/ports.ts
Page-: done

PR #1241, opened 2020-03-31
Move to an alpine base image and drop i386-nlp support

CameronDiver:
We also remove the Makefile to go to a simpler build system, as long with
the `retry_docker_push.sh` file. We remove the rest of the automation tools
as they're no longer used and update the circle.yml file.

We also remove debug builds, as these aren't needed moving forward, and were
only to enable livepush builds.

We remove `sync-debug.js` as there are no longer any debug builds, and move `sync.js` to `tools/` as this is no longer the preferred method of developing the supervisor.

We add a `npm run sync` command to the package.json, which integrates v3 of livepush, enabling live-cmds. This enables sub-1s code changes being reflected on device (when using an intel nuc).

Change-type: major


automation/build.sh
```diff
@@ -25,99 +24,97 @@
...
-docker pull balenalib/raspberry-pi-node:10-run &
-docker pull balenalib/armv7hf-node:10-run &
-docker pull balenalib/aarch64-node:10-run &
-docker pull balenalib/amd64-node:10-run &
-docker pull balenalib/i386-node:10-run &
-docker pull balenalib/i386-nlp-node:6-jessie &
-wait
+function retryImagePush() {
+	local image=$1
+	local retries=0
```
xginn8:
```suggestion
	local -i retries
```

will ensure this is an int, and set to 0

automation/build.sh
```diff
@@ -25,99 +24,97 @@
...
 
-export DOCKER_BUILD_OPTIONS=${CACHE_FROM}
-export ARCH
-export MIXPANEL_TOKEN
+	while [ "$retries" -lt 3 ]; do
```
xginn8:
```suggestion
	while (( retries < 3 )); do
```

automation/build.sh
```diff
@@ -25,99 +24,97 @@
...
+		((retries=retries+1))
```
xginn8:
```suggestion
		retries+=1
```

automation/build.sh
```diff
@@ -25,99 +24,97 @@
...
+		docker push "${image}"
```
xginn8:
```suggestion
		if docker push "${image}";
				break
		fi
```

Dockerfile
```diff
@@ -1,151 +1,114 @@
...
-
-RUN npm run test-nolint \
-	&& npm run build
-
-##############################################################################
-
-# Build nodejs dependencies
-FROM $ARCH-node-base as node-deps
+ARG ARCH
+ARG NODE_VERSION
```
xginn8: redefined from above (line 2), intended?

Dockerfile
```diff
@@ -1,151 +1,114 @@
...
+	libstdc++ \
+	libuv
 
-ARG NPM_VERSION
-# Install only the production modules that have C extensions
-RUN curl -LOJ https://www.npmjs.com/install.sh && \
-	npm config set unsafe-perm true && \
-	npm_install="${NPM_VERSION}" npm_config_prefix=/tmp sh ./install.sh && \
-	JOBS=MAX /tmp/bin/npm ci --no-optional --unsafe-perm --production
+# Install node from baleanas prebuilt cache
```
xginn8:
```suggestion
# Install node from balena's prebuilt cache
```

Dockerfile
```diff
@@ -1,151 +1,114 @@
...
+
+RUN ["cross-build-start"]
+
+RUN apk add --no-cache \
+	ca-certificates \
+	kmod \
+	iptables \
+	ip6tables \
+	rsync \
+	curl \
```
xginn8: Do you need curl specifically for something? If `wget` ships with alpine, can save the install.

Dockerfile
CameronDiver: Yeah, we need to bring it into the stage, it's why it's redefined again in the latter stage

Dockerfile
CameronDiver: Nice, I didn't actually realise that wget came bundled as a busybox binary. I'll change the healthcheck to use this too.

Dockerfile
xginn8: I missed that line, thanks for clarifying

Dockerfile
```diff
@@ -1,151 +1,121 @@
...
-FROM amd64-node-base as rpi-node-build
-FROM amd64-node-base as amd64-node-build
-FROM amd64-node-base as armv7hf-node-build
-FROM amd64-node-base as aarch64-node-build
-FROM amd64-node-base as i386-node-build
-FROM balenalib/amd64-node:6-build as i386-nlp-node-build
-##############################################################################
-
-FROM $ARCH-node-build as node-build
+ARG NODE_VERSION=10.18.0
```
Page-, line 2: v10.19.0 would make more sense:
```suggestion
ARG NODE_VERSION=10.19.0
```

Dockerfile
```diff
@@ -1,151 +1,121 @@
...
-	ln -s /lib rootfs-overlay/lib64
+	binutils \
+	libgcc \
+	libstdc++ \
+	libuv
 
-COPY package.json package-lock.json /usr/src/app/
+# Install node from baleana's prebuilt cache
+RUN curl -SLO "${NODE_LOCATION}" \
+	&& curl  -SLO "${S3_BASE}/${SUMS}" \
```
Page-, line 38: We should really include the SHAs locally, since they're meant to verify we get the expected file but downloading the SHAs from the same place as the binary only means that they didn't get corrupted in transport and says nothing about them containing what we expect

Dockerfile
```diff
@@ -1,151 +1,121 @@
...
 
 WORKDIR /usr/src/app
 
-COPY --from=node-build /usr/src/app/dist ./dist
-COPY --from=node-deps /build/node_modules ./node_modules
-COPY --from=node-deps /build/rootfs-overlay/ /
+COPY --from=BUILD /usr/src/app/dist ./dist
+COPY --from=BUILD /usr/src/app/package.json ./
+COPY --from=BUILD /usr/src/app/node_modules ./node_modules
+COPY --from=BUILD /usr/local/bin/node /usr/local/bin/node
```
Page-, line 98: It's probably worth putting this one first as it's the most likely to be cached

Dockerfile
```diff
@@ -1,151 +1,121 @@
...
+RUN mkdir -p rootfs-overlay && \
+	(([ ! -d rootfs-overlay/lib64 ] && ln -s /lib rootfs-overlay/lib64) || true)
 
 ARG ARCH
 ARG VERSION=master
 ARG DEFAULT_MIXPANEL_TOKEN=bananasbananas
 ENV CONFIG_MOUNT_POINT=/boot/config.json \
 	LED_FILE=/dev/null \
-	SUPERVISOR_IMAGE=resin/$ARCH-supervisor \
+	SUPERVISOR_IMAGE=balena/$ARCH-supervisor \
```
Page-, line 110: Will this conflict with balenaCI?

automation/build.sh
```diff
@@ -25,99 +24,100 @@
...
+MASTER_IMAGE=balena/$ARCH-supervisor:master
+MASTER_BUILD_IMAGE=balena/$ARCH-supervisor:master-build
 
 CACHE_FROM=""
 function useCache() {
 	image=$1
 	# Always add the cache because we can't do it from
 	# a subshell and specifying a missing image is fine
-	CACHE_FROM="$CACHE_FROM --cache-from $image"
+	CACHE_FROM="--cache-from $image ${CACHE_FROM}"
```
Page-, line 56: Any reason for the reordering?

Dockerfile
CameronDiver: What specifically are you worried about?

Page-: balenaCI usually tries to push toe `balena/` so we might get race conditions where both are trying to push to the same repo and so the actual contents could differ based on that

CameronDiver: Ah, this isn't actually used for a push, this is simply for the supervisor itself, and honestly it's not even really necessary anymore - I might just remove it when I ensure that.

automation/build.sh
CameronDiver: Was debugging a weird shell issue, I've reverted.

Dockerfile
CameronDiver: Done.

Dockerfile
```diff
@@ -1,151 +1,121 @@
...
-	-o -name '*.tar.*'      -o -name '*.in'     -o -name '*.cc' \
-	-o -name '*.c'          -o -name '*.coffee' -o -name '*.eslintrc' \
-	-o -name '*.h'          -o -name '*.html'   -o -name '*.markdown' \
-	-o -name '*.md'         -o -name '*.patch'  -o -name '*.png' \
-	-o -name '*.yml'        -o -name "*.ts" \
-	-delete \
-	&& find . -type f -path '*/node_modules/sqlite3/deps*' -delete \
-	&& find . -type f -path '*/node_modules/knex/build*' -delete \
-	&& rm -rf node_modules/sqlite3/node.dtps
+# Install node from baleana's prebuilt cache
```
richbayliss, line 37:
```suggestion
# Install node from balena's prebuilt cache
```

Dockerfile
```diff
@@ -1,151 +1,121 @@
...
-	SUPERVISOR_IMAGE=resin/$ARCH-supervisor \
+	SUPERVISOR_IMAGE=balena/$ARCH-supervisor \
 	VERSION=$VERSION \
 	DEFAULT_MIXPANEL_TOKEN=$DEFAULT_MIXPANEL_TOKEN
+COPY avahi-daemon.conf /etc/avahi/avahi-daemon.conf
 
+VOLUME /data
 HEALTHCHECK --interval=5m --start-period=1m --timeout=30s --retries=3 \
-	CMD wget -qO- http://127.0.0.1:${LISTEN_PORT:-48484}/v1/healthy || exit 1
+	CMD wget http://127.0.0.1:${LISTEN_PORT:-48484}/v1/healthy -O - -q
```
richbayliss, line 124: why no more `|| exit 1`?

README.md
```diff
@@ -1,24 +1,105 @@
...
 
-The supervisor is a Node.js program.
+Using the [balenaEngine](https://balena.io/engine)'s (our
+IoT-centric container engine) remote API, balenaSupervisor
+will install, start, stop and monitor IoT applications,
+delivered and ran as [OCI](https://www.opencontainers.org/)
+compliant containers.
 
-## Running a supervisor locally
+balenaSupervisor is developed using node.js.
```
richbayliss, line 14:
```suggestion
balenaSupervisor is developed using Node.js.
```

README.md
```diff
@@ -1,24 +1,105 @@
...
+
+- Every PR _should_ have an associated issue, and the PR's opening comment should say "Fixes #issue" or "Closes #issue".
+- We use [Versionist](https://github.com/resin-io/versionist) to manage versioning (and in particular, [semantic versioning](semver.org)) and generate the changelog for this project.
+- At least one commit in a PR should have a `Change-Type: type` footer, where `type` can be `patch`, `minor` or `major`. The subject of this commit will be added to the changelog.
+- Commits should be squashed as much as makes sense.
+- Commits should be signed-off (`git commit -s`)
+
+## Developing the supervisor
+
+By far the most convenient way to develop the supervisor is
```
richbayliss, line 44:
```suggestion
By far the most convenient way to develop the supervisor is to
```

README.md
```diff
@@ -92,43 +173,55 @@ additional options, for instance, to see the logs from the supervisor service:
...
+A production balena image does not have an open docker
+socket, required for livepush to work. In this situation,
+the `tools/sync.js` script can be used. Note that this
+process is no longer actively developed, so your mileage may
+vary.
 
-This will build the supervisor code and sync it onto the running supervisor container inside the device, and then restart it.
+Bug reports and pull requests are still accepted for changes
+to `sync.js`, but the balenaSupervisor team will focus on
+`npm run sync` in the future.
```
richbayliss, line 195: I would drop this bit entirely - we should follow this PR up with removing the `sync.js` method

Dockerfile
CameronDiver: It's not needed, wget will `exit 1` if there's no reply, or bad response.

README.md
CameronDiver: There is still a valid time to use it, that we don't have a replacement for though. It's super rare, but I'd still rather have it around.

CameronDiver:
@balena-ci retest

Dockerfile
Page-: I mean the `||` kind of implies it already exited with `1` (or some other failure code, but I'd expect they'd all be handled the same as `1` anyway)

tools/dind/sync.js
```diff
@@ -10,8 +10,8 @@ Usage:
 The script will first build a non-optimized version of the js code and sync the resulting app.js
```
Page-, line 10: I'm not sure why you moved this file to `tools/dind`, it's not dind specific in any way (and from memory not normally used for the dind supervisor at all)

sync/sync.ts
```diff
@@ -0,0 +1,88 @@
...
+		await setupLogs(docker, containerId);
+		await startLivepush({
+			dockerfile,
+			containerId,
+			docker,
+			noinit: true,
+		});
+	} catch (e) {
+		console.log('Error:');
+		console.log(e.message);
```
Page-, line 86: These should probably be `.error`

sync/logs.ts
```diff
@@ -0,0 +1,46 @@
...
+	// This happens when a container is restarted
+	stream.on('end', () => {
+		setupLogs(docker, containerId, lastReadTimestamp);
+	});
+}
+
+function extractMessage(msgBuf: Buffer) {
+	// Non-tty message format from:
+	// https://docs.docker.com/engine/api/v1.30/#operation/ContainerAttach
+	if (_.includes([0, 1, 2], msgBuf[0])) {
```
Page-, line 36:
```suggestion
	if ([0, 1, 2,.includes(msgBuf[0])) {
```

sync/init.ts
```diff
@@ -0,0 +1,56 @@
...
+	dockerfile: Dockerfile;
+	nocache: boolean;
+	arch?: string;
+}
+
+export async function initDevice(opts: Opts) {
+	let arch = opts.arch;
+	if (!arch) {
+		arch = await device.getDeviceArch(opts.docker);
+	}
```
Page-, line 21:
```suggestion
	let arch = opts.arch ?? await device.getDeviceArch(opts.docker);
```

sync/device.ts
```diff
@@ -0,0 +1,177 @@
...
+export async function getCacheFrom(docker: Docker): Promise<string[]> {
+	const container = await getSupervisorContainer(docker);
+	return [container.Image];
+}
+
+// perform the build and return the image id
+export async function performBuild(
+	docker: Docker,
+	dockerfile: Dockerfile,
+	dockerOpts: { [key: string]: any },
```
Page-, line 69: Doesn't dockerode expose a type for this?

sync/device.ts
```diff
@@ -0,0 +1,177 @@
+import * as Docker from 'dockerode';
+import { Dockerfile } from 'livepush';
+import * as _ from 'lodash';
+import { Builder } from 'resin-docker-build';
+
+import { child_process, fs } from 'mz';
```
Page-, line 6: You can do
```suggestion
import { child_process } from 'mz';
import { promises as fs } from 'fs';
```
now that node 10 is the minimum

tools/dind/sync.js
CameronDiver: oh it's meant to be tools/ only, that's weird

sync/device.ts
CameronDiver: Unfortunately not.

Page-: You can get the types via
```suggestion
	dockerOpts: Parameters<typeof builder.createBuildStream>[0],
```

PR #1243, opened 2020-04-01
Add newTargetState event and use it for backup loading

CameronDiver:


src/api-binder.ts
```diff
@@ -578,6 +583,10 @@ export class APIBinder {
 		return Bluebird.using(this.lockGetTarget(), async () => {
 			const targetState = await this.getTargetState();
 			if (isFromApi || !_.isEqual(targetState, this.lastTarget)) {
+				// If the target state has changed, we emit an event
+				// with it
+				globalEventBus.getInstance().emit('newTargetState', targetState);
```
richbayliss, line 588:
```suggestion
				globalEventBus.getInstance().emit('targetStateChanged', targetState);
```

CameronDiver: lgtm

src/device-api/v2.ts
```diff
@@ -291,10 +292,14 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
...
 
+		// FIXME: We need to validate the target state here,
+		// before emitting it
+
 		// Now attempt to set the state
 		const force = req.body.force;
 		const targetState = req.body;
 		try {
+			globalEventBus.getInstance().emit('targetStateChanged', targetState);
 			await deviceState.setTarget(targetState, true);
```
Page-, line 303: Shouldn't the `deviceState.setTarget` do the emit? It seems like the more sensible place to put it because then it'll handle it for all callsites

src/api-binder.ts
```diff
@@ -578,6 +583,10 @@ export class APIBinder {
 		return Bluebird.using(this.lockGetTarget(), async () => {
 			const targetState = await this.getTargetState();
 			if (isFromApi || !_.isEqual(targetState, this.lastTarget)) {
+				// If the target state has changed, we emit an event
+				// with it
+				globalEventBus.getInstance().emit('targetStateChanged', targetState);
+
 				await this.deviceState.setTarget(targetState);
```
Page-, line 590: Same here for `deviceState.setTarget` handling the emit

src/api-binder.ts
```diff
@@ -207,11 +207,16 @@ export class APIBinder {
...
 
-		await loadBackupFromMigration(
-			this.deviceState,
-			await this.getTargetState(),
-			bootstrapRetryDelay,
-		);
+		// When we've provisioned, try to load the backup. We
+		// must wait for the provisioning because we need a
+		// target state on which to apply the backup
+		globalEventBus.getInstance().once('targetStateChanged', async state => {
```
Page-, line 213: I'd prefer this as
```suggestion
		globalEventBus.getSingleton().once('targetStateChanged', async state => {
```
because right now it's a bit misleading in thinking there can be multiple instances potentially

CameronDiver: We're bikeshedding now, and `getInstance` is pretty standard. There's nothing that says this has to be a singleton either, that's just how it's implemented right now.

src/device-api/v2.ts
CameronDiver: Nice, good point.

src/api-binder.ts
Page-: It's not much of a "global" event bus if it's not a singleton :P

PR #1246, opened 2020-04-02
Don't wrap UpdatesLockedErrors with a detailed error

CameronDiver:
We do this as there is tooling around the other parts of the engine which works upon these errors.

Change-type: patch

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -774,17 +774,16 @@ export class DeviceState extends (EventEmitter as new () => DeviceStateEventEmit
 					retryCount,
 				});
 			} catch (e) {
-				const detailedError = new Error(
+				if (e instanceof UpdatesLockedError) {
+					// Forward the UpdatesLockedError directly
+					throw e;
+				}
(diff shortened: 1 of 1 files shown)


src/device-state.ts
```diff
@@ -774,17 +774,15 @@ export class DeviceState extends (EventEmitter as new () => DeviceStateEventEmit
...
-				});
+				const detailedError = !(e instanceof UpdatesLockedError)
+					? new Error(
+							'Failed to apply state transition steps. ' +
+								e.message +
+								' Steps:' +
+								JSON.stringify(_.map(steps, 'action')),
+					  )
+					: e;
+				throw detailedError;
```
Page-, line 785: I think it'd be clearer as:
```suggestion
				 if (e instanceof UpdatesLockedError) {
					throw e;
				 }
				throw new Error(
					'Failed to apply state transition steps. ' +
						e.message +
						' Steps:' +
						JSON.stringify(_.map(steps, 'action')),
				);
```

PR #1250, opened 2020-04-06
Convert application-manager.coffee to javascript

Page-:
Change-type: patch


src/supervisor.ts
```diff
@@ -40,25 +40,24 @@ export class Supervisor {
...
 			eventTracker: this.eventTracker,
 			logger: this.logger,
+			apiBinder: this.apiBinder,
 		});
 
 		// FIXME: rearchitect proxyvisor to avoid this circular dependency
 		// by storing current state and having the APIBinder query and report it / provision devices
 		this.deviceState.applications.proxyvisor.bindToAPI(this.apiBinder);
-		// We could also do without the below dependency, but it's part of a much larger refactor
-		this.deviceState.applications.apiBinder = this.apiBinder;
```
Page-, line 61: I needed to refactor this so that ApplicationManager knew about its `apiBinder` property otherwise it wouldn't exist in the inferred types which caused a bunch of headaches, so that's where all the `apiBinder` related changes come from (required by improved types)

src/lib/validation.ts
```diff
@@ -22,7 +22,7 @@ const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
 export function checkInt(
 	s: unknown,
 	options: CheckIntOptions = {},
-): number | void {
+): number | undefined {
```
Page-, line 25: This allows for `== null` / `??` checks if `checkInt` fails

src/device-state/preload.ts
```diff
@@ -65,7 +65,7 @@ export async function loadTargetFromFile(
 					releaseId: app.releaseId,
 					appId,
 				};
-				images.push(await deviceState.applications.imageForService(svc));
+				images.push(deviceState.applications.imageForService(svc));
```
Page-, line 68: Turns out this doesn't return a promise

src/compose/images.ts
```diff
@@ -38,7 +38,7 @@ interface FetchProgressEvent {
 }
 
 export interface Image {
```
Page-, line 40: The changes to this are because of `imageForService` returning an "Image" without these properties

src/application-manager.js
```diff
@@ -0,0 +1,1727 @@
...
+
+import { validateTargetContracts } from './lib/contracts';
+import { DockerUtils as Docker } from './lib/docker-utils';
+import { LocalModeManager } from './local-mode';
+import * as updateLock from './lib/update-lock';
+import { checkTruthy, checkInt, checkString } from './lib/validation';
+import {
+	ContractViolationError,
+	NotFoundError,
+	InternalInconsistencyError,
```
Page-, line 20: I added a bunch of these `InternalInconsistencyError` in order to check for `null`s that the coffeescript version assumed would not exist, you may want to double check them but in all cases I saw I think it would have errored anyway but with a worse message

entry.sh
```diff
@@ -55,7 +55,7 @@ modprobe ip6_tables || true
 
 if [ "${LIVEPUSH}" -eq "1" ]; then
 	exec npx nodemon --watch src --watch typings --ignore tests \
-		 --exec node -r ts-node/register/transpile-only -r coffeescript/register src/app.ts
+		 --exec node -r ts-node/register/transpile-only src/app.ts
```
CameronDiver, line 58: :raised_hands:

PR #1251, opened 2020-04-07
🚀 Update supervisor to node12

CameronDiver:
This change needs more testing before merge.

Change-type: patch


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
 	"compilerOptions": {
+		"target": "ES2019",
```
Page-, line 3:
```suggestion
		"target": "es2019",
```

PR #1253, opened 2020-04-07
Don't sync test files with livepush

CameronDiver:
Change-type: patch

### DIFF
--- a/sync/livepush.ts
+++ b/sync/livepush.ts
@@ -37,7 +37,7 @@ export async function startLivepush(opts: {
 
 	chokidar
 		.watch('.', {
-			ignored: /((^|[\/\\])\..|node_modules.*|sync\/.*)/,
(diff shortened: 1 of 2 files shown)


sync/livepush.ts
```diff
@@ -37,7 +37,7 @@ export async function startLivepush(opts: {
 
 	chokidar
 		.watch('.', {
-			ignored: /((^|[\/\\])\..|node_modules.*|sync\/.*)/,
+			ignored: /((^|[\/\\])\..|node_modules.*|sync\/.*|test\/.*)/,
```
zvin, line 40: Why do we expect a slash after `sync` or `test` but not after `node_modules` ?

CameronDiver: I specifically didn't want files that had sync or test in the name to be excluded. I guess the same goes for node_modules but it's less likely. I'll just add it to match.

PR #1258, opened 2020-04-08
Link sqlite with a system sqlite for quicker builds

CameronDiver:
A fresh build goes from 1:40 -> ~30 mins

Change-type: patch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,9 @@ RUN apk add --no-cache \
 	binutils \
 	libgcc \
 	libstdc++ \
-	libuv
+	libuv \
+	sqlite-libs \
+	sqlite-dev
 
 COPY build-conf/node-sums.txt .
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -43,7 +45,7 @@ RUN curl -SLO "${NODE_LOCATION}" \
 
 COPY package*.json ./
 
-RUN npm ci
+RUN npm ci --build-from-source --sqlite=/usr/lib
```
Page-, line 48: Is it possible to only build sqlite from source? There might be other modules you're now unnecessarily building from source and avoiding those would make sense. In fact maybe you can just omit the `--build-from-source` altogether since if we can use the prebuilt sqlite that should be fine?

Page-:
@balena-ci rebase

Page-:
@balena-ci rebase

PR #1261, opened 2020-04-09
Use livepush commands for copying and running dev specific steps

CameronDiver:
Change-type: patch

### DIFF
--- a/sync/livepush.ts
+++ b/sync/livepush.ts
@@ -40,16 +40,18 @@ export async function startLivepush(opts: {
 			ignored: /((^|[\/\\])\..|(node_modules|sync|test)\/.*)/,
 			ignoreInitial: opts.noinit,
 		})
-		.on('add', path => livepushExecutor(path))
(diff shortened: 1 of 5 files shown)


sync/livepush.ts
```diff
@@ -40,16 +40,18 @@ export async function startLivepush(opts: {
...
 
 const getExecutor = (livepush: Livepush) => {
-	const changedFiles: string[] = [];
-	const deletedFiles: string[] = [];
+	let changedFiles: string[] = [];
+	let deletedFiles: string[] = [];
 	const actualExecutor = _.debounce(async () => {
 		await livepush.performLivepush(changedFiles, deletedFiles);
+		changedFiles = [];
+		deletedFiles = [];
```
Page-, line 54: If you just want to empty the existing ones you can use
```suggestion
		changedFiles.length = 0;
		deletedFiles = 0;
```
which is better in terms of allocations

PR #1262, opened 2020-04-09
Move from dbus-native to dbus

### LINKED ISSUE #1049: Supervisor prints  a new Buffer deprecation warning with node10
This comes from the following: https://github.com/sidorares/dbus-native/issues/271#issuecomment-580262777

The fix it seems would be to move away from dbus-native. A possible replacement is https://github.com/Shouqun/node-dbus which does not rely on `put`.

CameronDiver:
Mainly this is to fix #1049 but dbus-native is not very well maintained either: https://github.com/sidorares/dbus-native/issues/271

Closes: #1049
Change-type: patch


src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
+}
 
-export function restartService(serviceName: string) {
-	return systemdManagerMethodCall('RestartUnit', 'ss', [
-		`${serviceName}.service`,
-		'fail',
-	]);
+export async function restartService(serviceName: string) {
+	try {
+		const systemd = await getInterface();
```
zvin, line 38: This `getInterface` call should be moved outside of the `try`, otherwise it can throw a `new DbusError(new DbusError(e))`.

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
 }
 
-export function startService(serviceName: string) {
-	return systemdManagerMethodCall('StartUnit', 'ss', [
-		`${serviceName}.service`,
-		'fail',
-	]);
+export async function startService(serviceName: string) {
+	try {
+		const systemd = await getInterface();
```
zvin, line 47: Same comment as above: double DBusError

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
 }
 
-export function stopService(serviceName: string) {
-	return systemdManagerMethodCall('StopUnit', 'ss', [
-		`${serviceName}.service`,
-		'fail',
-	]);
+export async function stopService(serviceName: string) {
+	try {
+		const systemd = await getInterface();
```
zvin, line 56: same

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
 
-export function enableService(serviceName: string) {
-	return systemdManagerMethodCall('EnableUnitFiles', 'asbb', [
-		[`${serviceName}.service`],
-		false,
-		false,
-	]);
+export async function enableService(serviceName: string) {
+	try {
+		const systemd = await getInterface();
```
zvin, line 65: same

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
 }
 
-export function disableService(serviceName: string) {
-	return systemdManagerMethodCall('DisableUnitFiles', 'asb', [
-		[`${serviceName}.service`],
-		false,
-	]);
+export async function disableService(serviceName: string) {
+	try {
+		const systemd = await getInterface();
```
zvin, line 74: same

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
-		() =>
-			systemdManagerMethodCall('PowerOff', 'b', [false], {
-				path: '/org/freedesktop/login1',
-				destination: 'org.freedesktop.login1',
-				interface: 'org.freedesktop.login1.Manager',
-			}),
-		1000,
-	),
+	setTimeout(async () => {
+		const systemd = await getInterface({
```
zvin, line 94: NIT: we could have a getLoginManager method

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
+export async function disableService(serviceName: string) {
+	try {
+		const systemd = await getInterface();
+		systemd.DisableUnitFiles([`${serviceName}.service`], false);
+	} catch (e) {
+		throw new DbusError(e);
+	}
 }
 
 export const reboot = Bluebird.method(() =>
```
zvin, line 81: Why is `Bluebird.method`needed here ?
Also this returns undefined and the caller seems to be awaiting on it.
This could be 
```typescript
await delay(1000);
const systemd = await ...
systemd.Reboot(false);
```
and the await in the caller can be removed.

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
+		const systemd = await getInterface({
+			path: '/org/freedesktop/login1',
+			destination: 'org.freedesktop.login1',
+			interface: 'org.freedesktop.login1.Manager',
+		});
+		systemd.Reboot(false);
+	}, 1000),
 );
 
 export const shutdown = Bluebird.method(() =>
```
zvin, line 92: Same remark as for `reboot`

webpack.config.js
```diff
@@ -19,6 +19,7 @@ var externalModules = [
 	'oracledb',
```
Page-, line 19: If you're going to lint this file it would make sense to also remove the special casing for it from .lintstagedrc

.lintstagedrc
```diff
@@ -5,7 +5,7 @@
 	"*.ts": [
 		"balena-lint --typescript --fix",
 	],
-	"*!(*webpack.config).js": [
+	"*.js": [
```
zvin, line 8: NIT: You can have typescript webpack config files

CameronDiver: That's cool, we're not specifically targeting it here though, we just aren't not-targeting it :P

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
-		destination: string;
-		interface: string;
-	} = defaultPathInfo,
-) => {
-	if (signature == null) {
-		signature = '';
+const bus = dbus.getBus('system');
+const getInterfaceAsync = Bluebird.promisify(bus.getInterface, {
+	context: bus,
+});
```
zvin, line 12: NIT: you could use something like this to avoid repeating the try/catch below:

```typescript
async function getInterface(...args: Parameters<getInterfaceAsync>) {
    try {
        return await getInterfaceAsync(...args);
    } catch (e) {
        throw new DbusError(e);
    }
}
```

src/lib/systemd.ts
```diff
@@ -1,113 +1,130 @@
...
+		throw new DbusError(e);
 	}
-	return invokeAsync({
-		...pathInfo,
-		member: method,
-		signature,
-		body,
-	});
-};
+}
```
zvin, line 37: A helper like this could make the functions below shorter:

```typescript
async function callSystemdMethod(serviceName: string, ...args: any[]) {
    const systemd = await getSystemdInterface();
    try {
        systemd[serviceName](...args);
    } catch (e) {
        throw new DbusError(e);
    }
}
```

Also, maybe `const systemd = await getSystemdInterface();` could be called only once and stored somewhere.

PR #1265, opened 2020-04-14
Keep locks on purge

### LINKED ISSUE #1164: Deleting wifi credentials during purge breaks device if update lock is set

mwohlert:
When purging the device keep the preexisting locks, when using force in place to make sure, that the locks are able to persist through the purge.
This prevents a scheduled update from happening after the purge leading to non deterministic behavior.

Fixes #1164


mwohlert:
@CameronDiver I redid the PR and tested it on my raspberry pi 3.

CameronDiver:
@mwohlert this still keeps on spuriously taking the lock for me. To test this, have an application which does not take the lock, and push another update. 
```
[debug]   Replacing container for service main because of config changes:
[debug]     Non-array fields:  {"added":{},"deleted":{},"updated":{"environment":{"BALENA_SUPERVISOR_VERSION":"11.1.3","RESIN_SUPERVISOR_VERSION":"11.1.3"}}}
[info]    Updates are locked, retrying in ~1s...
[info]    Applying target state
[debug]   Replacing container for service main because of config changes:
[debug]     Non-array fields:  {"added":{},"deleted":{},"updated":{"environment":{"BALENA_SUPERVISOR_VERSION":"11.1.3","RESIN_SUPERVISOR_VERSION":"11.1.3"}}}
[info]    Updates are locked, retrying in ~2s...
```

I made sure there wasn't a lockfile present before this, so it looks like the change is adding one when there wasn't previously a lockfile there.

mwohlert:
@CameronDiver I tried replicating this but for me if there is no lock phyisically on the disk the update won't be blocked.
Would you mind sharing the application with me so I can reproduce the issue?

PR #1266, opened 2020-04-14
Upgrade knex to avoid CVE-2019-10757 

CameronDiver:
This PR replaces https://github.com/balena-io/balena-supervisor/pull/1124


src/proxyvisor.js
```diff
@@ -693,14 +693,12 @@ export class Proxyvisor {
...
-			devices: this.db
-				.models('dependentDeviceTarget')
-				.select()
-				.map(this.normaliseDependentDeviceTargetFromDB),
+			apps: Promise.resolve(this.db.models('dependentAppTarget').select()).map(
+				this.normaliseDependentAppFromDB,
+			),
+			devices: Promise.resolve(
+				this.db.models('dependentDeviceTarget').select(),
+			).map(this.normaliseDependentDeviceTargetFromDB),
```
Page-, line 701:
```suggestion
			devices: Promise.map(
				this.db.models('dependentDeviceTarget').select(),
				this.normaliseDependentDeviceTargetFromDB
			),
```

src/proxyvisor.js
```diff
@@ -693,14 +693,12 @@ export class Proxyvisor {
...
-				.models('dependentAppTarget')
-				.select()
-				.map(this.normaliseDependentAppFromDB),
-			devices: this.db
-				.models('dependentDeviceTarget')
-				.select()
-				.map(this.normaliseDependentDeviceTargetFromDB),
+			apps: Promise.resolve(this.db.models('dependentAppTarget').select()).map(
+				this.normaliseDependentAppFromDB,
+			),
```
Page-, line 698:
```suggestion
			apps: Promise.map(
				this.db.models('dependentAppTarget').select(),
				this.normaliseDependentAppFromDB,
			),
```

src/proxyvisor.js
```diff
@@ -510,16 +510,16 @@ export class Proxyvisor {
...
 
 	getCurrentStates() {
 		return Promise.join(
-			this.db
-				.models('dependentApp')
-				.select()
-				.map(this.normaliseDependentAppFromDB),
+			Promise.resolve(this.db.models('dependentApp').select()).map(
+				this.normaliseDependentAppFromDB,
+			),
```
Page-, line 522:
```suggestion
			Promise.map(
				this.db.models('dependentApp').select(),
				this.normaliseDependentAppFromDB,
			),
```

src/migrations/20181907164000-endpoint-specific-logschannel.js
```diff
@@ -1,8 +1,9 @@
+const Bluebird = require('bluebird');
 const fs = require('fs');
 const configJsonPath = process.env.CONFIG_MOUNT_POINT;
 
-exports.up = function(knex, Promise) {
-	return new Promise(resolve => {
+exports.up = function(knex) {
+	return new Bluebird(resolve => {
```
Page-, line 6: You could keep this as 
```suggestion
	return new Promise(resolve => {
```
and just return a native promise - they're definitely supported in node 12

src/db.ts
```diff
@@ -27,42 +26,38 @@ export class DB {
...
-			})
-			.then(() => {
-				return this.knex.migrate.latest({
-					directory: path.join(__dirname, 'migrations'),
-				});
-			});
+	public async init(): Promise<void> {
+		try {
+			await this.knex('knex_migrations_lock').update({ is_locked: 0 });
+		} catch (e) {
```
Page-, line 32:
```suggestion
		} catch {
```

CameronDiver: TIL you can omit the parameters, nice.

src/migrations/20181907164000-endpoint-specific-logschannel.js
CameronDiver: This is for the `.tap` lower down. I could convert these to async/await, but I'd rather leave the existing migrations alone tbh.

src/db.ts
```diff
@@ -14,7 +13,7 @@ export type Transaction = Knex.Transaction;
 
 export class DB {
 	private databasePath: string;
-	private knex: Knex;
+	public knex: Knex;
```
zvin, line 16: NIT: Why make it public ?

src/migrations/20181907164000-endpoint-specific-logschannel.js
Page-: Ah I couldn't see it, fair enough

src/db.ts
Page-: I also wonder about this @CameronDiver ?

CameronDiver: Ah, leftover debugging. Good catch.

PR #1268, opened 2020-04-17
Remove unnecessary config.json keys

Page-:
Change-type: patch


circle.yml
```diff
@@ -58,7 +58,7 @@ version: 2
 jobs:
   generic:
     docker:
-      - image: balenalib/amd64-alpine-node:10
+      - image: balenalib/amd64-alpine-node:12
```
Page-, line 61: This is required for the tests to pass since the supervisor requires node 12 now - no idea how this didn't fail on previous PRs

Page-, line 61: Not strictly necessary as it turns out but it makes sense to run the tests on node 12 since that's what we run on

Page-:
@balena-ci rebase

circle.yml
20k-ultra: When I first started using supervisor I had some issues when running on node 10 so I think this is a good change.

20k-ultra:
Commit is missing Signed-off-by footer. I guess none of the CI jobs caught that?

Page-:
@balena-ci rebase

Page-:
Depends on #1309 to fix the CI build

Issue #1249, opened 2020-04-03
Support api key in http header


nghiant2710:
Currently, the supervisor API only support `ApiKey` header (https://github.com/balena-io/balena-supervisor/blob/20e339317e53a592e17f9936bc70b36d8dd17a4b/src/supervisor-api.ts#L15-L27).

Can we support `Bearer <API_KEY>` header (the same as Authorization header for the API) as well?

PR #1279, opened 2020-04-28
Support Bearer scheme in Authorization header

20k-ultra:
Closes: #1249
Change-type: minor

### DIFF
--- a/src/supervisor-api.ts
+++ b/src/supervisor-api.ts
@@ -12,18 +12,21 @@ import { checkTruthy } from './lib/validation';
 
 import log from './lib/supervisor-console';
 
-function getKeyFromReq(req: express.Request): string | null {
-	const queryKey = req.query.apikey;
-	if (queryKey != null) {
-		return queryKey;
+function getKeyFromReq(req: express.Request): string | undefined {
+	// Check query for key
+	if (req.query.apikey) {
+		return req.query.apikey;
 	}
-	const maybeHeaderKey = req.get('Authorization');
-	if (!maybeHeaderKey) {
-		return null;
+	// Get Authorization header to search for key
+	const authHeader = req.get('Authorization');
+	// Check header for key
+	if (!authHeader) {
+		return undefined;
 	}
-
-	const match = maybeHeaderKey.match(/^ApiKey (\w+)$/);
-	return match != null ? match[1] : null;
+	// Check authHeader with various schemes
+	const match = authHeader.match(/^(?:ApiKey|Bearer) (\w+)$/i);
+	// Return key from match or undefined
+	return match?.[1];
 }
 
 function authenticate(config: Config): express.RequestHandler {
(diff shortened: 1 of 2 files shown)


20k-ultra:
While I was developing this feature I used a different test script so I could run 1 test file quickly. I also periodically would run the normal test command mentioned in the README and also passed.

The different script can be found https://github.com/balena-io/balena-supervisor/issues/1278#issuecomment-620321950.

When I stash the changes to my package.json and run `npm test` my tests fail. When I stash apply the package.json with my script `npm test` passes. It's really bizarre. I'll spend some more time looking into this.

src/supervisor-api.ts
```diff
@@ -13,17 +13,24 @@ import { checkTruthy } from './lib/validation';
...
+	const authHeader = req.get('Authorization');
+	// Check header for key
+	if (!authHeader) {
 		return null;
 	}
-
-	const match = maybeHeaderKey.match(/^ApiKey (\w+)$/);
-	return match != null ? match[1] : null;
+	// Check authHeader with various schemes
+	return ['ApiKey', 'Bearer'].reduce((apiKey, scheme) => {
```
CameronDiver, line 27: _.find would be clearer here

Page-: I think it'd make more sense to match what was here before:
```suggestion
	const match = authHeader.match(/^(?:ApiKey|Bearer) (\w+)$/);
	return match?.[1];
```
although it might also make sense to make it case-insensitive to make life easier:
```suggestion
	const match = authHeader.match(/^(?:ApiKey|Bearer) (\w+)$/i);
	return match?.[1];
```

src/supervisor-api.ts
```diff
@@ -12,7 +12,7 @@ import { checkTruthy } from './lib/validation';
 
 import log from './lib/supervisor-console';
 
-function getKeyFromReq(req: express.Request): string | null {
+function getKeyFromReq(req: express.Request): string | undefined {
```
20k-ultra, line 15: The return value is only used in 1 place in the codebase and it just checks if it's truthy before comparing to another string so switching to undefined will result in the same behavior with the benefit of being able to use match?.[1] which returns undefined if match is falsey.

src/supervisor-api.ts
```diff
@@ -87,6 +90,14 @@ export class SupervisorAPI {
 
 	private api = express();
 	private server: Server | null = null;
+	// Holds the function which should apply iptables rules
+	private applyRules: SupervisorAPI['applyListeningRules'] =
+		process.env.TEST === '1'
+			? () => {
+					// don't try to alter iptables
```
CameronDiver, line 97: NIT: The indentation is super weird here, can you make it indented from the opening brace by one please please? I'm not sure why prettier didn't change this

20k-ultra: Will add this change before merging.

Issue #1242, opened 2020-04-01
The supervisor should add a random offset to it's update interval


CameronDiver:
This update should be stored, and applied to every update delay. The offset should be in the range:
```
[0, updateInterval]
```
 This will define a point in time that all devices will update at:
```
n * UPDATE_INTERVAL + offset
```
This should be recalculated on interval change. This then makes the `updateInterval` a minimum value, and the maximum is `2*updateInterval`.

PR #1281, opened 2020-04-28
Add a random constant offset to every poll timeout

CameronDiver:
This allows us to space out the requests from a fleet, but also ensures
that the pollInterval is a minimum, rather than the centre value between
0.5* and 1.5*

Closes: #1242
Change-type: patch

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -608,9 +608,10 @@ export class APIBinder {
 			instantUpdates,
 		} = await this.config.getMany(['appUpdatePollInterval', 'instantUpdates']);
 
-		// We add jitter to the poll interval so that it's between 0.5 and 1.5 times
-		// the configured interval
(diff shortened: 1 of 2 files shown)


src/config/index.ts
```diff
@@ -305,26 +305,40 @@ export class Config extends (EventEmitter as new () => ConfigEventEmitter) {
...
+		if (unmanaged) {
+			return;
+		}
+
+		if (!deviceApiKey) {
+			await this.set({ deviceApiKey: this.newUniqueKey() });
+		}
+
+		if (!pollIntervalOffset) {
+			await this.set({ pollIntervalOffset: _.random(appUpdatePollInterval) });
```
Page-, line 340: Am I reading that the offset is going from a random value up to +/-50% of the poll interval calculated per poll, to a random value up to +100% calculated once? So this means a device set at 15m poll interval could in reality use a fixed 30m poll interval, or for a 24h poll interval it could be 48h?

PR #1285, opened 2020-04-29
Added more documentation to help new contributors start developing

20k-ultra:
Change-type: patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,22 @@ Here's a few guidelines to make the process easier for everyone involved.
 - Commits should be squashed as much as makes sense.
 - Commits should be signed-off (`git commit -s`)
 
+## Setup
+
+To get the codebase setup on your development machine follow these steps. For running the supervisor on a device see [Developing the supervisor](#developing-the-supervisor) or [Using balenaOS-in-container](#using-balenaos-in-container).
+
+```sh
+# Clone the repo
+git clone git@github.com:balena-io/balena-supervisor.git
+
+# Install dependencies
+npm ci
+```
+
+We explicitly use `npm ci` over `npm install` to ensure the correct package versions are installed. More documentation for this can be found [here](https://docs.npmjs.com/cli/ci) on the npm cli docs.
+
+You're now ready to start developing. If you get stuck at some point please reference the [troubleshooting](#troubleshooting) section before creating an issue.
+
 ## Developing the supervisor
 
 By far the most convenient way to develop the supervisor is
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -231,13 +247,28 @@ You can run some unit tests with:
...
 which ensures that the environment is exactly the same.
 
+## Troubleshooting
+
+Make sure you are running atleast:
+
+```sh
+node -v       # >= 12.16.2
+npm -v        # >= 6.14.0
+docker -v     # >= 19.03.0
```
CameronDiver, line 264: There is no need to specify docker and git

README.md
```diff
@@ -231,13 +247,28 @@ You can run some unit tests with:
...
+git --version # >= 2.13.0
+```
+
+If you experience errors running any of the steps mentioned here and, don't know how to proceed, try deleting your `node_modules` folder and reinstalling the packages.
```
CameronDiver, line 268: I'd really prefer this was not in the README

README.md
20k-ultra: I saw logs from Husky saying I need Git >= 2.13 when installing dependencies. I added docker since this was in the troubleshooting section so it was helpful to add versions that are confirmed to be working.

```sh
> husky@4.2.3 install /home/mig/Projects/balena.io/balena-supervisor/node_modules/husky
> node husky install

husky > Setting up git hooks
Husky requires Git >=2.13.0. Got v2.11.0.
husky > Failed to install
```

README.md
20k-ultra: It was a technique that resolved weird issues for me twice now so I thought it was worth adding but I understand it's a weird tip for troubleshooting.

CameronDiver: The reason I'm not keen on adding it is it implies that we don't control our module versions, but this isn't true, and in fact I think maybe it's worth rewording it to something like "ensure you use `npm ci`" or something as I expect this is the issue that you were hitting.

README.md
CameronDiver: Fair enough, although that docker version for sure isn't required, we would work with *much* earlier versions. I'd prefer to simply specify that docker is required, or to add the actual version of docker that would support things, but that's too much of a yakshave IMO.

README.md
20k-ultra: I've changed the wording for this to remind people to use npm ci instead of deleting modules folder.

README.md
20k-ultra: Removed docker version since it would just complicate this PR to find out the oldest version needed.

Issue #1277, opened 2020-04-27
Move misleading log message


20k-ultra:
The [applyListeningRules](https://github.com/balena-io/balena-supervisor/blob/40f7ba6b2f603131975eb87f8c6f2f5bdd57b5b8/src/supervisor-api.ts#L197) function's catch block logs "stopping API". 

https://github.com/balena-io/balena-supervisor/blob/40f7ba6b2f603131975eb87f8c6f2f5bdd57b5b8/src/supervisor-api.ts#L202-L216

This is misleading since 

https://github.com/balena-io/balena-supervisor/blob/40f7ba6b2f603131975eb87f8c6f2f5bdd57b5b8/src/supervisor-api.ts#L219-L223

is what performs that action and within that function it only stops the service if the server is not null. 

Since the function that calls applyListeningRules is before the server value being set it always returns null therefore not actually stopping the server despite the log saying the API is being stopped.

I purpose that the "stopping API." section simply be moved to the stop function.

CameronDiver:
It's better to report that we're not starting up the API. This has happened in the past, before these logs existed, and users had no idea why the supervisor wasn't responding to endpoints. Definitely the wording could be clearer, but let's make sure users know what's going on.

PR #1286, opened 2020-04-30
Move SupervisorAPI state change logs to appropriate functions

20k-ultra:
Closes: #1277
Change-type: patch

### DIFF
--- a/src/supervisor-api.ts
+++ b/src/supervisor-api.ts
@@ -82,6 +82,10 @@ interface SupervisorAPIConstructOpts {
 	healthchecks: Array<() => Promise<boolean>>;
 }
 
+interface SupervisorAPIStopOpts {
+	errored: boolean;
+}
+
 export class SupervisorAPI {
 	private config: Config;
 	private eventTracker: EventTracker;
@@ -200,8 +204,15 @@ export class SupervisorAPI {
 			}
 		});
 
-		this.server = this.api.listen(port);
-		this.server.timeout = apiTimeout;
+		return new Promise(resolve => {
+			this.server = this.api.listen(port, () => {
+				log.info(`Supervisor API successfully started on port ${port}`);
+				if (this.server) {
+					this.server.timeout = apiTimeout;
+				}
+				return resolve();
+			});
+		});
 	}
 
 	private async applyListeningRules(
@@ -218,17 +229,25 @@ export class SupervisorAPI {
 				log.debug('Supervisor API listening on allowed interfaces only');
 			}
 		} catch (err) {
-			log.error(
-				'Error on switching supervisor API listening rules - stopping API.\n',
-				err,
-			);
-			this.stop();
+			log.error('Error on switching supervisor API listening rules', err);
+			return this.stop({ errored: true });
 		}
 	}
 
-	public stop() {
+	public async stop(options?: SupervisorAPIStopOpts): Promise<void> {
 		if (this.server != null) {
-			this.server.close();
+			return new Promise((resolve, reject) => {
+				this.server?.close((err: Error) => {
+					if (err) {
+						log.error('Failed to stop Supervisor API');
+						return reject(err);
+					}
+					options?.errored
+						? log.error('Stopped Supervisor API')
+						: log.info('Stopped Supervisor API');
+					return resolve();
+				});
+			});
 		}
 	}
 }
--- a/test/21-supervisor-api.spec.ts
+++ b/test/21-supervisor-api.spec.ts
@@ -1,0 +1,110 @@
+import { expect } from 'chai';
+import { fs } from 'mz';
+import { spy } from 'sinon';
+
+import Config from '../src/config';
+import Database from '../src/db';
+import EventTracker from '../src/event-tracker';
+import Log from '../src/lib/supervisor-console';
+import SupervisorAPI from '../src/supervisor-api';
+
+const mockedOptions = {
+	listenPort: 12345,
+	timeout: 30000,
+	dbPath: './test/data/supervisor-api.sqlite',
+};
+
+const ALLOWED_INTERFACES = ['lo']; // Only need loopback since this is for testing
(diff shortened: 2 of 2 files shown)


20k-ultra:
I created this PR as a draft first because it is failing the tests but I wanted to get feedback on the code changes. The code failing is related to the test file and not the code changes I made. When I run `npm test` it fails but passes when I was developing with `npm run test:fast -- --grep "SupervisorAPI". Will have update and request review when passing.

src/supervisor-api.ts
```diff
@@ -79,6 +79,10 @@ interface SupervisorAPIConstructOpts {
 	healthchecks: Array<() => Promise<boolean>>;
 }
 
+interface SupervisorAPIStopOpts {
```
20k-ultra, line 82: I created this interface so in the future if we want to pass more options for when we are closing the SupervisorAPI we can. Otherwise I could have just passed a boolean to the close function but thought this was more robust.

src/supervisor-api.ts
```diff
@@ -176,22 +188,29 @@ export class SupervisorAPI {
...
 					changedConfig.localMode || false,
 					port,
 					allowedInterfaces,
 				);
 			}
 		});
 
-		this.server = this.api.listen(port);
-		this.server.timeout = apiTimeout;
+		return new Promise(resolve => {
```
20k-ultra, line 205: the listen function offers a callback which wasn't being used. I refactored this to return a promise so code calling this can wait for the function to fully finish.

src/supervisor-api.ts
```diff
@@ -208,17 +227,28 @@ export class SupervisorAPI {
...
-				err,
-			);
-			this.stop();
+			log.error('Error on switching supervisor API listening rules', err);
+			return this.stop({ errored: true });
 		}
 	}
 
-	public stop() {
+	public async stop(options?: SupervisorAPIStopOpts): Promise<void> {
```
20k-ultra, line 235: The stop function offers a callbacks so I return a promise that allows the caller to know when this async operation is fully complete. I also am returning the error so that the person calling it knows how to handle the unsuccessful close.

src/supervisor-api.ts
```diff
@@ -176,22 +188,29 @@ export class SupervisorAPI {
...
 					port,
 					allowedInterfaces,
 				);
 			}
 		});
 
-		this.server = this.api.listen(port);
-		this.server.timeout = apiTimeout;
+		return new Promise(resolve => {
+			this.server = this.api.listen(port, () => {
```
20k-ultra, line 206: The listen function uses the callback when it is done trying to bind to a port. If an error occurs such as the port is already in use that would not be thrown here. In order to catch an error like that we must use something like

```js
server.on('error', (e) => {
  if (e.code === 'EADDRINUSE') {
    console.log('Address in use, retrying...');
    setTimeout(() => {
      server.close();
      server.listen(PORT, HOST);
    }, 1000);
  }
});
```

[server.listen() docs](https://nodejs.org/api/net.html#net_server_listen)

src/supervisor-api.ts
```diff
@@ -87,6 +91,14 @@ export class SupervisorAPI {
 
 	private api = express();
 	private server: Server | null = null;
+	// Holds the function which should apply iptables rules
+	private applyRules: SupervisorAPI['applyListeningRules'] =
+		process.env.TEST === '1'
+			? () => {
+					// don't try to alter iptables
```
CameronDiver, line 98: Did prettier do this?

src/supervisor-api.ts
```diff
@@ -208,17 +227,28 @@ export class SupervisorAPI {
...
 		}
 	}
 
-	public stop() {
+	public async stop(options?: SupervisorAPIStopOpts): Promise<void> {
 		if (this.server != null) {
-			this.server.close();
+			options?.errored
+				? log.error('Stopping Supervisor API...')
+				: log.debug('Stopping Supervisor API...');
```
CameronDiver, line 239: Honestly I don't feel that both a stopping and stopped message are needed, just a success and failure

src/supervisor-api.ts
20k-ultra: I copied that code from a patch you sent me when we worked on another issue. I did not write it and prettier is ok with the format. What's the issue? 

The original code I copied is here: https://github.com/balena-io/balena-supervisor/pull/1279/commits/32379d925576a45616150d44d06dfb2030eb4b05#diff-67d01b95ff8ac8912ffcf19a8c9dcb20

src/supervisor-api.ts
20k-ultra: I added that log because the supervisor logs when it's starting the API so I thought it was appropriate to add when it's stopping in addition to started and stopped message if successful.

https://github.com/balena-io/balena-supervisor/blob/ce39921c7ba6bb80f63e5f51750b432b7c6a17ee/src/supervisor.ts#L120-L125

20k-ultra: I removed these logs because I think it's more appropriate for the function calling the stop to log that it is stopping the supervisor api like in the above example when it starts the supervisor api.

src/supervisor-api.ts
```diff
@@ -200,8 +204,15 @@ export class SupervisorAPI {
 			}
 		});
 
-		this.server = this.api.listen(port);
-		this.server.timeout = apiTimeout;
+		return new Promise(resolve => {
+			this.server = this.api.listen(port, () => {
+				log.debug(`Supervisor API successfully started on port ${port}`);
```
CameronDiver, line 209:
```suggestion
				log.info(`Supervisor API successfully started on port ${port}`);
```

It's a small thing, but eventually debug output won't be shown by default, but IMO this should always be shown

20k-ultra: I've applied this change and followed this behavior with the message when supervisor shutdowns as well.

Issue #1180, opened 2020-01-17
CheckInt does not check if the number is valid


CameronDiver:
`parseInt` which is used in `checkInt` will still give a number for something like
```
parseInt('000123.45notanumber', 10)
123
```

Relevent code is here: https://github.com/balena-io/balena-supervisor/blob/master/src/lib/validation.ts#L22

PR #1287, opened 2020-05-04
Correctly check if value is a valid Integer

20k-ultra:
Closes: #1180
Change-type: patch

### DIFF
--- a/src/lib/validation.ts
+++ b/src/lib/validation.ts
@@ -12,6 +12,7 @@ export interface CheckIntOptions {
 
 const ENV_VAR_KEY_REGEX = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
+const NUMERALS_REGEX = /^-?[0-9]+\.?0*$/; // Allows trailing 0 decimals
 
 /**
  * checkInt
@@ -23,15 +24,14 @@ export function checkInt(
 	s: unknown,
 	options: CheckIntOptions = {},
 ): number | undefined {
-	if (s == null) {
+	// Check for non-numeric characters
+	if (!NUMERALS_REGEX.test(s as string)) {
 		return;
 	}
 
-	// parseInt will happily take a number, but the typings won't accept it,
-	// simply cast it here
-	const i = parseInt(s as string, 10);
+	const i = Number(s);
 
-	if (isNaN(i)) {
+	if (!Number.isInteger(i)) {
 		return;
 	}
 
--- a/test/07-validation.spec.ts
+++ b/test/07-validation.spec.ts
@@ -59,21 +59,30 @@ describe('validation', () => {
 	describe('checkInt', () => {
 		it('returns an integer for a string that can be parsed as one', () => {
 			expect(validation.checkInt('200')).to.equal(200);
+			expect(validation.checkInt('200.00')).to.equal(200); // Allow since no data is being lost
 			expect(validation.checkInt('0')).to.equal(0);
 			expect(validation.checkInt('-3')).to.equal(-3);
 		});
 
 		it('returns the same integer when passed an integer', () => {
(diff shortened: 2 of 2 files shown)


test/07-validation.spec.ts
```diff
@@ -196,7 +201,7 @@ describe('validation', () => {
 					},
 				},
 			};
-			expect(validation.isValidAppsObject(apps)).to.equal(true);
```
20k-ultra, line 199: This test was previously expecting an incorrect value so it had to be changed.

test/07-validation.spec.ts
```diff
@@ -61,19 +61,24 @@ describe('validation', () => {
 			expect(validation.checkInt('200')).to.equal(200);
 			expect(validation.checkInt('0')).to.equal(0);
 			expect(validation.checkInt('-3')).to.equal(-3);
+			expect(validation.checkInt(50.55)).to.equal(50); // Fractional digits
+			expect(validation.checkInt('0x11')).to.equal(17); // Hexadecimal
```
CameronDiver, line 65: I don't think we should add this extra input support. The supervisor expects integers to be provided in decimal, and I don't want any unexpected behaviour by supporting anything else.

20k-ultra: Do we want to return undefined for the above cases mentioned (fractional, hex, oct, binary) ? Or leave the current code and remove these test cases as they aren't a specification we are trying to support.

CameronDiver: Covering them is still a good idea I think, so we should return undefined, and let that be handled downstream

20k-ultra: Since you mentioned to let values that aren't Integers or can't be parsed into one without data lose be handled downstream I have added these cases to the test suite that makes sure they return undefined.

src/lib/validation.ts
```diff
@@ -12,6 +12,7 @@ export interface CheckIntOptions {
 
 const ENV_VAR_KEY_REGEX = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
+const NUMERALS_REGEX = /^-?[0-9]*\.?0*$/; // Allows trailing 0 decimals
```
CameronDiver, line 15: This feels very adhoc to me, and something we shouldn't really be implementing. I think that https://lodash.com/docs/4.17.15#toNumber should cover what we want, could you check that please?

20k-ultra: The problem is all these conversion functions are pretty generous with what they accept. We however want to allow only a very specific input so the regex was the easiest way to test that then use the conversion functions to only verify and cast to Number type. 

If I remove that regex then I have to write a regex to catch cases like '0x11', '0b11', '0o11' since [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) will accept them and convert to int. Lodash's implementation behaves the same.

src/lib/validation.ts
```diff
@@ -12,6 +12,7 @@ export interface CheckIntOptions {
 
 const ENV_VAR_KEY_REGEX = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
+const NUMERALS_REGEX = /^-?[0-9]+\.?0*$/; // Allows trailing 0 decimals
```
CameronDiver, line 15: Why not `\d` ?

20k-ultra: `\d` will match any unicode-defined digit. Since there are many notations for number systems, I think it's safest to explicitly have what values we want instead of thinking of all the edge cases possible.

src/lib/validation.ts
20k-ultra: We agreed to stick with regex as that was how others (mdn specifically) implemented it as well.

PR #1302, opened 2020-05-06
Use lstat rather than stat to avoid error with symlinks in sync

CameronDiver:
Change-type: patch
Closes: #1301

### DIFF
--- a/sync/device.ts
+++ b/sync/device.ts
@@ -99,20 +99,21 @@ async function tarDirectory(
 		const entries = await fs.readdir(path);
 		for (const entry of entries) {
 			const newPath = Path.resolve(path, entry);
-			const stat = await fs.stat(newPath);
+			// Here we filter the things we don't want
+			if (
+				newPath.includes('node_modules/') ||
+				newPath.includes('.git/') ||
+				newPath.includes('build/') ||
+				newPath.includes('coverage/')
+			) {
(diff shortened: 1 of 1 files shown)


sync/device.ts
```diff
@@ -99,20 +99,21 @@ async function tarDirectory(
 		const entries = await fs.readdir(path);
 		for (const entry of entries) {
 			const newPath = Path.resolve(path, entry);
-			const stat = await fs.stat(newPath);
+			// Here we filter the things we don't want
```
richbayliss, line 102: What if a user *wants* these directories to be included (for whatever reason) -- I feel like this might be a case of us over-reaching and being too opinionated. Isn't this what `.dockerignore` is for?

CameronDiver: Did you see the repo? :P We are the user, and you don't want to sync these things across to the supervisor container as you're not testing what would be running in production.

I think there's a valid point that we should have some way of sending over node_modules in some situations, for example testing changes in a satellite repo, but there is ways around that, for example adding a different version to the package.json

CameronDiver: I also don't want to bring dockerignore into the mix as it's been a massive pain point for the CLI team since integrating.

CameronDiver:
@balena-ci retest

Issue #1293, opened 2020-05-05
Dbus error when updating while openvpn-resin.service is not available on host


20k-ultra:
**UPDATE**: I have been able to create reproducible steps to get this state to occur as well as I have defined a scope of what's impacted and how it's resolved until the device gets back into the same state. I've yet to figure out what the problem actually is though and what service is responsible but one thing for certain is that it has so far only happened on Supervisor v11.2.0 regardless of balenaOS and where it was provisioned from (production or BoB).

**environment:**
 Supervisor v11.2.0
 balenaOS 2020.04.0 (from production) and balenaOS 2.47.0+rev1 (from BoB)

**Steps to reproduce:**
Provision a device then SSH to it and run `update-resin-supervisor -t v11.2.0` then do the following:

-  reboot the device either by SSHing to it or via API call
OR
- shutting down the device by SSHing to it or via API call then turning it back on

**How to verify you're in the broken state:**

1. SSH to the machine and observe docker container logs from supervisor. You will see the error message related to Dbus 
2. You can also use one of the endpoints that appear to be affected listed below

Examples of logs:

```sh
[error]   Error on POST /v2/applications/1/purge:  Error: Unit openvpn-resin.service not loaded.
[error]         at new DBusError (/usr/src/app/node_modules/dbus/lib/error.js:9:9)
[error]       at createError (/usr/src/app/node_modules/dbus/lib/bus.js:243:9)

[error]   Scheduling another update attempt in 64000ms due to failure:  Error: Unit openvpn-resin.service not loaded.
[error]         at new DBusError (/usr/src/app/node_modules/dbus/lib/error.js:9:9)
[error]       at createError (/usr/src/app/node_modules/dbus/lib/bus.js:243:9)

[error]   Device state apply error Error: Unit openvpn-resin.service not loaded.
[error]         at new DBusError (/usr/src/app/node_modules/dbus/lib/error.js:9:9)
[error]       at createError (/usr/src/app/node_modules/dbus/lib/bus.js:243:9)
```

**Effects of issue:**

Some endpoints are affected and others aren't. The affected endpoints will throw an exception with the dbus error and respond with a 503.

**Affected areas:**

From the example logs above we can see Device state apply functionality and update scheduler is throwing the Dbus exception.

The following endpoints will error with 503 and log the Dbus exception.

- [POST /v1/purge](https://www.balena.io/docs/reference/supervisor/supervisor-api/#post-v1purge) 
[...]

PR #1303, opened 2020-05-06
Don't use the openvpn alias to check VPN status

CameronDiver:
This still maintains compatibility with post-MC balenaOS versions, as all of those versions used `openvpn` as the service name proper. 

Closes: #1293
Change-type: patch


CameronDiver:
It was your catch @xginn8 ;)

PR #1311, opened 2020-05-08
Add device VPN information endpoint

### LINKED ISSUE #1068: User would like to use the supervisor API to check if device connected to VPN
Looking at the existing supervisor API responses, I cannot see any that would be able to tell the user if the device is connected to the VPN or not. Am I missing something and it's possible?

Or if not possible, would that answer fit one of the existing endpoints to add to the response / or figure out a new endpoint that could be used for this?

20k-ultra:
This PR provides a new V2 HTTP endpoint that allows querying the device for VPN information. Currently the only information provided back is if the VPN is connected but I tried to make it extendable so we can add more information.

In this PR I also created a lot of ground work for easily creating further test suites against the device API. I try very hard to use the standard lib as much as possible so me adding the new dependency was a decision I took very seriously but the benefits are great if we are to have extensive testing for the HTTP API.

The appropriate docs have also been updated to include this new endpoint.

Closes #1068


20k-ultra:
Thoughts on adding `enabled: boolean` to the `vpnInfo` interface ?

https://github.com/balena-io/balena-supervisor/blob/2fec4a3104b9400c94f5241de6b7476e941d59a9/src/device-api/v2.ts#L463-L465

 I saw this code that I could use to get that info and I think would be very useful since `connected: boolean` could always be false if it's not enabled.

https://github.com/balena-io/balena-supervisor/blob/d21366aa51a833ac88f20280627e57bf8c5745fa/src/device-config.ts#L574-L584

src/device-api/v2.ts
```diff
@@ -459,6 +458,22 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
 		}
 	});
 
+	router.get('/v2/device/vpn', async (_req, res) => {
+		// TODO: move all response interfaces to single location
+		interface VPNInfo {
```
CameronDiver, line 463: There's no real need to define this interface if it's only used here; typescript will infer it well enough (unless you want to add to it later in the function).

test/21-supervisor-api.spec.ts
```diff
@@ -1,68 +1,115 @@
...
 		});
+	});
 
-		beforeEach(async () => {
-			api = new SupervisorAPI({
-				config: mockedConfig,
-				eventTracker: new EventTracker(),
-				routers: [],
-				healthchecks: [],
+	describe('V1 endpoints', () => {
```
CameronDiver, line 61:
```suggestion
	describe.skip('V1 endpoints', () => {
```

src/device-api/v2.ts
20k-ultra: I added it to indicate I was planning for it to be extended in the future but can remove for now.

CameronDiver: Ideally we'd skip this kind of premature scaffolding, a lot of the time these can very easily be added when the time comes.

CameronDiver:
@20k-ultra Adding the enabled flag is fine for me.

src/device-api/v2.ts
20k-ultra: ok sounds good

test/21-supervisor-api.spec.ts
20k-ultra: will push in new commit

Issue #1310, opened 2020-05-08
Allow users to configure the retention limit on persistent logs for OS versions where persistent logs are on the data partition


dt-rush:
Ever since, in `meta-balena`, the commit

```
commit f69ff379ffac39dffcc126ac374b50533dcdef31
Author: Will Boyce <me@willboyce.com>
Date:   Fri Oct 25 12:10:46 2019 +0200

    resin-mounts: Move persistent logs from state to data partition
    
    Connects-to: #1646
    Change-type: minor
    Changelog-entry: Move persistent logs from state to data partition
    Signed-off-by: Will Boyce <will@balena.io>
```

(which went into `v2.45.0` of the OS), we have had the possibility of users configuring `RuntimeMaxUse=` in `/etc/systemd/journald.conf` to be much larger than the default (32M or 64M, depending on OS version). However, this can only be done by hand at present; as well, this file is on the root partition and will be gone with host OS update swapping in the default file as cooked by yocto into the OS image, each time.

It should be possible to have the supervisor detect a variable and apply it to this file via sed, should the user configure their own desired persistent logging retention limit.

PR #1312, opened 2020-05-09
allow customizing persistent logs retention size by var

dt-rush:
Change-type: minor
Connects-to: #1310
Signed-off-by: dt-rush <nickp@balena.io>


dt-rush:
NOTE: this currently fails since although it doesn't appear so at first glance, `/etc/systemd/journald.conf` is on a readonly FS.

src/lib/journald.ts
```diff
@@ -45,3 +45,11 @@ export function spawnJournalctl(opts: {
 
 	return journald;
 }
+
+export function setJournalRotationSize(size: string)  => {
+	execSync('sed', [
```
CameronDiver, line 50: Why use `sed`? It would be better to use typescript, rather than assuming people are happy with sed.

dt-rush: Just re-pushed with a simple `readFileSync/writeFileSync`

(however the filesystem itself being ro is a problem still)

dt-rush:
Ah, found out why: 

```
none / aufs ro,relatime,si=61e6a321e8f41828,dio,dirperm1 0 0
```

and the filesystem for `/etc/systemd/` is of course `none /`. 

There are however many other `/etc/` files/dirs which are mounted rw from state, I believe we should be able to add `journald.conf` to these:

```
root@264815e:/etc/systemd# cat /proc/mounts | grep etc
/dev/disk/by-label/resin-state /etc/machine-id ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/openvpn ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/docker ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/resin-supervisor ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/udev/rules.d ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/ssh/hostkeys ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/hostname ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/NetworkManager/conf.d ext4 rw,relatime 0 0
/dev/disk/by-label/resin-state /etc/NetworkManager/system-connections ext4 rw,relatime 0 0
```

What say ye, @CameronDiver ?

CameronDiver:
@dt-rush we should 100% be using `async` functions, there's no reason here to use `*Sync`. Adding a requirement on the host OS to mount this like so makes me uncomfortable, as we work very hard to avoid these kind of requirements and keep the compatibility range wide. If there's another way to do this that would be much preferred (perhaps this is best taken to an arch call?) otherwise we would have to version gate this feature within the supervisor, with correct messaging for a better UX.

Issue #1280, opened 2020-04-28
Send an event when we poll with details of whether it gave us a new target state


CameronDiver:
Using this information, we can decrease the poll interval if we find that it's rarely helpful, but currently we have no information on it.

PR #1313, opened 2020-05-11
Fixed stubs for test suite

20k-ultra:
This PR fixes a test suite that was incorrectly passing with an uncaught exception due to DBus operations.

Closes: #1280
Change-type: patch

### DIFF
--- a/test/18-startup.ts
+++ b/test/18-startup.ts
@@ -1,10 +1,52 @@
+import { SinonStub, stub } from 'sinon';
+
+import APIBinder from '../src/api-binder';
+import { ApplicationManager } from '../src/application-manager';
+import DeviceState from '../src/device-state';
+import * as constants from '../src/lib/constants';
+import { DockerUtils as Docker } from '../src/lib/docker-utils';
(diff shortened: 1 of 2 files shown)


test/18-startup.ts
```diff
@@ -1,10 +1,34 @@
...
+	after(() => {
+		initClientStub.restore();
+		startStub.restore();
+		vpnStatusPathStub.restore();
+	});
+
+	it('should startup correctly', async () => {
 		const supervisor = new Supervisor();
-		expect(supervisor.init()).to.not.throw;
+		expect(await supervisor.init()).to.not.throw;
```
CameronDiver, line 31:
```suggestion
		expect(async () => await supervisor.init()).to.not.throw;
```

20k-ultra: When the expect is written like that the exception thrown inside that closure doesn't trigger the test to fail. If this comment was related to the failing tests the issue was my local machine had docker socket open so the test passed but in CI docker is not there so it fails. Just another thing to stub.

PR #1319, opened 2020-05-12
Update engine information in package.json

CameronDiver:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "sqlite3": "^4.1.1"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "sqlite3": "^4.1.1"
   },
   "engines": {
-    "node": "^6.13.1"
+    "node": "^12.16.2"
   },
```
20k-ultra, line 37: We can also add "npm": "^6.14.0" so warnings are shown for npm as well.

Issue #1318, opened 2020-05-12
Convert all test files from Coffeescript to Typescript


20k-ultra:
Convert all the Coffeescript test files to Typescript. Remove any linting/transpilers needed after.

This will help to speed up test times and remove unneeded dependencies.

PR #1320, opened 2020-05-13
Remove CoffeeScript tests and all CoffeeScript tools

20k-ultra:
This makes the repo 100% CoffeeScript free! Exceptions for dependencies. This converts the last CoffeeScript code we had to TypeScript and removes all build/test tools previously used to do things like lint, test, build CoffeeScript.

Closes: #1318
Change-Type: patch


test/fast-mocha.opts
```diff
@@ -1,5 +1,4 @@
 --exit
 --require ts-node/register/transpile-only
---require coffeescript/register
 --timeout 30000
-test/*.{ts,coffee}
+test/*.{ts,js}
```
20k-ultra, line 4: This configuration previously would not allow you to run *.js test files with the test:fast command. Now that js is added you can run them.

test/13-device-config.spec.ts
```diff
@@ -0,0 +1,514 @@
...
+	it('properly reads a real config.txt file', function() {
+		Sinon.stub(fs, 'readFile').resolves(`\
+dtparam=i2c_arm=on\n\
+dtparam=spi=on\n\
+dtparam=audio=on\n\
+enable_uart=1\n\
+disable_splash=1\n\
+avoid_warnings=1\n\
+gpu_mem=16\n\
+				`);
```
20k-ultra, line 73: The only way to format this so it is indented would require doing a bunch of string concats and have those individual strings spaced with tabs. If we add tabs to this text directly it will break how deviceConfig interprets the values.

package.json
```diff
@@ -11,19 +11,17 @@
...
     "test:fast": "TEST=1 mocha --opts test/fast-mocha.opts",
     "test:debug": "npm run test:build && TEST=1 mocha --inspect-brk",
     "prettify": "balena-lint -e ts -e js --typescript --fix src/ test/ typings/",
     "typescript:test-build": "tsc --project tsconfig.json",
     "typescript:release": "tsc --project tsconfig.release.json && cp -r build/src/* build && rm -rf build/src",
-    "coffeescript:test": "coffee -m -c -o build .",
     "packagejson:copy": "cp package.json build/",
     "testitems:copy": "cp -r test/data build/test/",
-    "lint:coffee": "balena-lint test/",
     "lint:typescript": "balena-lint -e ts -e js --typescript src/ test/ typings/ build-utils/ && tsc --noEmit && tsc --noEmit --project tsconfig.js.json",
```
Page-, line 25: I'd say you can fold this into just `lint` now since there's no longer the coffee vs typescript diff

test/data/mnt/boot/extlinux/extlinux.conf
```diff
@@ -2,6 +2,6 @@ DEFAULT primary
 TIMEOUT 30
 MENU TITLE Boot Options
 LABEL primary
-      MENU LABEL primary Image
-      LINUX /Image
-      APPEND ${cbootargs} ${resin_kernel_root} ro rootwait
\ No newline at end of file
+MENU LABEL primary Image
+LINUX /Image
+APPEND ${cbootargs} ${resin_kernel_root} ro rootwait isolcpus=2
```
Page-, line 7: What are these changes?

test/13-device-config.spec.ts
CameronDiver: You can use `stripIndent` from `commonTags` - I always meant to add this. It is technically out of the scope for this PR so I'm not super worried.

test/data/mnt/boot/extlinux/extlinux.conf
20k-ultra: Any changes to the test files were by mistake. When adding types to fix the stubs some of the functions actually executed and wrote to the test files.

package.json
20k-ultra: I've removed any mention of :typescript since we are only using typescript now.

test/13-device-config.spec.ts
20k-ultra: When I added stripIndent the tests would fail. I tried troubleshooting it for awhile but can't figure out how adding that is causing the strings not to match what it was before..

PR #1323, opened 2020-05-15
Update typescript to v3.9

CameronDiver:
Also fix a bug in the typings which has been there for a while and only now was picked up.

Change-type: patch

### DIFF
--- a/src/compose/network.ts
+++ b/src/compose/network.ts
@@ -105,7 +105,7 @@ export class Network {
 
 		Network.validateComposeConfig(network);
 
-		const ipam: Partial<ComposeNetworkConfig['ipam']> = network.ipam || {};
+		const ipam: ComposeNetworkConfig['ipam'] = network.ipam || {};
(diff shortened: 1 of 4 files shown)


20k-ultra:
LGTM. I reviewed v3.9 changelog and looking forward to see if we get a reduction in compile time!

src/compose/network.ts
```diff
@@ -105,7 +105,7 @@ export class Network {
```
Page-, line 105: A few lines up
```ts
		network: Partial<Omit<ComposeNetworkConfig, 'ipam'>> & {
			ipam?: Partial<ComposeNetworkConfig['ipam']>;
		},
```
should be changed to
```ts
		network: Partial<ComposeNetworkConfig>
```
I think

20k-ultra:
closing in favour of https://github.com/balena-os/balena-supervisor/pull/1664

PR #1324, opened 2020-05-15
Update dependencies

Page-:
Change-type: patch


src/event-tracker.ts
```diff
@@ -92,10 +92,8 @@ export class EventTracker {
 		(event: string) => {
 			// Call this function at maximum once every minute
 			return _.throttle(
-				properties => {
-					if (this.client != null) {
-						this.client.track(event, properties);
-					}
+				(properties: EventTrackProperties | Error) => {
```
Page-, line 95: Updated typescript picked up that this was an implicit any

src/event-tracker.ts
```diff
@@ -92,10 +92,8 @@ export class EventTracker {
...
+					this.client?.track(event, properties);
```
Page-, line 96: This change wasn't strictly necessary but is part of incremental improvement

src/compose/network.ts
```diff
@@ -217,7 +217,9 @@ export class Network {
 	}
 
 	private static validateComposeConfig(
-		config: Partial<ComposeNetworkConfig>,
+		config: Partial<Omit<ComposeNetworkConfig, 'ipam'>> & {
+			ipam?: Partial<ComposeNetworkConfig['ipam']>;
+		},
```
Page-, line 222: Updated typescript picked up that this is the actual type we call this with so I had to amend it here to accept it

20k-ultra:
Do we want to include updating modules that contain fixes for vulnerabilities via `npm audit fix` in this PR?

20k-ultra:
I've fixed the conflicts with master that the current commit had. I'll be adding another commit with the updates from `npm audit fix` so that we can revert separately if there are any issues.

20k-ultra:
Ready for review @Page- @CameronDiver

CameronDiver:
@balena-ci rebase

Issue #1294, opened 2020-05-05
 /v2/applications/:appId/state returns state even if appId doesn't exist


20k-ultra:
[GET /v2/applications/:appId/state](https://www.balena.io/docs/reference/supervisor/supervisor-api/#get-v2applicationsappidstate) will return the state of the application on my device even if the appId does not exist.

```sh
$ curl "$BALENA_SUPERVISOR_ADDRESS/v2/applications/144/state?apikey=$BALENA_SUPERVISOR_API_KEY"
{"local":{"1":{"services":{"1":{"status":"Running","releaseId":1,"download_progress":null}}}},"dependent":{},"commit":"localrelease"}
```

This is confusing since the docs say "Use this endpoint to get the state of a single application, given the appId." I would have expect an error to return.

PR #1329, opened 2020-05-15
Improved handling of invalid appId in V2 state endpoint

20k-ultra:
This PR changes how the /v2/applications/:appId/state endpoint handles invalid or missing applications. Discussing the status codes used might be beneficial if the approach I chose does not follow previous status code responses but I personally believe what I used is the most accurate usage.

This PR also includes additional improvements to the mocked device API. We can now add services, networks, images, and volumes so that when we test endpoints we can more accurately see how the code performs on a device.

Closes: #1294
Change-type: patch

### DIFF
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -220,14 +220,41 @@ export function createV2Api(router: Router, applications: ApplicationManager) {
 
 	router.get(
 		'/v2/applications/:appId/state',
-		(_req: Request, res: Response, next: NextFunction) => {
-			// Get all services and their statuses, and return it
-			applications
-				.getStatus()
-				.then((apps) => {
-					res.status(200).json(apps);
-				})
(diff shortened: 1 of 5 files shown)


src/application-manager.d.ts
```diff
@@ -35,7 +35,7 @@ declare interface Options {
 }
 
 // TODO: This needs to be moved to the correct module's typings
-declare interface Application {
+export declare interface Application {
```
CameronDiver, line 38: Is this definitely needed?

20k-ultra: ah actually that was when I stubbed the ApplicationManager and wanted to use the correct types. I later stubbed the ServiceManager, NetworkManager, VolumeManager, etc so no longer needed this export because I was not stubbing ApplicationManager.

I will remove this change!

PR #1330, opened 2020-05-16
Avoid querying `instantUpdates` on every state poll

Page-:
It's only needed on the very first call so I moved the querying outside to where we start the polling to avoid the unnecessary work (which I believe includes disk io)

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -354,7 +354,16 @@ export class APIBinder {
 					'Trying to start poll without initializing API client',
 				);
 			}
-			this.pollTargetState(true);
+			this.config
+				.get('instantUpdates')
+				.catch(() => {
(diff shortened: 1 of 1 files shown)


src/api-binder.ts
```diff
@@ -354,7 +354,16 @@ export class APIBinder {
 					'Trying to start poll without initializing API client',
 				);
 			}
-			this.pollTargetState(true);
+			this.config
+				.get('instantUpdates')
+				.catch(() => {
+					// Default to skipping the initial update if we couldn't fetch the setting
+					// which should be the exceptional case
+					return false;
```
Page-, line 362: The current behavior if we can't fetch it is that we never poll the target state at all, so this is definitely an improvement in the exceptional circumstance that we do hit it

PR #1332, opened 2020-05-16
Isolate target state

Page-:


src/device-state/target-state.ts
```diff
@@ -0,0 +1,157 @@
...
+	}).finally(() => {
+		lastFetch = process.hrtime();
+	});
+};
+
+let targetStateFetchErrors = 0;
+const poll = async (skipFirstGet: boolean = false): Promise<void> => {
+	try {
+		// TODO: I'd prefer to get this only once and then add an event listener for updates
+		const appUpdatePollInterval = await config.get('appUpdatePollInterval');
```
CameronDiver, line 101: This should definitely be handled before merge.

src/device-state/target-state.ts
```diff
@@ -0,0 +1,157 @@
...
+import { getRequestInstance } from '../lib/request';
+import * as url from 'url';
+import * as config from '../config';
+import type { Headers } from 'request';
+import { delay } from 'bluebird';
+import constants = require('../lib/constants');
+import * as _ from 'lodash';
+import { writeLock } from '../lib/update-lock';
+import Bluebird = require('bluebird');
+import type StrictEventEmitter from 'strict-event-emitter-types';
```
CameronDiver, line 13: Can you split the local and external dependencies into groups please?

20k-ultra: I've added this change.

20k-ultra:
Before this PR the lastFetch variable was declared right away. In this PR lastFetch is initialized but only set after a successful update. 

https://github.com/balena-io/balena-supervisor/blob/b3b800299e0801015a44b45f696a9144d0e4e787/src/device-state/target-state.ts#L35-L38

I have reverted this change so that we are maintaining the previous behavior. I realize setting lastFetch to a time by default without successfully performing a target state fetch is incorrect but this change should be discussed in a separate PR since it could have additional side affects.

src/device-state/target-state.ts
20k-ultra: Currently, if poll cannot get `appUpdatePollInterval` it will wait 10 seconds and then recursively call itself. Getting this config value is the first thing the function does so effectively poll will keep trying to read the config every 10 seconds until successful then continue.

What I have implemented maintains this exact behavior but provides the benefit of best case only querying the config once. I did this by moving all (1) config queries to startPoll. This function already had a try catch for getting `instantUpdates` which would default to a value if there was an error. The default value is basically useless though since the poll function will be stuck in a loop trying to read `appUpdatePollInterval` from config.

PR #1333, opened 2020-05-18
Fix leftover spurious return from typescript conversion

CameronDiver:
Change-type: patch


src/device-state.ts
```diff
@@ -338,7 +338,7 @@ export class DeviceState extends (EventEmitter as new () => DeviceStateEventEmit
 
 		await this.applications.init();
 		if (!conf.initialConfigSaved) {
-			return this.saveInitialConfig();
+			this.saveInitialConfig();
```
Page-, line 341: Should this be
```suggestion
			await this.saveInitialConfig();
```
?

CameronDiver: Good catch, yea

PR #1338, opened 2020-05-19
Check for 409 status code, rather than string matching uuid conflicts

CameronDiver:
Change-type: patch


CameronDiver:
We're ready to go here, I want to get this released ASAP as a customer is blocked on this release @20k-ultra @Page- can you take a look please?

Issue #1292, opened 2020-05-05
Provide logs for failed health checks


20k-ultra:
When interacting with the supervisor API's health check endpoint I get a 500. When looking at the logs the only output is:

```sh
[error]   Healthcheck failed
[api]     GET /v1/healthy 500 - 14.246 ms
```

I think it would be extremely valuable to have some log.debug to show the operator why the health check is failing.

PR #1340, opened 2020-05-20
log healthcheck failure

20k-ultra:
This PR makes sure whatever causes the healthchecks used by GET /v1/healthy to not pass by logged via Log.info so that debugging this can be easier.

The PR includes added test coverage for the API endpoint as well as the healthcheck functions themselves that re called from this endpoint.

Closes: #1292

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -1,5 +1,6 @@
 import * as Bluebird from 'bluebird';
 import * as bodyParser from 'body-parser';
+import { stripIndent } from 'common-tags';
 import * as express from 'express';
 import { isLeft } from 'fp-ts/lib/Either';
 import * as t from 'io-ts';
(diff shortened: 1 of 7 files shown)


20k-ultra:
I found the grouped together conditions to be pretty confusing but didn't change them as I think that's out of this PR's scope. It did however make writing the log message more challenging as I didn't clearly understand them.

https://github.com/balena-io/balena-supervisor/blob/05540fe0cbbb9d278f9645d739121acd508c9d60/src/device-state.ts#L307-L310

The above condition is confusing to me because I'm unsure how a state being applied (applyInProgress) is a indicator of being unhealthy

https://github.com/balena-io/balena-supervisor/blob/05540fe0cbbb9d278f9645d739121acd508c9d60/src/api-binder.ts#L145-L148

In this above example it will say it is healthy if it's not connected ? Since this.deviceState.connected is if the device is connected but it's inverted.

CameronDiver:
@20k-ultra great work here. I'm not sure what the original intent was with the `!this.deviceState.connected` value here, but I'm pretty sure that the supervisor doesn't get killed with a lack of connection.

I checked the code, and it does indeed appear that is the case though: [We start a connectivity check](https://github.com/balena-io/balena-supervisor/blob/6972dcd92d36576517681685a129e89a8a06f722/src/device-state.ts#L392) and the code has been like this since before I converted it from coffeescript: https://github.com/balena-io/balena-supervisor/commit/01ed7bb103b4df8fb0679cf858220db42d4a0b92#diff-92e4f7a81c9dc7195cda50a5cdfdb0abL67

Can you please give this a little test when you get time? 

For the `applyInProgress` variable, we try to avoid killing the supervisor while applying target state changes, because we can end up killing downloads and corrupting directories. In addition, it's very rare that an error in the state application would be fixed with a restart.

20k-ultra:
I will test what happens to the supervisor if there is no network connection.

> For the `applyInProgress` check, we try to avoid killing the supervisor while applying target state changes, because we can end up killing downloads and corrupting directories. In addition, it's very rare that an error in the state application would be fixed with a restart.

Can you clarifying how this is related to health checks. Does the supervisor query the health checks before deciding to terminate itself ? That's currently how I'm interpreting this.

CameronDiver:
The healthcheck endpoint is called by systemd, and after a configurable amount of healthcheck failures, systemd will kill the supervisor and start a new instance.

20k-ultra:
I didn't perform any tests because I'm not sure exactly what scenario we want to try but I can pretty confidently theorize what would happen if we tested what the supervisor will do if there is no network connection.

We wanted to understand why the healthCheck will return true with no network connection since that seems counter intuitive. If I create the conditions to satisfy this test all that will happen is healthCheck returns true thus preventing the Supervisor from restarting.

I believe this happens for 2 reasons; The first is if the there is no network connection then the healthCheck will constantly report false and potentially cause the Supervisor to be stuck in a reboot loop until the connection is re-established. The second reason is because if the Supervisor is not connected then restarting might perhaps break some running services and so letting it run is better then restarting due to no network connection and having to physically then go to the device.

Either way, changing these conditions is outside the scope of this PR as we just wanted to learn the WHAT failed not WHY.

20k-ultra:
@balena-ci I self-certify!

PR #1344, opened 2020-05-26
Support ResponseErrors from balena-register-device

CameronDiver:
Change-type: patch

### DIFF
--- a/src/lib/register-device.ts
+++ b/src/lib/register-device.ts
@@ -2,6 +2,8 @@ import factory = require('balena-register-device');
 import * as Bluebird from 'bluebird';
 import { getRequestInstance } from './request';
 
+export const { ApiError } = factory;
+
 export const { generateUniqueKey, register } = factory({
 	request: {
 		send: Bluebird.method(async (options: {}) => {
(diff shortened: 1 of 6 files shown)


src/lib/register-device.ts
```diff
@@ -2,6 +2,9 @@ import factory = require('balena-register-device');
 import * as Bluebird from 'bluebird';
 import { getRequestInstance } from './request';
 
+const { ApiError } = factory;
+export { ApiError };
```
Page-, line 6:
```suggestion
export const { ApiError } = factory;
```

typings/balena-register-device.d.ts
```diff
@@ -1,6 +1,14 @@
...
 // TODO: Upstream types to the repo
 declare module 'balena-register-device' {
+	import TypedError = require('typed-error');
+	import { Response } from 'request';
+
+	factory.ApiError = class ApiError extends TypedError {
+		response: Response;
+	};
+
 	function factory({
```
Page-, line 12: It'd be better to put this avoid the `factory.ApiError =` just to avoid confusing people

CameronDiver: I'm not sure what you mean by this?

Page-: The first `avoid` was meant to be `above` - but you did it anyway so it's all cool

CameronDiver:
@balena-ci I self-certify!

PR #1345, opened 2020-05-28
Make the db module a singleton

CameronDiver:
We were treating the database class as a singleton, but still having to pass
around the db instance. Now we can simply require the db module and have
access to the database handle.

Change-type: patch


test/lib/prepare.ts
```diff
@@ -1,18 +1,30 @@
...
+try {
+	fs.unlinkSync(process.env.DATABASE_PATH!);
+} catch (e) {
+	/* ignore */
+}
+
+export = async function () {
+	// // @ts-ignore
+	// constants.databasePath = process.env.DATABASE_PATH;
+	// delete require.cache[require.resolve('../../src/db')];
```
Page-, line 14: You can delete this

test/02-db.spec.ts
```diff
@@ -45,36 +45,29 @@ async function createOldDatabase(path: string) {
...
 
-	it('creates a database at the path passed on creation', () => {
-		const db2 = new DB({ databasePath: process.env.DATABASE_PATH_2 });
-		const promise = db2
-			.init()
-			.then(() => fs.stat(process.env.DATABASE_PATH_2!));
-		return expect(promise).to.be.fulfilled;
+	it('creates a database at the path passed on creation', async () => {
+		const databasePath = process.env.DATABASE_PATH_2!;
+		// @ts-ignoreas
```
Page-, line 54:
```suggestion
		// @ts-ignore
```

PR #1356, opened 2020-06-04
Make the event-tracker module a singleton

CameronDiver:
Change-type: patch

### DIFF
--- a/src/event-tracker.ts
+++ b/src/event-tracker.ts
@@ -1,21 +1,13 @@
-import * as Bluebird from 'bluebird';
 import mask = require('json-mask');
 import * as _ from 'lodash';
 import * as memoizee from 'memoizee';
 import * as mixpanel from 'mixpanel';
 
-import { ConfigType } from './config';
+import * as config from './config';
 import log from './lib/supervisor-console';
 import supervisorVersion = require('./lib/supervisor-version');
 
 export type EventTrackProperties = Dictionary<any>;
-
-interface InitArgs {
-	uuid: ConfigType<'uuid'>;
-	unmanaged: ConfigType<'unmanaged'>;
-	mixpanelHost: ConfigType<'mixpanelHost'>;
-	mixpanelToken: ConfigType<'mixpanelToken'>;
-}
 
 // The minimum amount of time to wait between sending
 // events of the same type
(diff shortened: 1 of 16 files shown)


CameronDiver:
@balena-ci rebase

test/11-api-binder.spec.ts
```diff
@@ -65,6 +60,14 @@ const mockProvisioningOpts = {
 describe('ApiBinder', () => {
 	let server: Server;
 
+	beforeEach(() => {
+		stub(eventTracker, 'track');
+	});
+	afterEach(() => {
+		// @ts-expect-error Restoring a non-stub type function
```
Page-, line 67: This is more of a `ts-ignore` because if the typings were magical enough then this would be valid whereas `@ts-expect-error` is more if you'd always expect an error, eg `const s: string = new Date()` - for reference if you did
```ts
// @ts-expect-error
const s: string = 'valid string'
```
then it would actually error as it expects it to fail and the fact that it doesn't is treated as an issue

test/09-event-tracker.spec.ts
```diff
@@ -1,172 +1,251 @@
...
-			eventTracker.track('test6', { d: 4 });
-			eventTracker.track('test7', { e: 5 });
+			await eventTracker.track('test3', { a: 1 });
+			await eventTracker.track('test4', { b: 2 });
+			await eventTracker.track('test5', { c: 3 });
+			await eventTracker.track('test6', { d: 4 });
+			await eventTracker.track('test7', { e: 5 });
 
-			// @ts-ignore
+			// @ts-expect-error possibly null object
```
Page-, line 248: You can use `!.` or `?.` for possibly null cases

src/event-tracker.ts
```diff
@@ -32,85 +24,89 @@ const mixpanelMask = [
...
+			eventDebounceTime,
+			{ leading: true },
+		);
+	},
+	{ primitive: true },
+);
+
+export function logEvent(...args: string[]) {
+	log.event(...args);
+}
```
Page-, line 106: This can now be changed to
```suggestion
export const logEvent = log.event.bind(log);
```
if you want

src/event-tracker.ts
```diff
@@ -32,85 +24,89 @@ const mixpanelMask = [
 	'stateDiff/local(os_version,superisor_version,ip_address,apps/*/services)',
 ].join(',');
 
-export class EventTracker {
-	private defaultProperties: EventTrackProperties | null;
-	private client: mixpanel.Mixpanel | null;
+let defaultProperties: EventTrackProperties;
+export let client: mixpanel.Mixpanel | null = null;
```
Page-, line 28: Why is this exported now?

src/event-tracker.ts
CameronDiver: Honestly I'm not sure why it's even still there, thinking about it.

Issue #1357, opened 2020-06-04
Should report the device mac_address(es) to the API


thgreasi:
Similarly to the ip_addess field, we should report a space delimited list of the mac addresses of the interfaces of the device.
HQ: https://github.com/balena-io/balena/issues/2195
Depends-on: https://github.com/balena-io/balena-api/issues/2625

PR #1361, opened 2020-06-05
state: Report device MAC address to the API

richbayliss:
When reporting device information, send the MAC address of any
interfaces on the system. Also expose in the Supervisor API at
the route GET /v1/device.

Related-to: #1357 
Change-type: patch


test/lib/mocked-balena-api.ts
```diff
@@ -47,6 +47,11 @@ api.balenaBackend = {
 	},
 };
 
+api.patch('/device/v2/:uuid/state', (req, res) => {
+	console.log(`/device/v2/${req.params.uuid}/state called with `, req.body);
+	return res.status(200).end();
```
richbayliss, line 52: Can drop this now and just return the `Promise.resolve()` since the arg value is checked in a stub test.

test/lib/mocked-device-api.ts
```diff
@@ -81,9 +81,15 @@ async function create(): Promise<SupervisorAPI> {
 	// Create SupervisorAPI
 	const api = new SupervisorAPI({
 		eventTracker,
-		routers: [buildRoutes(appManager)],
+		routers: [deviceState.router, buildRoutes(appManager)],
```
richbayliss, line 84: without this, the `/v1/device` endpoint isn't registered...

test/lib/mocked-device-api.ts
```diff
@@ -81,9 +81,15 @@ async function create(): Promise<SupervisorAPI> {
...
 		healthchecks: [deviceState.healthcheck, apiBinder.healthcheck],
 	});
+
+	const conf = await config.getMany(['macAddress']);
+	deviceState.reportCurrentState({
+		mac_address: conf.macAddress,
+	});
```
richbayliss, line 91: without this, the device state isn't aware of the MAC address and will not report it...

src/lib/mac-address.ts
```diff
@@ -0,0 +1,20 @@
+import * as fs from 'mz/fs';
+import * as path from 'path';
+
+export async function getAll(sysClassNet: string): Promise<string | undefined> {
+	if (!(await fs.exists(sysClassNet))) {
+		return '';
```
CameronDiver, line 6: why not `undefined`, given it's part of the function signature?

src/lib/mac-address.ts
```diff
@@ -0,0 +1,20 @@
+import * as fs from 'mz/fs';
+import * as path from 'path';
+
+export async function getAll(sysClassNet: string): Promise<string | undefined> {
```
CameronDiver, line 4: Can you wrap this whole function in a try catch please, it means we can wrap any errors in a `MacAddress` error so it makes things more clear when this code fails.

test/lib/mocked-device-api.ts
```diff
@@ -81,9 +81,15 @@ async function create(): Promise<SupervisorAPI> {
 	// Create SupervisorAPI
 	const api = new SupervisorAPI({
 		eventTracker,
-		routers: [buildRoutes(appManager)],
+		routers: [deviceState.router, buildRoutes(appManager)],
 		healthchecks: [deviceState.healthcheck, apiBinder.healthcheck],
 	});
+
+	const conf = await config.getMany(['macAddress']);
```
CameronDiver, line 88:
```suggestion
        const macAddress = await config.get('macAddress');
```

src/lib/mac-address.ts
richbayliss: Oversight. Corrected.

src/lib/mac-address.ts
richbayliss: Nice. Done.

src/lib/mac-address.ts
```diff
@@ -0,0 +1,28 @@
...
+			return;
+		}
+
+		// read the directories in the sysfs...
+		const interfaces = await fs.readdir(sysClassNet);
+		const addresses = await (
+			await Promise.all(
+				interfaces
+					.map((intf) => path.join(sysClassNet, intf, 'address'))
+					.filter((addr) => fs.existsSync(addr))
```
CameronDiver, line 19: I think typically the right thing to do here is attempt to read it without checking that it exists. Additionally, please don't use any `Sync` functions under normal operation.

richbayliss: My concern here was avoiding this scenario in the `Promise.all`:
> It rejects with the reason of the first promise that rejects

I could wrap the inner in a try/catch too but felt this might be less readable/messy. I will try and find some middle ground.

src/device-state.ts
```diff
@@ -400,8 +403,10 @@ export class DeviceState extends (EventEmitter as new () => DeviceStateEventEmit
 		log.debug('Starting periodic check for IP addresses');
 
 		await network.startIPAddressUpdate()(async (addresses) => {
+			const macAddress = await config.get('macAddress');
```
CameronDiver, line 406: You should be able to move this outside of the update loop, right? We don't expect this to change throughout runtime do we?

src/lib/mac-address.ts
```diff
@@ -0,0 +1,33 @@
...
+		// read the directories in the sysfs...
+		return (
+			await Promise.all(
+				interfaces.map(async (intf) => {
+					try {
+						const file = path.join(sysClassNet, intf, 'address');
+						const addr = await fs.readFile(file, 'utf8');
+						return addr.split('\n')[0].trim().toUpperCase();
+					} catch {
+						return '';
```
CameronDiver, line 22: Is this expected to happen a lot, or is it an exceptional circumstance? Basically I'm wondering if it points to other issues on-device, and we should report it in the logs.

src/device-state.ts
richbayliss: yes we do - if a user add/removes a network interface, for example.

test/21-supervisor-api.spec.ts
```diff
@@ -113,6 +113,18 @@ describe('SupervisorAPI', () => {
...
+		describe('GET /v1/device', () => {
+			it('returns MAC address', async () => {
+				await request
+					.get('/v1/device')
+					.set('Accept', 'application/json')
+					.set('Authorization', `Bearer ${VALID_SECRET}`)
+					.expect(200)
+					.then((response) => {
+						expect(response.body).to.have.property('mac_address').not.empty;
+					});
```
Page-, line 125:
```suggestion
				const response = await request
					.get('/v1/device')
					.set('Accept', 'application/json')
					.set('Authorization', `Bearer ${VALID_SECRET}`)
					.expect(200);
				expect(response.body).to.have.property('mac_address').not.empty;
```

test/18-startup.spec.ts
```diff
@@ -52,5 +58,24 @@ describe('Startup', () => {
 		expect(anySupervisor.logger).to.not.be.null;
 		expect(anySupervisor.deviceState).to.not.be.null;
 		expect(anySupervisor.apiBinder).to.not.be.null;
+
+		const macAddresses: string[] = [];
+		reportCurrentStateStub.getCalls().forEach((call) => {
+			const m: string = call.args[0]['mac_address'];
+			if (!m || m === '') {
```
Page-, line 65: This is equivalent to
```suggestion
			if (!m) {
```

src/lib/mac-address.ts
```diff
@@ -0,0 +1,49 @@
...
+						const addr = await fs.readFile(addressFile, 'utf8');
+						return addr.split('\n')[0].trim().toUpperCase();
+					} catch {
+						return '';
+					}
+				}),
+			),
+		)
+			.filter((addr) => addr !== '')
+			.sortedUniq()
```
Page-, line 43: This requires the array to be sorted or otherwise it won't work - I don't see anything that guarantees the array will be sorted:
```ts
_.sortedUniq([1, 1, 2]); // [1, 2]
_.sortedUniq([1, 2, 1]); // [1, 2, 1]

File: src/lib/mac-address.ts [hunk 13]
```diff
@@ -0,0 +1,49 @@
...
+			await Promise.all(
+				interfaces.map(async (intf) => {
+					try {
+						const [addressFile, typeFile, masterFile] = [
+							'address',
+							'type',
+							'master',
+						].map((f) => path.join(sysClassNet, intf, f));
+						const intfType = (await fs.readFile(typeFile, 'utf8')).trim();
+						const intfHasMaster = await fs.exists(masterFile);
```
Page-, line 25: Can you either do these fs accesses in parallel or make use of short-circuiting to avoid the second access if the first is invalid

test/18-startup.spec.ts
```diff
@@ -52,5 +58,24 @@ describe('Startup', () => {
...
+			const m: string = call.args[0]['mac_address'];
+			if (!m) {
+				return;
+			}
+
+			m.split(' ').forEach((a) => {
+				if (!macAddresses.includes(a)) {
+					macAddresses.push(a);
+				}
+			});
```
Page-, line 73:
```suggestion
			macAddresses = _.union(macAddresses, m.split(' '));
```

test/18-startup.spec.ts
```diff
@@ -20,6 +21,10 @@ describe('Startup', () => {
 		initClientStub = stub(APIBinder.prototype as any, 'initClient').returns(
 			Promise.resolve(),
 		);
+		reportCurrentStateStub = stub(
+			DeviceState.prototype as any,
+			'reportCurrentState',
+		).returns(Promise.resolve());
```
Page-, line 27: I think this can be
```suggestion
		).resolves();
```

test/21-supervisor-api.spec.ts
```diff
@@ -113,6 +113,17 @@ describe('SupervisorAPI', () => {
...
 		// TODO: add tests for V1 endpoints
+		describe('GET /v1/device', () => {
+			it('returns MAC address', async () => {
+				const response = await request
+					.get('/v1/device')
+					.set('Accept', 'application/json')
+					.set('Authorization', `Bearer ${VALID_SECRET}`)
+					.expect(200);
+
+				expect(response.body).to.have.property('mac_address').not.empty;
```
Page-, line 124:
```suggestion
				expect(response.body).to.have.property('mac_address').that.is.not.empty;
```

src/lib/mac-address.ts
```diff
@@ -0,0 +1,53 @@
+import * as _ from 'lodash';
+import * as fs from 'mz/fs';
```
CameronDiver, line 2:
```suggestion
import { promises as fs } from 'fs';
```

We're slowly moving to this for everything in the supervisor

src/lib/mac-address.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as path from 'path';
+
+import { shouldReportInterface } from '../network';
+import log from './supervisor-console';
+
+export class MacAddressError extends Error {}
```
CameronDiver, line 8:
```suggestion
export class MacAddressError extends TypedError {}
```

src/lib/mac-address.ts
```diff
@@ -0,0 +1,53 @@
...
+						].map((f) => path.join(sysClassNet, intf, f));
+
+						const [intfType, intfHasMaster] = await Promise.all([
+							fs.readFile(typeFile, 'utf8'),
+							fs.exists(masterFile),
+						]);
+
+						if (
+							!shouldReportInterface(intf) ||
+							intfType.trim() !== '1' ||
```
CameronDiver, line 32: Can you add a comment detailing what this check is please? As it stands I have no idea.

src/lib/mac-address.ts
```diff
@@ -0,0 +1,53 @@
...
+							intfHasMaster
+						) {
+							return '';
+						}
+
+						const addr = await fs.readFile(addressFile, 'utf8');
+						return addr.split('\n')[0].trim().toUpperCase();
```
CameronDiver, line 39:
```suggestion
						return addr.split('\n')[0]?.trim()?.toUpperCase() ?? '';
```

src/lib/mac-address.ts
CameronDiver: Yeah I think in that case we want to add a log as breadcrumbs for why this failed.

src/lib/mac-address.ts
richbayliss: this will break `fs.exists()` which is used below... Happy to change this, but would you rather have 2 imports, or import it like this?

PR #1362, opened 2020-06-05
Add label to expose gpu to container

robertgzr:
In the absence of an upstream implementation of the DeviceRequest API introduced
as part of Docker API v1.40 we roll our own using a feature label.

As per my comment in the code, we fall back to the default behavior of
docker cli's `--gpu` and request single device with the `gpu` capabilty.
The only implementation at the moment is the NVIDIA driver; here:
https://github.com/balena-os/balena-engine/blob/master/daemon/nvidia_linux.go

Background on the composefile implementation:
https://github.com/compose-spec/compose-spec/issues/74
https://github.com/docker/compose/issues/6691

Change-type: patch
Connects-to: https://github.com/balena-os/balena-jetson/pull/75
Signed-off-by: Robert Günzler <robertg@balena.io>


robertgzr:
fixed the signoff :)

src/compose/utils.ts
```diff
@@ -373,6 +373,15 @@ export function addFeaturesFromLabels(
 		'io.balena.features.sysfs': () => service.config.volumes.push('/sys:/sys'),
 		'io.balena.features.procfs': () =>
 			service.config.volumes.push('/proc:/proc'),
+		'io.balena.features.gpu': () =>
```
richbayliss, line 376: should this be device-type dependent @robertgzr ?

robertgzr: now that you mention it. yeah probably should be...

robertgzr: it's also specific to docker api v1.40, but not sure if we care about that?

robertgzr:
can you have a look. not sure if this super basic test is enough...
basically all we need is the supervisor to add a device request to the container create request, as if we were doing `docker run --gpu`

I also think Rich is right above, in that we should probably guard against this on older devices (incompatible device types?)

src/compose/utils.ts
```diff
@@ -373,6 +373,15 @@ export function addFeaturesFromLabels(
...
+			// TODO once the compose-spec has an implementation we
+			// should probably follow that, for now we copy the
+			// bahavior of docker cli
+			// https://github.com/balena-os/balena-engine-cli/blob/19.03-balena/opts/gpus.go#L81-L89
+			service.config.deviceRequests.push({
+				Count: 1,
+				Capabilities: [['gpu']] as string[][],
```
CameronDiver, line 383:
```suggestion
				Capabilities: [['gpu']],
```

test/04-service.spec.ts
```diff
@@ -464,6 +464,29 @@ describe('compose/service', () => {
...
+			},
+			{ appName: 'test' } as any,
+		);
+
+		expect(s.config)
+			.to.have.property('deviceRequests')
+			.that.deep.equals([
+				{
+					Count: 1,
+					Capabilities: [['gpu']] as string[][],
```
CameronDiver, line 485:
```suggestion
					Capabilities: [['gpu']],
```

CameronDiver:
Ready to approve with those changes, no need to cast.

robertgzr:
cool. removed those and squashed the tests into the main commit

robertgzr:
for anybody coming across this:

this work on it's own will not enable gpu integration on balenaOS. It depends on a few OS level additions similar to the WIP here: https://github.com/balena-os/balena-jetson/pull/75

We are working on this on a per-device basis with the current milestone being https://github.com/balena-os/meta-balena/pull/1987

PR #1365, opened 2020-06-05
Convert all test files to TS and add .spec to all filenames

20k-ultra:
This ensures all test files are typescript and contains a .spec extension. I purposelessly did not remove the ability to test JS files in fast-mocha.opts because I didn't want to force any new behavior, just normalize our existing files since it was a mix of js and ts.

The addition of .spec extension is also just to keep files consistent but also mocha can automatically run files with .spec in the filename so technically we no longer have to specify where the test files are.

Change-type: patch


20k-ultra:
@balena-ci I self-certify!

test/11-api-binder.spec.ts
```diff
@@ -24,8 +25,7 @@ const initModels = async (obj: Dictionary<any>, filename: string) => {
 	// @ts-ignore
 	config.configJsonBackend = new ConfigJsonConfigBackend(schema, filename);
 	config.generateRequiredFields();
-	// @ts-expect-error using private properties
-	config.configJsonBackend.cache = await config.configJsonBackend.read();
+	(config.configJsonBackend as any).cache = await (config.configJsonBackend as any).read();
```
CameronDiver, line 28: I'd rather a ts-expect-error than `any`. If something changes I'd rather get informed via a compile time breakage than not at all.

test/11-api-binder.spec.ts
```diff
@@ -307,6 +307,51 @@ describe('ApiBinder', () => {
...
 
+	describe('local mode', () => {
+		const components: Dictionary<any> = {};
+
+		before(() => {
+			return initModels(components, '/config-apibinder.json');
+		});
+
+		after(async () => {
+			// @ts-ignore
```
CameronDiver, line 318: Can you use `ts-expect-error` with information about why/what you're expecting please?

test/12-logger.spec.ts
```diff
@@ -14,7 +14,7 @@ describe('Logger', function () {
 		this._req.end = sinon.spy();
 
 		this._req.body = '';
-		this._req.pipe(zlib.createGunzip()).on('data', (chunk) => {
+		this._req.pipe(zlib.createGunzip()).on('data', (chunk: any) => {
```
CameronDiver, line 17: I think this should be a `Buffer`. In general I'd like to remove the vast majority of `any`s from the code.

test/11-api-binder.spec.ts
CameronDiver: I see this was just from an extra file moved here. Feel free to ignore then.

PR #1366, opened 2020-06-08
Make logger module a singleton

CameronDiver:
Change-type: patch

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -16,261 +16,252 @@ import {
 import LogMonitor from './logging/monitor';
 
 import * as globalEventBus from './event-bus';
-import log from './lib/supervisor-console';
-
-interface LoggerSetupOptions {
-	apiEndpoint: config.ConfigType<'apiEndpoint'>;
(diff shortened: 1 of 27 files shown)


src/logger.ts
```diff
@@ -16,261 +16,252 @@ import {
...
+		'localMode',
+	]);
+
+	balenaBackend = new BalenaLogBackend(apiEndpoint, uuid, deviceApiKey);
+	localBackend = new LocalLogBackend();
+	backend = localMode ? localBackend : balenaBackend;
+	backend.unmanaged = unmanaged;
+	backend.publishEnabled = loggingEnabled;
+
+	if (!balenaBackend.isIntialised()) {
```
Page-, line 55:
```suggestion
	if (!balenaBackend.isInitialised()) {
```

CameronDiver: :S why wasn't this picked up by typescript. That's concerning, I'll have to check it out.

CameronDiver: Ah, consistent typos :)

PR #1369, opened 2020-06-09
fix: API auth missing on state GET/PATCH

richbayliss:
When performing a state GET/PATCH the authentication header was being
missed off.

Change-type: patch
Signed-off-by: Rich Bayliss

### DIFF
--- a/src/device-state/target-state.ts
+++ b/src/device-state/target-state.ts
@@ -1,6 +1,5 @@
 import { EventEmitter } from 'events';
 import * as url from 'url';
-import type { Headers } from 'request';
 import { delay } from 'bluebird';
 import * as _ from 'lodash';
 import Bluebird = require('bluebird');
@@ -9,9 +8,13 @@ import type StrictEventEmitter from 'strict-event-emitter-types';
 import type { TargetState } from '../types/state';
 import { InternalInconsistencyError } from '../lib/errors';
 import { getRequestInstance } from '../lib/request';
+import { CoreOptions } from 'request';
 import * as config from '../config';
 import { writeLock } from '../lib/update-lock';
 import constants = require('../lib/constants');
+import log from '../lib/supervisor-console';
+
+export class ApiResponseError extends Error {}
 
 interface TargetStateEvents {
 	'target-state-update': (
@@ -70,10 +73,16 @@ export const update = async (
 ): Promise<void> => {
 	await config.initialized;
 	return Bluebird.using(lockGetTarget(), async () => {
-		const { uuid, apiEndpoint, apiTimeout } = await config.getMany([
+		const {
+			uuid,
+			apiEndpoint,
+			apiTimeout,
+			deviceApiKey,
+		} = await config.getMany([
 			'uuid',
 			'apiEndpoint',
 			'apiTimeout',
+			'deviceApiKey',
 		]);
 
 		if (typeof apiEndpoint !== 'string') {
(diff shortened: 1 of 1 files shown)


src/device-state/target-state.ts
```diff
@@ -85,33 +93,42 @@ export const update = async (
...
 			// There's no change so no need to update the cache or emit a change event
 			return;
 		}
 
+		if (statusCode < 200 || statusCode >= 300) {
+			log.error(`Error from the API: ${statusCode}`);
+			throw new ApiResponseError(`Error from the API: ${statusCode}`);
+		}
+
+		const targetState = JSON.parse(body);
```
Page-, line 119: This shouldn't be here, it's because of a bug above

src/device-state/target-state.ts
```diff
@@ -85,33 +93,42 @@ export const update = async (
 		const endpoint = url.resolve(apiEndpoint, `/device/v2/${uuid}/state`);
 		const request = await getRequestInstance();
 
-		const params: Headers = {
-			json: true,
+		const params: any = {
+			headers: {
+				Authorization: `Bearer ${currentApiKey}`,
+				json: true,
```
Page-, line 99: This isn't a header, it's a top level param to `request` that means we request a json response and will automatically parse it as well

src/device-state/target-state.ts
```diff
@@ -85,33 +93,42 @@ export const update = async (
 		const endpoint = url.resolve(apiEndpoint, `/device/v2/${uuid}/state`);
 		const request = await getRequestInstance();
 
-		const params: Headers = {
-			json: true,
+		const params: any = {
```
Page-, line 96: Can you type this properly please

src/device-state/target-state.ts
```diff
@@ -85,33 +93,42 @@ export const update = async (
...
+			headers: {
+				Authorization: `Bearer ${currentApiKey}`,
+				json: true,
+				['If-None-Match']:
```
Page-, line 100:
```suggestion
				'If-None-Match':
```

src/device-state/target-state.ts
```diff
@@ -85,33 +93,42 @@ export const update = async (
...
 			.timeout(apiTimeout);
 
 		if (statusCode === 304) {
 			// There's no change so no need to update the cache or emit a change event
 			return;
 		}
 
+		if (statusCode < 200 || statusCode >= 300) {
+			log.error(`Error from the API: ${statusCode}`);
+			throw new ApiResponseError(`Error from the API: ${statusCode}`);
```
Page-, line 116: Do we need to both log and throw here? Just wondering if it'll be caught and logged twice further up

src/device-state/target-state.ts
```diff
@@ -12,6 +11,9 @@ import { getRequestInstance } from '../lib/request';
 import * as config from '../config';
 import { writeLock } from '../lib/update-lock';
 import constants = require('../lib/constants');
+import log from '../lib/supervisor-console';
+
+class ApiResponseError extends Error {}
```
Page-, line 16: If you're using a specific error class like this you should export it so that users can catch and check for the specific error with
```ts
try {
	...
} catch (err) {
	if (err instanceof ApiResponseError) {
		...
	}
}

Comment (richbayliss) [reply src/device-state/target-state.ts:16]: fair point

File: src/device-state/target-state.ts (hunk 5 above)
Comment (richbayliss) [reply src/device-state/target-state.ts:116]: It isn't logged so it doesn't show up and the logs just look like nothing is wrong...

File: src/device-state/target-state.ts [hunk 7]
```diff
@@ -70,10 +73,16 @@ export const update = async (
 ): Promise<void> => {
 	await config.initialized;
 	return Bluebird.using(lockGetTarget(), async () => {
-		const { uuid, apiEndpoint, apiTimeout } = await config.getMany([
+		const {
+			uuid,
+			apiEndpoint,
+			apiTimeout,
+			currentApiKey,
```
CameronDiver, line 80:
```suggestion
			deviceApiKey,
```
`currentApiKey` was a hack that will be removed at some point.

richbayliss: good to know

src/device-state/target-state.ts
Page-: Ok, just checking 👍

src/device-state/target-state.ts
```diff
@@ -85,27 +94,29 @@ export const update = async (
 		const endpoint = url.resolve(apiEndpoint, `/device/v2/${uuid}/state`);
 		const request = await getRequestInstance();
 
-		const params: Headers = {
+		const params: CoreOptions = {
 			json: true,
+			headers: {
+				Authorization: `Bearer ${deviceApiKey}`,
+				'If-None-Match':
+					typeof cache?.etag === 'string' ? cache.etag : undefined,
```
Page-, line 102: This should actually be fine as
```suggestion
					cache?.etag,
```
since it can only be a string if it exists and so the end result is the same

PR #1370, opened 2020-06-09
Make images module a singleton

CameronDiver:
Change-type: patch


src/compose/images.ts
```diff
@@ -52,584 +47,599 @@ type NormalisedDockerImage = Docker.ImageInfo & {
...
+		}
+	});
+})();
+
+export async function triggerFetch(
+	image: Image,
+	opts: FetchOptions,
+	onFinish = _.noop,
+	serviceName: string,
+): Promise<null> {
```
20k-ultra, line 94: Consider returning `Promise<void>` because triggerFetch is a command rather then query and the return type is the give away for the difference. 

```suggestion
): Promise<void> {
```

test/05-device-state.spec.ts
```diff
@@ -230,24 +234,30 @@ describe('deviceState', () => {
 			Promise.resolve('172.17.0.1'),
 		);
 
-		stub(deviceState.applications.images, 'inspectByName').callsFake(() => {
-			const err: any = new Error();
+		// @ts-expect-error Assigning to a RO property
```
20k-ultra, line 237: We recently agreed to start using ts-expect-error however, @Page- had mentioned some valid points about the use of ts-expect-error vs ts-ignore. I am convinced for situations such as this where we are just suppressing an error to achieve the right conditions for a test then ts-ignore is more suitable. This directive is found in multiple places in the PR so if you make the change I'd apply to all the other places as well.

src/application-manager.js
```diff
@@ -28,7 +28,7 @@ import { TargetStateAccessor } from './device-state/target-state-cache';
 
 import { ServiceManager } from './compose/service-manager';
 import { Service } from './compose/service';
-import { Images } from './compose/images';
+import * as images from './compose/images';
```
20k-ultra, line 31: I'm in favour of making all imports of the image module to be `Images` not that it is a singleton. This first letter uppercase convention for singletons is something I've seen before and provides an indicator that it is a singleton rather then a variable that exists as multiple copies.

Doing that would also lets us resolve naming conflicts that had to be fixed like `images` being `imgs`:
https://github.com/balena-io/balena-supervisor/blob/abeb33ae1fd9b14e014c267df9c903368bbe064c/src/application-manager.js#L264

Examples like this are found throughout the PR so if implemented we should make sure to follow this convention.

src/compose/images.ts
```diff
@@ -52,584 +47,599 @@ type NormalisedDockerImage = Docker.ImageInfo & {
...
+		throw e;
 	}
+}
 
-	public async removeByDockerId(id: string): Promise<void> {
-		const image = await this.getByDockerId(id);
-		await this.remove(image);
-	}
+export async function getByDockerId(id: string): Promise<Image> {
+	return await db.models('image').where({ dockerImageId: id }).first();
```
20k-ultra, line 196: We can remove the await and just return. In examples such as: https://github.com/balena-io/balena-supervisor/blob/abeb33ae1fd9b14e014c267df9c903368bbe064c/src/compose/images.ts#L413-L415 we have return await so the rejection gets caught in the try rather then returned. There are a few instances in the PR we can remove await in the return.

src/compose/images.ts
```diff
@@ -52,584 +47,598 @@ type NormalisedDockerImage = Docker.ImageInfo & {
...
+		(dockerImage) =>
+			matchesTagOrDigest(image, dockerImage) ||
+			image.dockerImageId === dockerImage.Id,
+	);
+}
 
-		return images;
-	}
+export async function getAvailable(): Promise<Image[]> {
+	const images = await withImagesFromDockerAndDB(
```
richbayliss, line 259:
```suggestion
	return withImagesFromDockerAndDB(
```

... just a NIT but I think this is equivailent, no?

Issue #1207, opened 2020-02-19
Allow switching dtb in extlinux.conf for nvidia devices


acostach:
This feature is related to switching dtbs from dashboard , in a similar way to how it's done for the RPI. Therefore a similar property
called DTB needs to be added to the device section in the dashboard. User will fill in the dtb file name in the dashboard, and the supervisor needs to check
if the file exists in the current active rootfs in /boot/, and modify /boot/extlinux/extlinux.conf
to specify the dtb. Example:

LABEL primary
  MENU LABEL primary Image
  LINUX /boot/Image
  FDT /boot/mycustomdtb.dtb <-- this path should be added, or replaced in extlinux.conf
  APPEND ${cbootargs} ${resin_kernel_root} ro rootwait sdhci_tegra.en_boot_part_access=1 ${os_cmdline}

The extlinux.conf file and the dtbs are located in the current active rootfs. This feature is added in L4T 32.3.1, and is available for all jetson device types.

HQ: https://github.com/balena-io/balena/issues/2166

AeroNotix:
Will this feature allow users to upload a custom DTB into the dashboard and have that set as the device-tree?

CameronDiver:
@AeroNotix At least not in the first iteration, the dtb must be present in the host OS I believe.

AeroNotix:
From what @acostach mentioned it seemed like that could be the case that users would be able to upload their own.

Even if this isn't the initial implementation it would be a great addition - the DTB file the user uploads could be injected into the img.zip file. For example, there's a feature I haven't used myself, but is documented to work where the dtb is just appended to the Linux zImage, and Linux will use this appended device-tree.

In addition to the above feature, it would be very useful if somehow the DTS/DTSI files were hosted alongside the DTB files such that users could understand from where the resulting DTB files were created as it would be a big service to the TX2 community as a whole because Nvidia is..... lets just say "lacking", in this area, with many (including myself at one point) scouring documentation and vague forum posts to build the proper device tree.

thgreasi:
After checking with Rich, it seems that we are going to call it `RESIN_HOST_EXTLINUX_fdt`.

AeroNotix:
@acostach How's this feature going? Interested to start testing it.

CameronDiver:
Hey @AeroNotix we have a plan to start moving forward with this, and should have more news fairly soon.

acostach:
Adding discussion from the new arch call topic and in fd here for consistency:

[...]

PR #1371, opened 2020-06-10
Support setting FDT entry in extlinux.conf

20k-ultra:
- [x] Ensure test coverage for previous functionality
- [x] Separate config backends into their own files
- [x] Implement sorting target config variables by their entry type
- [x] Implement a APPEND interface so APPEND entry type variables can be set (uses existing code)
- [x] Implement a FDT interface so FDT entry type variables can be set

Closes: #1207 


src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface Directive {
+	[directive: string]: string;
+}
+
+export abstract class ConfigurableDirective {
+	// Parses the values for this directive
+	public abstract parse(directives: Directive): ConfigOptions;
+	// Return the value to be set for this directive using the provided ConfigOptions
+	public abstract value(opts: ConfigOptions, existingValue: string): string;
```
CameronDiver, line 23: At the risk of bike-shedding here, I'm not super fond of this method name. Something a little more applicable, such as `generate` or `compose` might be nice (although compose might be confused with docker-compose).

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface Label {
+	[labelName: string]: Directive;
+}
+
+export interface Directive {
+	[directive: string]: string;
+}
+
+export abstract class ConfigurableDirective {
```
CameronDiver, line 19: Given that we were talking about the ease of creating new directives, could you add a couple of comments here detailing that this is the base class of the directives, to make a new one, implement this etc etc.

Basically just an overview of what its for for any future contributors.

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+	labels: Label;
+	globals: Directive;
+}
+
+export interface Label {
+	[labelName: string]: Directive;
+}
+
+export interface Directive {
+	[directive: string]: string;
```
CameronDiver, line 16: We could more strongly type this, right? 
```
export interface Directive {
    [key in 'APPEND' | 'FDT']: string
```

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+				'Could not find APPEND directive in default extlinux.conf boot entry',
+			);
+		}
+
+		const configOpts: ConfigOptions = {};
+		const values = appendEntry.split(' ');
+
+		for (const value of values) {
+			const parts = value.split('=');
+			if (parts.length !== 2) {
```
CameronDiver, line 41: Did you look at whether the values could contain a `=` too? How is that handled. Given we're testing for example 2 entries, that could cause issues. I expect they're just not allowed, but it's worth checking.

src/config/backends/extlinux.ts
```diff
@@ -248,4 +232,31 @@ export class ExtlinuxConfigBackend extends DeviceConfigBackend {
 		});
 		return ret;
 	}
+
+	private static findDefaultLabel(file: ExtlinuxFile): string {
+		const defaultLabel = _.find(file.globals, (_v, l) => {
+			if (l === 'DEFAULT') {
```
CameronDiver, line 238:
```suggestion
			return label === 'DEFAULT'
```

src/lib/errors.ts
```diff
@@ -103,6 +103,12 @@ export class ContractViolationError extends TypedError {
 	}
 }
 
+export class ExtLinuxParseError extends TypedError {
+	public constructor(parseError: string) {
```
CameronDiver, line 107: You can omit the constructor when it only calls `super`.

src/config/backends/raspberry-pi.ts
```diff
@@ -0,0 +1,157 @@
...
+		return _.startsWith(envVar, RPiConfigBackend.bootConfigVarPrefix);
+	}
+
+	public processConfigVarName(envVar: string): string {
+		return envVar.replace(RPiConfigBackend.bootConfigVarRegex, '$2');
+	}
+
+	public processConfigVarValue(key: string, value: string): string | string[] {
+		if (_.includes(RPiConfigBackend.arrayConfigKeys, key)) {
+			if (!_.startsWith(value, '"')) {
```
Page-, line 145: If we know it's a string then we can use the native version
```suggestion
			if (!value.startsWith('"')) {
```
- the same goes for the other usages elsewhere

src/config/backends/raspberry-pi.ts
```diff
@@ -0,0 +1,157 @@
...
+	public isBootConfigVar(envVar: string): boolean {
+		return _.startsWith(envVar, RPiConfigBackend.bootConfigVarPrefix);
+	}
+
+	public processConfigVarName(envVar: string): string {
+		return envVar.replace(RPiConfigBackend.bootConfigVarRegex, '$2');
+	}
+
+	public processConfigVarValue(key: string, value: string): string | string[] {
+		if (_.includes(RPiConfigBackend.arrayConfigKeys, key)) {
```
Page-, line 144: Same here, if we know it's an array we can use the native version
```suggestion
		if (RPiConfigBackend.arrayConfigKeys.includes(key)) {
```
- the same goes for the other usages elsewhere

src/config/backends/raspberry-pi.ts
```diff
@@ -0,0 +1,157 @@
...
+			if (keyValue != null) {
+				const [, key, value] = keyValue;
+				if (!_.includes(RPiConfigBackend.arrayConfigKeys, key)) {
+					conf[key] = value;
+				} else {
+					if (conf[key] == null) {
+						conf[key] = [];
+					}
+					const confArr = conf[key];
+					if (!_.isArray(confArr)) {
```
Page-, line 88:
```suggestion
					if (!Array.isArray(confArr)) {
```

src/config/backends/raspberry-pi.ts
```diff
@@ -0,0 +1,157 @@
...
+ * 	- {BALENA|RESIN}_HOST_CONFIG_device_tree_overlay = value | "value" | "value1","value2"
+ * 	- {BALENA|RESIN}_HOST_CONFIG_gpio = value | "value" | "value1","value2"
+ */
+
+export class RPiConfigBackend extends DeviceConfigBackend {
+	private static bootConfigVarPrefix = `${constants.hostConfigVarPrefix}CONFIG_`;
+	private static bootConfigPath = `${bootMountPoint}/config.txt`;
+
+	public static bootConfigVarRegex = new RegExp(
+		'(' + _.escapeRegExp(RPiConfigBackend.bootConfigVarPrefix) + ')(.+)',
```
Page-, line 29: Since you only seem to care about the second group you could make the first non-capturing like
```suggestion
		'(?:' + _.escapeRegExp(RPiConfigBackend.bootConfigVarPrefix) + ')(.+)',
```
and then use `$1` in the replacement instead

src/config/backends/extlinux.ts
```diff
@@ -0,0 +1,262 @@
...
+	private static bootConfigPath = `${bootMountPoint}/extlinux/extlinux.conf`;
+	private static supportedConfigValues = ['isolcpus', 'fdt'];
+	private static supportedDirectives = ['APPEND', 'FDT'];
+
+	// These are the directives we can configure
+	private appendDirective = new AppendDirective();
+	private fdtDirective = new FDTDirective();
+
+	public static bootConfigVarRegex = new RegExp(
+		'(' + _.escapeRegExp(ExtlinuxConfigBackend.bootConfigVarPrefix) + ')(.+)',
```
Page-, line 39: Since you only seem to care about the second group you could make the first non-capturing like
```suggestion
		'(?:' + _.escapeRegExp(ExtlinuxConfigBackend.bootConfigVarPrefix) + ')(.+)',
```
and then use `$1` in the replacement instead

src/config/backends/raspberry-pi.ts
```diff
@@ -0,0 +1,157 @@
+import * as _ from 'lodash';
+import { fs } from 'mz';
```
Page-, line 2: This should be able to be
```suggestion
import { promises as fs } from 'fs';
```
now that the supervisor uses nodejs 12 and work the same, except using the in-built promise interface (it was added in nodejs 10)
- the same goes for other usage as well

src/config/backends/extlinux.ts
```diff
@@ -0,0 +1,262 @@
...
+		return ret;
+	}
+
+	private static findDefaultLabel(file: ExtlinuxFile): string {
+		const defaultLabel = _.find(file.globals, (_v, l) => {
+			if (l === 'DEFAULT') {
+				return true;
+			}
+			return false;
+		});
```
Page-, line 242: I think this can be
```suggestion
		const defaultLabel = file.globals.DEFAULT;
```
?

src/config/backends/extlinux.ts
```diff
@@ -0,0 +1,262 @@
...
+	private static parseExtlinuxFile(confStr: string): ExtlinuxFile {
+		const file: ExtlinuxFile = {
+			globals: {},
+			labels: {},
+		};
+
+		// Firstly split by line and filter any comments and empty lines
+		let lines = confStr.split(/\r?\n/);
+		lines = _.filter(lines, (l) => {
+			const trimmed = _.trimStart(l);
```
Page-, line 183: You could technically include this trim in the split above with
```ts
let lines = confStr.split(/\r?\n\s*/);
```
I'm not sure I'm a big fan of it, but I found it interesting to note.

And actually I just realized you could replace the entire filter with
```ts
let lines = confStr.split(/(?:\r?\n[\s#]*)+/);
```

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+	public parse(directives: Directive): ConfigOptions {
+		// NOTE: We normalize FDT to lowercase fdt
+		return directives.FDT ? { fdt: directives.FDT } : {};
+	}
+
+	public value(opts: ConfigOptions): string {
+		if (opts.fdt.length === 0) {
+			throw new Error('Cannot set FDT of an empty value.');
+		}
+		return opts.fdt as string;
```
Page-, line 84:
```suggestion
		if (typeof opts.fdt !== 'string') {
			throw new Error($errorMessage);
		}
		if (opts.fdt.length === 0) {
			throw new Error('Cannot set FDT of an empty value.');
		}
		return opts.fdt;
```
to avoid the need for the cast, especially as I think this can come from user input?

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+				// Example: rootwait (config without a value)
+				appendLine.push(`${key}`);
+			} else {
+				// Example: isolcpus=2,3 (config with a value)
+				appendLine.push(`${key}=${value}`);
+			}
+		});
+
+		// Return modified line as a string
+		return appendLine.join(' ');
```
Page-, line 70: I prefer a map/concat style as it makes it clear that we're mapping over/transforming the array in a direct 1-1 style
```suggestion
		return appendLine.concat(_.map(opts, (value, key) => {
			if (!value) {
				// Example: rootwait (config without a value)
				return `${key}`;
			} else {
				// Example: isolcpus=2,3 (config with a value)
				return `${key}=${value}`;
			}
		})).join(' ');
```

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,86 @@
...
+				configOpts[parts[0]] = parts[1];
+			}
+		}
+
+		return configOpts;
+	}
+
+	public value(opts: ConfigOptions, existingValue: string): string {
+		// Parse current append line and remove already set values so we don't have duplicates
+		const appendLine = _.filter(existingValue.split(' '), (entry) => {
```
Page-, line 53:
```suggestion
		const appendLine = existingValue.split(' ').filter((entry) => {
```

src/config/backends/config-fs.ts
```diff
@@ -0,0 +1,231 @@
...
+			return {};
+		}
+	}
+
+	private async writeConfigJSON(config: ConfigfsConfig): Promise<void> {
+		await remountAndWriteAtomic(this.ConfigFilePath, JSON.stringify(config));
+	}
+
+	private async loadConfiguredSsdt(config: ConfigfsConfig): Promise<void> {
+		if (_.isArray(config['ssdt'])) {
```
Page-, line 124:
```suggestion
		if (Array.isArray(config['ssdt'])) {
```

src/config/backends/config-fs.ts
```diff
@@ -0,0 +1,231 @@
...
+				fs.readFile(path.join(amlDstPath, 'oem_id'), 'utf8'),
+				fs.readFile(path.join(amlDstPath, 'oem_table_id'), 'utf8'),
+				fs.readFile(path.join(amlDstPath, 'oem_revision'), 'utf8'),
+			]);
+
+			log.info(
+				`AML: ${oemId.trim()} ${oemTableId.trim()} (Rev ${oemRevision.trim()})`,
+			);
+		} catch (e) {
+			log.error(e);
```
Page-, line 97: Can you add a string message to go with `e`, ie
```suggestion
			log.error($errorMessage, e);
```
It makes it far easier to find where a log message is coming from in the case that `e` itself is not very useful, or could be coming from multiple places

src/config/backends/extlinux-file.ts
20k-ultra: I totally agree. Compose is perfect but in the off chance it can be confused with docker-compose I'll opt for generate.

src/config/backends/extlinux-file.ts
20k-ultra: I've added documentation to the class abstraction and each directive class.

src/lib/errors.ts
20k-ultra: fixed!

src/config/backends/extlinux-file.ts
20k-ultra: To my knowledge a key within the APPEND value cannot have more then one =.

APPEND ro rootwait isolcpu=2,3 other=4 // OK
APPEND ro rootwait isolcpu=2,3other=4 // WRONG

Therefore your comment has made me aware of an issue where the parse wouldn't error on these invalid values. I have modified it to catch any config options that have more then a KEY and VALUE.

```js
const [KEY, VALUE, ...more] = appendValue.split('=');
if (more.length > 0) {
// APPEND value is not formatted correctly
}
```

src/config/backends/extlinux-file.ts
20k-ultra: I have been able to achieve a stronger typing for Directive but it actually just complicated things more. I may need some help to figure out a way to do this.

src/config/backends/extlinux.ts
20k-ultra: added.

src/config/backends/raspberry-pi.ts
20k-ultra: I've replaced any lodash implementations of startsWith to use standard lib startsWith

src/config/backends/raspberry-pi.ts
20k-ultra: I've replaced any lodash implementations of includes to use standard lib includes

src/config/backends/raspberry-pi.ts
20k-ultra: I've replaced all Lodash instance of Array checking with standard lib's version

src/config/backends/extlinux.ts
20k-ultra: Yes, and I was able to make it even simpler by just check if the value is falsey else return it. No need to find.

src/config/backends/raspberry-pi.ts
20k-ultra: added

src/config/backends/extlinux.ts
20k-ultra: added

src/config/backends/extlinux-file.ts
20k-ultra: added

src/config/backends/extlinux-file.ts
20k-ultra: I agree and added this implementation.

src/config/backends/extlinux.ts
20k-ultra: Yup this works great!

20k-ultra:
I have added most of the changes commented except for the 2 that are currently unresolved. I also added 2 new tests files specifically for each directive. This let me find a lot of potential bugs so it was a great addition.

src/config/backends/raspberry-pi.ts
```diff
@@ -0,0 +1,157 @@
...
+			if (key === 'initramfs') {
+				confStatements.push(`${key} ${value}`);
+			} else if (Array.isArray(value)) {
+				confStatements = confStatements.concat(
+					value.map((entry) => `${key}=${entry}`),
+				);
+			} else {
+				confStatements.push(`${key}=${value}`);
+			}
+		});
```
Page-, line 124:
```suggestion
		const confStatements = _.flatMap(opts, (value, key) => {
			if (key === 'initramfs') {
				return `${key} ${value}`;
			} else if (Array.isArray(value)) {
				return value.map((entry) => `${key}=${entry}`);
			} else {
				return `${key}=${value}`;
			}
		});
```

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,166 @@
...
+		});
+		// Add new configurations values to the provided append line
+		return appendLine
+			.concat(
+				_.map(opts, (value, key) => {
+					if (key.includes('=') || value.includes('=')) {
+						throw new AppendDirectiveError(
+							`One of the values being set contains an invalid character: [ value: ${value}, key: ${key} ]`,
+						);
+					} else if (!value) {
```
Page-, line 113: What are you looking for with `!value`? If it can be null then the `value.includes('=')` would need to be `value?.includes('=')`

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,166 @@
...
+	 * Example:
+	 * 	generate({ isolcpus: '0,4' })
+	 * 	-> 'isolcpus=0,4'
+	 *
+	 */
+	public generate(opts: ConfigOptions, existingValue: string = ''): string {
+		// Parse current append line and remove whitelisted values
+		// We remove whitelisted values to avoid duplicates
+		const appendLine = existingValue.split(' ').filter((entry) => {
+			const lhs = entry.split('=');
```
Page-, line 102: It's worth limiting this to return just the first result since that's all we care about:
```suggestion
			const lhs = entry.split('=', 1);
```

src/config/backends/extlinux-file.ts
```diff
@@ -0,0 +1,166 @@
...
+			);
+		}
+		// Parse all the key and values into ConfigOptions
+		return directives.APPEND.split(' ').reduce(
+			(configOpts: ConfigOptions, appendValue: string) => {
+				// Break this append config into key and value
+				const [KEY, VALUE = '', ...more] = appendValue.split('=');
+				if (!KEY) {
+					return configOpts; // No value to set so return
+				} else if (more.length > 0) {
```
Page-, line 73: You can limit to a max of 3 returned and also avoid building up the whole array of junk results by doing:
```suggestion
				const [KEY, VALUE = '', more] = appendValue.split('=', 3);
				if (!KEY) {
					return configOpts; // No value to set so return
				} else if (more != null) {
```
since `more` will either be a string or nullish:
```ts
'a=b='.split('=') // ["a", "b", ""]
'a=b'.split('=') // ["a", "b"]

'a=b==='.split('=') // ["a", "b", "", "", ""]
'a=b==='.split('=', 3) // ["a", "b", ""]

File: src/config/backends/config-fs.ts [hunk 24]
```diff
@@ -0,0 +1,231 @@
...
+
+	private async writeConfigJSON(config: ConfigfsConfig): Promise<void> {
+		await remountAndWriteAtomic(this.ConfigFilePath, JSON.stringify(config));
+	}
+
+	private async loadConfiguredSsdt(config: ConfigfsConfig): Promise<void> {
+		if (Array.isArray(config['ssdt'])) {
+			log.info('Loading configured SSDTs');
+			for (const aml of config['ssdt']) {
+				await this.loadAML(aml);
```
Page-, line 127: Should this be done in serial or in parallel?

src/config/backends/config-fs.ts
```diff
@@ -0,0 +1,231 @@
...
+			if (loadedTables.indexOf(aml) < 0) {
+				await fs.mkdir(amlDstPath);
+			}
+
+			log.info(`Loading AML ${aml}`);
+			// we use `cat` here as this didn't work when using `cp` and all
+			// examples of this loading mechanism use `cat`.
+			await child_process.exec(
+				`cat ${amlSrcPath} > ${path.join(amlDstPath, 'aml')}`,
+			);
```
Page-, line 85: Any reason not to handle this in nodejs, eg
```ts
fs.createReadStream(amlSrcPath).on('error', ...).pipe(fs.createWriteStream(path.join(amlDstPath, 'aml'))).on('error', ...);
```
it just feel a bit weird to be calling out to another binary to handle this

src/config/backends/raspberry-pi.ts
20k-ultra: I've created https://github.com/balena-io/balena-supervisor/issues/1384 to address this in another PR.

src/config/backends/raspberry-pi.ts
20k-ultra: added this change

src/config/backends/extlinux-file.ts
20k-ultra: good catch. I had thought includes would return false since that character is not included in null/undefined but actually includes is only for string prototype so it errors...

src/config/backends/extlinux-file.ts
20k-ultra: I like this philosophy of specifying the exact number of splits to return. I'll be sure to use it more. added.

src/config/backends/config-fs.ts
CameronDiver: @richbayliss should know the reason exactly, but I remember none of nodes tools seemed to work (streams, async and sync).

src/config/backends/extlinux-file.ts
20k-ultra: We agreed to move on from this for now as there's no decent solution I liked.

src/config/backends/config-fs.ts
20k-ultra: This is from existing functionality.

PR #1372, opened 2020-06-11
Move database app processing out to its own module

CameronDiver:
This is part of the work to make the application-manager module much
less monolithic, in preparation for system apps and more generally
multi-app.

Change-type: patch

### DIFF
--- a/src/application-manager.js
+++ b/src/application-manager.js
@@ -24,7 +24,7 @@ import {
 } from './lib/errors';
 import { pathExistsOnHost } from './lib/fs-utils';
 
-import { TargetStateAccessor } from './device-state/target-state-cache';
+import * as targetStateCache from './device-state/target-state-cache';
 
 import { ServiceManager } from './compose/service-manager';
 import { Service } from './compose/service';
@@ -185,8 +185,6 @@ export class ApplicationManager extends EventEmitter {
 		this._targetVolatilePerImageId = {};
 		this._containerStarted = {};
 
-		this.targetStateWrapper = new TargetStateAccessor(this);
-
 		this.actionExecutors = compositionSteps.getExecutors({
 			lockFn: this._lockingIfNecessary,
 			services: this.services,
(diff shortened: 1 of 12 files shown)


src/application-manager.js
```diff
@@ -225,34 +209,26 @@ export class ApplicationManager extends EventEmitter {
...
+		// This has the advantage that if for some reason a container is removed while the
+		// supervisor is down, we won't have zombie entries in the db
+
+		// Once a day
+		setInterval(cleanup, 1000 * 60 * 60 * 24);
+		// But also run it in on startup
+		await cleanup();
+		await this.localModeManager.init();
+		await this.services.attachToRunning();
+		this.services.listenToEvents();
```
richbayliss, line 231: should this not be a `return ...` ?

CameronDiver: Nah, it should be awaited though, strange, the editor said it didn't need it for whatever reason.

CameronDiver:
@balenaci rebase

src/device-state/db-format.ts
```diff
@@ -0,0 +1,185 @@
...
+			};
+		}),
+	);
+	await targetStateCache.setTargetApps(dbApps, trx);
+}
+
+function getDBEntry(): Promise<targetStateCache.DatabaseApp[]>;
+function getDBEntry(appId: number): Promise<targetStateCache.DatabaseApp>;
+async function getDBEntry(appId?: number) {
+	await config.initialized;
```
Page-, line 179: This should be necessary as this function doesn't reference `config` at all

src/device-state/db-format.ts
```diff
@@ -0,0 +1,185 @@
...
+) {
+	const cloned = _.cloneDeep(apps);
+
+	const dbApps = await Promise.all(
+		Object.keys(cloned).map(async (appIdStr) => {
+			const appId = checkInt(appIdStr)!;
+
+			const app = cloned[appId];
+			const services = await Promise.all(
+				_.map(app.services, async (s, sId) => ({
```
Page-, line 151: I know this is an pre-existing issue, but it does mean that services in the db format are an array whereas in the api version they're an object, which is fairly confusing and makes the types awkward as some places expect one style and other expect the other

src/device-state/db-format.ts
```diff
@@ -0,0 +1,185 @@
...
+		networks,
+	};
+}
+
+export async function setApps(
+	apps: { [appId: number]: TargetApplication },
+	source: string,
+	trx?: db.Transaction,
+) {
+	const cloned = _.cloneDeep(apps);
```
Page-, line 143: This can be avoided, as far as I can see the only possible mutation of `apps` this function might do is in `await images.normalise(s.image)` - unless you're worried about the caller mutating it during the async parts of this function (which is possible but I would expect unlikely)

src/device-state/db-format.ts
```diff
@@ -0,0 +1,185 @@
...
+					// Try to fill the image id if the image is downloaded
+					let imageInfo: ImageInspectInfo | undefined;
+					try {
+						imageInfo = await images.inspectByName(svc.image);
+					} catch (e) {
+						if (NotFoundError(e)) {
+							imageInfo = undefined;
+						} else {
+							throw e;
+						}
```
Page-, line 105: This can be
```ts
						if (!NotFoundError(e)) {
							throw e;
						}
```
since `imageInfo` will already be `undefined`

src/device-state/db-format.ts
```diff
@@ -0,0 +1,185 @@
...
+export async function getApp(id: number): Promise<InstancedApp> {
+	const dbApp = await getDBEntry(id);
+	return await buildApp(dbApp);
+}
+
+export async function getApps(): Promise<InstancedAppState> {
+	const dbApps = await getDBEntry();
+	const apps: InstancedAppState = {};
+	for (const app of dbApps) {
+		apps[app.appId] = await buildApp(app);
```
Page-, line 41: Did you want all of these to be done in serial?

src/device-state/db-format.ts
CameronDiver: Yeah, I was thinking about making that change to the database, but I'd rather do this first.

PR #1373, opened 2020-06-11
Make network-manager, volume-manager and service-manager singletons

CameronDiver:


test/lib/mocked-device-api.ts
```diff
@@ -133,20 +132,25 @@ function buildRoutes(appManager: ApplicationManager): Router {
...
 function setupStubs() {
-	stub(ServiceManager.prototype, 'getStatus').resolves(STUBBED_VALUES.services);
-	stub(NetworkManager.prototype, 'getAllByAppId').resolves(
-		STUBBED_VALUES.networks,
-	);
-	stub(VolumeManager.prototype, 'getAllByAppId').resolves(
-		STUBBED_VALUES.volumes,
-	);
+	// @ts-expect-error Assigning to a RO property
+	networkManager.getAllByAppId = () => Promise.resolve(STUBBED_VALUES.networks);
```
Page-, line 140: These would be nicer imo as
```suggestion
	networkManager.getAllByAppId = async () => STUBBED_VALUES.networks;
```

src/compose/service-manager.ts
```diff
@@ -32,626 +32,618 @@ type ServiceManagerEventEmitter = StrictEventEmitter<
...
-				this.volatileState[containerId] = {};
-				_.merge(this.volatileState[containerId], status);
-			} else if (this.volatileState[containerId] != null) {
-				delete this.volatileState[containerId];
-			}
+function reportChange(containerId?: string, status?: Partial<Service>) {
+	if (containerId != null) {
+		if (status != null) {
+			volatileState[containerId] = {};
+			_.merge(volatileState[containerId], status);
```
Page-, line 491: I think this can be
```suggestion
			volatileState[containerId] = { ...status };
```
since there's nothing to merge into. And if you don't need the cloning side-effect then just `= status` would work instead

src/compose/service-manager.ts
```diff
@@ -32,626 +32,618 @@ type ServiceManagerEventEmitter = StrictEventEmitter<
...
+		});
+	};
+
+	Bluebird.resolve(listen())
+		.catch((e) => {
+			log.error('Error listening to events:', e, e.stack);
+		})
+		.finally(() => {
+			listening = false;
+			setTimeout(() => listenToEvents(), 1000);
```
Page-, line 447:
```suggestion
			setTimeout(listenToEvents, 1000);
```

src/compose/service-manager.ts
```diff
@@ -32,626 +32,618 @@ type ServiceManagerEventEmitter = StrictEventEmitter<
...
+		try {
+			await container.start();
+		} catch (e) {
+			// Get the statusCode from the original cause and make sure it's
+			// definitely an int for comparison reasons
+			const maybeStatusCode = PermissiveNumber.decode(e.statusCode);
+			if (isLeft(maybeStatusCode)) {
+				shouldRemove = true;
+				err = new Error(`Could not parse status code from docker error:  ${e}`);
+				throw err;
```
Page-, line 312: There doesn't seem to be any reason to assign here
```suggestion
				throw new Error(`Could not parse status code from docker error:  ${e}`);
```

src/compose/service-manager.ts
```diff
@@ -32,626 +32,616 @@ type ServiceManagerEventEmitter = StrictEventEmitter<
...
+			if (statusCode === 304) {
+				alreadyStarted = true;
+			} else if (
+				statusCode === 500 &&
+				_.isString(message) &&
+				message.trim().match(/exec format error$/)
+			) {
+				// Provide a friendlier error message for "exec format error"
+				const deviceType = await config.get('deviceType');
+				throw new Error(
```
Page-, line 327: I just realized the reason for the previous `err =`, it's because you use it in the `finally` so it does need to be assigned (ideally with a comment somewhere about why we're always assigning the err) or the code needs to be reworked to handle it differently

src/compose/service-manager.ts
```diff
@@ -32,626 +32,616 @@ type ServiceManagerEventEmitter = StrictEventEmitter<
...
+export function getAllByAppId(appId: number) {
+	return getAll(`app-id=${appId}`);
+}
 
-	public async remove(service: Service) {
-		logger.logSystemEvent(LogTypes.removeDeadService, { service });
-		const existingService = await this.get(service);
+export async function stopAllByAppId(appId: number) {
+	for (const app of await getAllByAppId(appId)) {
+		await kill(app, { removeContainer: false });
```
Page-, line 224: Is this intended to be done in serial?

Page-: bump

CameronDiver: yeah sorry I hadn't submitted my comments.

CameronDiver:
@balena-ci rebase

src/compose/service-manager.ts
Page-: Right, but is it intended to be done in serial or is it a mistake that passed through previous code review?

PR #1377, opened 2020-06-15
Db-format module code fixups

CameronDiver:
Change-type: patch

### DIFF
--- a/src/device-state/db-format.ts
+++ b/src/device-state/db-format.ts
@@ -37,9 +37,11 @@ export async function getApp(id: number): Promise<InstancedApp> {
 export async function getApps(): Promise<InstancedAppState> {
 	const dbApps = await getDBEntry();
 	const apps: InstancedAppState = {};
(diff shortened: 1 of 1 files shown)


src/device-state/db-format.ts
```diff
@@ -37,9 +37,9 @@ export async function getApp(id: number): Promise<InstancedApp> {
 export async function getApps(): Promise<InstancedAppState> {
 	const dbApps = await getDBEntry();
 	const apps: InstancedAppState = {};
-	for (const app of dbApps) {
-		apps[app.appId] = await buildApp(app);
-	}
+	await Promise.all(
+		dbApps.map(async (app) => (apps[app.appId] = await buildApp(app))),
```
Page-, line 41: I think this would be clearer as
```suggestion
		dbApps.map(async (app) => {
			apps[app.appId] = await buildApp(app);
		}),
```

PR #1378, opened 2020-06-15
Use API v6

dfunckt:
open-balena-api still only exposes v5, but I'd like to get the ball rolling as it'll take a while until the new Supervisor makes it into a new stable balenaOS release.

See: https://github.com/balena-io/open-balena/issues/80
Change-type: minor

### DIFF
--- a/src/api-binder.ts
+++ b/src/api-binder.ts
@@ -153,7 +153,7 @@ export class APIBinder {
 			return;
 		}
 
-		const baseUrl = url.resolve(apiEndpoint, '/v5/');
+		const baseUrl = url.resolve(apiEndpoint, '/v6/');
 		const passthrough = _.cloneDeep(await request.getRequestOptions());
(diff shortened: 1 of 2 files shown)


dfunckt:
It's also worth mentioning that the Supervisor should move away from the OData API. AFAIK it currently only uses it to update the host config variables, but we should consider making an explicit endpoint on the API instead. There are several reasons why the endpoint is better:

- all communication with the API is explicit (ie. no more "AFAIK"s going forward :)
- it's much easier to maintain compatibility with old Supervisors that way from the API
- we won't have to bump versions like here

CameronDiver:
@dfunckt when open switches to v6, will v5 supervisor versions stop working?

dfunckt:
Yes @CameronDiver

dfunckt:
@balena-ci retest

PR #1380, opened 2020-06-16
Add host firewall

richbayliss:


src/lib/firewall.ts
```diff
@@ -0,0 +1,198 @@
...
+
+import * as dbFormat from '../device-state/db-format';
+
+export const initialised = (async () => {
+	await config.initialized;
+	await applyFirewall();
+
+	// apply firewall whenever relevant config changes occur...
+	config.on('change', async ({ firewallMode, localMode }) => {
+		if (firewallMode || localMode !== undefined) {
```
CameronDiver, line 16:
```suggestion
		if (firewallMode || localMode != null) {
```
I always encourage this until everything in TS in the supervisor (and even then I'd probably continue to).

src/lib/firewall.ts
```diff
@@ -0,0 +1,198 @@
...
+			applyFirewall({ firewallMode, localMode });
+		}
+	});
+})();
+
+const BALENA_FIREWALL_CHAIN = 'BALENA-FIREWALL';
+const LOG_PREFIX = '\u{1F525}';
```
CameronDiver, line 23: Is this the fire emoji? You can just add it directly

src/lib/firewall.ts
```diff
@@ -0,0 +1,198 @@
...
+];
+
+const standardPolicy: iptables.Rule[] = [
+	{
+		comment: 'Locally-sourced traffic',
+		action: '-I',
+		matches: ['-m addrtype', '--src-type LOCAL'],
+		target: 'ACCEPT',
+	},
+	{
```
CameronDiver, line 70: What's this rule for?

src/lib/firewall.ts
```diff
@@ -0,0 +1,198 @@
...
+			supervisorAccessRules.push({
+				comment: 'Supervisor API',
+				action: '-A',
+				proto: 'tcp',
+				matches: [`--dport ${port}`, `-i ${intf}`],
+				target: 'ACCEPT',
+				applyIfExists: false,
+			});
+		});
+	}
```
CameronDiver, line 111:
```suggestion

			supervisorAccessRules.push({
				comment: 'Supervisor API',
				action: '-A',
				proto: 'tcp',
				matches: [`--dport ${port}`].concat(localMode? [] : [`-i ${intf}`']),
				target: 'ACCEPT',
				applyIfExists: false,
			});
	}
```

src/lib/firewall.ts
```diff
@@ -0,0 +1,198 @@
...
+		'listenPort',
+		'firewallMode',
+		'localMode',
+	]);
+
+	// populate missing config elements...
+	const { listenPort, firewallMode, localMode } = _.defaults(
+		opts,
+		currentConfig,
+	);
```
CameronDiver, line 136:
```suggestion
	const { listenPort, firewallMode, localMode } = {
		...opts,
		...currentConfig,
	};
```

src/lib/firewall.ts
```diff
@@ -0,0 +1,198 @@
...
+		localMode,
+		constants.allowedInterfaces,
+		listenPort,
+	);
+
+	// apply the firewall rules...
+	await exports.applyFirewallMode(firewallMode ?? '');
+}
+
+export const AllowedModes = ['on', 'off', 'auto'];
```
CameronDiver, line 149: Either `ALLOWED_MODES` or `allowedModes`, only type names should be capitalized and camel case.

src/lib/iptables.ts
```diff
@@ -1,66 +1,170 @@
...
+	if (rule.comment) {
+		args.push('-m comment');
+		args.push(`--comment "${rule.comment}"`);
+	}
+	if (rule.target) {
+		args.push(`-j ${rule.target}`);
+	}
+
+	if (!rule.family) {
+		rule.family = 4;
```
CameronDiver, line 62: Weren't we going to apply to both by default?

src/lib/iptables.ts
```diff
@@ -1,66 +1,170 @@
...
+				...rule,
+				...{
+					family: 4,
+				},
+			},
+			adaptor,
 		);
+	}
+
+	if (rule.applyIfExists !== undefined) {
```
CameronDiver, line 132:
```suggestion
	if (rule.applyIfExists != null) {
```

src/lib/iptables.ts
```diff
@@ -1,66 +1,170 @@
...
-		),
-	)
-		.then(() =>
-			clearAndAppendIptablesRule(
-				`OUTPUT -p tcp --sport ${port} -m state --state ESTABLISHED -j ACCEPT`,
-			),
-		)
-		.then(() =>
-			clearAndAppendIptablesRule(`INPUT -p tcp --dport ${port} -j REJECT`),
+async function applyRule(rule: Rule, adaptor: RuleAdaptor) {
```
CameronDiver, line 118: Can you either split this function or add some comments please? Following what's going on here isn't too easy.

src/lib/firewall.ts
richbayliss: Locally-sourced traffic -- traffic coming into the device from itself; stuff like DNS lookups that use 127.0.0.2

src/lib/firewall.ts
richbayliss: there's also a difference in that the non-localmode path has to create this rule for each interface

richbayliss: also, you cannot pass them in one hit -- `multiple -i flags not allowed`

src/lib/iptables.ts
richbayliss: We do, just not here -- this is the adaptor which applies the rule and it _needs_ a family is one is missing. However, we don't call this externally, we have another function which wraps/sanitises the *Rule* object and it handles the extra apply if it's missing a family 👍

src/lib/iptables.ts
```diff
@@ -1,66 +1,184 @@
...
+import { child_process } from 'mz';
 
-function clearIptablesRule(rule: string): Bluebird<void> {
-	return applyIptablesArgs(`-D ${rule}`);
+export class IPTablesRuleError extends Error {}
+
+export interface Rule {
+	id?: number;
+	family?: 4 | 6;
+	action?: '-L' | '-I' | '-A' | '-D' | '-C' | '-N' | '-F' | '-X';
```
wrboyce, line 9: I'd be tempted to use an Enum here

```typescript
enum RuleAction {
    List = '-L',
    Insert = '-I',
    // etc
}
```

PR #1381, opened 2020-06-16
Update pinejs-client-request to 7.x

Page-:
Update pinejs-client-request from 5.2.0 to 7.0.0

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test:debug": "npm run test:build && TEST=1 mocha --inspect-brk",
     "prettify": "balena-lint -e ts -e js --typescript --fix src/ test/ typings/ build-utils/ webpack.config.js",
     "test-build": "tsc --project tsconfig.json",
-    "release": "tsc --project tsconfig.release.json && cp -r build/src/* build && rm -rf build/src",
+    "release": "tsc --project tsconfig.release.json && mv build/src/* build",
     "packagejson:copy": "cp package.json build/",
     "testitems:copy": "cp -r test/data build/test/",
     "sync": "ts-node sync/sync.ts"
@@ -101,7 +101,7 @@
     "mz": "^2.7.0",
     "network-checker": "^0.1.1",
     "nodemon": "^2.0.2",
-    "pinejs-client-request": "^5.2.0",
+    "pinejs-client-request": "^7.0.0",
     "pretty-ms": "^5.1.0",
     "request": "^2.88.2",
     "resin-cli-visuals": "^1.5.2",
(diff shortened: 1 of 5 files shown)


src/api-binder.ts
```diff
@@ -283,9 +283,9 @@ export class APIBinder {
 		}
 
 		try {
-			const res = (await this.balenaApi
-				.get(reqOpts)
-				.timeout(timeout)) as Device[];
+			const res = (await Bluebird.resolve(this.balenaApi.get(reqOpts)).timeout(
```
Page-, line 286: One of the breaking changes is using native promises and async/await in pinejs-client-request now, hence this wrapping for the `.timeout`

webpack.config.js
```diff
@@ -78,6 +78,10 @@ module.exports = function(env) {
 		},
 		resolve: {
 			extensions: ['.js', '.ts', '.json'],
+			alias: {
+				// Use the es2018 build instead of the default es2015 build
+				'pinejs-client-core': 'pinejs-client-core/es2018',
```
Page-, line 83: As the comment says - and es2018 is nice as it lets us use a build with native async/await rather than transpiled

package.json
```diff
@@ -20,7 +20,7 @@
     "test:debug": "npm run test:build && TEST=1 mocha --inspect-brk",
     "prettify": "balena-lint -e ts -e js --typescript --fix src/ test/ typings/ build-utils/",
     "test-build": "tsc --project tsconfig.json",
-    "release": "tsc --project tsconfig.release.json && cp -r build/src/* build && rm -rf build/src",
+    "release": "tsc --project tsconfig.release.json && mv build/src/* build",
```
CameronDiver, line 23: Why have you made this change? We've still got `src/` there.

Page-: I can remove the empty dir if you want but it won't affect anything.  
As for why mv instead of cp, it's twofold:
1. it works better on windows so running `npx tsc --project tsconfig.release.json; mv build/src/* build` can work whereas `cp ..; rm -rf build/src` won't (the way to pass the `-f` flag is slightly different because with just `-f` it's ambiguous with another flag)
2. it should be faster as it's not having to copy all the contents and can instead change just the file pointers - on some filesystems with the right options passed to cp and the right versions of things then it may do a COW and end up equivalent, but that's a lot of ifs when a mv makes it simple :P

CameronDiver: No that's fine, it was just a bit of a strange change so wanted further clarification.

Page-:
@balena-ci rebase

PR #1392, opened 2020-07-10
bug: Fix unhandled promise rejection

richbayliss:
When invoking iptables-restore it can fail. This wasn't handled
and this makes sure that it fails gracefully.

Closes; https://github.com/balena-io/balena-supervisor/issues/1393
Change-type: patch

### DIFF
--- a/src/lib/firewall.ts
+++ b/src/lib/firewall.ts
@@ -4,6 +4,7 @@ import * as config from '../config/index';
 import * as constants from './constants';
 import * as iptables from './iptables';
 import { log } from './supervisor-console';
+import { logSystemMessage } from '../logger';
 
 import * as dbFormat from '../device-state/db-format';
 
@@ -12,7 +13,7 @@ export const initialised = (async () => {
 	await applyFirewall();
 
 	// apply firewall whenever relevant config changes occur...
-	config.on('change', async ({ firewallMode, localMode }) => {
+	config.on('change', ({ firewallMode, localMode }) => {
 		if (firewallMode || localMode != null) {
 			applyFirewall({ firewallMode, localMode });
 		}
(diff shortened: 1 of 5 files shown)


src/lib/iptables.ts
```diff
@@ -98,10 +102,11 @@ export function convertToRestoreRulesFormat(rules: Rule[]): string {
...
 				}
-				if (rule.comment) {
-					args.push('-m comment');
-					args.push(`--comment "${rule.comment}"`);
-				}
+				// TODO: Once comments are universally supported in meta-balena we can enable this
+				// if (rule.comment) {
+				// 	args.push('-m comment');
+				// 	args.push(`--comment "${rule.comment}"`);
+				// }
```
Page-, line 109: Is there any way to feature detect this?

richbayliss: @Page- we want to but we want to think through the best way to do the detection in more "general" way; perhaps contracts etc

src/lib/firewall.ts
```diff
@@ -137,56 +137,65 @@ async function applyFirewall(
 export const ALLOWED_MODES = ['on', 'off', 'auto'];
 
 export async function applyFirewallMode(mode: string) {
-	// only apply valid mode...
-	if (!ALLOWED_MODES.includes(mode)) {
-		log.warn(`Invalid firewall mode: ${mode}. Reverting to state: off`);
-		mode = 'off';
-	}
+	try {
```
20k-ultra, line 140: Can we remove all the variable declaration outside of the try block so it isn't as large ?

src/lib/firewall.ts
```diff
@@ -14,7 +14,7 @@ export const initialised = (async () => {
 	// apply firewall whenever relevant config changes occur...
 	config.on('change', async ({ firewallMode, localMode }) => {
 		if (firewallMode || localMode != null) {
-			applyFirewall({ firewallMode, localMode });
+			await applyFirewall({ firewallMode, localMode });
```
CameronDiver, line 17: I'm not sure you need this

src/lib/firewall.ts
```diff
@@ -137,56 +137,65 @@ async function applyFirewall(
...
+							}),
+					),
+			)
+			.apply(ruleAdaptor);
+
+		// all done!
+		log.success(`${LOG_PREFIX} Firewall mode applied`);
+	} catch (err) {
+		if (err instanceof iptables.IPTablesRuleError) {
+			log.debug(`Ruleset:\r\n${err.ruleset}`);
```
CameronDiver, line 195: It might make more sense to add this after the other error printouts. Also it's probably worth the user knowing on the dashboard that the firewall failed to apply.

test/29-firewall.spec.ts
```diff
@@ -7,22 +7,33 @@ import * as sinon from 'sinon';
...
+import * as logger from '../src/logger';
 import * as iptablesMock from './lib/mocked-iptables';
 import * as targetStateCache from '../src/device-state/target-state-cache';
 
 import constants = require('../src/lib/constants');
 import { RuleAction } from '../src/lib/iptables';
+import { log } from '../src/lib/supervisor-console';
 
 describe('Host Firewall', function () {
+	const dockerStubs: _.Dictionary<sinon.SinonStub> = {};
```
CameronDiver, line 19:
```suggestion
	const dockerStubs: Dictionary<sinon.SinonStub> = {};
```

Issue #1385, opened 2020-06-23
Set FDT and cmdline args with extra_uEnv.txt


20k-ultra:
**Update**: Through various discussions it has been agreed that moving forward the Supervisor will no longer directly modify the devices `extlinux.conf` file. To simplify how kernal configuration values like FDT and additional cmdline args are managed the Supervisor will now be using `extra_uEnv.txt`. 

This allows us to get/set values and each device will intercept the values it supports from there.  For example, we just recently added support for setting FDT directive in extlinux.conf which required extending the [extlinux](https://github.com/balena-io/balena-supervisor/blob/master/src/config/backends/extlinux.ts) backend. In this new scenario we would simple take the configuration value from the target state provided by the cloud and set it in extra_uEnv.txt using the agreed key (in this case custom_fdt_file).

To achieve this we will just need to add a new config backend but we must keep in mind what will happen if a user is running an old balenaOS version but decides to upgrade the Supervisor. Since old versions do not support `extra_uEnv.txt` their target state won't be applied. 

---

As of balenaOS version v2.47.0+rev1 for Jetson devices the extlinux.conf has been moved to a read only partition and so the current strategy for modifying extlinux backend does not work for these versions. The change can be found in this [commit](https://github.com/balena-os/balena-jetson/commit/00c0a0a22b45552d998ca95af2628ced2955322b).

As a result, https://github.com/balena-os/balena-jetson/pull/79 has been created. It works by having the Supervisor append key values to `extra_uEnv.txt` in the boot partition. The currently purposed variable naming is:

```
custom_fdt_file=mycustom.dtb
extra_cmdline_args=ro rootwait isolcpus=3,4
```

Therefore, we must 
- (1) create a new strategy that will detect when Supervisor is running in this environment so that it can then 
- (2) append target state values (fdt, kernal cmds) in extra_uEnv.txt.

For environment detection it cannot rely on just checking the release version because balenaOS ESR does not use semver versioning.

20k-ultra:
Updated post to use new file and variable names.

20k-ultra:
Updated original post to contain new plan for solving this issue.

acostach:
@20k-ultra we need to also note that the supervisor should NOT reboot the board due to any configuration change while HUP is running, or right after a reboot into the new OS, if the rollback files were not cleared yet.

20k-ultra:
[...]

PR #1397, opened 2020-07-17
Support setting device/fleet configuration in extra_uEnv.txt

20k-ultra:
Closes: #1385
Change-Type: minor


20k-ultra:
I'd like to hear feedback on how I used [versionRequirements.ts](https://github.com/balena-io/balena-supervisor/blob/2dfd14c3d51c13f3e233a74ccf60e7fdc68f808d/src/lib/versionRequirements.ts) and why I made it. There are no places in the codebase that I know of that check versions to execute different code paths and I really dislike that it was introduced but not sure how else to approach the problem.

**Why:** I needed a way check if the device supported extra_uEnv in 2 places (extlinux, extra_uenv backends) so wanted them both to reference the same static value rather then us managing 2 static versions. These 2 places need to check if this feature is supported so they know if they should execute.

**How:** I created this centralized version requirement file that could be extended to include additional versions of other features we ant to check for.

Let me know your thoughts.

src/config/backends/extra-uEnv.ts
```diff
@@ -0,0 +1,261 @@
...
+		const [, KEY, VALUE] = optionArray;
+		// Check if this key's value is a collection
+		if (ExtraUEnvConfigBackend.entries[KEY as EntryKey]?.collection) {
+			// Return split collection of options
+			return ExtraUEnvConfigBackend.parseOptionCollection(VALUE);
+		}
+		// Find the option that belongs to this entry
+		const optionKey = _.findKey(
+			ExtraUEnvConfigBackend.supportedConfigs,
+			(_v, k) => ExtraUEnvConfigBackend.supportedConfigs[k].key === KEY,
```
CameronDiver, line 166:
```suggestion
			(config) => config.key === KEY,
```

src/config/backends/extra-uEnv.ts
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	private static configToString(configs: ConfigOptions): string {
+		// Get Map of ConfigOptions object
+		const configMap = ExtraUEnvConfigBackend.configToMap(configs);
+		// Iterator over configMap and concat to configString
+		let configString = '';
+		for (const [key, value] of configMap) {
+			if (!value) {
+				log.warn(`Could not find value for ${key} in configMap.`);
```
CameronDiver, line 225: This shouldn't be possible, right?

src/config/backends/extra-uEnv.ts
```diff
@@ -0,0 +1,261 @@
...
+				// Check if we have to build the value for the entry
+				if (ENTRY_IS_COLLECTION) {
+					return configMap.set(
+						ENTRY_KEY,
+						// Append existing values if already set
+						(configMap.has(ENTRY_KEY) ? `${configMap.get(ENTRY_KEY)} ` : '') +
+							// Add new config's key
+							configKey +
+							// Append the config's value if it has one
+							(configValue !== '' ? `=${configValue}` : ''),
```
CameronDiver, line 251: There's a little bit too much going on here for my liking, I think even just another variable with a descriptive name would clear this up a little

src/config/backends/raspberry-pi.ts
```diff
@@ -50,7 +50,10 @@ export class RPiConfigBackend extends DeviceConfigBackend {
 		'avoid_safe_mode',
 	];
 
-	public matches(deviceType: string): boolean {
+	public matches(
+		deviceType: string,
+		_metaRelease: string | undefined,
+	): boolean {
```
CameronDiver, line 56:
```suggestion
	public matches(deviceType: string): boolean {
```

src/lib/os-release.ts
```diff
@@ -59,6 +59,12 @@ export function getOSSemver(path: string): Promise<string | undefined> {
 	return getOSReleaseField(path, 'VERSION');
 }
 
+export async function getMetaRelease(
```
CameronDiver, line 62:
```suggestion
export async function getMetaOSRelease(
```

src/lib/versionRequirements.ts
```diff
@@ -0,0 +1,7 @@
+const VERSIONS = {
+	EXTRA_SUPPORT: {
```
CameronDiver, line 2: This name isn't really descriptive enough for me. Looking at this file I (and without context) I don't really have any idea what this means, other than *something* to do with the jetson at v2.47

src/config/backends/extra-uEnv.ts
20k-ultra: ahh... Why did I write that... thank you for catching that.

src/config/backends/extra-uEnv.ts
20k-ultra: Very unlikely unless something had messed with the file. I was being extra verbose with this feature as I think more logs explaining why stuff doesn't happen is good.

src/config/backends/extra-uEnv.ts
20k-ultra: Refactored it into another function.

src/lib/versionRequirements.ts
20k-ultra: I've removed this file and instead added a single value to constants.ts

https://github.com/balena-io/balena-supervisor/blob/cbe4fc4e5b276e9965402f105069d43ab2d3fc11/src/lib/constants.ts#L68-L69

acostach:
In the current form, this PR works fine with a 2.52.7+rev1 build (custom dtb & isolcpus) and a 2.45.1+rev3 dashboard image (isolcpus).

20k-ultra:
> In the current form, this PR works fine with a 2.52.7+rev1 build (custom dtb & isolcpus) and a 2.45.1+rev3 dashboard image (isolcpus).

I would have expected this branch to still apply FDT for 2.45.1+rev3. If this is the case then I'll have to investigate this.

src/config/backends/backend.ts
```diff
@@ -52,7 +52,7 @@ export abstract class DeviceConfigBackend {
 	): string | string[];
 
 	// Return the env var name for this config option
```
CameronDiver, line 54: It'd be good if you could add a comment here that says what returning `null` actually means.

20k-ultra:
@balena-ci I self-certify!

Issue #1398, opened 2020-07-17
Restarting device from dashboard gets 503 error from supervisor: 


saintaardvark:
A forum user has reported that they are unable to use the "Restart" button on the dashboard on at least one of their devices; they get an error message in the dashboard saying `Request error: [object Object]`.  The device has been shared with support, and I've confirmed this behaviour.  

When I try to use the "Restart" button, I see this show up in the supervisor logs from the device:

```
[event]   Event: Restart container (v1) {"appId":[xxx]}
[error]   Device state apply error TypeError: Cannot read property 'ipam' of undefined
[error]         at Network.isEqualConfig (/usr/src/app/dist/app.js:2:960631)
[error]       at /usr/src/app/dist/app.js:22:13749
[error]       at arrayFilter (/usr/src/app/dist/app.js:2:86532)
[error]       at Function.lodash.filter (/usr/src/app/dist/app.js:2:175795)
[error]       at ApplicationManager._compareNetworksOrVolumesForUpdate (/usr/src/app/dist/app.js:22:13680)
[error]       at ApplicationManager.compareNetworksForUpdate (/usr/src/app/dist/app.js:22:13959)
[error]       at ApplicationManager._nextStepsForAppUpdate (/usr/src/app/dist/app.js:22:20489)
[error]       at /usr/src/app/dist/app.js:22:27294
[error]       at tryCatcher (/usr/src/app/dist/app.js:2:223577)
[error]       at Function.Promise.attempt.Promise.try (/usr/src/app/dist/app.js:6:70875)
[error]       at ApplicationManager._inferNextSteps (/usr/src/app/dist/app.js:22:26251)
[error]       at ApplicationManager.getRequiredSteps (/usr/src/app/dist/app.js:22:30834)
[error]       at /usr/src/app/dist/app.js:10:7618
[info]    Applying target state
[error]   Error on POST /v1/restart:  TypeError: Cannot read property 'ipam' of undefined
[error]         at Network.isEqualConfig (/usr/src/app/dist/app.js:2:960631)
[error]       at /usr/src/app/dist/app.js:22:13749
[error]       at arrayFilter (/usr/src/app/dist/app.js:2:86532)
[error]       at Function.lodash.filter (/usr/src/app/dist/app.js:2:175795)
[error]       at ApplicationManager._compareNetworksOrVolumesForUpdate (/usr/src/app/dist/app.js:22:13680)
[...]

PR #1400, opened 2020-07-20
Fix v1 invocations of restart and purge

CameronDiver:
The v1 endpoints had two problems:
* Purge was still expecting to pick up instances of singletons from within the application-manager, but this had changed recently. Unfortunately this wasn't picked up by the compiler as this was still javascript.
* Restart and Purge were both passing plain javascript objects to code which expected instances, and again, because there were plain javascript, this was not caught. We now clone the instances correctly, and also make sure to remove docker generated aliases that are added on container creation (as this confuses the supervisor when we set a previous current state to a new target state - like we do in restart and purge).

NB: Why do we set a previous current state as a new target state in restart and purge? Well when bringing the containers back up after both methods, the containers must be started in the correct order (due to depends_on). Rather than re-implement this, we allow the state engine to use its existing code to do this. 

The implementation is not great, but this is for the v1 endpoints, which are very soon going to be deprecated and right now it's about maintaining behavior. We should look at replacing the v1/restart and v1/purge endpoints with equivalent v2 endpoints, and remove their usage from the dashboard.

Closes: #1398

### DIFF
--- a/src/device-api/common.js
+++ b/src/device-api/common.js
@@ -10,11 +10,11 @@ export function doRestart(applications, appId, force) {
 
 	return _lockingIfNecessary(appId, { force }, () =>
 		deviceState.getCurrentForComparison().then(function (currentState) {
-			const app = currentState.local.apps[appId];
+			const app = safeAppClone(currentState.local.apps[appId]);
 			const imageIds = _.map(app.services, 'imageId');
 			applications.clearTargetVolatileForServices(imageIds);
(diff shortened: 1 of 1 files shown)


20k-ultra:
LGTM. I thought the exact same thing about how the compiler didn't catch this. We should try to chip away at these remaining parts of the codebase that are still JS. Application Manager is a rather large one.

PR #1402, opened 2020-07-21
log detection of changes to VPN status

dt-rush:
inspiration: https://jel.ly.fish/4cc20e2c-f7f3-4350-99ca-1cc6c451dcf1

Change-type: patch
Signed-off-by: dt-rush <nickp@balena.io>

### DIFF
--- a/src/network.ts
+++ b/src/network.ts
@@ -43,12 +43,14 @@ export function enableCheck(enable: boolean) {
 }
 
 export async function isVPNActive(): Promise<boolean> {
+	let active: boolean = true;
 	try {
 		await fs.lstat(`${constants.vpnStatusPath}/active`);
-	} catch {
-		return false;
+	} catch (e) {
+		active = false;
 	}
-	return true;
+	log.info(`VPN connection is ${active ? 'active' : 'not active'}.`);
+	return active;
 }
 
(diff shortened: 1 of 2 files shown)


20k-ultra:
Thanks @dt-rush. I think the place you decided to add logs provides much more insight because if we wanted to now we can query [/v2/device/vpn](https://www.balena.io/docs/reference/supervisor/supervisor-api/#device-vpn-information). This leaves a trail in the scenario that someone is actively monitoring the VPN connection with this endpoint the logs will show the status without having for them to persist the network responses. It also will be logged when the VPN active file is modified and the Supervisor checks it.

This might be redundant as the VPN service will probably log disconnects however, being able to reference that with what the Supervisor thinks about the VPN status could be useful. I don't think there's any harm in having more insight.

Here's the JF thread with our conversation as the support ticket itself doesn't have us talking about this change: https://www.flowdock.com/app/rulemotion/public-s-premium/threads/8Lo3QYx97HDbSmtFg1eqFXktv0o

20k-ultra:
I extended the PR a bit to include some sets that checks for these logs and that the `isVPNActive` function returns correct boolean.

 Let me know what you think @dt-rush then we'll merge.

dt-rush:
Looking great @20k-ultra !

PR #1408, opened 2020-07-23
Refactor to singletons

richbayliss:
Refactor to singleton pattern;
- ApiBinder
- Application Manager
- DeviceState


test/lib/mocked-dockerode.ts
```diff
@@ -1,4 +1,49 @@
...
+}
+
+// default overrides needed to startup...
+registerOverride('listImages', async () => []);
+registerOverride(
+	'getEvents',
+	async () =>
+		new Stream.Readable({
+			read: () => {
+				return _.noop() as any;
```
Page-, line 34: Is this cast necessary? I think the return type accepts `any`

test/lib/mocked-dockerode.ts
```diff
@@ -1,4 +1,49 @@
+process.env.DOCKER_HOST = 'unix:///your/dockerode/mocks/are/not/working';
+
 import * as dockerode from 'dockerode';
+import { Stream } from 'stream';
+import _ = require('lodash');
+
+const overrides: Dictionary<(...args: any[]) => Resolvable<any>> = {};
+
+type DockerodeFunction = keyof dockerode;
+for (const fn of Object.getOwnPropertyNames(dockerode.prototype)) {
```
Page-, line 10: Do you specifically want non-enumerable properties as well as enumerable?

test/lib/mocked-balena-api.ts
```diff
@@ -52,6 +52,12 @@ api.get(/\/v6\/device\(uuid=%27([0-9a-f]+)%27\)/, (req, res) =>
 	api.balenaBackend!.getDeviceHandler(req, res, _.noop),
 );
 
+api.get(/\/v6\/device/, (req, res) => {
+	const [, uuid] = /uuid eq '([0-9a-f]+)'/i.exec(req.query['$filter']) ?? [];
```
Page-, line 56: Where does the supervisor make a request like this?

test/05-device-state.spec.ts
```diff
@@ -406,46 +379,58 @@ describe('deviceState', () => {
...
+		// 		stripIndent`
+		// 		Healthcheck failure - Atleast ONE of the following conditions must be true:
+		// 			- No applyInProgress      ? false
+		// 			- fetchesInProgress       ? false
+		// 			- cycleTimeWithinInterval ? false`,
+		// 	);
+		// 	// Restore value
+		// 	deviceState.__set__('applyInProgress', previousValue);
+		// 	deviceState.__set__('maxPollTime', previousPollTime);
+		// });
```
Page-, line 434: What's happening with this commented code?

src/proxyvisor.js
```diff
@@ -693,6 +693,8 @@ export class Proxyvisor {
 	}
 
 	imagesInUse(current, target) {
+		// FIXME: Do not merge without first fixing this
+		return [];
```
Page-, line 697: Fix this

src/lib/api-helper.ts
```diff
@@ -0,0 +1,203 @@
...
+			opts.registered_at = Date.now();
+		} else if (opts.provisioningApiKey != null) {
+			log.debug(
+				'Device is registered but we still have an apiKey, attempting key exchange',
+			);
+			device = await exchangeKeyAndGetDevice(balenaApi, opts);
+		}
+
+		if (!device) {
+			// TODO: Type this?
```
Page-, line 185: Did you plan to do this before merging?

src/lib/api-helper.ts
```diff
@@ -0,0 +1,203 @@
...
+		opts.deviceId == null ||
+		opts.provisioningApiKey != null
+	) {
+		if (opts.registered_at != null && opts.deviceId == null) {
+			log.debug(
+				'Device is registered but no device id available, attempting key exchange',
+			);
+
+			device =
+				(await exchangeKeyAndGetDeviceOrRegenerate(balenaApi, opts)) || null;
```
Page-, line 160: Is this `|| null` necessary?

src/device-api/common.js
```diff
@@ -33,25 +45,33 @@ export function doRestart(applications, appId, force) {
...
 		'Purge data',
 	);
-	return _lockingIfNecessary(appId, { force }, () =>
-		deviceState.getCurrentForComparison().then(function (currentState) {
-			if (currentState.local.apps[appId] == null) {
+	return lockingIfNecessary(appId, { force }, () =>
+		deviceState.getCurrentState().then(function (currentState) {
+			const allApps = currentState.local.apps;
+
+			if (allApps == null || allApps[appId] == null) {
```
Page-, line 63:
```suggestion
			if (allApps?.[appId] == null) {
```

src/compose/types/service.ts
```diff
@@ -184,7 +184,7 @@ export interface ConfigMap {
 // is typescript
 export interface DeviceMetadata {
 	imageInfo?: Dockerode.ImageInspectInfo;
-	uuid: string;
+	uuid: Nullable<string>;
```
Page-, line 187: Is this better as
```suggestion
	uuid?: string | null;
```
?

src/compose/service.ts
```diff
@@ -688,6 +743,14 @@ export class Service {
 		service: Service,
 		currentContainerIds: Dictionary<string>,
 	): boolean {
+		// if (this.imageId !== service.imageId) {
+		// 	return false;
+		// }
+
+		// if (this.releaseId !== service.releaseId) {
+		// 	return false;
+		// }
```
Page-, line 752: Commented code

src/compose/service.ts
```diff
@@ -560,22 +583,54 @@ export class Service {
...
+		// 	healthcheck: {
+		// 		...this.config.healthcheck,
+		// 		interval: this.config.healthcheck.interval?.toString(),
+		// 		timeout: this.config.healthcheck.timeout?.toString(),
+		// 		startPeriod: this.config.healthcheck.startPeriod?.toString(),
+		// 	},
+		// 	stopGracePeriod: this.config.startGracePeriod?.toString(),
+		// 	memLimit: this.config.memLimit?.toString(),
+		// 	memReservation: this.config.memReservation?.toString(),
+		// };
```
Page-, line 633: I think most of this comment could be moved into jsdoc for the function where it'd be available for/show up in intellisense

src/compose/network.ts
```diff
@@ -175,16 +175,24 @@ export class Network {
...
-				.getNetwork(Network.generateDockerName(this.appId, this.name))
-				.remove(),
-		).tapCatch((error) => {
-			logger.logSystemEvent(logTypes.removeNetworkError, {
-				network: { name: this.name, appId: this.appId },
-				error,
+		const networkName = Network.generateDockerName(this.appId, this.name);
+
+		return Bluebird.resolve()
+			.then(() => docker.listNetworks())
```
Page-, line 181:
```suggestion
		return Bluebird.resolve(docker.listNetworks())
```

src/compose/images.ts
```diff
@@ -263,15 +282,10 @@ export async function getAvailable(): Promise<Image[]> {
...
-export function getDownloadingImageIds() {
-	return Bluebird.try(() =>
-		_(volatileState)
-			.pickBy({ status: 'Downloading' })
-			.keys()
-			.map(validation.checkInt)
-			.value(),
-	);
+export function getDownloadingImageIds(): number[] {
+	return _.keys(_.pickBy(volatileState)).map((i) =>
```
Page-, line 286: Is this trying to match what was here before? If so it needs to be
```suggestion
	return _.keys(_.pickBy(volatileState, { status: 'Downloading' })).map((i) =>
```

Page-, line 286:
```suggestion
	return Object.keys(_.pickBy(volatileState)).map((i) =>
```

src/compose/application-manager.ts
```diff
@@ -0,0 +1,845 @@
...
+}
+
+async function getAppContainerIds(currentApps: InstancedAppState) {
+	const containerIds: { [appId: number]: Dictionary<string> } = {};
+	await Promise.all([
+		_.map(currentApps, async (_app, appId) => {
+			const intAppId = parseInt(appId, 10);
+			containerIds[intAppId] = await serviceManager.getContainerIdMap(intAppId);
+		}),
+	]);
```
Page-, line 736: This looks broken to me, is it meant to be
```suggestion
	await Promise.all(
		_.map(currentApps, async (_app, appId) => {
			const intAppId = parseInt(appId, 10);
			containerIds[intAppId] = await serviceManager.getContainerIdMap(intAppId);
		}),
	);
```
?

src/compose/app.ts
```diff
@@ -0,0 +1,811 @@
...
+			return Volume.fromComposeObject(name, app.appId, conf);
+		});
+
+		const networks = _.mapValues(
+			JSON.parse(app.networks) ?? {},
+			(conf, name) => {
+				if (conf == null) {
+					conf = {};
+				}
+				return Network.fromComposeObject(name, app.appId, conf);
```
Page-, line 732:
```suggestion
				return Network.fromComposeObject(name, app.appId, conf ?? {});
```

test/lib/mocked-dockerode.ts
richbayliss: How do you mean? I want every function on the prototype...

test/lib/mocked-balena-api.ts
richbayliss: https://github.com/balena-io/balena-supervisor/blob/ca8e9f8f6f3773896f614b8aa50e5653dc20dfa8/src/lib/api-helper.ts#L25

richbayliss: https://github.com/balena-io/balena-supervisor/blob/ca8e9f8f6f3773896f614b8aa50e5653dc20dfa8/src/lib/api-helper.ts#L84-L96

test/05-device-state.spec.ts
richbayliss: It needs dropping out, good spot.

src/proxyvisor.js
richbayliss: 100% needs fixing

src/lib/api-helper.ts
richbayliss: Yep. Will do.

src/lib/api-helper.ts
richbayliss: I don't think so. I followed the code path and unless we throw an exception then you get a `Device`

src/compose/images.ts
richbayliss: @CameronDiver to confirm, but I would say that without this extra then it is returning any image IDs irrespective of `status`.

test/lib/mocked-balena-api.ts
richbayliss: @Page- thinking about this, we shouldn't be fetching the device model even if it is only used to check a key. We should, if this is part of the provisioning flow, do a permissions check on a out-of-model endpoint, like the state endpoint, so that the supervisor isn't tied to an API model version. CC @wrboyce as this might be relevant to your work.

test/lib/mocked-device-api.ts
```diff
@@ -109,7 +109,7 @@ async function createAPIOpts(): Promise<SupervisorAPIOpts> {
 	const deviceState = new DeviceState({
 		apiBinder: null as any,
 	});
-	const apiBinder = new APIBinder();
+	const apiBinder = APIBinder;
 	return {
 		deviceState,
 		apiBinder,
```
CameronDiver:
```suggestion
		apiBinder: APIBinder
```

src/supervisor.ts
```diff
@@ -32,25 +32,24 @@ const startupConfigFields: config.ConfigKey[] = [
...
 		// by storing current state and having the APIBinder query and report it / provision devices
-		this.deviceState.applications.proxyvisor.bindToAPI(this.apiBinder);
+		this.deviceState.applications.proxyvisor.bindToAPI(apiBinder);
 
 		this.api = new SupervisorAPI({
-			routers: [this.apiBinder.router, this.deviceState.router],
+			routers: [apiBinder.router, this.deviceState.router],
 			healthchecks: [
-				this.apiBinder.healthcheck.bind(this.apiBinder),
+				apiBinder.healthcheck.bind(apiBinder),
```
CameronDiver:
```suggestion
				apiBinder.healthcheck,
```

src/lib/api-helper.ts
```diff
@@ -0,0 +1,208 @@
...
+				Authorization: `Bearer ${apiKey}`,
+			},
+		},
+	};
+
+	try {
+		const res = (await Bluebird.resolve(balenaApi.get(reqOpts)).timeout(
+			timeout,
+		)) as Device[];
+		return res[0];
```
CameronDiver, line 56: I think we also need to check if it's an empty array here too.

src/lib/api-helper.ts
```diff
@@ -0,0 +1,208 @@
...
+import * as deviceRegister from './register-device';
+import {
+	DeviceNotFoundError,
+	ExchangeKeyError,
+	FailedToProvisionDeviceError,
+	InternalInconsistencyError,
+	isHttpConflictError,
+} from './errors';
+import log from './supervisor-console';
+import Bluebird = require('bluebird');
```
CameronDiver, line 16:
```suggestion
import * as Bluebird from 'bluebird';
```

src/api-binder.ts
```diff
@@ -48,390 +53,470 @@ interface DeviceTag {
...
+	if (deviceId == null) {
+		throw new Error('Attempt to retrieve device tags before provision');
+	}
+	const tags = (await balenaApi.get({
+		resource: 'device_tag',
+		options: {
+			$select: ['id', 'tag_key', 'value'],
+			$filter: { device: deviceId },
+		},
+	})) as Array<Dictionary<unknown>>;
```
CameronDiver, line 307: I think the pine typings should cover this now?

src/api-binder.ts
```diff
@@ -48,390 +53,470 @@ interface DeviceTag {
...
+	const conf = await config.getMany([
+		'deviceId',
+		'apiTimeout',
+		'apiEndpoint',
+		'uuid',
+		'localMode',
+	]);
+
+	const stateDiff = getStateDiff();
+	if (_.size(stateDiff) === 0) {
```
CameronDiver, line 415: It's worth doing this before the config fetch, as with the return, we can avoid a whole config.getMany

src/api-binder.ts
```diff
@@ -48,390 +53,470 @@ interface DeviceTag {
...
-	// without overwriting any variables that are already set.
-	private async reportInitialEnv(
-		apiEndpoint: string,
-		deviceId: number,
-		initialName?: string,
-	) {
-		if (this.balenaApi == null) {
-			throw new InternalInconsistencyError(
-				'Attempt to report initial environment without an API client',
+		const releaseId = _.get(release, '[0].id');
```
CameronDiver, line 511:
```suggestion
		const releaseId = release?.[0]?.id
```

src/api-binder.ts
```diff
@@ -443,309 +528,259 @@ export class APIBinder {
...
-				varValue = 'false';
-			}
-			// We never want to disable VPN if, for instance, it failed to start so far
-			if (key === 'SUPERVISOR_VPN_CONTROL') {
-				varValue = 'true';
-			}
+	const targetConfigUnformatted = _.get(
+		await TargetState.get(),
+		'local.config',
+	);
```
CameronDiver, line 559:
```suggestion
	const targetConfigUnformatted = (await TargetState.get())?.local?.config
```

src/api-binder.ts
```diff
@@ -443,309 +528,259 @@ export class APIBinder {
...
-			'apiKey',
-			'pinDevice',
-		]);
-
-		if (!conf.provisioned || conf.apiKey != null || conf.pinDevice != null) {
-			await this.provisionOrRetry(conf.bootstrapRetryDelay);
-			globalEventBus.getInstance().emit('deviceProvisioned');
-			return;
-		}
+	await balenaApi.patch({
```
CameronDiver, line 697: Let's wrap this in a  try/catch because we don't want to break if for some reason we couldn't set the device name

src/api-binder.ts
```diff
@@ -443,309 +528,259 @@ export class APIBinder {
...
+					res.sendStatus(202);
+				}
+			})
+			.catch(next);
+	} else {
+		res.sendStatus(202);
+	}
+});
+
+// export interface ApiBinder {
```
CameronDiver: Commented code

src/device-state.ts
```diff
@@ -216,677 +217,672 @@ type DeviceStateStep<T extends PossibleStepTargets> =
...
-				log.success('Device state apply success');
-				// We also let the device-config module know that we
-				// successfully reached the target state and that it
-				// should clear any rate limiting it's applied
-				return deviceConfig.resetRateLimits();
-			}
-		});
-		this.applications.on('change', (d) => this.reportCurrentState(d));
-		this.router = createDeviceStateRouter(this);
+// export class DeviceState extends (EventEmitter as new () => DeviceStateEventEmitter) {
```
CameronDiver: Remove

src/device-state.ts
```diff
@@ -216,677 +217,672 @@ type DeviceStateStep<T extends PossibleStepTargets> =
...
+	globalEventBus.getInstance().emit('targetStateChanged', target);
+
+	const apiEndpoint = await config.get('apiEndpoint');
+
+	await usingWriteLockTarget(async () => {
+		await db.transaction(async (trx) => {
+			await config.set({ name: target.local.name }, trx);
+			await deviceConfig.setTarget(target.local.config, trx);
+
+			if (localSource || apiEndpoint == null) {
```
CameronDiver, line 471: We could de-dupe this code and use a ternary

src/lib/migration.ts
```diff
@@ -248,7 +258,7 @@ export async function normaliseLegacyDatabase(
 	await serviceManager.killAllLegacy();
 	log.debug('Migrating legacy app volumes');
 
-	const targetApps = await application.getTargetApps();
+	const targetApps = await deviceState.applications.getTargetApps();
```
CameronDiver: In the next commit we should make this the actual applications singleton import

src/compose/application-manager.ts
```diff
@@ -0,0 +1,847 @@
...
+
+	imageManager.on('change', reportCurrentState);
+	serviceManager.on('change', reportCurrentState);
+})();
+
+export function getDependentState() {
+	return proxyvisor.getCurrentStates();
+}
+
+// FIXME: Data type here
```
CameronDiver, line 149: This FIXME

src/compose/application-manager.ts
```diff
@@ -0,0 +1,847 @@
...
+	if (localMode) {
+		ignoreImages = localMode;
+	}
+
+	const currentAppIds = Object.keys(currentApps).map((i) => parseInt(i, 10));
+	const targetAppIds = Object.keys(targetApps).map((i) => parseInt(i, 10));
+
+	let steps: CompositionStep[] = [];
+
+	// First check if we need to create the superviosr network
```
CameronDiver, line 199:
```suggestion
	// First check if we need to create the supervisor network
```

src/device-api/v1.js
```diff
@@ -25,37 +28,41 @@ export const createV1Api = function (router, applications) {
 		if (appId == null) {
 			return res.status(400).send('Missing app id');
 		}
-		return applications
-			.getCurrentApp(appId)
-			.then(function (app) {
-				let service = app?.services?.[0];
-				if (service == null) {
+
+		// FIX-ME: This cannot work, since we will not know which appId is the one to use
```
CameronDiver, line 32: I don't think this comment is true, because we get the app id from the request (also I tested it  :P)

src/device-state.ts
```diff
@@ -878,11 +898,21 @@ export function triggerApplyTarget({
...
 ) {
-	intermediateTarget = _.cloneDeep(intermediate);
+	// TODO: Make sure we don't accidentally overwrite this
+	intermediateTarget = intermediate;
 	return applyTarget({ intermediate: true, force, skipLock }).then(() => {
 		intermediateTarget = null;
 	});
 }
+
+// This is useful as some of the api endpoints will take a current state,
```
CameronDiver, line 911: Remove this

src/compose/app.ts
```diff
@@ -0,0 +1,808 @@
...
+			);
+		}
+
+		if (needsDownload && context.downloading.includes(target?.imageId!)) {
+			// The image needs to be downloaded, and it's currently downloading. We simply keep
+			// the application loop alive
+			return generateStep('noop', {});
+		}
+
+		if (target && current?.isEqualConfig(target, context.containerIds)) {
```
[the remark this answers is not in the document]
richbayliss: this is lowest form of check; `isEqual` includes running state && `isEqualConfig` so this is not including the running state as a check.

CameronDiver (dismissed):

CameronDiver (approved):

PR #1415, opened 2020-07-29
Fixing v1 start/stop endpoint issue with service access

nitishagar:


test/21-supervisor-api.spec.ts
```diff
@@ -95,7 +95,6 @@ describe('SupervisorAPI', () => {
 				await request
 					.get('/v1/healthy')
 					.set('Accept', 'application/json')
-					.set('Authorization', `Bearer ${VALID_SECRET}`)
```
nitishagar, line 98: The auth step is only applied after `ping` and `healthy` endpoint. We would also need the doc to correctly update the `ping` endpoint as it currently shows `apiKey` as required param.

Ref: https://github.com/balena-io/balena-supervisor/blob/master/docs/API.md#http-api-reference

20k-ultra: good catch. The documentation for the API is located in https://github.com/balena-io/balena-supervisor/blob/master/docs/API.md. Once that file is modified we can notify the production team to release it to the site. Feel free to include that change in this PR.

nitishagar:
@20k-ultra Thanks for your input. I clean-up the 2 commits and forced push with `Change-Type: minor` as this to me looks bigger than a minor patch - given the `start/stop` for v1 were not functioning.

Also, `start/stop` stem from a single function where this bug was. `restart` has its own code path and not affected by this change. I have written a test case specifically to cover the issue here. We can add cases to cover `start` and or `restart` for code path coverage later. This to me looks an important change to get out the door. Thoughts?

20k-ultra:
Looks good @nitishagar. Given the following description from semantic versionings website:
```
    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards compatible manner, and
    PATCH version when you make backwards compatible bug fixes.
```

I think the Change-type should actually be `patch` rather then `minor` since we only fixed a bug.

nitishagar:
Done. Moved the `Change-Type` to `Patch`.

PR #1420, opened 2020-07-30
Update pinejs-client-request and make use of a named key

Page-:
Update pinejs-client-request from 7.0.0 to 7.1.0


Page-:
@balena-ci rebase

PR #1422, opened 2020-07-31
bug: Fix undefined containerId object

richbayliss:
Change-type: patch

### DIFF
--- a/src/device-api/common.js
+++ b/src/device-api/common.js
@@ -135,7 +135,7 @@ export function safeAppClone(app) {
 	const containerIdForService = _.fromPairs(
 		_.map(app.services, (svc) => [
 			svc.serviceName,
-			svc.containerId.substr(0, 12),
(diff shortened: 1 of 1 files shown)


src/device-api/common.js
```diff
@@ -135,7 +135,7 @@ export function safeAppClone(app) {
 	const containerIdForService = _.fromPairs(
 		_.map(app.services, (svc) => [
 			svc.serviceName,
-			svc.containerId.substr(0, 12),
+			svc.containerId ? svc.containerId.substr(0, 12) : '',
```
CameronDiver, line 138:
```suggestion
			svc.containerId != null ? svc.containerId.substr(0, 12) : '',
```
Just to make the intent clear

Page-: What about
```suggestion
			svc.containerId?.substr(0, 12) ?? '',
```

PR #1423, opened 2020-08-04
Refactor extra_uEnv backend to match with more devices

20k-ultra:
There are are more devices that support extra_uEnv configuration backend that were not known until @acostach told me. This PR includes those devices so we can set their FDT and kernal arguments.


20k-ultra:
@balena-ci I self-certify!

PR #1425, opened 2020-08-05
Update resumable-request

Page-:
Update resumable-request from 2.0.0 to 2.0.1

Change-type: patch


package.json
```diff
@@ -105,9 +105,8 @@
     "pinejs-client-request": "^7.1.0",
     "pretty-ms": "^5.1.0",
     "request": "^2.88.2",
-    "resin-cli-visuals": "^1.5.2",
```
Page-, line 108: I removed this as it's completely unused

package.json
```diff
@@ -105,9 +105,8 @@
...
     "resin-docker-build": "^1.1.4",
-    "resumable-request": "^2.0.0",
+    "resumable-request": "^2.0.1",
```
Page-, line 109: Aside from this everything else is only used at compile/bundle time

PR #1431, opened 2020-08-07
Allow multiple config backends

20k-ultra:
As a part of implementing https://github.com/balena-io/balena-supervisor/issues/1206 it was beneficial/required to apply more then just 1 configuration backend. For example, with this change devices can now match with Extlinux and Config.txt if there was a device that needed this. I plan to leverage this functionality to allow Jetson device to configure Extra_uEnv for setting FDT and a new backend for setting ODMDATA. 

This PR also contains some styling changes to the naming of classes.

### DIFF
--- a/src/config/utils.ts
+++ b/src/config/utils.ts
@@ -1,41 +1,23 @@
 import * as _ from 'lodash';
+import * as Bluebird from 'bluebird';
 
+import * as config from '../config';
 import * as constants from '../lib/constants';
 import { getMetaOSRelease } from '../lib/os-release';
 import { EnvVarObject } from '../lib/types';
-import { Extlinux } from './backends/extlinux';
-import { ExtraUEnv } from './backends/extra-uEnv';
-import { ConfigTxt } from './backends/config-txt';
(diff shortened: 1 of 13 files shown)


src/config/utils.ts
```diff
@@ -1,51 +1,42 @@
...
 import { EnvVarObject } from '../lib/types';
-import { ExtlinuxConfigBackend } from './backends/extlinux';
-import { ExtraUEnvConfigBackend } from './backends/extra-uEnv';
-import { RPiConfigBackend } from './backends/raspberry-pi';
-import { ConfigfsConfigBackend } from './backends/config-fs';
-import { ConfigOptions, DeviceConfigBackend } from './backends/backend';
+import { Extlinux } from './backends/extlinux';
+import { ExtraUEnv } from './backends/extra-uEnv';
+import { ConfigTxt } from './backends/config-txt';
+import { ConfigFs } from './backends/config-fs';
```
CameronDiver, line 11: Probably it's worth adding an `index.ts` at this point which exports all of these.

PR #1436, opened 2020-08-18
added support for configuring ODMDATA

20k-ultra:
ODMDATA is a configuration used to switch the hardware configuration for Jetson devices. This PR allows users to select between Configuration Modes 1 to 6 which will be available in the dashboard with https://github.com/balena-io/open-balena-api/pull/409.

The config name it looks for is `RESIN_HOST_ODMDATA_configuration`.

HQ issue: https://github.com/balena-io/balena-io/issues/2166

### DIFF
--- a/src/config/backends/odmdata.ts
+++ b/src/config/backends/odmdata.ts
@@ -1,0 +1,247 @@
+import * as _ from 'lodash';
+import { promises as fs } from 'fs';
+
+import { ConfigOptions, ConfigBackend } from './backend';
+import * as constants from '../../lib/constants';
+import log from '../../lib/supervisor-console';
+import { ODMDataError } from '../../lib/errors';
+
+/**
+ * A backend to handle ODMDATA configuration
+ *
+ * Supports:
+ * 	- {BALENA|RESIN}_HOST_CONFIG_odmdata_configuration = value | "value"
+ */
+
+export class Odmdata extends ConfigBackend {
+	private static bootConfigVarPrefix = `${constants.hostConfigVarPrefix}ODMDATA_`;
+	private static bootConfigPath = `${constants.rootMountPoint}/dev/mmcblk0boot0`;
+	private static bootConfigLockPath = `${constants.rootMountPoint}/sys/block/mmcblk0boot0/force_ro`;
+	private static supportedConfigs = ['configuration'];
+	private BYTE_OFFSETS = [1659, 5243, 18043];
+	private CONFIG_BYTES = [
+		0x0 /* Config Option #1 */,
+		0x1 /* Config Option #2 */,
+		0x6 /* Config Option #3 */,
+		0x7 /* Config Option #4 */,
+		0x2 /* Config Option #5 */,
+		0x3 /* Config Option #6 */,
+	];
+	private CONFIG_BUFFER = Buffer.from(this.CONFIG_BYTES);
+
+	public static bootConfigVarRegex = new RegExp(
+		'(?:' + _.escapeRegExp(Odmdata.bootConfigVarPrefix) + ')(.+)',
+	);
+
+	public async matches(deviceType: string): Promise<boolean> {
+		return deviceType.endsWith('-tx2');
+	}
+
+	public async getBootConfig(): Promise<ConfigOptions> {
+		// Get config buffer from bootConfigPath
+		const confBuffer = await this.readBootConfigPath();
+		// Parse ConfigOptions from bootConfigPath buffer
+		return this.parseOptions(confBuffer);
+	}
+
+	public async setBootConfig(opts: ConfigOptions): Promise<void> {
+		log.info('Attempting to configure ODMDATA.');
+		// Filter out unsupported options
+		const supportedOptions = _.pickBy(opts, (value, key) => {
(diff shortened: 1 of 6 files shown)


20k-ultra:
I'm currently aware of an issue with the tests not passing but it's something that I don't want to block a review. The failing tests is due to stubbing not any logic so once this stub is fixed the logic will still be the same.

src/config/backends/odmdata.ts
```diff
@@ -0,0 +1,242 @@
...
+				return false;
+			}
+			return true;
+		});
+		// Check that there is a configuration mode
+		if (!supportedOptions.configuration) {
+			log.info('No changes made to ODMDATA configuration.');
+			return;
+		}
+		// Check that confiugration mode is supported
```
CameronDiver, line 62: Typo

src/config/backends/odmdata.ts
```diff
@@ -0,0 +1,242 @@
...
+		} catch (e) {
+			log.error(`Issue reading '${Odmdata.bootConfigPath}'`);
+			throw e;
+		} finally {
+			await fileHandle.close();
+		}
+		return valuesBuffer;
+	}
+
+	private async setReadOnly(value: boolean): Promise<void> {
```
CameronDiver, line 227: Please name this parameter more accurately.

src/config/backends/odmdata.ts
```diff
@@ -0,0 +1,242 @@
...
+		const mode = parseInt(supportedOptions.configuration as string, 10);
+		if (mode < 1 || mode > 6) {
+			log.error(`Configuration mode of: ${mode} is not supported.`);
+			log.info('No changes made to ODMDATA configuration.');
+			return;
+		}
+		// Get the byte for this configuration mode
+		const byte = this.CONFIG_BYTES[mode - 1];
+		log.info(`Setting ODMDATA Configuration Mode #${mode}.`);
+		// Begin writing to bootConfigPath by unlocking RO access
```
CameronDiver, line 72: These comments aren't really very helpful. What I would like to see is a description of *why* we're doing what we're doing, not just a description of the code. The naming of functions et al should be enough to discern the what, but for example, I have no idea as a casual observer why we might be unlocking RO access. Unlocking access to what? Why do we need to unlock? etc

src/config/backends/odmdata.ts
```diff
@@ -0,0 +1,242 @@
...
+		await this.setReadOnly(false);
+		try {
+			// Write this byte to each BYTE_OFFSETS in bootConfigPath
+			for (const offset of this.BYTE_OFFSETS) {
+				await this.setByteAtOffset(byte, offset);
+			}
+		} catch (e) {
+			log.error('Failed to set configuration mode.');
```
CameronDiver, line 80: I would log the error too, although it likely will be further up the stack. I guess have a play and make sure that if somebody came to support with some supervisor logs and no working ODMData that we'd be able to debug it easily.

src/config/backends/odmdata.ts
20k-ultra: I think this is acceptable because the name of the function is simple and clear on what the single parameter is for. If there were more parameters passed I would change `value` to something more clear. Thoughts?

src/config/backends/odmdata.ts
20k-ultra: I went threw the file and added the error to 4 or 5 more places where I log a vague message like "Couldn't do X or Y for unknown reason".

20k-ultra:
I resolved the failing test and tested this on a real Jetson device and everything looks good! Only thing I have to change is the device will get stuck in a boot loop if the user doesn't set a configuration mode in the API.

20k-ultra:
I have resolved the issue with devices getting stuck in a loop if a ODMDATA configuration value is not present in target state and confirmed this on a real device.

src/device-config.ts
```diff
@@ -352,23 +353,39 @@ export function bootConfigChangeRequired(
...
-				if (currentBootConfig[key] !== value) {
-					const err = `Attempt to change blacklisted config value ${key}`;
-					logger.logSystemMessage(
-						err,
-						{ error: err },
-						'Apply boot config error',
-					);
-					throw new Error(err);
-				}
+		// Check if the only difference is the targetBootConfig not containing a special case
```
20k-ultra, line 369: I'd like to put a spotlight on this specific change. This code is a necessary evil for devices that upgrade their Supervisor version and have not had their ODMDATA configuration mode sent to the cloud which happens on provision. The logic to detect this case can be quiet the rabbit hole and I concluded the most robust way and understand way to make sure we catch this case is to do what I did rather then if (condition_1 && condition_2) because it gets messy quickly.

src/device-config.ts
```diff
@@ -352,23 +353,39 @@ export function bootConfigChangeRequired(
...
+		const SPECIAL_CASE = 'configuration'; // ODMDATA Mode for TX2 devices
```
CameronDiver, line 370: This is a little concerning, after we strip the namespace we're left with simply `configuration`. I could see this causing name clashes in the future, and definitely something that we should be aware of.

CameronDiver (approved):

Issue #1437, opened 2020-08-18
Empty config returned if no Backends loaded


codewithcheese:
The following block returns an empty object when no backend's are loaded. This has the knock on effect of setting the targetState to an empty object, effectively resetting the state to defaults.

https://github.com/balena-io/balena-supervisor/blob/61811dd27a083ecff7ec83e88d019c205a8295a0/src/device-config.ts#L301-L313

`formatConfigKeys` is used by the `setTarget`, when `formatConfigKeys` returns `{}` then `deviceConfig.targetValues` is to `{}`

```
export async function setTarget(
    target: Dictionary<string>,
    trx?: db.Transaction,
): Promise<void> {
    const $db = trx ?? db.models.bind(db);

    const formatted = await formatConfigKeys(target);
    // check for legacy keys
    if (formatted['OVERRIDE_LOCK'] != null) {
        formatted['SUPERVISOR_OVERRIDE_LOCK'] = formatted['OVERRIDE_LOCK'];
    }

    const confToUpdate = {
        targetValues: JSON.stringify(formatted),
    };

    await $db('deviceConfig').update(confToUpdate);
}
```

If `targetValues` is set to `{}` then the target state is lost and defaults will be applied by `getTarget`
```
export async function getTarget({
	initial = false,
}: { initial?: boolean } = {}) {
	const [unmanaged, [devConfig]] = await Promise.all([
		config.get('unmanaged'),
		db.models('deviceConfig').select('targetValues'),
	]);

	let conf: Dictionary<string>;
	try {
		conf = JSON.parse(devConfig.targetValues);
	} catch (e) {
		throw new Error(`Corrupted supervisor database! Error: ${e.message}`);
	}
	if (initial || conf.SUPERVISOR_VPN_CONTROL == null) {
		conf.SUPERVISOR_VPN_CONTROL = 'true';
	}
	if (unmanaged && conf.SUPERVISOR_LOCAL_MODE == null) {
		conf.SUPERVISOR_LOCAL_MODE = 'true';
	}

	_.defaults(
		conf,
		_(configKeys).mapKeys('envVarName').mapValues('defaultValue').value(),
	);

	return conf;
}
```

PR #1440, opened 2020-08-18
Preventing removing all configurations if device has no backends

20k-ultra:
**Problem:** 

When https://github.com/balena-io/balena-supervisor/pull/1431 was merged it introduced a regression that would make devices ignore all their configurations set in the dashboard such as firewall, localmode, etc. The cause of this issue was because in the following code we iterate over the list of matched backends and run [configUtils.formatConfigKeys](https://github.com/balena-io/balena-supervisor/blob/61811dd27a083ecff7ec83e88d019c205a8295a0/src/config/utils.ts#L54) with the backend as the first parameter.

https://github.com/balena-io/balena-supervisor/blob/61811dd27a083ecff7ec83e88d019c205a8295a0/src/device-config.ts#L301-L313

When the PR that introduced this regression was written I had gotten the impression that this function should only run if you have a backend otherwise return an empty config object. However, that function formats all configuration options even if a null backend is passed. So when you don't have a backend that function still needs to run to return the configs for firewall, localmode, etc.

**Solution:**

This PR resolves this issue by always running formatConfigKeys and modifying the original logic of "does this configVar match this backend" and instead try to match that configVar on any backend. The previous logic worked because the device only used 1 backend to configure at a time but now we allow multiple backends to configure and so matching the configVar must be applied to all backends and if any returns true then it is valid. We also gain the benefit of no longer having any async IO operations when running this code path. In addition this PR also optimizes that fact that each backend matched would format the same vars multiple times and so this only formats the variables once.

----

Closes: #1437
Change-type: patch

### DIFF
--- a/src/config/backends/index.ts
+++ b/src/config/backends/index.ts
@@ -1,11 +1,17 @@
+import * as _ from 'lodash';
+
 import { Extlinux } from './extlinux';
 import { ExtraUEnv } from './extra-uEnv';
 import { ConfigTxt } from './config-txt';
 import { ConfigFs } from './config-fs';
 
-export default [
+export const allBackends = [
 	new Extlinux(),
 	new ExtraUEnv(),
 	new ConfigTxt(),
 	new ConfigFs(),
 ];
+
+export function matchesAnyBootConfig(envVar: string): boolean {
+	return allBackends.some((a) => a.isBootConfigVar(envVar));
+}
--- a/src/device-config.ts
+++ b/src/device-config.ts
@@ -2,17 +2,17 @@ import * as _ from 'lodash';
 import { inspect } from 'util';
 
 import * as config from './config';
-import { SchemaTypeKey } from './config/schema-type';
 import * as db from './db';
 import * as logger from './logger';
-
-import { ConfigOptions, ConfigBackend } from './config/backends/backend';
-import * as configUtils from './config/utils';
 import * as dbus from './lib/dbus';
-import { UnitNotLoadedError } from './lib/errors';
 import { EnvVarObject } from './lib/types';
+import { UnitNotLoadedError } from './lib/errors';
 import { checkInt, checkTruthy } from './lib/validation';
 import { DeviceStatus } from './types/state';
+import * as configUtils from './config/utils';
+import { SchemaTypeKey } from './config/schema-type';
+import { matchesAnyBootConfig } from './config/backends';
+import { ConfigOptions, ConfigBackend } from './config/backends/backend';
 
 const vpnServiceName = 'openvpn';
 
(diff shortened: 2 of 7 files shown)


src/config/backends/index.ts
```diff
@@ -3,9 +3,20 @@ import { ExtraUEnv } from './extra-uEnv';
...
 
-export default [
+export const all = [
 	new Extlinux(),
 	new ExtraUEnv(),
 	new ConfigTxt(),
 	new ConfigFs(),
 ];
+
+export function matchesAnyBootConfig(envVar: string): boolean {
```
20k-ultra, line 13: Is there a better way to write this ? I don't like how I have a variable outside the loop and the loop updates the value to true if one of the conditions passes. I want a way to pass a single value and a list of compare function. It's like the inverse of filter where it takes a list of values and compare function

src/config/backends/index.ts
```diff
@@ -3,9 +3,20 @@ import { ExtraUEnv } from './extra-uEnv';
 import { ConfigTxt } from './config-txt';
 import { ConfigFs } from './config-fs';
 
-export default [
+export const all = [
```
CameronDiver, line 6:
```suggestion
export const allBackends = [
```

src/config/backends/index.ts
CameronDiver: `_.some` would work here

src/device-config.ts
```diff
@@ -298,18 +298,8 @@ export async function getCurrent(): Promise<Dictionary<string>> {
...
-	const formattedKeys: Dictionary<any> = {};
-	for (const backend of backends) {
-		_.assign(
-			formattedKeys,
-			configUtils.formatConfigKeys(backend, validKeys, conf),
-		);
-	}
-	return formattedKeys;
+export function formatConfigKeys(conf: Dictionary<string>): Dictionary<any> {
+	return configUtils.formatConfigKeys(validKeys, conf);
```
CameronDiver, line 302: Just call this function directly from the caller of `formatConfigKeys`, no need to have this wrapper function.

src/config/backends/index.ts
20k-ultra: It looks like `_.some` is used as "here is a list of values, check if one of them returns true with my function".

src/config/backends/index.ts
20k-ultra: I was trying to avoid being too verbose. Ie import `backend` module and using `allBackends` variable has backend twice but I am not entirely convinced being generic is *better*...I'll revert to being more specific.

src/device-config.ts
20k-ultra: I thought about this as well but `configUtils.formatConfigKeys` takes 2 parameters instead of one like device-config's version. That first parameter (validKeys) is a value set in device-config and once I began changing the code to directly call configUtils function it required moving validKeys to configUtils and it just began to feel wrong. Having this function in device-config adds extra information that config utilities needs.

CameronDiver (approved):

src/device-config.ts
20k-ultra: great suggestion @codewithcheese! I just added this change and it really cleaned stuff up.

Issue #1404, opened 2020-07-22
Disabling vc4-fkms-v3d on raspberry pi 4


Razikus:
The issue is that i have to disable vc4-fkms-v3d, but i can't, because supervisor is adding this overlay always in case of raspberry pi 4
https://github.com/balena-io/balena-supervisor/blob/de80aacc2f15e9cf290af9798c22cf00634889bb/src/device-config.ts#L593-L594

Related #1171 

I need to disable it, because without it raspberry pi 4 64bit correctly show output on monitor.
Problems with vc4-fkms-v3d are common

It's not always needed.

Forum topic
https://forums.balena.io/t/disable-vc4-kms-v3d-for-raspberry-pi-4/159547

CameronDiver:
Thanks for the report @Razikus, this was actually a feature request from @chrisys so I'd like to find out the reasoning behind both use cases. The context back then was that this was always necessary for raspberry pi 4, but I'd like to work out the correct behavior here.

@20k-ultra Can you work with Chris to find the best method to handle this please?

Razikus:
Hi, any info on this?
We can't proceed without workaround, our monitors are not working without this.

So we can't change rpi3 to rpi4

CameronDiver:
@Razikus I've created https://github.com/balena-io/balena-supervisor/pull/1441 which when merged will make its way into the next balenaOS release. If you need a supervisor update in the meantime (we currently have an OS in testing so this would have to make the subsequent one, which would be anywhere up to a month before release) I can provide the steps to show you how to upgrade the supervisor.

PR #1441, opened 2020-08-19
Don't enforce the vc4-fkms-v3d dtoverlay on rpi4

CameronDiver:
Change-type: patch
Closes: #1404

### DIFF
--- a/src/device-config.ts
+++ b/src/device-config.ts
@@ -600,13 +600,8 @@ function checkBoolChanged(
 // Modifies conf
 // exported for tests
 export function ensureRequiredOverlay(deviceType: string, conf: ConfigOptions) {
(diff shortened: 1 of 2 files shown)


CameronDiver:
@balena-ci rebase

src/device-config.ts
```diff
@@ -600,13 +600,8 @@ function checkBoolChanged(
 // Modifies conf
 // exported for tests
 export function ensureRequiredOverlay(deviceType: string, conf: ConfigOptions) {
```
20k-ultra, line 602: We should remove the export of this function since it is only there so that tests don't freak out from accessing a private value. Then in the tests add the right error suppression.

richbayliss (approved):

PR #1446, opened 2020-08-26
Revert "Don't enforce the vc4-fkms-v3d dtoverlay on rpi4"

CameronDiver:
This was fixed in balenaOS 2.44.0+rev9

This reverts commit de4c1b25385630b6f9fadb7a362d5556c2cc5008.
Change-type: patch


20k-ultra:
We made this PR because of some confusion that made it seem like we should revert but keeping this change is actually desired.

PR #1448, opened 2020-08-26
bug: Resolve mDNS API URLs

richbayliss:
Due to switching to Alpine the ability to resolve mDNS
hostnames was lost. This patch overrides the lookup and
manually resolves the names.

Change-type: patch

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,3 +1,24 @@
+import { NOTFOUND } from 'dns';
+import * as mdnsResolver from 'mdns-resolver';
+
+class DnsLookupError extends Error {
+	public constructor(public code: string = NOTFOUND) {
+		super();
+	}
+}
+
+interface DnsLookupCallback {
+	(err: any): void;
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -7,10 +28,46 @@
...
+		if (name.endsWith('.local')) {
+			return Promise.all([
+				mdnsResolver.resolve4(name).catch(() => {
+					return '';
+				}),
+				mdnsResolver.resolve6(name).catch(() => {
+					return '';
+				}),
+			]).then(([v4, v6]) => {
+				const allAddresses = [v4, v6].filter((addr) => addr !== '');
```
CameronDiver, line 50: Shouldn't we avoid doing a lookup for both addresses if we aren't going to need it?

richbayliss: Agreed. I have remedied this.

CameronDiver (approved):

PR #1451, opened 2020-09-03
Add device system information to state endpoint patch

CameronDiver:
Could do with a little bit of code cleanup and tests, but PRing now to have something to work with as part of 03/08/20 release party.

Change-type: minor

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -11,6 +11,8 @@ import DeviceState from '../device-state';
 import { CoreOptions } from 'request';
 import * as url from 'url';
 
+import * as sysInfo from '../lib/system-info';
+
 // The exponential backoff starts at 15s
 const MINIMUM_BACKOFF_DELAY = 15000;
 
@@ -121,6 +123,39 @@ const getStateDiff = (): DeviceStatus => {
 			.value(),
 	};
 
+	const shouldReportSysInfo = (type: string, past: number, now: number) => {
+		if (!past) {
+			return true;
+		}
+		// TODO: Deduplicate this code
+		switch (type) {
+			case 'cpu_usage':
+				// The bucket size of cpu usage is 20
+				return Math.floor(past / 20) !== Math.floor(now / 20);
+			case 'cpu_temp':
+				return Math.floor(past / 5) !== Math.floor(now / 5);
+			case 'memory_usage':
+				return Math.floor(past / 10) !== Math.floor(now / 10);
+		}
+		return true;
+	};
+
+	const toOmit: string[] = [];
+	_.each(diff.local, (value, key) => {
+		// if we have some system information that has changed, we check that it's
+		// within a certain range before reporting
+		if (
+			!shouldReportSysInfo(
+				key,
+				(lastReportedLocal as any)[key],
+				value as number,
+			)
+		) {
+			toOmit.push(key);
+		}
+	});
+
+	diff.local = _.omit(diff.local, toOmit);
 	return _.omitBy(diff, _.isEmpty);
 };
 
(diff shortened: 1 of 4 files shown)


src/lib/system-info.ts
```diff
@@ -0,0 +1,114 @@
...
+		storage_total: storage.storageTotal,
+		storage_block_device: storage.blockDevice,
+		cpu_temp: temp,
+		cpu_id: cpuid,
+		is_undervolted: undervoltage,
+	};
+}
+
+function bytesToMb(bytes: number) {
+	// TODO: This is technically mibibyte, maybe change this
```
Page-, line 112:
```suggestion
	// TODO: This is technically mebibyte, maybe change this
```

Page-, line 112: We should solve this before merging, because once it's deployed we can't easily change it

src/lib/system-info.ts
```diff
@@ -0,0 +1,114 @@
...
+export async function getMemoryInformation() {
+	const mem = await systeminformation.mem();
+	return {
+		used: bytesToMb(mem.used),
+		total: bytesToMb(mem.total),
+	};
+}
+
+export async function getCpuTemp() {
+	return Math.round((await systeminformation.cpuTemperature()).main);
```
Page-, line 63: The api will accept decimals, if we aren't going to report decimals then we should probably update the api, cc @richbayliss

src/lib/system-info.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export function getCpuUsage() {
+	return new Promise((resolve) => {
+		osUtils.cpuUsage((percent) => {
+			resolve(Math.round(percent * 100));
+		});
+	});
+}
+
+const blockDeviceRegex = /(\/dev\/.*)p\d+/;
```
Page-, line 13: What's the `p`?

src/device-state/current-state.ts
```diff
@@ -121,6 +123,39 @@ const getStateDiff = (): DeviceStatus => {
...
+	};
+
+	const toOmit: string[] = [];
+	_.each(diff.local, (value, key) => {
+		// if we have some system information that has changed, we check that it's
+		// within a certain range before reporting
+		if (
+			!shouldReportSysInfo(
+				key,
+				(lastReportedLocal as any)[key],
```
Page-, line 150: Is this `any` required?

src/lib/system-info.ts
```diff
@@ -0,0 +1,114 @@
+import * as systeminformation from 'systeminformation';
+import * as osUtils from 'os-utils';
+import { fs, child_process } from 'mz';
+
+export function getCpuUsage() {
+	return new Promise((resolve) => {
+		osUtils.cpuUsage((percent) => {
```
Page-, line 7: I think you might be able to replace this with `systeminformation.currentLoad` and have just a single module

src/supervisor.ts
```diff
@@ -15,6 +15,8 @@ import version = require('./lib/supervisor-version');
 import * as avahi from './lib/avahi';
 import * as firewall from './lib/firewall';
 
+import * as sysInfo from './lib/system-info';
```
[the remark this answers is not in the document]
CameronDiver: Outdated

src/lib/system-info.ts
CameronDiver: I think we're ok with keeping mebibyte, we chatted on-stream.

src/lib/system-info.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export async function getSysInfoToReport() {
+	const [cpu, mem, temp, cpuid, storage, undervoltage] = await Promise.all([
+		getCpuUsage(),
+		getMemoryInformation(),
+		getCpuTemp(),
+		getCpuId(),
+		getStorageInfo(),
+		undervoltageDetected(),
+	]);
```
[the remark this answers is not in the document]
CameronDiver: They don't, nicely.

src/lib/system-info.ts
CameronDiver: Matches against the partition

src/lib/system-info.ts
CameronDiver: load and cpu usage is different @Page-

richbayliss (approved):
Some optimisations could possibly be had, but this is OK I think.

src/lib/system-info.ts
Page-: @CameronDiver that's sd/emmc specific, it won't work for nucs or other devices that don't use that format

src/lib/system-info.ts
Page-: @CameronDiver I know "load average" is different but the docs for that call definitely imply it's cpu load/usage:
image

src/lib/system-info.ts
CameronDiver: It will work for an nvme nuc, will do further testing.

Page-: Could you add a comment about what we expect to match - it could help a lot later on when looking back at this code

PR #1461, opened 2020-09-16
Implement scoped Supervisor API keys

richbayliss:
Each service, when requesting access to the Supervisor API, will now get an individual key which can be scoped to specific resources. In this iteration the scopes will be hardcoded to the application that the service belongs to.

Change-type: patch


src/lib/api-keys.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+export function isScoped(
+	resources: Partial<ScopedResources>,
+	scopes: Scope[],
+): boolean {
+	return (
+		scopes.filter((scope) =>
+			scopeChecks[scope.type](resources, (scope as unknown) as any),
+		).length >= 1
```
richbayliss, line 63: `scopes.some(...)` might be nicer

src/lib/api-keys.ts
```diff
@@ -0,0 +1,290 @@
...
+
+				// which resources are scoped...
+				req.auth.isScoped = (resources) => isScoped(resources, req.auth.scopes);
+
+				return next();
+			}
+		}
+
+		// we do not have a valid key...
+		return res.sendStatus(401);
```
richbayliss, line 146: This covers the instance of an unknown key, rather than an unscoped key

src/lib/api-keys.ts
```diff
@@ -0,0 +1,290 @@
...
+	} catch (err) {
+		console.error(err);
+		res.status(503).send(`Unexpected error: ${err}`);
+	}
+};
+
+function isEqualScope(a: Scope, b: Scope): boolean {
+	return _.isEqual(a, b);
```
richbayliss, line 154: `Scope` types are (de)serialised to JSON so have to be flat objects, which makes this possible. Might be an issue if a scope becomes more complex.

src/lib/api-keys.ts
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	// Check authHeader with various schemes
+	const match = authHeader.match(/^(?:ApiKey|Bearer) (\w+)$/i);
+
+	// Return key from match or undefined
+	return match?.[1];
+}
+
+interface DbApiSecret {
```
richbayliss, line 178: Move this up to the top with the other definitions

src/lib/api-keys.ts
```diff
@@ -0,0 +1,290 @@
...
+		scopes: serialiseScopes(scopes),
+	});
+
+	// return the new key
+	return key;
+}
+
+export type GenerateKeyOptions = { force: boolean; scopes: Scope[] };
+
+// TODO: Memoize this
```
richbayliss, line 202: Todo...

src/lib/api-keys.ts
```diff
@@ -0,0 +1,290 @@
...
+	// null means the key wasn't known...
+	if (secret == null) {
+		return null;
+	}
+
+	const { scopes } = secret;
+	return deserialiseScopes(scopes);
+}
+
+// TODO: Memoize this
```
richbayliss, line 220: Todo...

src/device-state.ts
```diff
@@ -346,11 +345,11 @@ async function saveInitialConfig() {
 
 export async function loadInitialState() {
 	await applicationManager.initialized;
+	await apiKeys.initialized;
```
richbayliss, line 348: as long as this is called...

src/device-state.ts
```diff
@@ -374,7 +373,7 @@ export async function loadInitialState() {
 	log.info('Reporting initial state, supervisor version and API info');
 	reportCurrentState({
 		api_port: conf.listenPort,
-		api_secret: conf.apiSecret,
+		api_secret: apiKeys.cloudApiKey,
```
richbayliss, line 376: ... then this will be set to a valid key

src/supervisor-api.ts
```diff
@@ -124,9 +77,19 @@ export class SupervisorAPI {
...
+			// force a new cloud key
+			await apiKeys.generateCloudKey(true);
+
+			// report the new key to the cloud API
+			deviceState.reportCurrentState({
+				api_secret: apiKeys.cloudApiKey,
+			});
+
+			// return the value of the new key to the caller
+			res.status(200).send(apiKeys.cloudApiKey);
```
richbayliss, line 92: Is this a possible security risk? The `cloudApiKey` is scoped to `global` so it can do anything, and it's returned to the client who has used a non-global scoped key. We should think about this @CameronDiver

richbayliss: reading this: https://www.balena.io/docs/reference/supervisor/supervisor-api/#post-v1regenerate-api-key

... it seems there is 2 things going on here;
- the scoped key should be forcibly regenerated and returned
- the key is sent to the API

In the new scoped key world the key in the cloud is different to the key in the app. So I propose the following:
- regenerate the scoped key and return it to the client
- **IF** the key is the cloud key, send the key to the API also

src/compose/utils.ts
```diff
@@ -388,11 +396,12 @@ export function addFeaturesFromLabels(
 			} as Dockerode.DeviceRequest),
 	};
 
-	_.each(features, (fn, label) => {
-		if (checkTruthy(service.config.labels[label])) {
-			fn();
+	for (const feature of Object.keys(features) as [keyof typeof features]) {
```
CameronDiver, line 399:
```suggestion
	for (const feature: [keyof typeof features] of Object.keys(features)) {
```

src/supervisor-api.ts
CameronDiver: Sounds good to me

CameronDiver (commented):
Happy to approve with these changes

src/compose/utils.ts
richbayliss: @CameronDiver `The left-hand side of a 'for...of' statement cannot use a type annotation.ts(2483)`

CameronDiver: Ah, that's a surprise. Ok no worries then

CameronDiver (approved):

richbayliss:
@balena-ci rebase

Issue #1434, opened 2020-08-11
Not deleting images with the same name


20k-ultra:
The following steps describe a way to create a release that will not delete images from previous releases. The only way to remove them is to manually prune the system with `balena system prune -a`.

Conditions required :
 - 2 or more containers using same image
 - containers using the `build` and `image` directives to reference the same image

In the following docker-compose example we have `app_1` and `app_2` that use the same image `company/custom-python-runtime` but apply their own configurations at build time. Per the Compose file [documentation](https://docs.docker.com/compose/compose-file/compose-file-v2/#build), If you specify image as well as build, then Compose names the built image with the value passed to image directive.

```yaml
version: '2.1'
services:
  app_1:
    build:
      context: .
      dockerfile: ./python_runtime/Dockerfile.template
      args:
        SPECIAL_ARG: ${app_1_arg}
    image: company/custom-python-runtime
  app_2:
    build:
      context: .
      dockerfile: ./python_runtime/Dockerfile.template
      args:
        SPECIAL_ARG: ${app_2_arg}
    image: company/custom-python-runtime
```

This causes multiple images to be created with the same name.

The issue is that when a new release is deployed the Supervisor does not remove these containers for some reason. So if you continue to deploy over time the device's disk space will be consumed by these images until you prune the system. 

# Reproduction

We have a working reproduction of this issue at this repo: https://github.com/balena-io-playground/remove-images-bug-reproduction.

# Workaround

I have discovered that this can be resolved by adding a uniquee tag to each image (example image: company/custom-python-runtime:app_2), so their names are unique.  In [our repo demonstrating the issue](https://github.com/balena-io-playground/remove-images-bug-reproduction), here's the workaround in `docker-compose.yml`:

```
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,10 +2,10 @@ version: '2.1'
 services:
   app_1:
     build: ./noisy-1
-    image: noisy1
+    image: noisy1:noisy1
   app_2:
     build: ./noisy-1
-    image: noisy1
+    image: noisy1:noisy2
   app_3:
     build: ./noisy-1
-    image: noisy1
+    image: noisy1:noisy3
```

(However, the SV should still not allow this to happen.)

20k-ultra:
This is the function responsible for removing images: https://github.com/balena-io/balena-supervisor/blob/b656aa73045c426d4e6d553a6b8b6f2fa3624f23/src/compose/images.ts#L504

[...]

PR #1462, opened 2020-09-17
Handle delete of multiple images with same dockerImageId

pipex:
A docker-compose.yml with the following structure

```
version: '2.1'
services:
  app_1:
    build: ./noisy-1
    image: noisy1
  app_2:
    build: ./noisy-1
    image: noisy1
  app_3:
    build: ./noisy-1
    image: noisy1
```

Will lead to the supervisor creating multiple image database entries
with the same dockerId (this is because of how the engine handles this
particular case). This case is not handled by the removal process
leading to image pile up and increased disk usage.

Change-type: patch
Connects-to: #1434

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -535,6 +535,33 @@ async function removeImageIfNotNeeded(image: Image): Promise<void> {
 				logger.logSystemEvent(LogTypes.deleteImage, { image });
 				docker.getImage(img.dockerImageId).remove({ force: true });
 				removed = true;
+			} else if (imagesFromDb.length > 1 && hasDigest(img.name)) {
+				const [dockerRepo] = img.name.split('@');
+				const dockerImage = await docker.getImage(img.dockerImageId).inspect();
+				const matchingTags = dockerImage.RepoTags.filter((tag) => {
+					const [tagRepo] = tag.split(':');
+					return tagRepo === dockerRepo;
+				});
+
+				reportChange(
+					image.imageId,
+					_.merge(_.clone(image), { status: 'Deleting' }),
+				);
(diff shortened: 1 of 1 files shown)


20k-ultra:
let's make this PR ready for review. We have confirmed this PR fixes the issue.

pipex:
When do you have some time? We can pair program some tests, which is what I suppose it's missing

20k-ultra:
This PR is being blocked due to lack of testing. I have been working on adding these but since this PR touches code that uses the knex database and dockerode there is a lot of mocking required to correctly test. I really want this merged so I want to approve and created https://github.com/balena-io/balena-supervisor/issues/1492 to come back right after to add testing. I believe this is acceptable because there isn't any test coverage for the current functionality which means the argument for avoiding merging due to lack of testing is already lost.

pipex:
That's similar to what I experienced trying to test the database migrations, that's why I created this https://github.com/balena-io/balena-supervisor/issues/1475

If you want we can work together to try to tackle the database testing stuff @20k-ultra

20k-ultra (dismissed):
This PR has been tested on real devices for few weeks so he lack of testing is fine for now but should be added soon. For future reference the timeout to avoid race condition is due to that these container operations such as remove, download, start etc are all done in parallel. If this timeout was not present the code will still work but an exception is logged and then it runs again but with the timeout it's enough time that it never throws the exception. Not ideal but works right now.

20k-ultra:
@balena-ci retest

20k-ultra:
@balena-ci retest please

20k-ultra:
@balena-ci test

20k-ultra (approved):

PR #1463, opened 2020-09-17
Clarify location for enabling update lock override

20k-ultra:
Previous text says to add a config value which is not possible because the dashboard will prevent you from adding the config value needed to toggle this feature. This is because we have added it as a UI input via an action in the actions list.


20k-ultra:
@balena-ci I self-certify!

Issue #1428, opened 2020-08-05
`No delay specified in scheduledApply` error while livepushing


xginn8:
```
[Debug]   Sending request to http://192.168.3.215:48484/v2/local/target-state
No delay specified in scheduledApply

BadRequestDeviceAPIError: No delay specified in scheduledApply
    at Request._callback (/usr/lib/node_modules/balena-cli/build/utils/device/api.js:145:39)
    at Request.self.callback (/usr/lib/node_modules/balena-cli/node_modules/request/request.js:185:22)
    at Request.emit (events.js:315:20)
    at Request.EventEmitter.emit (domain.js:483:12)
    at Request.<anonymous> (/usr/lib/node_modules/balena-cli/node_modules/request/request.js:1154:10)
    at Request.emit (events.js:315:20)
    at Request.EventEmitter.emit (domain.js:483:12)
    at IncomingMessage.<anonymous> (/usr/lib/node_modules/balena-cli/node_modules/request/request.js:1076:12)
    at Object.onceWrapper (events.js:421:28)
    at IncomingMessage.emit (events.js:327:22)
    at IncomingMessage.EventEmitter.emit (domain.js:483:12)
    at endReadableNT (_stream_readable.js:1220:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

ref. https://www.flowdock.com/app/rulemotion/r-supervisor/threads/ZrRWOyEHrSg39e-xEyvZVOUN57b

supervisor `v11.11.7`
balena-CLI `12.12.1`

Timple:
Having the same issue. Did you get around this?

20k-ultra:
hey @xginn8 and @Timple, could you provide more info on when this happens ? I'm hoping to get some reproducible steps or at least a better idea of what I can do to trigger this to happen. Should I be triggering restarts on my service or does this just happen without any input from you ? What is the frequency in which this happens ? 

So far, running Supervisor `v11.14.0` and CLI `12.20.0` hasn't given me the issue but I haven't messed around with it much.

Timple:
Yesterday it reproduced fine. I'll try to reproduce today. Maybe it's related to this error I was also constantly getting when doing a local push:

```
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Error: No native build was found for platform=linux arch=x64 runtime=node abi=64 uv=1 libc=glibc node=10.21.0
[...]

PR #1466, opened 2020-09-22
Correctly evaluate if scheduledApply.delay is not set

20k-ultra:
[triggerApplyTarget](https://github.com/balena-io/balena-supervisor/blob/fbc2875951afbfe4735e09708b55c90520fc2dbd/src/device-state.ts#L838) function in device-state uses a default value of 0 for a delay. When this value is checked if it is undefined/not set the condition can fail given the default value. This PR checks if the value is undefined or not a number rather then if it's falsey.

Closes: #1428
Change-type: patch

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -854,7 +854,10 @@ export function triggerApplyTarget({
 			// If a delay has been set it's because we need to hold off before applying again,
 			// so we need to respect the maximum delay that has
 			// been passed
-			if (!scheduledApply.delay) {
+			if (scheduledApply.delay === undefined || isNaN(scheduledApply.delay)) {
+				log.debug(
+					`Tried to apply target with invalid delay: ${scheduledApply.delay}`,
+				);
 				throw new InternalInconsistencyError(
 					'No delay specified in scheduledApply',
 				);


src/device-state.ts
```diff
@@ -854,7 +854,7 @@ export function triggerApplyTarget({
 			// If a delay has been set it's because we need to hold off before applying again,
 			// so we need to respect the maximum delay that has
 			// been passed
-			if (!scheduledApply.delay) {
+			if (scheduledApply.delay === undefined || isNaN(scheduledApply.delay)) {
```
CameronDiver, line 857: This could be:
```suggestion
			if (isNaN(scheduledApply.delay)) {
```
and we should also show what the value itself is in the error message

20k-ultra: I agree this is the desired logic but the **delay** value can be undefined as well and isNaN expects a number so we must check first so the compiler can infer it is a number when passing it to isNaN. The typing seems kinda off doesn't it ? You'd expect isNaN to accept type of any.

Also I was confident the value being logged will be undefined and isNaN was used just to be more strict. I don't see how it will ever be anything but a number or undefined however I will never say no to adding more insight on whats going on with the device :p

20k-ultra:
Added a log to convey what the delay value was when it was considered invalid.

CameronDiver (approved):

Issue #1465, opened 2020-09-21
Supervisor gets stuck into loop when trying to apply HOST_CONFIGFS_ssdt


pipex:
When trying to apply SSDT overlays in Up Boards, the supervisor appears to get stuck in a loop trying to apply the target state. 

```
21.09.20 16:04:35 (+0000) Applying boot config: {"HOST_CONFIGFS_ssdt":["spidev1.0"]}
21.09.20 16:04:35 (+0000) Applied boot config: {"HOST_CONFIGFS_ssdt":["spidev1.0"]}
21.09.20 16:04:35 (+0000) Applying boot config: {"HOST_CONFIGFS_ssdt":["spidev1.0"]}
21.09.20 16:04:35 (+0000) Applied boot config: {"HOST_CONFIGFS_ssdt":["spidev1.0"]}
21.09.20 16:04:36 (+0000) Applying boot config: {"HOST_CONFIGFS_ssdt":["spidev1.0"]}
```

Possibly the supervisor is failing to determine the correct state of the overlay application. 
From dmesg log

```
[425631.012994] ACPI: Host-directed Dynamic ACPI Table Load:
[425631.013021] ACPI: SSDT 0xFFFF9C70A7423840 0000B7 (v05 INTEL  SPIDEV0  00000001 INTL 20170303)
[425631.013207] ACPI: Executed 1 blocks of module-level executable AML code
[425631.020057] spidev spi-SPT0001:00: do not use this driver in production systems!
[427705.663280] ACPI configfs: table already loaded
[427727.625072] ACPI configfs: table already loaded
[430143.823475] ACPI configfs: table already loaded
[430166.332069] ACPI configfs: table already loaded
[432582.562157] ACPI configfs: table already loaded
[432604.572864] ACPI configfs: table already loaded
```

Further dicussion [here](https://www.flowdock.com/app/rulemotion/resin-devices/threads/Bh6t9mEHFyV-OMwvTZ4FVXbUVMS) 
Related to PR #1187

PR #1467, opened 2020-09-22
Fix config checks for ConfigFS backend

pipex:
When trying to apply SSDT overlays in Up Board, the supervisor currently
gets stuck in a loop trying to apply target state. See #1465

This was due to a bug in parsing the configuration, which lead to
the method bootConfigChangeRequired returning true when no change was
needed.

Change-type: patch
Connects-to: #1465

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ tools/dind/config/
 tools/dind/config.json*
 tools/dind/apps.json
 tools/dind/backup.tgz
-test/data/config*.json
+test/data/**/config*.json
 test/data/*.sqlite
 test/data/led_file
 /coverage/
--- a/src/config/backends/config-fs.ts
+++ b/src/config/backends/config-fs.ts
@@ -94,7 +94,7 @@ export class ConfigFs extends ConfigBackend {
 				`AML: ${oemId.trim()} ${oemTableId.trim()} (Rev ${oemRevision.trim()})`,
 			);
 		} catch (e) {
-			log.error('Issue while loading AML ${aml}', e);
+			log.error(`Issue while loading AML ${aml}`, e);
 		}
 		return true;
 	}
@@ -214,7 +214,7 @@ export class ConfigFs extends ConfigBackend {
 					return [value];
 				} else {
 					// or, it could be parsable as the content of a JSON array; "value" | "value1","value2"
-					return value.split(',').map((v) => v.replace('"', '').trim());
+					return value.split(',').map((v) => v.replace(/"/g, '').trim());
 				}
 			default:
 				return value;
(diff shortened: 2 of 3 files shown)


pipex:
@20k-ultra just realized I named the branch for this PR wrong. Will that affect the behavior of versionbot?

20k-ultra:
> @20k-ultra just realized I named the branch for this PR wrong. Will that affect the behavior of versionbot?

It's ok. The branches get deleted after the merge and only the commit messages are persisted in the [Changelog](https://github.com/balena-io/balena-supervisor/blob/master/CHANGELOG.md)

src/config/backends/config-fs.ts
```diff
@@ -117,6 +117,7 @@ export class ConfigFs extends ConfigBackend {
 	}
 
 	private async writeConfigJSON(config: ConfigfsConfig): Promise<void> {
+		console.log(this.ConfigFilePath);
```
20k-ultra, line 120:
```suggestion
```

I usually do a search for console.log before submitting to find any debugging code :p

20k-ultra (changes requested):
Looks great especially with the additional test coverage on a backend we previously didn't have any for..besides the commented out stuff.

Speaking of, does the commented test suites pass ? If not you could try to recreate whatever cases it was trying to cover and delete that big block of commented code.

src/config/backends/config-fs.ts
pipex: 😳 thanks!

pipex:
@20k-ultra no, the commented test suite does not pass, I think it was commented as part of the singleton work by @richbayliss and @CameronDiver, I think the new tests cover the same stuff, but I'd have to check

20k-ultra:
@pipex if you remove the commented out block we can approve and get this merged

pipex:
@20k-ultra sorry Miguel, I'm just making sure that the new tests cover everything from the old test.

pipex:
@20k-ultra, changes done

.gitignore
```diff
@@ -10,7 +10,7 @@ tools/dind/config/
 tools/dind/config.json*
 tools/dind/apps.json
 tools/dind/backup.tgz
-test/data/config*.json
+test/data/**/config*.json
```
20k-ultra, line 13: What's this addition for ?

20k-ultra (approved):
Sorry didn't mean to rush. LGTM!

.gitignore
pipex: The new test creates a configfs.json in `test/data/mnt/boot` which we don't want in the repo

20k-ultra: ah ok. I was just making sure since it had the wildcard match.

Issue #1447, opened 2020-08-26
Drop `(BALENA|RESIN)_SUPERVISOR_VERSION` env vars


xginn8:
Exposing env vars with the supervisor version requires a restart of each user container when the supervisor is upgraded, which makes upgrades themselves less transparent. This may require a major change, but should be transparent otherwise as we already expose the version info itself as part of https://www.balena.io/docs/reference/supervisor/supervisor-api/#get-v2version.

PR #1469, opened 2020-09-23
version: drop SUPERVISOR_VERSION env var

xginn8:
In order to make supervisor upgrades more transparent, lets move away
from this env var since it requires a container restart any time the supervisor
is upgraded. We should ultimately move towards providing the supervisors
set of capabilities, but that can come later

Connects-to: #1447
Change-type: major


20k-ultra:
much agreed. I don't know the full scope of this variable but I have seen it trigger container restarts before.

xginn8:
I don't understand how this is failing in CI but not locally, as both tests succeed when I run them on my machine.

20k-ultra:
@xginn8 There are 2 situations I can think of:

 - When you switch between branches and run tests and then go to another branch those tests from the first branch will actually run again. The test script only builds the tests and src but does not remove existing tests. This is not the case here but I do a whole repo purge if I'm not sure using `rm -rf node_modules build && npm ci`.
- Most likely the cause given the pipeline is logging an error related to a docker test and my machine also passes the tests...if we don't stub everything needed for a test like docker daemon interactions then the test uses our machines and passes because we most likely have docker installed but when it runs in the pipeline docker is not running so it fails. I create https://github.com/balena-io/balena-supervisor/issues/1355 to improve this.

Otherwise your code changes had nothing to do with docker so it's pretty weird to fail.

CameronDiver (approved):

xginn8:
cc @balenalabs/hardware-hackers in case we rely on this env var in any of our projects, we should migrate away from that in favor of querying the supervisor API directly.

CameronDiver (approved):

Issue #1468, opened 2020-09-23
Supervisor 11.14.4 fails to start on upgrade


pipex:
Tested with livepush. When starting and there is data in the database, supervisor gets stuck at 

```
Supervisor v11.14.4 starting up...
```

20k-ultra:
What does upgrade mean ? You went from an older Supervisor version to v11.14.4 or greater and are getting stuck with this error ?

As we've spoken before about this, that is the path I took and experienced this issue. When I freshly provisioned a device (no upgrade) v11.14.4 booted.

pipex:
Sorry, probably should be more descriptive here. The problem is when there is data on the database and the migration in 11.14.4 is ran. In that case `db.initialized` runs into a deadlock. I'm preparing a PR now

20k-ultra:
nice! thanks for clarifying that for me. Here is a screenshot of my error if it helps:

![image](https://user-images.githubusercontent.com/3946250/94055533-0f879480-fdab-11ea-93c4-bb4778efd93c.png)

20k-ultra:
Is there a workaround @pipex  ?

```
Sep 29 00:51:40 8f07331 resin-supervisor[3232]: [info]    Supervisor v11.14.7 starting up...
Sep 29 00:51:41 8f07331 resin-supervisor[3232]: Can't take lock to run migrations: update `knex_migrations_lock` set `is_locked` = 1 where `is_locked` = 0 - SQLITE_BUSY: database is locked
Sep 29 00:51:41 8f07331 resin-supervisor[3232]: If you are sure migrations are not running you can release the lock manually by deleting all the rows = require(migrations lock table: knex_migrations_lock
Sep 29 00:51:41 8f07331 resin-supervisor[3232]: Can't take lock to run migrations: update `knex_migrations_lock` set `is_locked` = 1 where `is_locked` = 0 - SQLITE_BUSY: database is locked
Sep 29 00:51:41 8f07331 resin-supervisor[3232]: If you are sure migrations are not running you can release the lock manually by deleting all the rows = require(migrations lock table: knex_migrations_lock
[...]

PR #1474, opened 2020-09-29
Fix supervisor deadlock during migration

pipex:
Due to the singleton work, when performing migration M00005 and there
are apps with services created in the database, a deadlock occurs
during database initialization due to a circular
dependency for generating scoped keys.

Change-type: patch
Connects-to: #1468


pipex:
Created pull request without tests to fix the issue for now, will create an issue for testing migrations.

CameronDiver (approved):

PR #1477, opened 2020-10-01
log networks that do not match

20k-ultra:
In a few instances users are unclear why their services are restarted because the SV only logs "Network changes detected". This makes it really challenging to know where to look next. This PR aims to show what networks changes have triggered their services to restart.


20k-ultra:
@balena-ci I self-certify!

PR #1479, opened 2020-10-07
Add the 'io.balena.features.early-run' label to bind mount the early /run filesystem

kaisoz:
Some daemons (such udev or systemd) use the early stage `/run` filesystem for runtime data. This label makes the supervisor mount the host `/run` folder on the `/run` folder of the container

Change-type: minor
Signed-off-by: Tomás Tormo <tomast@balena.io>

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -383,6 +383,8 @@ export async function addFeaturesFromLabels(
 			);
 		},
 		'io.balena.features.sysfs': () => service.config.volumes.push('/sys:/sys'),
+		'io.balena.features.early-run': () =>
+			service.config.volumes.push('/run:/run'),
 		'io.balena.features.procfs': () =>
 			service.config.volumes.push('/proc:/proc'),
 		'io.balena.features.gpu': () =>


src/compose/utils.ts
```diff
@@ -383,6 +383,8 @@ export async function addFeaturesFromLabels(
 			);
 		},
 		'io.balena.features.sysfs': () => service.config.volumes.push('/sys:/sys'),
+		'io.balena.features.early-run': () =>
```
CameronDiver, line 386: Why `early-run` and not `runfs`, for example?

kaisoz: This was for the hack week. We needed for our 'Predictive maintenance' project, which uses telegraf to get iostat data, and one of telegraf's requirements is to have access to the host `/run` fs within the container.

I named the label like that because it refers to the `/run` fs which is set up for early daemons (like systemd or udev) instead of the `/var/run`. Just wanted to make it clear in the label.

Actually... now I think the naming is incorrect, should be something like `early-runfs`, in case you still want this to be named like that (and you want this PR merged)

PR #1480, opened 2020-10-07
Populate config.json from EC2 metadata

ab77:
* Hackathon 2k20 project

Change-type: patch

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -1,0 +1,34 @@
+import * as constants from './lib/constants';
+import * as request from './lib/request';
+import { readFileSync, writeFileSync } from 'fs';
+import { reboot } from './device-state';
+import log from './lib/supervisor-console';
+
+export const initialized = (async () => {
+	if (constants.configJsonNonAtomicPath != null) {
+		const configJsonPath = constants.configJsonNonAtomicPath;
+		const config = JSON.parse(readFileSync(configJsonPath, 'utf-8'));
+		if (
+			!config.hasOwnProperty('unmanaged') &&
+			!config.hasOwnProperty('registered_at') &&
+			!config.hasOwnProperty('applicationId') &&
+			!config.hasOwnProperty('deviceId')
+		) {
+			const opts: request.requestLib.CoreOptions = {};
+
+			log.debug('Attempting to initialise from cloud metadata');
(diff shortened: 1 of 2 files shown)


src/metadata.ts
```diff
@@ -0,0 +1,34 @@
+import * as constants from './lib/constants';
```
CameronDiver, line 1: This filename (metadata.ts) needs to be more specific.

src/metadata.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as request from './lib/request';
+import { readFileSync, writeFileSync } from 'fs';
```
CameronDiver, line 3: Don't ever use `Sync` functions in the supervisor, use the promise based ones:

```suggestion
import { promises as fs } from 'fs';
```

and then 
`fs.readFile`, fs.writeFile`

src/metadata.ts
```diff
@@ -0,0 +1,34 @@
...
+import { reboot } from './device-state';
+import log from './lib/supervisor-console';
+
+export const initialized = (async () => {
+	if (constants.configJsonNonAtomicPath != null) {
+		const configJsonPath = constants.configJsonNonAtomicPath;
```
CameronDiver, line 9: This variable is useless

src/metadata.ts
```diff
@@ -0,0 +1,34 @@
...
+		if (
+			!config.hasOwnProperty('unmanaged') &&
+			!config.hasOwnProperty('registered_at') &&
+			!config.hasOwnProperty('applicationId') &&
+			!config.hasOwnProperty('deviceId')
+		) {
+			const opts: request.requestLib.CoreOptions = {};
+
+			log.debug('Attempting to initialise from cloud metadata');
+			const url = 'http://169.254.169.254/latest/user-data';
```
CameronDiver, line 20: What is this IP?

CameronDiver (changes requested):
Can I get a little more information about what this PR is for please? It should be in the PR description, but as it stands I don't know what this is for, and more specifically I don't know that it should be added to the supervisor

src/metadata.ts
ab77: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html

Issue #1471, opened 2020-09-25
Inconsistent memory reporting for node containers


pipex:
When running node containers, dashboard RAM usage shows a different value that what is reported by the 
`top` and `balena stats` commands. 

In a [reported case](https://www.flowdock.com/app/rulemotion/r-supervisor/threads/UvX4-J4fQkvosmG_SkGtlvLN5eH) for a rpi3, the dashboard and  the `free` command show 700MB in use, while `top` and `balena stats` show 200MB usage. 

This requires further investigation to find the source of truth of memory usage and report the correct value to the user.

20k-ultra:
A user has provided a description of what seems to be the current cause of this issue.

```
The new added dashboard widget is giving wrong information, the memory usage doesn't take into account memory used for cache, giving a false impression that the device is running out of memory when in reality linux tries to use as much memory as possible for cache that can be take into account as free memory, ie. you can check how htop displays the memory usage. Then the disk usage one, at least in one device, is showing 2.8G of 2.8G used and if we look at df in the device, we can see that 1.6G is free.
```

@xginn8 looked into how we are retrieving the memory usage of the device and had the following to say:

```
In terms of memory usage, the supervisor uses a module called systeminformation which in turn calculates used memory as totalmem - freemem from /proc/meminfo (seen here: https://github.com/sebhildebrandt/systeminformation/blob/master/lib/memory.js#L154).
```

Therefore, we should see if this module is calculating with the cached memory.

xginn8:
This is pretty well outlined in the kernel manpage: https://man7.org/linux/man-pages/man5/proc.5.html

PR #1487, opened 2020-10-12
Improve calculation for used system memory

pipex:
The memory information reported by the supervisor currently
estimates the value of used memory as `MemTotal - MemFree`.
However, linux systems will try to cache and buffer as much
memory as possible, which will affect the output of `MemFree`
(from /proc/meminfo) and in consequence the memory usage seen
by the user on the dashboard, which will appear much greater than
it is.

The correct calculation should be `MemTotal - MemFree - Buffers - Cached`,
which the calculation performed by `htop` and the `free` commands.

Change-type: patch
Connects-to: #1471

### DIFF
--- a/src/lib/system-info.ts
+++ b/src/lib/system-info.ts
@@ -62,7 +62,7 @@ export async function getMemoryInformation(): Promise<{
 }> {
 	const mem = await systeminformation.mem();
 	return {
-		used: bytesToMb(mem.used),
+		used: bytesToMb(mem.used - mem.cached - mem.buffers),
 		total: bytesToMb(mem.total),
 	};
 }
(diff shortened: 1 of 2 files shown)


test/35-system-info.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+function kbToMb(kilobytes: number) {
+	return Math.floor(kilobytes / 1024);
+}
+
+describe('System information', function () {
+	it('Returns the correct memory values', async () => {
+		// Stub the output of the systeminformation module
+		stub(systeminformation, 'mem').resolves({
+			total: 763472 * 1024,
```
xginn8, line 15: can we clean up some of these magic numbers since they seem to be used in multiple places? it's a little painful on the eyes and hopefully will prevent any fat fingering down the road.

pipex: Yeah, good point, my original idea was to mock the output of `/proc/meminfo` hence the magic numbers.

CameronDiver (approved):
Nice!

Issue #1472, opened 2020-09-25
Stop fetching device-type from config.json


CameronDiver:
We should instead use the device-type.json. We use this file here https://github.com/balena-io/balena-supervisor/blob/master/src/config/functions.ts#L43 for the device architecture, and we should also fetch the device-type from there. 

We should use the `BOOT_MOUNTPOINT` supervisor variable to access the file in addition.

20k-ultra:
The device-type is populate from config.json in [provisioningOptions](https://github.com/balena-io/balena-supervisor/blob/7ca1f837ff98990435333488f8312965420c0f34/src/config/functions.ts#L65) and [extendedEnvOptions](https://github.com/balena-io/balena-supervisor/blob/7ca1f837ff98990435333488f8312965420c0f34/src/config/functions.ts#L104). Also device architecture has its own function that you provided but in the above 2 functions I just linked they also provide a deviceArch value which pulls from config.json.

Should we make all references to deviceType and deviceArch be pulled from device-type.json ?

20k-ultra:
There's no test coverage for this file so we should make them when we make this change.

CameronDiver:
@20k-ultra yeah all references to deviceType and deviceArch should come from device-type.Jason

PR #1490, opened 2020-10-15
Change source of deviceType to device-type.json

pipex:
The source of truth for the device-type should be
device-type.json instead of config.json

Change-type: minor
Connects-to: #1472

### DIFF
--- a/src/config/functions.ts
+++ b/src/config/functions.ts
@@ -1,5 +1,6 @@
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
+import * as memoizee from 'memoizee';
 import { fs } from 'mz';
 import { URL } from 'url';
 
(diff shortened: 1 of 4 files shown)


src/config/functions.ts
```diff
@@ -38,12 +38,12 @@ export const fnSchema = {
 	macAddress: () => {
 		return macAddress.getAll(constants.macAddressPath);
 	},
-	deviceArch: async () => {
+	deviceArch: _.memoize(async () => {
```
Page-, line 41: `_.once` makes more sense if you never plan to pass arguments:
```suggestion
	deviceArch: _.once(async () => {
```
There is the issue of if the promise fails whether you want to remember that failure and forever fail (in this case the try/catch should avoid that though). If you don't want to remember the failure then an easy solution is to use `memoizee` with `promise: true`, and the supervisor already uses memoizee elsewhere so it won't increase the size

src/config/functions.ts
```diff
@@ -53,7 +53,23 @@ export const fnSchema = {
 			log.error(`Unable to get architecture: ${e}`);
 			return 'unknown';
 		}
-	},
+	}),
+	deviceType: _.memoize(async () => {
```
Page-, line 57: `_.once` makes more sense if you never plan to pass arguments:
```suggestion
	deviceType: _.once(async () => {
```
There is the issue of if the promise fails whether you want to remember that failure and forever fail (in this case the try/catch should avoid that though). If you don't want to remember the failure then an easy solution is to use `memoizee` with `promise: true`, and the supervisor already uses memoizee elsewhere so it won't increase the size

src/config/functions.ts
pipex: Oh, good catch @Page-, hadn't thought of that case. Thanks!

src/config/index.ts
```diff
@@ -87,11 +87,23 @@ export async function get<T extends SchemaTypeKey>(
...
+				const defaultValue = schemaTypes[key].default;
+				if (defaultValue instanceof t.Type) {
+					const maybeDecoded = (defaultValue as t.Type<any>).decode(undefined);
+
+					return (
+						checkValueDecode(maybeDecoded, key, undefined) && maybeDecoded.right
+					);
+				}
+				return defaultValue as SchemaReturn<T>;
+			});
```
pipex, line 106: This change makes me kind of nervous, I don't think I understand the schema types entirely yet. 

The reason for this is to allow functions to throw. In the case of `deviceArch` and `deviceType`, this allows to memoize only if there was no error.

Does this make sense @Page-, @20k-ultra ?

pipex: I researched a bit more and I'm fairly confident that this is the correct implementation. In the rare case that a function throws, the catch will try to use the default value. Most functions seem to use t.never, which should lead to the check failing and another exception to be thrown. In the case of deviceType and deviceArch, the default value is the string `'unknown'` which is what should be returned. Since the function throws an error, the result will not be memoized.

CameronDiver (approved):

Issue #1476, opened 2020-09-30
Restart policy not enforced on v11.14.1 and up


20k-ultra:
If you use the [restart](https://docs.docker.com/compose/compose-file/#restart) key in your docker-compose service and run Supervisor version `v11.14.1` or greater this value is not respected. 

I created https://github.com/balena-io-playground/restart-policy-regression to make it easy to recreate this issue.

I have been told that the Suprevisor is not in charge of this restart behaviour and in fact this value is just forwarded to the docker daemon. So we must figure out why in `v11.14.1` the Supervisor is not passing on the correct information to the daemon.

20k-ultra:
We have discovered that the restart policy is working but the issue happens when the target state is being applied. When the Supervisor is told to apply a new target state a list of steps to get there is compiled. The SV will perform the steps such as download image and start container but since it exits right away for some reason that start container step thinks it must run again. Since the container keeps exiting while the Supervisor tries to finish applying the target state it gets stuck in this loop. When we added a timeout of 2 minutes (arbitrary but point is to allow target state steps to complete) then the container will remain stopped.

We are not investigating what changes were made in https://github.com/balena-io/balena-supervisor/commit/96c68166a1e1d9b400fca84d33910d526009ecfd commit that causes this. (We have isolated the regression to come from this commit).

20k-ultra:
Below we have the code before the commit. It computes what steps are required to achieve the target. We can see that if the target config wants the container running then a start step is added..

https://github.com/balena-io/balena-supervisor/blob/13241bf6e7aaa2d29b3fc9937052f2d685afc087/src/application-manager.js#L690-L701

However below we see different logic where if the target running state is not equal to the current running state then we add the step that is opposite to the current state.

https://github.com/balena-io/balena-supervisor/blob/30c352f4f696a0231d8f7e0c04590ed7dcbb078d/src/compose/app.ts#L560-L574

20k-ultra:
Although I created a PR that fixes the issue it did not address why the old code was not impacted when it did not check the restart policy. I have now concluded that the issue resides within the new function [compareServices](https://github.com/balena-io/balena-supervisor/blob/30c352f4f696a0231d8f7e0c04590ed7dcbb078d/src/compose/app.ts#L255) which replaced [compareServicesForUpdate](https://github.com/balena-io/balena-supervisor/blob/13241bf6e7aaa2d29b3fc9937052f2d685afc087/src/application-manager.js#L388).

[...]

PR #1495, opened 2020-10-21
Prevent getting stuck in loop when trying to apply target state

20k-ultra:
Prevent the Supervisor from getting stuck in a loop when trying to start new containers from target state that exit right away.

Closes: #1476
Change-type: patch

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -565,9 +565,13 @@ export class App {
 		) {
 			return generateStep('updateMetadata', { current, target });
 		} else if (target.config.running !== current.config.running) {
-			if (target.config.running) {
+			// Since the running states don't match that means they are inversed!
+			// We can achieve the target running state by doing the opposite then of what is current.
+			// However, for starting a container we must first check the restart policy.
+			if (!current.config.running && target.config.restart === 'always') {
+				// Only start the container if is stopped and wants to be restarted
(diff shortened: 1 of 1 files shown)


src/compose/app.ts
```diff
@@ -565,9 +565,13 @@ export class App {
...
+			// Since the running states don't match that means they are inversed!
+			// We can achieve the target running state by doing the opposite then of what is current.
+			// However, for starting a container we must first check the restart policy.
+			if (!current.config.running && target.config.restart === 'always') {
+				// Only start the container if is stopped and wants to be restarted
 				return generateStep('start', { target });
-			} else {
+			} else if (current.config.running) {
 				return generateStep('stop', { current });
 			}
```
Page-, line 576: I'd change this slightly to be
```ts
			if (current.config.running) {
				// If it's running and should be stopped then stop it
				return generateStep('stop', { current });
			} else if (target.config.restart === 'always') {
				// Otherwise only start the container if wants to be restarted
				return generateStep('start', { target });
			}
```
to avoid having to check `current.config.running` twice

20k-ultra: great optimization @Page-. I might remove the restart condition check entirely as some new questions have come to mind. If I leave it I'll definitely add this change.

Issue #1494, opened 2020-10-17
'io.balena.features.balena-socket' label not creating volume for Intel Nuc


20k-ultra:
If you add 'io.balena.features.balena-socket' label to your service in most device the balena engine socket will be mounted as a volume inside the service. This still works on the latest versions for rpi4 but it has been reported for broken for the following versions on Intel Nuc:
 - 2.44.0+rev1 with Supervisor 10.3.7
 - 2.50.1+rev1 with Supervisor 11.4.10

This is on the same device that was updated to try and fix but did not.

We confirmed this issue exists by checking if the socket is in `/var/run` within the service.

ab77:
Running `systemd` in the container breaks `balena-engine.socket` bind mount. To replicate, deploy https://github.com/balena-io-examples/balenalib-systemd-example to a device with updated `docker-compose.yml` mapping balena-engine socket.

PR #1497, opened 2020-10-21
Mount docker socket under `/host/run` for services

pipex:
Currently, when the label `io.balena.features.balena-socket` is set,
the balena engine socket is mounted under `/run/balena-engine.sock`.

This causes a problem when using systemd inside the container, since
this service remounts `/run` and `/run/lock` as tmpfs, causing the
socket to become unavailable.

Making a mount of the socket into `/host/run` solves this issue. This is
the same approach taken with DBUS.

Change-type: patch
Connects-to: #1494

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -345,12 +345,18 @@ export async function addFeaturesFromLabels(
 				: null,
 		'io.balena.features.balena-socket': () => {
 			service.config.volumes.push(
+				`${constants.dockerSocket}:${constants.containerDockerSocket}`,
+			);
+
+			// Maintain the /var/run mount for backwards compatibility
+			service.config.volumes.push(
 				`${constants.dockerSocket}:${constants.dockerSocket}`,
 			);
+
 			if (service.config.environment['DOCKER_HOST'] == null) {
 				service.config.environment[
 					'DOCKER_HOST'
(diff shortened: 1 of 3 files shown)


CameronDiver (commented):
Can we add information about the variable passed in to the container here please as well? https://www.balena.io/docs/reference/supervisor/docker-compose/#labels

src/compose/utils.ts
```diff
@@ -344,13 +344,19 @@ export async function addFeaturesFromLabels(
 				? service.config.volumes.push('/lib/firmware:/lib/firmware')
 				: null,
 		'io.balena.features.balena-socket': () => {
+			// Mount in /host/run to avoid clashing with systemd
+			const dstDockerSocket = '/host/run/balena-engine.sock';
```
CameronDiver, line 348: We can probably stick this in `./src/lib/constants.ts` I think

pipex: I thought about it, but `constants.ts` seems to be more related to values related to the host. I'll move it

pipex:
> Can we add information about the variable passed in to the container here please as well? https://www.balena.io/docs/reference/supervisor/docker-compose/#labels

@CameronDiver https://github.com/balena-io/docs/pull/1512

CameronDiver (approved):

PR #1498, opened 2020-10-22
Fixed evaluating if updates are needed to reach target state

20k-ultra:
This PR extends the logic that checks if an updated is needed to check for services that are a "run once" style which will exit right away. This is needed because the Supervisor will get stuck in a loop trying to start these services but since they keep exiting right away it can never finish applying the target state.

I am able to add support for this condition without modifying previous conditions by adding another check for services that are not running and if the Supervisor should start those which is done by checking the restart policy. I created some more functions to try and make the boolean condition be easier to read.

Closes: #1476
Change-type: patch

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -307,46 +307,78 @@ export class App {
 			}
 		}
 
-		const alreadyStarted = (serviceId: number) => {
-			const equalExceptForRunning = currentByServiceId[
-				serviceId
-			].isEqualExceptForRunningState(
-				targetByServiceId[serviceId],
-				containerIds,
-			);
-			if (!equalExceptForRunning) {
-				// We need to recreate the container, as the configuration has changed
+		/**
+		 * Checks that the config for the current and target services matches, ignoring their run state.
+		 * @param serviceCurrent
+		 * @param serviceTarget
+		 */
+		const isEqualExceptForRunningState = (
+			serviceCurrent: Service,
+			serviceTarget: Service,
+		) =>
+			serviceCurrent.isEqualExceptForRunningState(serviceTarget, containerIds);
+
+		/**
+		 * Checks if a service is running, if we tracked it as being started, if the config matches the desired config, and if we actually want it to ever be started.
+		 * @param serviceCurrent
+		 * @param serviceTarget
+		 */
+		const shouldBeStarted = (
+			serviceCurrent: Service,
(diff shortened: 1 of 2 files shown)


src/compose/app.ts
```diff
@@ -307,42 +307,43 @@ export class App {
...
 		const needUpdate = maybeUpdate.filter(
+			// Needs an update if NOT...
 			(serviceId) =>
 				!(
-					currentByServiceId[serviceId].isEqual(
+					alreadyStarted(currentByServiceId[serviceId]) && // 1 - Been started before
+					// 2 - Running states match OR Service stopped AND Supervisor should not restart
+					(runningStatesMatch(serviceId) ||
+						(!currentByServiceId[serviceId].config.running &&
+							doNotRestart(currentByServiceId[serviceId]))) &&
```
richbayliss, line 338: @20k-ultra I really don't think this is a check we should be doing; if the `alreadyStarted` check considers if it's been started in the past then this check is moot and just adds to the complexity. What am I missing?

20k-ultra: confirmed we don't to check restart policy.

src/compose/app.ts
```diff
@@ -307,46 +307,78 @@ export class App {
...
+			serviceTarget: Service,
+		) => {
+			// check that we want to stop it, and that it isn't stopped
+			if (
+				serviceTarget.config.running === false &&
+				applicationManager.containerStarted[serviceCurrent.containerId!] == null
+			) {
+				return true;
+			}
+		};
```
CameronDiver, line 366: There's no else here, so it will return undefined if the if case does not evaluate to true. Let's just return the predicate directly and I added a comment about the second check:

```suggestion
		const shouldBeStopped = (
			serviceCurrent: Service,
			serviceTarget: Service,
		) => {
			// check that we want to stop it, and that it isn't stopped
				return serviceTarget.config.running === false &&
				// When we issue a stop step, we remove the containerId from this structure. We check if the container has been removed first, so that we can ensure we're not providing multiple stop steps.
				applicationManager.containerStarted[serviceCurrent.containerId!] == null
		};
```

CameronDiver (commented):
Just self-certify after making that suggestion change

20k-ultra:
@balena-ci I-self-certify!

20k-ultra:
@balena-ci I self-certify!

Issue #1492, opened 2020-10-16
Add testing for PR #1462


20k-ultra:
PR #1462 is being blocked due to lack of testing however while writing the testing a lot of challenges came up when mocking various components to write a successful test suite. To unblock this PR to get the fix out I'm going to merge it and this issue is created to write tests for this PR and make other tweaks to the tests.

PR #1505, opened 2020-10-30
Add more test coverage for compose/images

20k-ultra:
The tests here add some general functionality to all testing with the addition of mocked-database which lets you easily return mocked data. It could be improved since you have to stub basically each type of database call in the exact way the code you're testing assembles its query builder...I couldn't find a simply solution to stubbing Knex that doesn't break other tests that use the database  (I tried mock-knex).

The tests themselves are for a very complicated function so open to hearing others improvements as well. Tbh, writing these tests just shows to me that the Images remove function needs to be broken down more to allow for better testing. Overall these test provide some coverage that we didn't have before.

Closes: #1492
Change-type: patch

### DIFF
--- a/test/00-init.spec.ts
+++ b/test/00-init.spec.ts
@@ -6,9 +6,6 @@ process.env.DATABASE_PATH_2 = './test/data/database2.sqlite';
 process.env.DATABASE_PATH_3 = './test/data/database3.sqlite';
 process.env.LED_FILE = './test/data/led_file';
 
-import * as dbus from 'dbus';
-import { DBusError, DBusInterface } from 'dbus';
-import { stub } from 'sinon';
 import * as fs from 'fs';
 
 // Make sure they are no database files left over from
@@ -33,31 +30,7 @@ fs.writeFileSync(
 	fs.readFileSync('./test/data/testconfig.json'),
 );
 
-stub(dbus, 'getBus').returns({
-	getInterface: (
-		_serviceName: string,
-		_objectPath: string,
-		_interfaceName: string,
-		interfaceCb: (err: null | DBusError, iface: DBusInterface) => void,
-	) => {
-		interfaceCb(null, {
-			Get: (
-				_unitName: string,
-				_property: string,
-				getCb: (err: null | Error, value: unknown) => void,
-			) => {
-				getCb(null, 'this is the value');
-			},
-			GetUnit: (
-				_unitName: string,
-				getUnitCb: (err: null | Error, unitPath: string) => void,
-			) => {
-				getUnitCb(null, 'this is the unit path');
-			},
-		} as any);
-	},
-} as any);
-
+import './lib/mocked-dbus';
 import './lib/mocked-dockerode';
 import './lib/mocked-iptables';
 import './lib/mocked-event-tracker';
--- a/test/39-compose-images.spec.ts
+++ b/test/39-compose-images.spec.ts
@@ -1,0 +1,133 @@
+import * as _ from 'lodash';
+
+import { docker } from '../src/lib/docker-utils';
+import { expect } from './lib/chai-config';
+import * as Images from '../src/compose/images';
+import * as mockedDockerode from './lib/mocked-dockerode';
+import * as mockedDatabase from './lib/mocked-database';
+import * as db from '../src/db';
+import * as sampleImageData from './data/compose-image-data.json';
+
+describe('compose/images', () => {
+	before(() => {
+		mockedDatabase.create();
+	});
+
+	after(() => {
+		try {
+			mockedDatabase.restore();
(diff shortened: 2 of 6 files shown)


test/00-init.spec.ts
```diff
@@ -33,31 +30,7 @@ fs.writeFileSync(
 	fs.readFileSync('./test/data/testconfig.json'),
 );
 
-stub(dbus, 'getBus').returns({
```
20k-ultra, line 36: I moved this code to its own file just like all the other init code the runs.

test/39-compose-images.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+			/* noop */
+		}
+	});
+
+	afterEach(() => {
+		// Clear Dockerode actions recorded for each test
+		mockedDockerode.resetHistory();
+	});
+
+	it('Removes a legacy Image', async () => {
```
20k-ultra, line 28: I modified this crafted these "legacy" payloads from real data since I've not seen legacy images but I'm used the legacy logic to make these.

test/39-compose-images.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+			// Check that docker remove was called once
+			const removeSteps = _(mockedDockerode.actions)
+				.pickBy({ name: 'remove' })
+				.map()
+				.value();
+			expect(removeSteps).to.have.lengthOf(1);
+		});
+	});
+
+	it('Removes a single Image', async () => {
```
20k-ultra, line 123: This uses real data that causes this bug to actually happen. We are adding this test coverage precisely because it caused us to write a new PR to fix that bug and this is the exact payloads that will cause it.

test/39-compose-images.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+			// Check that docker remove was called once
+			const removeSteps = _(mockedDockerode.actions)
+				.pickBy({ name: 'remove' })
+				.map()
+				.value();
+			expect(removeSteps).to.have.lengthOf(1);
+		});
+	});
+
+	it('Removes an Image with digests', async () => {
```
20k-ultra, line 327: This uses real data that causes this bug to actually happen. We are adding this test coverage precisely because it caused us to write a new PR to fix that bug and this is the exact payloads that will cause it.

test/lib/mocked-dockerode.ts
```diff
@@ -15,6 +15,24 @@ export class NotFoundError extends TypedError {
 
 const overrides: Dictionary<(...args: any[]) => Resolvable<any>> = {};
 
+interface Action {
+	name: string;
+	parameters: Dictionary<any>;
+}
+
+export let actions: Action[] = [];
```
20k-ultra, line 23: I couldn't think of a good way to add a Spy on this so had to create this...if anyone knows a better way let me know.

test/lib/mocked-dockerode.ts
```diff
@@ -62,18 +80,28 @@ export interface TestData {
...
 		return {
 			inspect: async () => {
+				addAction('inspect', {});
 				return data.networks[id];
 			},
 		} as dockerode.Network;
 	};
 
 	mockedDockerode.getImage = (name: string) => {
+		addAction('getImage', { name });
```
20k-ultra, line 93: I couldn't think of a good way to add a Spy on this so had to create this...if anyone knows a better way let me know.

test/lib/mocked-database.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export function setImages(images: Image[]) {
+	MOCKED_IMAGES = images;
+	return {
+		stub: stubImages,
+	};
+}
+
+function stubImages() {
```
20k-ultra, line 38: This unfortunately requires you to write out each object that the code you're testing will create using a QueryBuilder. If this mocked-database tool grows we will definitely look into making this more elegant.

20k-ultra (commented):
Please see my comments on specific code sections.

test/39-compose-images.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+					'sha256:a29f45ccde2ac0bde957b1277b1501f471960c8ca49f1588c6c885941640ae60',
+				Labels: null,
+				ParentId: '',
+				RepoDigests: null,
+				RepoTags: ['balena-healthcheck-image:latest'],
+				SharedSize: -1,
+				Size: 9136,
+				VirtualSize: 9136,
+			},
+		};
```
CameronDiver, line 498: This is great, the only thing I'd request is moving these huge data payloads out to separate JSON files, to keep the file a little cleaner.

20k-ultra:
@balena-ci I self-certify!

PR #1506, opened 2020-10-30
Add pipex to codeowners

20k-ultra:
welcome to the jungle


20k-ultra:
@balena-ci I self-certify!

PR #1511, opened 2020-11-04
Allow storing commits against their appIds

CameronDiver:
This paves the way for running multiple applications and storing
information related to the application against the application itself. A
couple of hacks have been added to v1 and v2 endpoints to maintain
compatability but these should eventually be removed with the addition
of a v3 api.

Change-type: minor

### DIFF
--- a/src/compose/commit.ts
+++ b/src/compose/commit.ts
@@ -1,0 +1,32 @@
+import * as db from '../db';
+
+const cache: { [appId: number]: string } = {};
+
+export async function getCommitForApp(
+	appId: number,
+): Promise<string | undefined> {
+	if (cache[appId] != null) {
+		return cache[appId];
+	}
+
+	const commit = await db
+		.models('currentCommit')
+		.where({ appId })
+		.select('commit');
+
+	if (commit?.[0] != null) {
+		cache[appId] = commit[0].commit;
+		return commit[0].commit;
+	}
+
+	return;
+}
+
+export function upsertCommitForApp(
+	appId: number,
+	commit: string,
+	trx?: db.Transaction,
+): Promise<void> {
+	cache[appId] = commit;
+	return db.upsertModel('currentCommit', { commit, appId }, { appId }, trx);
+}
--- a/src/device-api/v1.ts
+++ b/src/device-api/v1.ts
@@ -1,4 +1,3 @@
-import * as Promise from 'bluebird';
 import * as express from 'express';
 import * as _ from 'lodash';
 
@@ -9,6 +8,7 @@ import { doRestart, doPurge } from './common';
 
 import * as applicationManager from '../compose/application-manager';
 import { generateStep } from '../compose/composition-steps';
+import * as commitStore from '../compose/commit';
 import { AuthorizedRequest } from '../lib/api-keys';
 
 export function createV1Api(router: express.Router) {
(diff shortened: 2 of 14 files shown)


src/device-api/v1.ts
```diff
@@ -106,52 +106,54 @@ export function createV1Api(router: express.Router) {
...
+			}
+
+			// handle the case where the appId is out of scope
+			if (!req.auth.isScoped({ apps: [app.appId] })) {
+				res.status(401).json({
+					status: 'failed',
+					message: 'Application is not available',
+				});
+				return;
+			}
```
richbayliss, line 137: This check should happen BEFORE the service count check. It's not a huge issue since a _valid_ key is required to get this far, but technically we're leaking information to authenticated, but unauthorised, clients.

src/device-api/v2.ts
```diff
@@ -327,8 +328,11 @@ export function createV2Api(router: Router) {
 					delete apps.local[app];
 				}
 			}
+
+			const commit = await commitStore.getCommitForApp(appId);
+
 			// Return filtered applications
-			return res.status(200).json(apps);
+			return res.status(200).json({ commit, ...apps });
```
richbayliss, line 335: The commit from `apps` is going to override `commit` from the store? Just checking...

src/device-api/v2.ts
```diff
@@ -505,13 +510,26 @@ export function createV2Api(router: Router) {
...
 		}
 
+		if (_.uniq(appIds).length > 1) {
+			// We can't accurately return the commit value each app without changing
+			// the shape of the data, and instead we'd like users to use the new v3
+			// endpoints, which will come with multiapp
+			// If we're going to return information about more than one app, error out
+			return res.status(405).json({
+				status: 'failed',
+				message: `Cannot use /v2/ endpoints when a device is running multiple applications`,
```
richbayliss, line 520:
```suggestion
				message: `Cannot use /v2/ endpoints with a key that is scoped to multiple applications`,
```

richbayliss: If you have a single-app scoped key this would be fine, or no?

src/compose/commit.ts
```diff
@@ -0,0 +1,32 @@
+import * as db from '../db';
+
+const cache: { [appId: number]: string } = {};
+
+export async function getCommitForApp(
+	appId: number,
```
richbayliss, line 6: We potentially pass `undefined` to this method, would this impact the `cache[appId]` part?

src/device-state.ts
```diff
@@ -514,11 +515,28 @@ export async function getStatus(): Promise<DeviceStatus> {
...
+	// Fortunately this won't become a problem until we have system apps, and then
+	// at that point we can filter non-system apps leaving a single user app.
+	// After this, for true multi-app, we will need to report our status back in a
+	// different way, meaning this function will no longer be needed
+	const appIds = Object.keys(theState.local!.apps).map((strId) =>
+		parseInt(strId, 10),
+	);
+
+	const appId: number | undefined = appIds[0];
+	const commit = await commitStore.getCommitForApp(appId);
```
richbayliss, line 536: Possible passing of `undefined` as `appId`

richbayliss (commented):
Some questions for you if you'd like to discuss. Otherwise its good.

src/device-state.ts
```diff
@@ -514,11 +515,28 @@ export async function getStatus(): Promise<DeviceStatus> {
...
+
+	if (commit != null && !applyInProgress) {
```
pipex, line 538: We had a user asking about this on the forums the other day. If by any chance they poll the device endpoint while apply is in progress, the commit does not show in the property list. Not sure if that is what we want.

src/device-api/v2.ts
CameronDiver: There's now no longer any commit returned from the `getStatus` function

src/compose/commit.ts
CameronDiver: Good catch.

richbayliss (approved):
LGTM

Issue #1512, opened 2020-11-05
supervisor v12.1.1 not working well with livepush


pdcastro:
As originally reported by Shaun here:
https://www.flowdock.com/app/rulemotion/i-cli/threads/NDIXJP3vrEsNHSz57tuO9UneDNw

```
$ balena push 192.168.1.36 -s ~/balena-io-examples/balena-node-hello-world
$ /usr/local/bin/balena push 192.168.1.36 -s ~/balena/balena-io-examples/balena-node-hello-world
[Info]    Starting build on device 192.168.1.36
[Info]    No "docker-compose.yml" file found at "/Users/paulo/balena/balena-io-examples/balena-node-hello-world"
[Info]    Creating default composition with source: "/Users/paulo/balena/balena-io-examples/balena-node-hello-world"
[Build]   [main] Step 1/10 : FROM balenalib/raspberrypi3-node:10-stretch-run
[Build]   [main]  ---> 258e60f37899
[Build]   [main] Step 2/10 : WORKDIR /usr/src/app
[Build]   [main]  ---> Using cache
[Build]   [main]  ---> bfb39d98772a
[Build]   [main] Step 3/10 : COPY package.json package.json
[Build]   [main]  ---> Using cache
[Build]   [main]  ---> 0a7a4a9bca9d
[Build]   [main] Step 4/10 : RUN JOBS=MAX npm install --production --unsafe-perm && npm cache verify && rm -rf /tmp/*
[Build]   [main]  ---> Using cache
[Build]   [main]  ---> 226a41c0745d
[Build]   [main] Step 5/10 : COPY . ./
[Build]   [main]  ---> c7c658933754
[Build]   [main] Step 6/10 : ENV UDEV=1
[Build]   [main]  ---> Running in ca5a7bf2f0f5
[Build]   [main] Removing intermediate container ca5a7bf2f0f5
[Build]   [main]  ---> 9c31ef7a2176
[Build]   [main] Step 7/10 : ENV test=7
[Build]   [main]  ---> Running in 4a4b11ad0969
[Build]   [main] Removing intermediate container 4a4b11ad0969
[Build]   [main]  ---> 4f06c82b84db
[Build]   [main] Step 8/10 : CMD ["npm", "start"]
[Build]   [main]  ---> Running in d2b18cd715e4
[Build]   [main] Removing intermediate container d2b18cd715e4
[Build]   [main]  ---> 5af76ac9b26d
[...]

PR #1515, opened 2020-11-11
Patched scoped api keys to allow current CLI to use livepush

20k-ultra:
When the CLI uses livepush to allow for real time development of containers on your device it runs the containers on your device by building them locally but then also via livepush it can magically update your container on device with changes you made on your machine. To do this livepush needs a container Id.

When we added scoped api keys to all the /v2 API endpoints in https://github.com/balena-io/balena-supervisor/commit/c08de8701e49f417d2fc5849d237b9ed69b438e2 we caused the CLI to not be able to get the container Id that it wants to reload. This is because the CLI queries the `/v2/state/status` endpoint and then gets the container by serviceName. Since the CLI does not know about scoped api keys and it doesn't pass any key at all, the API returns an empty container list.

This PR is a patch to allow the current CLI to continue working until it can implement scoped api keys. At which we would then remove this patch.

Closes: #1512
Change-type: patch

### DIFF
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -510,7 +510,12 @@ export function createV2Api(router: Router) {
 			overallDownloadProgress = downloadProgressTotal / downloads;
 		}
 
-		if (_.uniq(appIds).length > 1) {
+		// TODO: This check is to allow the CLI to continue using livepush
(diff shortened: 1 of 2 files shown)


richbayliss:
Resolved in #1516

Issue #1455, opened 2020-09-08
Does not always reboot device with fleet custom configuration variables


majorz:
Supervisor does not always reboot a device when it is provisioned for the first time with custom configuration variables applied to the whole fleet. Steps that I used to reproduce (in sync with the customer's use-case):

1. Create a new application for with a RPi 4 device type
2. Define the following variables fleet-wise
```
BALENA_HOST_CONFIG_gpu_mem_1024 = 448
BALENA_HOST_CONFIG_gpu_mem_256 = 192
BALENA_HOST_CONFIG_gpu_mem_512 = 256
RESIN_HOST_CONFIG_start_x = 1
```
3. Flash an SD card with an OS image for this application
4. Power up the device
5. The device is not rebooted on first boot although config.txt does contain the necessary definitions written by supervisor.

I needed like 8 attempts to reproduce this, so it does not happen each time.

Useful command to see whether the changes were applied is `vcgencmd get_mem gpu` which gives `gpu=448M` if all worked correctly, or `gpu=16M` if not. The dashboard logs are also a good indication of whether the device was rebooted.

pipex:
@majorz hey Zahari, did you by any chance happen to note the supervisor version for this issue?

hnnsgstfssn:
I've seen this reproduced on at least v11.4.10 and v11.9.9, @pipex.

PR #1518, opened 2020-11-11
Ensure the first target state update is applied

pipex:
During first time run of the supervisor, the target state is queried
by `reportInitialEnv`. Since this happens early on the initialization
process, this target state report is missed by any listeners and this
can lead to the initial target state not beeing applied (see #1455).

This PR ensures that target state is re-emitted if there were no
listeners setup on call to update.

Change-type: patch
Connects-to: #1455

### DIFF
--- a/test/40-target-state.spec.ts
+++ b/test/40-target-state.spec.ts
@@ -1,0 +1,235 @@
+import { SinonStub, stub, spy, SinonSpy } from 'sinon';
+import { Promise } from 'bluebird';
+
+import * as _ from 'lodash';
+
+import { expect } from './lib/chai-config';
+import * as TargetState from '../src/device-state/target-state';
+import * as request from '../src/lib/request';
+
+const stateEndpointBody = {
+	local: {
+		name: 'solitary-bush',
+		config: {},
+		apps: {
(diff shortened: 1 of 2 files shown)


20k-ultra:
@balena-ci retest

test/40-target-state.spec.ts
```diff
@@ -0,0 +1,235 @@
+import { SinonStub, stub, spy, SinonSpy } from 'sinon';
+import { Promise } from 'bluebird';
```
20k-ultra, line 2: Should be able to use native Promises

20k-ultra (approved):
LGTM with great testing! However, it seems there's a flacky test that will prevent this from passing. When I checked out this branch locally the tests passed....

pipex:
Yeah, the tests pass on my branch as well. I will merge and see if they pass

test/40-target-state.spec.ts
pipex: It doesn't :(  I have to mock the `getAsync` method with is the promisified version of `get` from the `request` module, and it must return a Bluebird Promise.

PR #1520, opened 2020-11-11
fix: Scoped keys breaking livepush with existing cloud images on the …

richbayliss:
…device

Change-type: patch

### DIFF
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -510,25 +510,22 @@ export function createV2Api(router: Router) {
 			overallDownloadProgress = downloadProgressTotal / downloads;
 		}
 
-		if (_.uniq(appIds).length > 1) {
-			// We can't accurately return the commit value each app without changing
-			// the shape of the data, and instead we'd like users to use the new v3
-			// endpoints, which will come with multiapp
-			// If we're going to return information about more than one app, error out
-			return res.status(405).json({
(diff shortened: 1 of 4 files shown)


src/device-api/v2.ts
```diff
@@ -510,26 +510,23 @@ export function createV2Api(router: Router) {
...
-		const commit = await commitStore.getCommitForApp(appIds[0]);
+		// This endpoint does not support multi-app but the device might be running multiple apps
+		// We must return information for only 1 application so use the first one in the list
+		const APP_ID = appIds[0];
+		// Get the commit for this application
+		const COMMIT = await commitStore.getCommitForApp(APP_ID);
+		// Filter containers by this application
+		const APP_CONTAINERS = containerStates.filter((c) => c.appId === APP_ID);
+		// Filter images by this application
+		const APP_IMAGES = imagesStates.filter((i) => i.appId === APP_ID);
```
CameronDiver, line 521: Why are these all in caps?

20k-ultra: naming convention for constants. However, I think the Supervisor codebase only uses this styling for compile time constants not runtime like values from a config file vs values set on function execution. I'll revert it to be lowercase.

CameronDiver (approved):

20k-ultra:
@balena-ci retest

PR #1522, opened 2020-11-16
Add tri-app support

CameronDiver:
We store the uuid on the individual docker components so that when
building a current state, we can match them with the target state.

Change-type: minor


pipex:
Superseded by #1765

Issue #1509, opened 2020-11-03
/v1/restart throws 503 and downloads release again


20k-ultra:
Provisioned a RPI4 which has the latest versions in production as of writing this issue which are:

**Versions:**

Host OS version: balenaOS 2.58.6+rev1 (development)
Supervisor version: 12.1.4

**Steps to reproduce:**

1. Freshly provisioned device with a release (mine is a multi-container release with 3 services)
2. Click the restart button on the dashboard
3. See a toaster message (red bar with Request Error text) on the dashboard and can view errors in Supervisor journal logs.

 When looking at the journal logs I see a lot of issues.

1. My release was removed then downloaded again (the dashboard shows download progress as well)
2. Why do the steps contain fetch related steps and what is undefined ?
3. The log at the end says device state apply success and "Device state apply error Error" in the same second

```
Nov 03 19:05:29 13b8587 resin-supervisor[5200]: [event]   Event: Restart container (v1) {"appId":1658654}
Nov 03 19:05:29 13b8587 resin-supervisor[5200]: [event]   Event: Service kill {"service":{"appId":1658654,"serviceId":640680,"serviceName":"frontend","releaseId":1581185}}
Nov 03 19:05:29 13b8587 resin-supervisor[5200]: [event]   Event: Service kill {"service":{"appId":1658654,"serviceId":640681,"serviceName":"proxy","releaseId":1581185}}
Nov 03 19:05:29 13b8587 resin-supervisor[5200]: [event]   Event: Service kill {"service":{"appId":1658654,"serviceId":640682,"serviceName":"data","releaseId":1581185}}
Nov 03 19:05:30 13b8587 resin-supervisor[5200]: [event]   Event: Service exit {"service":{"appId":1658654,"serviceId":640680,"serviceName":"frontend","releaseId":1581185}}
Nov 03 19:05:30 13b8587 resin-supervisor[5200]: [event]   Event: Service stop {"service":{"appId":1658654,"serviceId":640680,"serviceName":"frontend","releaseId":1581185}}
Nov 03 19:05:31 13b8587 resin-supervisor[5200]: [event]   Event: Service exit {"service":{"appId":1658654,"serviceId":640682,"serviceName":"data","releaseId":1581185}}
Nov 03 19:05:31 13b8587 resin-supervisor[5200]: [event]   Event: Service exit {"service":{"appId":1658654,"serviceId":640681,"serviceName":"proxy","releaseId":1581185}}
[...]

PR #1524, opened 2020-11-19
1509 failed state transition steps

20k-ultra:
The balena dashboard has a button that uses the `/v1/restart` endpoint to restart all your containers by the appId provided in the request. 

This PR adds some big refactors to our testing to make it more manageable by separately the V1 and V2 test suites into their own files as seen in the first commit of this PR 1110629c2a73bcd7683c9244e45907a818a85010.

Secondly, since this endpoint did not have existing test cases I wrote test cases that would exhibit the behavior we wanted from this endpoint. Seen with the second commit ad0787ffd94f8c2b853bd7061d232a4422a67b18.

Then finally I wrote the implementation that meets this test case. I look at how the /v2 endpoint restarts a service and used the exact same methods. Some of refactors to the test files were made as improvements (additional tests for other endpoints and modified mocked-dbus to fix an error being logged about a function not existing). eb83f96fa36bffb320cb1d4fb97013aabde735f5


test/41-device-api-v1.spec.ts
```diff
@@ -178,7 +199,31 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 	});
 
 	describe('GET /v1/apps/:appId', () => {
+		it('does not return information for an application when there is more than 1 container', async () => {
```
pipex, line 202: Nice, this goes more in line with the documentation here https://www.balena.io/docs/reference/supervisor/supervisor-api/#http-api-reference

src/device-api/common.js
```diff
@@ -27,20 +28,13 @@ export async function doRestart(appId, force) {
...
-						.applyIntermediateTarget(currentState, { skipLock: true })
-						.then(function () {
-							app.services = currentServices;
-							return deviceState.applyIntermediateTarget(currentState, {
-								skipLock: true,
-							});
-						}),
-				)
-				.finally(() => deviceState.triggerApplyTarget());
+			return deviceState.pausingApply(async () => {
```
pipex, line 31: I understand this is the fix for #1509 @20k-ultra. I'm wondering why it was being done using intermediate targets before, was it to serialize the container restarts? What about doing it that way was causing the failure? 

I don't understand too much about intermediate targets so that might be cause of my confusion. 

Regarding the code itself, using `Promise.each` instead of `for` might make it more consistent with the rest of the code, but I don't see other issues.

pipex:
Nice PR, I like all the new tests. I will try to test this on my computer and finish my review. 

Just one thing, I see your first commit has the footer but I'm not sure if the footer has to go on the PR or the commits.

src/device-api/common.js
20k-ultra: Tbh, I couldn't understand it much and it was written so differently from how V2 endpoints that I didn't want to invest the time to learn it as I was just applying how V2 performs a restart which is much easier to follow.

20k-ultra:
ready for review again. added the Bluebird.each change, thanks @pipex

src/device-api/common.js
```diff
@@ -27,20 +28,12 @@ export async function doRestart(appId, force) {
...
+				return Bluebird.each(app.services, async (service) => {
```
pipex, line 32: This looks better @20k-ultra 

I'm still a bit curious if this is equivalent to the implementation with `applyIntermediateTarget`. Maybe @CameronDiver can shed a bit more light into that since I don't fully understand that yet.

My main concern was whether some race condition could occur, but I suppose that is the purpose of the `pausingApply`

pipex:
Github doesn't let me approve @20k-ultra 😅

src/device-api/common.js
20k-ultra: `applyIntermediateTarget` sounds just like another way to apply a target which means the SV figures out what is needed to reach the tagret and then executes those steps. The V2 builds a list of steps with the only step being restart from the [composition steps](https://github.com/balena-io/balena-supervisor/blob/f6ba69144b926accc279ee9d6172ba3266beef3e/src/compose/composition-steps.ts#L197) file.

pipex: Yeah, I understand that, I was thinking mostly in the sense of the 'chesterton fence'. You replaced a fairly complex piece of code, with a two line fix that does the same. I'm wondering if there are any reasons the old piece of code was made that way. 

In any case I have tested the whole PR on my device and it appears to be working well. 

I don't have permissions to approve the PR, so you might want to wait for Cameron. I'm waiting on this PR to be merged to finish the work on PR #1529

pipex (approved):
Maybe I can approve. Lets try this.

Issue #1523, opened 2020-11-16
/v2/.../start-service API endpoint throws exception if lock exists


pipex:
If a lock is left after a service crash, `/v2/applications/:appId/start-service` will fail with the error `UnhandledPromiseRejectionWarning: UpdatesLockedError: Updates are locked: EEXIST: file already exists, open '/mnt/root/tmp/balena-supervisor/services/1680126/hello-world/updates.lock'`

This is a probably a bug, since a lock should not prevent the start of a service (as a restart does)

Steps for replication

1. Install any service on a device such as https://github.com/balena-io-examples/balena-node-hello-world
2. SSH into the service and create a [lockfile](https://www.balena.io/docs/learn/deploy/release-strategy/update-locking/#application-update-locks) in `/tmp/balena/updates.lock`
3. Simulate a service crash with `balena stop <containerId>`
4. Check that the dashboard service state is updated, it should report `Exited` as state for the service.
5. Click on the start button for the service. The journal will report the error above

The workaround for now is enabling lock overrides and retrying.

20k-ultra:
fixed in 12.4.10

PR #1529, opened 2020-11-23
Improve supervisor API behavior when locks are set

pipex:
This PR adds the following

* Supervisor v1 API application actions now return HTTP status code 423 when locks
  are preventing the action to be performed. Previously this resulted in a
  503 error
* Supervisor API v2 service actions now returns HTTP status code 423 when locks are
  preventing the action to be performed. Previously, this resulted in an
  exception logged by the supervisor and the API query timing out
* Supervisor API `/v2/applications/:appId/start-service` now does not
  check for a lock to allow a crashed service to be restarted. 
Lock handling in v2 service actions is now performed by each step executor
* /v1/apps/:appId/start now queries the target state and uses that
  information to execute the start step (as v2 does). Previously start
  resulted in `cannot get appId from undefined`
* Add tests for API methods

Change-type: patch
Connects-to: #1523

### DIFF
--- a/src/device-api/v1.ts
+++ b/src/device-api/v1.ts
@@ -10,6 +10,7 @@ import * as applicationManager from '../compose/application-manager';
 import { generateStep } from '../compose/composition-steps';
 import * as commitStore from '../compose/commit';
 import { AuthorizedRequest } from '../lib/api-keys';
+import { getApp } from '../device-state/db-format';
 
 export function createV1Api(router: express.Router) {
 	router.post('/v1/restart', (req: AuthorizedRequest, res, next) => {
@@ -46,9 +47,8 @@ export function createV1Api(router: express.Router) {
 			return res.status(400).send('Missing app id');
 		}
 
-		return applicationManager
-			.getCurrentApps()
-			.then(function (apps) {
+		return Promise.all([applicationManager.getCurrentApps(), getApp(appId)])
+			.then(([apps, targetApp]) => {
 				if (apps[appId] == null) {
 					return res.status(400).send('App not found');
 				}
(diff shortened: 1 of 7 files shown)


src/device-api/v1.ts
```diff
@@ -70,12 +70,23 @@ export function createV1Api(router: express.Router) {
 					return res.status(401).send('Unauthorized');
 				}
 
+				// Get the service from the target state (as we do in v2)
+				// TODO: what if we want to start a service belonging to the current app?
+				const targetService = _.find(
+					targetApp.services,
+					(svc) => svc.serviceName === service.serviceName,
```
CameronDiver, line 77:
```suggestion
					{ serviceName: service.serviceName }
```

src/device-api/v2.ts
```diff
@@ -59,66 +59,64 @@ export function createV2Api(router: Router) {
...
+						(svc) => svc.imageId === imageId,
+					);
+				} else {
+					service = _.find(
+						app.services,
+						(svc) => svc.serviceName === serviceName,
+					);
+					targetService = _.find(
+						targetApp.services,
+						(svc) => svc.serviceName === serviceName,
```
CameronDiver, line 93: All of these can use the shorthand too

src/device-api/v2.ts
```diff
@@ -59,66 +59,64 @@ export function createV2Api(router: Router) {
...
 					});
-					return applicationManager
-						.executeStep(
-							generateStep(action, {
-								current: service,
-								target: targetService,
-								wait: true,
-							}),
-							{
-								skipLock: true,
```
CameronDiver, line 113: What's the effective difference between using force now compared to skiplock here?

pipex: `skipLock` ignores the lock and calls the funcion directly, `force` checks for a lock,  removes it if exists, and then sets its own lock to perorm the function call. In this case, since the `lockingIfNecessary` call above had the lock, it was ok to skip the lock for the `executeStep` function

src/device-api/v2.ts
```diff
@@ -59,66 +59,55 @@ export function createV2Api(router: Router) {
...
-						targetService = _.find(
-							targetApp.services,
-							(svc) => svc.serviceName === serviceName,
-						);
-					}
-					if (service == null) {
-						res.status(404).send(serviceNotFoundMessage);
-						return;
+				// Work if we have a service name or an image id
+				if (imageId == null) {
```
CameronDiver, line 72: i would probably just collapse these two checks down into one

CameronDiver (approved):

PR #1562, opened 2020-12-16
Revert  "Add dependabot configuration"

pipex:
Change-type: patch


20k-ultra:
What's the idea behind only limiting dependabot to security updates ? I realize there are some issues with the current config since it's trying to PR major version changes but otherwise why not allow minor/patch changes for bug fixes that aren't security related ?

20k-ultra (approved):

PR #1566, opened 2020-12-21
Add support for `BALENA_HOST_SPLASH_IMAGE` config var

pipex:
Setting this this configuration variable to a base64 encoded string will replace the splash
image on the device by rewriting `/mnt/boot/splash/balena-logo.png`.
This will also make a copy of the default balena logo so the splash can
be restored if the variable is removed.

Change-type: minor

### DIFF
--- a/src/config/backends/index.ts
+++ b/src/config/backends/index.ts
@@ -5,6 +5,7 @@ import { ExtraUEnv } from './extra-uEnv';
 import { ConfigTxt } from './config-txt';
 import { ConfigFs } from './config-fs';
 import { Odmdata } from './odmdata';
+import { SplashImage } from './splash-image';
 
 export const allBackends = [
 	new Extlinux(),
@@ -12,6 +13,7 @@ export const allBackends = [
 	new ConfigTxt(),
 	new ConfigFs(),
 	new Odmdata(),
+	new SplashImage(),
 ];
 
 export function matchesAnyBootConfig(envVar: string): boolean {
--- a/src/config/backends/splash-image.ts
+++ b/src/config/backends/splash-image.ts
@@ -1,0 +1,206 @@
+import * as Bluebird from 'bluebird';
+import * as _ from 'lodash';
+import { fs } from 'mz';
+import * as path from 'path';
+
+import * as constants from '../../lib/constants';
+import log from '../../lib/supervisor-console';
+import {
+	bootMountPoint,
+	ConfigBackend,
+	ConfigOptions,
+	remountAndWriteAtomic,
+} from './backend';
+
+export class SplashImage extends ConfigBackend {
+	private static readonly BASEPATH = path.join(bootMountPoint, 'splash');
+	private static readonly DEFAULT = path.join(
+		SplashImage.BASEPATH,
+		'balena-logo-default.png',
+	);
+	private static readonly FILENAMES = ['balena-logo.png', 'resin-logo.png'];
+	private static readonly PREFIX = `${constants.hostConfigVarPrefix}SPLASH_`;
+	private static readonly CONFIGS = ['image'];
+	private static readonly DATA_URI_REGEX = /^data:(.+);base64,(.*)$/;
+
+	// Check the first 8 bytes of a buffer for a PNG header
+	// Source: https://github.com/sindresorhus/is-png/blob/master/index.js
+	private isPng(buffer: Buffer) {
+		if (!buffer || buffer.length < 8) {
+			return false;
+		}
+
+		return (
+			buffer[0] === 0x89 &&
+			buffer[1] === 0x50 &&
+			buffer[2] === 0x4e &&
+			buffer[3] === 0x47 &&
+			buffer[4] === 0x0d &&
+			buffer[5] === 0x0a &&
+			buffer[6] === 0x1a &&
+			buffer[7] === 0x0a
+		);
+	}
+
+	// Get the file path for the splash image for the underlying
+	// system
+	private async getSplashPath(): Promise<string> {
+		// TODO: this is not perfect, if the file was supposed to be
+		// resin-logo.png and for some reason is not present on the folder
+		// the supervisor will try to write new images from the API into
+		// balena-logo.png to no effect. Ideally this should be configurable
+		// as a constant that is provided on supervisor container launch.
+		const [
+			// If no logo is found, assume the file is `balena-logo.png`
+			splashFile = path.join(SplashImage.BASEPATH, 'balena-logo.png'),
+		] = (
+			await Bluebird.resolve(fs.readdir(SplashImage.BASEPATH))
+				// Read the splash dir (will throw if the path does not exist)
+				// And filter valid filenames
+				.filter((filename) => SplashImage.FILENAMES.includes(filename))
+		)
+			// Sort by name, so in case both files are defined, balena-logo will
+			// be chosen
+			.sort()
+			// Convert to full path
+			.map((filename) => path.join(SplashImage.BASEPATH, filename));
+
+		return splashFile;
+	}
+
+	// Returns the base64 contents of the splash image
+	private async readSplashImage(where?: string): Promise<string> {
+		// Read from defaultPath unless where is defined
+		where = where ?? (await this.getSplashPath());
+
+		// read the image file...
+		return (await fs.readFile(where)).toString('base64');
+	}
+
+	// Write a splash image provided as a base64 string
+	private async writeSplashImage(image: string, where?: string): Promise<void> {
+		// Write to defaultPath unless where is defined
+		where = where ?? (await this.getSplashPath());
+
+		const buffer = Buffer.from(image, 'base64');
+		if (this.isPng(buffer)) {
+			// Write the buffer to the given location
+			await remountAndWriteAtomic(where, buffer);
+		} else {
+			throw new Error('Splash image should be a base64 encoded PNG image');
+		}
+	}
+
+	private stripPrefix(name: string): string {
+		if (!name.startsWith(SplashImage.PREFIX)) {
+			return name;
+		}
+		return name.substr(SplashImage.PREFIX.length);
+	}
+
+	public createConfigVarName(name: string): string {
+		return `${SplashImage.PREFIX}${name}`;
+	}
+
+	public async getBootConfig(): Promise<ConfigOptions> {
+		try {
+			const defaultImg = await this.readSplashImage(SplashImage.DEFAULT);
+			const img = await this.readSplashImage();
+
+			// If the image is the same as the default image
+			// return nothing
+			if (img !== defaultImg) {
+				return {
+					image: `data:image/png;base64,${img}`,
+				};
+			}
+		} catch (e) {
+			log.warn('Failed to read splash image:', e);
+		}
+		return {};
+	}
+
+	public async initialise(): Promise<SplashImage> {
+		try {
+			await super.initialise();
+
+			// The default boot image file has already
+			// been created
+			if (await fs.exists(SplashImage.DEFAULT)) {
+				return this;
+			}
+
+			// read the existing image file...
+			const image = await this.readSplashImage();
+
+			// write the image to the DEFAULT path
+			await this.writeSplashImage(image, SplashImage.DEFAULT);
+
+			log.success('Initialised splash image backend');
+		} catch (error) {
+			log.warn('Could not initialise splash image backend', error);
+		}
+		return this;
+	}
+
+	public ensureRequiredConfig(_deviceType: string, conf: ConfigOptions) {
+		// If the value from the cloud is empty, it is the same as no definition
+		if (
+			!_.isUndefined(conf.image) &&
+			_.isEmpty((conf.image as string).trim())
+		) {
+			delete conf.image;
+		}
+		return conf;
+	}
+
+	public isSupportedConfig(name: string): boolean {
+		return SplashImage.CONFIGS.includes(this.stripPrefix(name).toLowerCase());
+	}
+
+	public isBootConfigVar(name: string): boolean {
+		return SplashImage.CONFIGS.includes(this.stripPrefix(name).toLowerCase());
+	}
+
+	public async matches(_deviceType: string): Promise<boolean> {
+		// all device types
+		return true;
+	}
+
+	public processConfigVarName(name: string): string {
+		return this.stripPrefix(name).toLowerCase();
+	}
+
+	public processConfigVarValue(
+		_name: string,
+		value: string,
+	): string | string[] {
+		// check data url regex
+		const matches = value.match(SplashImage.DATA_URI_REGEX);
+		if (!_.isNull(matches)) {
+			const [, media, data] = matches;
+			const [type] = media.split(';'); // discard mediatype parameters
+
+			return `data:${type};base64,${data}`;
+		}
+
+		if (!_.isEmpty(value.trim())) {
+			// Assume data is base64 encoded. If is not, setBootConfig will fail
+			return `data:image/png;base64,${value}`;
+		}
+		return value.trim();
+	}
+
+	public async setBootConfig(opts: ConfigOptions): Promise<void> {
+		// If no splash image is defined, revert to the default splash image
+		const value = !_.isEmpty(opts.image)
+			? (opts.image as string)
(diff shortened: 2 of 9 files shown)


src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+			buffer[7] === 0x0a
+		);
+	}
+
+	// Returns the base64 contents of the splash image
+	private async readSplashImage(where?: string): Promise<string> {
+		const [
+			// If no logo is defined, assume the file `balena-logo.png`
+			defaultPath = path.join(SplashImage.BASEPATH, 'balena-logo.png'),
+		] = await Bluebird.filter(SplashImage.PATHS, (logo) => fs.exists(logo));
```
20k-ultra, line 53: We could perform 1 IO operation if we check the contents of BASEPATH then filter on the files.

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export class SplashImage extends ConfigBackend {
+	private static readonly BASEPATH = path.join(bootMountPoint, 'splash');
+	private static readonly DEFAULT = path.join(
+		SplashImage.BASEPATH,
+		'balena-logo-default.png',
+	);
+	private static readonly PATHS = [
+		'balena-logo.png',
+		'resin-logo.png',
```
20k-ultra, line 24: Is this some legacy support ? We want to introduce any configurations that reference resin anymore.

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+	].map((logo) => path.join(SplashImage.BASEPATH, logo));
+	private static readonly PREFIX = `${constants.hostConfigVarPrefix}SPLASH_`;
+	private static readonly CONFIGS = ['image'];
+
+	// Check the header of the image for a png
+	// https://github.com/sindresorhus/is-png/blob/master/index.js
```
20k-ultra, line 30: Adding this citation is nice. We must not take pieces from modules even if it's small without mentioning it.

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+		// read the image file...
+		try {
+			const buffer = await fs.readFile(where);
+
+			return buffer.toString('base64');
+		} catch (err) {
+			log.error('Failed to read splash image.', err);
+		}
+
+		return '';
```
20k-ultra, line 67: I'm in favour of throwing an exception when we fail to read the splash image. This is because the backend has a few places where it does `!_.isEmpty(image)` and I those can be avoided if we just throw an exception and catch it where ever needed.

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+			buffer[3] === 0x47 &&
+			buffer[4] === 0x0d &&
+			buffer[5] === 0x0a &&
+			buffer[6] === 0x1a &&
+			buffer[7] === 0x0a
+		);
+	}
+
+	// Returns the base64 contents of the splash image
+	private async readSplashImage(where?: string): Promise<string> {
```
20k-ultra, line 49: I personally think a function so critical shouldn't have to worry about validating the parameters so making the where param not optional so that this logic can be handled else where makes this function cleaner.

20k-ultra, line 49: I'm in favour of making the where parameter required to simplify the function.

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+
+			return buffer.toString('base64');
+		} catch (err) {
+			log.error('Failed to read splash image.', err);
+		}
+
+		return '';
+	}
+
+	private async writeSplashImage(image: string, where?: string): Promise<void> {
```
20k-ultra, line 70: Same thing as the read, I'm in favour of requiring a write parameter.

src/config/backends/splash.ts
20k-ultra, line 24: Is this to support a legacy image ?

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+
+		if (!_.isEmpty(image)) {
+			try {
+				const buffer = new Buffer(image, 'base64');
+
+				if (this.isPng(buffer)) {
+					// Write the buffer to the given location
+					await remountAndWriteAtomic(where, buffer);
+				}
+			} catch (e) {
```
20k-ultra, line 87: I'm in favour of throwing the error up so the function calling can decide what to do. The way it's written now `writeSplashImage` writes an image and decides how failures are handled (just log a message).

src/config/backends/index.ts
```diff
@@ -5,13 +5,15 @@ import { ExtraUEnv } from './extra-uEnv';
 import { ConfigTxt } from './config-txt';
 import { ConfigFs } from './config-fs';
 import { Odmdata } from './odmdata';
+import { SplashImage } from './splash';
```
20k-ultra, line 8: I think changing the backend to splash-image makes it more clear that this backend only supports images in case splash could be interpreted as something else.

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
...
+		const image = await this.readSplashImage();
+
+		// If the the /mnt/boot splash exists it will set the correct
+		// options
+		if (!_.isEmpty(image)) {
+			options['image'] = image;
+		}
+
+		return options;
+	}
```
20k-ultra, line 117: If we throw the exceptions up then we can rewrite functions like this :p

```suggestion
	public async getBootConfig(): Promise<ConfigOptions> {
		try {
			return {
				image: await this.readSplashImage()
			}
		} catch (e) {
			// Return empty config
			return {};
		}
	}
```

src/config/backends/splash.ts
```diff
@@ -0,0 +1,187 @@
+import * as Bluebird from 'bluebird';
+import * as _ from 'lodash';
+import { fs } from 'mz';
```
20k-ultra, line 3: See if you can use the native fs.Promises like in https://github.com/balena-io/balena-supervisor/blob/0c7cb46d1551fb92ae7dbfe9f303181e72d739b7/src/config/backends/odmdata.ts#L2

20k-ultra (commented):
pretty good so far

src/config/backends/splash.ts
pipex: Makes sense

src/config/backends/splash.ts
pipex: Yeah, I probably will end up adding the dependency

src/config/backends/splash.ts
pipex: yep

src/config/backends/splash.ts
pipex: What is the advantage @20k-ultra ?

src/config/backends/splash.ts
pipex: yeah, that looks good

src/config/backends/splash.ts
20k-ultra: The mz module was made to bring new ECMAScript features to nodejs faster then the node project implements them. One example being Promise based fs module. Since the standard lib supports Promises in the fs module theirs no need to use mz for fs. I just checked out the repo too and it was last updated in 2017 so maybe all the features it meant to modernize have been implemented.

I'm also in favour of always trying to remove dependencies :p

pipex (commented):
No way to cancel a review?

src/config/backends/splash.ts
pipex: I'm for removing dependencies, but I didn't end up making this change since it will mess with the tests as well since other backends use 'mz' as dependency as well. I will create an issue to study the removal of the package

test/12-device-config.spec.ts
```diff
@@ -171,6 +174,54 @@ describe('Device Backend Config', () => {
 		(child_process.exec as SinonStub).restore();
 	});
 
+	it('ensures required fields are written to config.txt', async () => {
```
20k-ultra, line 177: This file places any tests that use a backend into their own suite except for config-txt. The first 5 cases in this file are using config-txt. I think it would be nice to tidy it up and put them into their own suite like the other backend specific device config cases

pipex: I don't love this either, but the counterargument I see is, these are more like integration tests from the device config side of the code, where the backend tests are checking the actual backend behavior. 

There probably should be a test for the config backend, then these tests could become much smaller, but probably not going to address that in this PR.

src/config/backends/splash-image.ts
```diff
@@ -0,0 +1,196 @@
...
+			// been created
+			if (await fs.exists(SplashImage.DEFAULT)) {
+				return this;
+			}
+
+			// read the existing image file...
+			const image = await this.readSplashImage();
+
+			// write the image to the DEFAULT path
+			await this.writeSplashImage(image, SplashImage.DEFAULT);
```
20k-ultra, line 127: Does this imply that there is always an image on the OS ? I am understanding this as..

1. run readSplashImage 
 a. No where passed so get the where value from getSplashPath function
 b. getSplashPath to me just returns balena-logo.png unless resin-logo.png exists so let's say it returns balena-logo.png
 c. balena-logo.png and now written to balena-logo-default.png

src/config/backends/splash-image.ts
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	// Get the file path for the splash image for the underlying
+	// system
+	private async getSplashPath(): Promise<string> {
+		const [
+			// If no logo is defined, assume the file is `balena-logo.png`
+			splashFile = path.join(SplashImage.BASEPATH, 'balena-logo.png'),
+			// Read the splash dir (will throw if the path does not exist)
+		] = await Bluebird.resolve(fs.readdir(SplashImage.BASEPATH))
```
20k-ultra, line 52: This function to me seems like it is basically just returning balena-logo.png unless resin-logo.png exists. This is because the splashFile variable defaults to balena-logo.png and then loads the files in the directory. These files are filtered against the FILENAMES array which is balena-logo.png and resin-logo.png. 

If balena-logo.png exists then splashFile was already set to that so it is now set to the same value. If the files array contains resin-logo.png then that is set to the splashFile. If the directory contains both then splashFile is whichever file is at index 0 which I'd guess is alphabetical order (I just did ls -l and my system lists in alphabetical order).

If this is correct then it should just check if resin-logo.png exists else return balena-logo.png to simplify it.

src/config/backends/splash-image.ts
pipex: Yes, there should always be a resin-logo.png or balena-logo.png on that directory. If there is not, the initialization will fail. On initialization, the backend makes a copy so it can remember the initial value when the the variable is unset. I actually haven't tested that case on a device, it might be checking that no weird loop occurs.

src/config/backends/splash-image.ts
pipex: No, that is not correct. The function looks at the directory and checks that either of the files in `FILENAMES` exist
- If both exist, it will return the first (which might be an error, since it might be implementation or OS dependent)
- If only one exists, it returns that file
- If none exist, assume is `balena-logo.png`.

A better way would probably be to check os version and always use `balena-logo` if >= 2.52 and `resin-logo` if < 2.52

src/config/backends/splash-image.ts
```diff
@@ -0,0 +1,196 @@
...
+
+			log.success('Initialised splash image backend');
+		} catch (error) {
+			log.warn('Could not initialise splash image backend', error);
+		}
+		return this;
+	}
+
+	public ensureRequiredConfig(_deviceType: string, conf: ConfigOptions) {
+		// If the value from the cloud is empty, it is the same as no definition
```
20k-ultra, line 137: This is actually a very useful thing to have in the backends! I might use this for the ODMDATA backend when there's no mode specified.

src/config/backends/splash-image.ts
```diff
@@ -0,0 +1,196 @@
...
+		where = where ?? (await this.getSplashPath());
+
+		// read the image file...
+		return (await fs.readFile(where)).toString('base64');
+	}
+
+	// Write a splash image provided as a base64 string
+	private async writeSplashImage(image: string, where?: string): Promise<void> {
+		// Read from defaultPath unless where is defined
+		where = where ?? (await this.getSplashPath());
```
20k-ultra, line 73: If resin-logo.png exists then images will continue to write to that location. We could take this opportunity to move on from that name and only write to balena-logo.png.

20k-ultra (commented):
Just had some clarifications I wanted to ask. Overall nothing major.

src/config/backends/splash-image.ts
pipex: It depends on the OS, newer versions of the OS support `balena-logo`, but we might want to have the supervisor to still work with older versions

src/config/backends/splash-image.ts
20k-ultra: Yeah, give that a test. Remember if the OS changes something beneath us we want to be able to easily observe what is causing the weird behavior. The logging in the backend has done a good job of that so far though.

src/config/backends/splash-image.ts
20k-ultra: Also noticed the comment says read from defaultPath which it should say Write to defaultPath :p

src/config/backends/splash-image.ts
20k-ultra: To me that sounds just like if resin-logo.png exists return that else return balena-.logo.png.

src/config/backends/splash-image.ts
20k-ultra: Adding the version check you mentioned would be useful then so we have something in place that is actively moving away from resin naming conventions.

20k-ultra: "A better way would probably be to check os version and always use balena-logo if >= 2.52 and resin-logo if < 2.52"

src/config/backends/splash-image.ts
pipex: no, because if both resin-logo and balena-logo exist you should default to balena-logo. I agree is confusing though

src/config/backends/splash-image.ts
pipex: This is the output of the supervisor in the case the directory or none of the files exist. I think it's acceptable

```
[debug]   Finished applying target state
[success] Device state apply success
[info]    Applying target state
[warn]    Could not initialise splash image backend Error: ENOENT: no such file or directory, open '/mnt/root/mnt/boot/splash/balena-logo.png'
[warn]    Failed to read splash image: Error: ENOENT: no such file or directory, open '/mnt/root/mnt/boot/splash/balena-logo-default.png'
[warn]    Could not initialise splash image backend Error: ENOENT: no such file or directory, open '/mnt/root/mnt/boot/splash/balena-logo.png'
[debug]   Finished applying target state
[success] Device state apply success
```

src/config/backends/splash-image.ts
pipex: I changed my mind on this. Checking os versions goes against the idea of the supervisor not being tied to balenaOS. Probably the best option would be to specify the splash image as a variable that can be modified on the container start. But probably not going to happen at this time. 

I am modifying the code so the result of readdir is sorted, to avoid any implementation traps.

src/config/backends/splash-image.ts
20k-ultra: Right, I remember stepping through the code and reading that exact situation actually. That is enough information for someone to know the next step is to figure out what that file is and why it's missing.

src/config/backends/splash-image.ts
20k-ultra: Oh I understand the distinction now...this wants to use balena-logo by default AND choose that over resin-logo if they both exist. However, the mechanism for choosing balena-logo over resin-logo isn't predicable or conveyed in the code well.

If you add the sort then that does fix that though!

20k-ultra:
@balena-ci retest

src/config/backends/splash-image.ts
20k-ultra: no longer needed since we are sorting the files to always use balena-logo if it exists thus achieving the affect I wanted here.

20k-ultra (approved):

PR #1572, opened 2021-01-09
Reset update delay when request is from API

20k-ultra:
The fix itself isn't much but the testing was really time consuming...there are a few unfortunate exports added to make the testing more sane but open to hearing any ideas how to prevent it (I really doubt there is). Added testing for the API and the module itself.


20k-ultra:
poke @pipex

20k-ultra:
@balena-ci retest

20k-ultra:
interesting... let's see if the tests fail again. On my local machine they pass.

```
  412 passing (1m)
  21 pending
  1 failing

  1) Splash image configuration
       setBootConfig
         should write the given image to resin-logo.png if set:
     AssertionError: expected readdir to have been called exactly once, but it was called twice
    readdir(test/data/mnt/boot/splash) => [Promise] {  } at unknown
    readdir(test/data/mnt/boot/splash) => [Promise] {  } at unknown
      at Context.<anonymous> (build/test/43-splash-image.spec.js:128:52)



npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! balena-supervisor@12.3.2 test-nolint: `npm run test:build && TEST=1 mocha`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the balena-supervisor@12.3.2 test-nolint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2021-01-14T21_33_32_362Z-debug.log
The command '/bin/sh -c npm run test-nolint 	&& npm run build' returned a non-zero code: 1

Exited with code exit status 1

CircleCI received exit code 1
```

20k-ultra:
I went to circleci and re-ran the tests because balena ci wasn't doing it. The tests are passing now without any changes.

pipex:
Sorry @20k-ultra, reviewing now

pipex (commented):
I'm not entirely sure I like the solution @20k-ultra, I thought when you talked about resetting the timers, you meant actually set them to 0 so that the state check happened inmediately. This essentially makes the supervisor re-check the target state (even if it hasnt changed) on each call to `/v1/update`. I'm wondering if that might have unintended consequences>

src/api-binder.ts
```diff
@@ -46,7 +46,8 @@ interface DeviceTag {
 	value: string;
 }
 
-let readyForUpdates = false;
+// Export for testing
+export let readyForUpdates = false;
```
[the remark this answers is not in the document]
20k-ultra: I think I was stuck in the moment of working on this feature without realizing I could just wrap that var in a function that just returns its value. The var remains in the module but now we have a function we can export and stub the results of. Just modify the remainder of the code to use that function to access the var.  The result is a var that is only write-able in the module but readable from external modules.

This approach is how I plan to implement any kind of state within modules purely for the fact it makes testing so much easier.

src/device-state/target-state.ts
```diff
@@ -127,6 +127,13 @@ export const update = async (
 			);
 		}
 
+		// If this update request came from the API reset emitted to cancel
+		// any processes that may be already applying the cached target release
+		// such as the back off timeout for when update locks are detected
+		if (isFromApi) {
```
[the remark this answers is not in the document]
20k-ultra: There is code that already exists which will cancel and immediately try to apply the target state again. It was just never used to my knowledge and this change uses that. I think this code was intended to always exist tbh since the apply target state cancellation seems to never be called.

src/device-state.ts
```diff
@@ -224,7 +224,8 @@ type DeviceStateStep<T extends PossibleStepTargets> =
 let currentVolatile: DeviceReportFields = {};
 const writeLock = updateLock.writeLock;
 const readLock = updateLock.readLock;
-let maxPollTime: number;
+// Exported for tests
+export let maxPollTime: number;
```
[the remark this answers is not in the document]
20k-ultra: I can wrap this in a get function as I describe in the other comment.

src/device-state/target-state.ts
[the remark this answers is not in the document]
20k-ultra: This doesn't make the `statusCode === 304` as it still compares the local target state and the one the server has.

20k-ultra:
> This makes the check for statusCode === 304 below redundant, essentially re-emitting the target state for every api request. 

Every time the [update](https://github.com/balena-io/balena-supervisor/blob/e6ebfca0b93ad5992ba87d4d2edaed0257ac48c2/src/device-state/target-state.ts#L105) function is called which is on every poll interval and `/v1/update` request the device will ask the cloud for the target state with an etag in the headers so that the server can respond with a 304 if they are the same. Setting emitted to false if isFromApi doesn't prevent the device from asking the API for target state. However, the state is emitted but that's how the mechanism for cancelling currently running updates works. 

> This essentially makes the supervisor re-check the target state (even if it hasnt changed) on each call to `/v1/update`.

If you start a delay by adding a lock for example and trying to apply an update, each time the function executes again (reaches 0 seconds) it starts from the very beginning of this whole process (minus asking the API for target state). This PR doesn't change that behavior.

https://github.com/balena-io/balena-supervisor/blob/e6ebfca0b93ad5992ba87d4d2edaed0257ac48c2/src/device-state.ts#L718-L724

TL;DR the only behavior change this has other then immediately retrying to apply the target state as if the delay has reached 0 seconds is that the same target state would be emitted using `target-state-update` event which per the name used that is not correct as there was no update to the target. The only issue I could see is if someone adds a non-idempotent function to execute on that event.

I might have found a way though to fix the issue and not emit the target state.

20k-ultra:
The way I wanted to fix this used the deviceState module which we don't want to include in the target state module.

20k-ultra:
@pipex so I was able to not need the export on the variables by using a module. I'd check it out because it was very useful!

Additionally, I refactored the fix to not emit a `target-state-update`. Instead, it has a new event `target-state-apply` which you can see the logic for is contained in the same `emitTargetState` function so I think that's a pretty clean approach.

Thank you for your previous review as it made this PR much better :)

test/lib/mocked-device-api.ts
```diff
@@ -186,12 +188,15 @@ function buildRoutes(): Router {
...
 
+// TO-DO: Create a cleaner way to restore previous values.
 const originalNetGetAll = networkManager.getAllByAppId;
 const originalVolGetAll = volumeManager.getAllByAppId;
 const originalSvcGetAppId = serviceManager.getAllByAppId;
 const originalSvcGetStatus = serviceManager.getStatus;
+const originalReadyForUpdates = apiBinder.__get__('readyForUpdates');
 
 function setupStubs() {
+	apiBinder.__set__('readyForUpdates', true);
```
pipex, line 199: I like this way to stub private variables with rewire

src/device-state/target-state.ts
```diff
@@ -76,16 +76,24 @@ const emitTargetState = (
...
 	) {
+		// CachedResponse has not been emitted before so emit as an update
 		emitter.emit(
 			'target-state-update',
 			_.cloneDeep(cachedResponse.body),
 			force,
 			isFromApi,
 		);
-
+		return true;
```
pipex, line 86: Don't think I'm being overcritical or overthinking stuff. Couldn't we get the same objective by emitting a `fresh` flag with the `target-state-update` event? Something like

```
const emitTargetState = (
	cachedResponse: CachedResponse,
	force = false,
	isFromApi = false,
): boolean => {
	if (
		!cachedResponse.emitted &&
		emitter.listenerCount('target-state-update') > 0
	) {
		// CachedResponse has not been emitted before so emit as an update
		emitter.emit(
			'target-state-update',
			_.cloneDeep(cachedResponse.body),
			force,
			isFromApi,
                        true // is fresh
		);
		return true;
	} 
        emitter.emit(
	     'target-state-update',
	     _.cloneDeep(cachedResponse.body),
	     force,
	     isFromApi,
             !!cache.emitted // fresh will be false if we already emitted
        );
       
	// Return the same emitted value but normalized to be a boolean
	return !!cache.emitted;
};
```

And then in the handler do

```
if (fresh) {
    // Only save fresh target
    await deviceState.setTarget(targetState);
}
deviceState.triggerApplyTarget({ force, isFromApi });
```

If just feel that the target-state module having to care whether the request comes from the API or not and know not to emit the target state is getting very complex. What do you think?

pipex: I mean this is fine, and probably we should not keep spending time on this. We can make a note to revisit this module to clean it up eventually

pipex (approved):
I don't want to keep delaying this fix. Let's make a note to revisit the target state module soon so we can move on to other stuff. LGTM!

src/device-state/target-state.ts
20k-ultra: I'm totally fine with discussing better ways to write the fix :p 

This is nice because we don't have to add a new emit type but then it sounds redundant. The name of this event implies that the data is fresh since it contains the word "update". I think a good solution is changing the event name to `target-state-notify` which then a listener can apply some business logic and emit `target-state-update` to run deviceState.setTarget and `target-state-apply` to run deviceState.triggerApplyTarget. 

It sounds like we just made it more complicated but it's more flexible now and easier to write new functionality that listens on specific events. Plus the event names are 100% exactly what's happening and not a combination of actions.

pipex: I think of it from the point of view of the single responsibilty principle. The target state module has the responsibility to query the target state endpoint, and report the results. It should not care where the request came from, it probably shouldn't even care if there is someone listening. That last one something I added to fix #1455, but now feels like a hack as well. Having the module now be aware of whether an target state should be applied or just notified feels outside scope.

I think this partly has to do with the supervisor design, where you start a bunch of services performing actions on a timer and sending messages and have sometimes to perform extra checks to ensure that events don't catch you in an inconsistent state. That is probably something we could look to improve in the medium term.

Issue #1581, opened 2021-01-27
Update SV from v7.16.6 to v12.3.0 - sv then fails to start after hup


acostach:
Steps:

1) Provision a Pi3 to version 2.15.1 on an App that has no releases
2) Execute Update OS in the dashboard, to the latest OS as of now, 2.67.3

After the update, the supervisor fails to start properly and we see in the logs:

```
an 27 11:02:08 6be700d resin-supervisor[1917]: resin_supervisor
Jan 27 11:02:08 6be700d resin-supervisor[1949]: active
Jan 27 11:02:13 6be700d resin-supervisor[1951]: Container config has not changed
Jan 27 12:32:26 6be700d resin-supervisor[2165]: [info]    Supervisor v12.3.0 starting up...
Jan 27 12:32:26 6be700d resin-supervisor[2165]: migration file "M00006.js" failed
Jan 27 12:32:26 6be700d resin-supervisor[2165]: migration failed with error: insert into `currentCommit` (`appId`, `commit`) values (121528, NULL) - SQLITE_CONSTRAINT: NOT NULL constraint failed: currentCommit.commit
```

```
> db.all('SELECT * FROM config;', console.log);
Database { open: true, filename: '/data/database.sqlite', mode: 65542 }
> null [
  {
    key: 'apiSecret',
    value: 'XXXXXX'
  },
  { key: 'initialConfigSaved', value: 'true' },
  {
    key: 'initialConfigReported',
    value: 'https://api.balena-staging.com'
  },
  { key: 'name', value: 'wispy-voice' },
  { key: 'deltaVersion', value: '3' },
  { key: 'currentCommit', value: null }
]

Discussed internally at: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/GkTxxg72dTIVw1WrvEx4t6SgmiC
```

PR #1584, opened 2021-02-03
Prevent inserting null commit during DB migration

20k-ultra:
The migration checks if the data it is about to insert is not null but the Supervisor creates a object with a null commit inside. This PR just extends the check to make sure the commit value being inserted isn't null.

Change-type: patch
Closes: #1581


20k-ultra:
@balena-ci I self-certify!

PR #1589, opened 2021-02-09
Complete V1 unit tests, modify PATCH /v1/device/host-config

cywang117:
Complete V1 unit tests for the following endpoints (as mentioned in [this issue](https://github.com/balena-io/balena-supervisor/issues/1327)):  
- GET/V1/device/host-config
- PATCH /V1/device/host-config
- POST /V1/blink
- POST /V1/regenerate-api-key
- POST /V1/purge

While writing unit tests, discovered that PATCH /v1/device/host-config implementation for updating host proxy configs behaved more like a PUT, in that all proxy configs were replaced with PATCH body, so I tweaked the implementation to allow partial PATCH requests, & wrote matching unit tests. 

Additionally, added user input validation for this endpoint to ensure the user gets proper feedback on malformed or invalid request bodies.


pipex (commented):
_Editing this because I thought I could comment on a particular commit_

For your last commit `Merge branch 'master' ...`

We generally try to avoid this kind of commit, because it adds noise in the git history. 

For updating the branch to master you should use `git rebase` instead of `git merge`

Not sure if you saw this already https://github.com/balena-io/process/blob/master/process/Commit_and_PR_Guidelines.md#pull-request-guidelines but it might be handy

To remove this from the PR you can reset to the previous commit, do a rebase and then use `git push --force-with-lease origin v1-unit-tests`

I will be adding more to the PR review during this week, this is just the first thing that called my attention. On a different note: love the very tidy commits!

cywang117:
> For your last commit `Merge branch 'master' ...`
> 
> We generally try to avoid this kind of commit, because it adds noise in the git history.
> 
> For updating the branch to master you should use `git rebase` instead of `git merge`
> 
> Not sure if you saw this already https://github.com/balena-io/process/blob/master/process/Commit_and_PR_Guidelines.md#pull-request-guidelines but it might be handy

Hey Felipe, thanks for bringing this to my attention, all fixed now :)

src/device-state.ts
```diff
@@ -128,14 +128,64 @@ function createDeviceStateRouter() {
...
-			.patch(req.body)
-			.then(() => res.status(200).send('OK'))
-			.catch((err) =>
-				res.status(503).send(err?.message ?? err ?? 'Unknown error'),
-			),
-	);
+	router.patch('/v1/device/host-config', async (req, res) => {
+		// Validate user input, returning 400 with malformed PATCH body
+		try {
+			if (!req.body.network) {
```
pipex, line 134: I like these new checks, I'm wondering if they should go in the `host-config` module, since they seem very specific to that. The same with the defined constants. What do you think @20k-ultra ?

src/lib/constants.ts
```diff
@@ -69,6 +69,15 @@ const constants = {
 	// (this number is used as an upper bound when generating
 	// a random jitter)
 	maxApiJitterDelay: 60 * 1000,
+	validRedsocksProxyTypes: ['socks4', 'socks5', 'http-connect', 'http-relay'],
```
pipex, line 72: Not sure if there is an advantage to having these constants here. They are very specific to host config, so it might make sense to put them inside the module

pipex: Also ping @20k-ultra, this is related to my comment above on the API route

src/device-state.ts
```diff
@@ -128,14 +128,64 @@ function createDeviceStateRouter() {
...
+
+			// If hostname is an empty string, return first 7 digits of device uuid
+			if (hostname === '') {
+				const uuid = await config.get('uuid');
+				req.body.network.hostname = uuid?.slice(0, 7);
+			}
+
+			// Validate proxy fields, if they exist
+			if (proxy && Object.keys(proxy).length) {
+				// TODO: Should probably support changing local_port to any value
```
pipex, line 148: Restricting the allowed configurations might break things for users @cywang117, what is your reasoning behind this? I found at least one instance where a user was using `local_ip` https://forums.balena.io/t/redsocks-proxy-not-working/220377

Unless there is a good reason I would allow everything redsocks allows. If we know redsocks doesn't crash with unknown fields we might not bother checking at all, since this is something we will need to keep maintaining if the OS redsocks version changes.

pipex: If we end up removing the validation, you can disregard my previous comment about moving the checks to `host-config`

test/41-device-api-v1.spec.ts
```diff
@@ -615,5 +628,593 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 		});
 	});
 
-	// TODO: add tests for V1 endpoints
+	describe('POST /v1/blink', () => {
```
pipex, line 631: We're getting close to 100% testing of the supervisor API 🎉

test/data/mnt/boot/system-proxy/redsocks.conf
```diff
@@ -10,8 +10,8 @@ redsocks {
     local_ip = 127.0.0.1;
     local_port = 12345;
     ip = example.org;
-	port = 1080;
+    port = 1080;
     type = socks5;
     login = "foo";
-	password = "bar";
+    password = "bar";
```
pipex, line 16: I would have probably just push-force to my PR branch if I missed some validation, but I understand it may not feel very comfortable

test/41-device-api-v1.spec.ts
```diff
@@ -615,5 +623,44 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
...
+		// Further blink function-specific testing located in 07-blink.spec.ts
+		it('responds with code 200 and empty body', async () => {
+			await request
+				.post('/v1/blink')
+				.set('Accept', 'application/json')
+				.set('Authorization', `Bearer ${apiKeys.cloudApiKey}`)
+				.expect(sampleResponses.V1.POST['/blink'].statusCode)
+				.then((response) => {
+					expect(response.body).to.deep.equal(
+						sampleResponses.V1.POST['/blink'].body,
```
pipex, line 641: This test won't run without the responses from the other commit right? I would recommend squashing both commits so you get a single passing change,

.editorconfig
```diff
@@ -13,3 +13,6 @@ trim_trailing_whitespace = false
 
 [*.json]
 indent_style = space
+
```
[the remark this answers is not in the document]
pipex: If you can get .prettierrc to work as I suggested on flowdock maybe you could remove this commit. I agree that we should improve the module configuration so someone creating their first PR doesn't get caught in formatting issues. Maybe adding prettify to `lint-staged` @20k-ultra ?

test/41-device-api-v1.spec.ts
```diff
@@ -725,5 +728,123 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 		});
 	});
 
+	describe('GET /v1/device/host-config', () => {
```
pipex: I would squash this commit with 708761e so if you checkout to the commit you get the change and the tests

test/41-device-api-v1.spec.ts
```diff
@@ -728,121 +729,429 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
...
-						sampleResponses.V1.GET['/device/host-config'].body,
-					);
-				});
-		});
-	});
-
-	describe('PATCH /v1/device/host-config', () => {
-		// TODO: add tests for proxy PATCH reqs - probably need to add test files for redsocks.conf or other
-
+	describe('/v1/device/host-config', () => {
```
pipex, line 733: Same here, is probably better to squash this commit with 18f03a0

test/41-device-api-v1.spec.ts
```diff
@@ -841,15 +841,15 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 				configSetStub = stub(config, 'set').callsFake(configSetFakeFn);
 			});
 
-			after(async () => {
+			after(() => {
```
pipex, line 844: You can also probably squash these changes into 18f03a0

pipex (changes requested):
Very nice first PR @cywang117. Most my observations are minor, suggestions about commit squashing are more about keeping things tidy.

src/device-state.ts
cywang117: Or, perhaps these v1 routes should be moved to device-api/v1.ts as commented in the [TODO](https://github.com/balena-io/balena-supervisor/blob/2fa8156b31c60d56d07bee28bd6ff5fab8cfece2/src/device-state.ts#L92) in this file? Logically grouping the routes together makes sense, and it would make it easier for devs new to the codebase to get acclimated.

.editorconfig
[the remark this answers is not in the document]
cywang117: Yeah, the suggestion of adding `.prettierrc` is gold, since `.editorconfig` doesn't actually tell the IDE to format things in the same way as is done when running `prettify`.  If it were included in the README under development, or just as a footnote in a potential "Intro to Supervisor" Q&A gist, that would be great.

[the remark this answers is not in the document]
cywang117: @pipex  Agreed, will remove this commit pending comment from @20k-ultra (since I'm not sure if comments on deleted commits remain in GH -- does anyone know?)

src/device-state.ts
cywang117: @pipex I was simply building on top of legacy code, [which sets `local_ip` and `local_port` as static fields](https://github.com/balena-io/balena-supervisor/blob/2fa8156b31c60d56d07bee28bd6ff5fab8cfece2/src/host-config.ts#L19). 

However I do agree that we should support all fields being changed. I figured it would make for a separate PR, but forgot to mention the topic. Since we are implementing feature parity on V2 from V1, this issue would be relevant up to and possible after V3 comes out. I've raised an issue [here - TODO]().

I haven't researched into redsocks deeply enough to know 100%, but AFAIK unknown fields would be ignored. There are also a lot more config options for redsocks that are not supported here, even if we allow unknown fields. The top one that comes to mind is [support for `redudp`](https://github.com/darkk/redsocks/blob/master/redsocks.conf.example#L117), which would be a separate bracket in the `.conf` file. Should we enable support for this sort of thing as well?

I do have more context-gathering questions for redsocks-related things, but I'll send them over in the FD supervisor thread.

src/lib/constants.ts
cywang117: @pipex While I do agree that these constants might be minor compared to some of the other ones in `constants.ts`, as a developer I (and potentially others) would feel more comfortable if I could reference and change constants from a single source of truth, instead of a line at the top of a module. 

In my opinion, it feels a bit more organized this way. There's also certainly no disadvantage to having constants stored in a single location that I'm aware of, unless I don't have the context for it. Thoughts?

test/data/mnt/boot/system-proxy/redsocks.conf
cywang117: Will do! I will say, I've had nightmares of accidentally force-pushing to production 😰

test/41-device-api-v1.spec.ts
cywang117: Done! (pending push)

src/host-config.ts
```diff
@@ -132,11 +139,23 @@ async function setProxy(maybeConf: ProxyConfig | null): Promise<void> {
 		const conf = maybeConf as ProxyConfig;
 		await mkdirp(proxyBasePath);
 		if (_.isArray(conf.noProxy)) {
-			await writeFileAtomic(noProxyPath, conf.noProxy.join('\n'));
+			conf.noProxy.length
+				? await writeFileAtomic(noProxyPath, conf.noProxy.join('\n'))
```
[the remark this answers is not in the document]
cywang117: Makes sense & saves a conditional check, especially if we're filtering empty strings from noProxy for the GET requests to this same endpoint -- I'll change it back.

test/41-device-api-v1.spec.ts
cywang117: Done!

test/data/mnt/boot/system-proxy/redsocks.conf
cywang117: Squashed into 12ce1fd2c23f606040afcc27f2b1e4c5dfd48590

test/41-device-api-v1.spec.ts
cywang117: Squashed into 12ce1fd2c23f606040afcc27f2b1e4c5dfd48590

test/41-device-api-v1.spec.ts
cywang117: & pushed :)

test/41-device-api-v1.spec.ts
cywang117: Done!

src/device-state.ts
pipex: You're right, I hadn't seen that part of the code before. Maybe blacklist proxy fields instead of whitelisting is the option?

> I was simply building on top of legacy code, which sets local_ip and local_port as static fields.

No, let's keep it as is for now, for that sort of change is better to go through a product discussion and for that is better to have user evidence, otherwise we end up with extra code we will forced to maintain.
> Should we enable support for this sort of thing as well?

src/lib/constants.ts
pipex: I agree that both can work, having it all contained has the advantage IMO that you don't have to go looking for the variable values elsewhere, and it makes it easy to split up the module if you eventually want to. 

Personally I think of constants.ts as a place to store values that will be used across the codebase, if the value is only used by a single module then i think it makes more sense to have the values in the module.

.editorconfig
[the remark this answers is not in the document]
pipex: Not sure, but you can remove the commit if you want, we can always take the conversation back to FD

src/device-state.ts
pipex: We should definitely do that, but not on this PR, I am working on some other PR that touches this codebase and I don't want it to break 😅.

.editorconfig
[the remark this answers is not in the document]
cywang117: Comments stay, apparently. Commit removed 👍🏼

src/device-state.ts
cywang117: For sure! I will open an issue ^^

src/lib/constants.ts
cywang117: I understand your point -- that `constants.ts` should be for more widely used constants. However, I *have* noticed multiple other constants within `constants.ts` that are only used in 1 module, plus the tests for that module, similar to how often `validRedsocksProxyTypes` is used. Maybe it's the importance or influence of a module that determines whether its constants should go in `constants.ts`?

If we're just talking about `validV1PatchProxyFields`, the other constant added, then I'd agree about removing it. Having thought about the broader scope of SV, and the fact that v1 is up for deprecation, adding this field in constants doesn't really make sense, and overall v1 is legacy & therefore not as important for the future, I suppose. However, I'm inclined to want to keep validRedsocksProxyTypes where it is, if possible, since it will also be used in the upcoming v2 endpoints (see: [v2 feature parity](https://github.com/balena-io/balena-supervisor/issues/1586)). 

I don't mean to belabor the point, of course. Eager to hear your thoughts on this compromise :)

src/device-state.ts
cywang117: > Maybe blacklist proxy fields instead of whitelisting is the option?

Sorry, could you clarify? I thought 

```typescript
if (
	Object.keys(proxy).filter(
		(key) => !constants.validV1PatchProxyFields.includes(key),
	).length > 0
) {...}
```

functioned as blacklisting? ([device-state.ts line 151 in PR](https://github.com/balena-io/balena-supervisor/pull/1589/commits/84eee4381aea81f932c7417245a378c64a99e76f#diff-28bbeff4b860b9fa9095237759c4b8891ef72dd96acfed6c7241641c964439c4R150))

Unless you're talking about adding local_ip and local_port support as well?

pipex: If I understand correctly, that code checks that the provided values match the values on a 'valid field list', so is more whitelisting. Blacklisting would be something like

```
if (
	Object.keys(proxy).filter(
		(key) => constants.disalllowedV1PatchProxyFields.includes(key),
	).length > 0
) {...}
```

Does this make more sense?

src/lib/constants.ts
pipex: Yeah, I agree there is currently not much consistency in what goes into `constants.ts` right now, going forward I think it will be good if we all get to a common set of good practices we'd like to see on the supervisor codebase.

I don't want to extend the discussion too much either for this though, I think the compromise you propose is good.

src/device-state.ts
cywang117: Thanks, makes sense. I've updated the commit :)

src/host-config.ts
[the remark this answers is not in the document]
cywang117: Changed back

src/lib/constants.ts
cywang117: For sure! I've updated the commit.

src/device-state.ts
cywang117: @pipex  I've also changed the route from throwing errors on invalid fields to logging in the supervisor console, per our discussion yesterday. Let me know if it works for you.

src/device-state.ts
```diff
@@ -41,6 +41,8 @@ import {
 import * as dbFormat from './device-state/db-format';
 import * as apiKeys from './lib/api-keys';
 
+const disallowedPatchFields = ['local_ip', 'local_port'];
```
pipex, line 44: Could you move this to host-config.ts? It doesn't make much sense here.

src/host-config.ts
```diff
@@ -134,9 +146,19 @@ async function setProxy(maybeConf: ProxyConfig | null): Promise<void> {
 		if (_.isArray(conf.noProxy)) {
 			await writeFileAtomic(noProxyPath, conf.noProxy.join('\n'));
 		}
-		const redsocksConf = `${redsocksHeader}${generateRedsocksConfEntries(
-			conf,
-		)}${redsocksFooter}`;
+
+		let currentConf: ProxyConfig | undefined;
+		try {
+			currentConf = await readProxy();
```
pipex, line 152: I thought we were going to remove this new read? If this fails `currentConf` will be undefined which will make the object spread fail on line 159.

src/device-state.ts
```diff
@@ -41,6 +41,8 @@ import {
 import * as dbFormat from './device-state/db-format';
```
pipex, line 41: Maybe we can merge this commit with with fb0cc12. It seems like tests on that might not pass without some stuff on this commit?

pipex (changes requested):
Almost there :p

src/device-state.ts
pipex: This comment refers to the commit 5d55b87, there is no way to make a comment on a commit (that i have found)

src/device-state.ts
cywang117: This array is being used locally in this module, in `router.patch`. I've modified it to be this way since we decided to remove it from constants.ts.

src/host-config.ts
cywang117: Sorry, don't remember discussing this, but you're right, it's an oversight on my part.

cywang117: @pipex Actually, I did a lil sandbox test with undefined spread in TS, and learned that [a falsey spread will be skipped over in TS](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#conditional-spreads-create-optional-properties)

So even if `currentConf` is undefined, `generateRedsocksConfEntries({ ...currentConf, ...conf })` on line 159 below will not error & will simply skip over currentConf.

cywang117: [Here](https://repl.it/@cywang117/OverdueWelcomeBytecode#index.ts) is my sandbox for reference. You can run `node index.js` in the integrated shell to see what's happening.

Thanks for this by the way! I too thought it would error after looking at it again, but as a result learned about this cool feature of TypeScript.

src/device-state.ts
cywang117: Done!

src/host-config.ts
cywang117: And apparently this is just a JS feature! https://www.flowdock.com/app/rulemotion/r-beginners/threads/f_bsL91RV1Zx6FgcQ8-IAuYgWh8

src/device-state.ts
pipex: Then maybe we should call it something `disallowedHostConfigPatchFields`?  Since this is (for now) in the device state module it might be better to be more verbose.

src/host-config.ts
pipex: I saw that thread, TIL for me as well. This looks good, I think I was confusing this change with something else.

src/device-state.ts
pipex: Only this is missing and we can get this merged :)

cywang117: Done!

pipex (approved):
Approved! You will have to rebase changes from master for it to merge. See https://github.com/balena-io/process/blob/master/process/Commit_and_PR_Guidelines.md#pull-request-guidelines 

I did not test this on a running device, I assume you did, right? :p

Good work!

Issue #1591, opened 2021-02-11
Journalctl process infinite spawn


20k-ultra:
In the following code we can see that whenever a spawned journalctl process receives any message from stderr a new process will be spawned:

https://github.com/balena-io/balena-supervisor/blob/17b6b007780132898e7f6f6d9ee504e7ec6a1dbb/src/logging/monitor.ts#L76-L82

This snippet is used in the monitor module to being watching for engine logs and is not for following container logs...there is another process for that called [backfill](https://github.com/balena-io/balena-supervisor/blob/17b6b007780132898e7f6f6d9ee504e7ec6a1dbb/src/logging/monitor.ts#L129) in the same module.

Another side affect of infinite spawning journalctl process is that eventually the system runs out of watch descriptors and so following logs is no longer possible. This would cause new logs to no longer be sent to the cloud.

The solution for this problem is to basically have better error handling on journalctl processes. If an error is received we should atleast stop the current one before spawning another however it has been seen that if you change the debug loglevel then debug messages will be sent to sterr and they aren't actually errors so stopping processes simply because a message was received isn't the best fix.

An example "error" message that was causing this issue:

```
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [debug]   Spawning journald with: chroot  /mnt/root journalctl -a -S 2021-02-10 00:16:30 -o json CONTAINER_ID_FULL=ac3ac505d32272be62784b0cde5274f6a3275daf806fecc1249c62752cf22546
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [error]   journalctl backfill error:  Considering root directory '/run/log/journal'.
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [error]     Root directory /run/log/journal added.
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [error]   Considering directory '/run/log/journal/503b9249a316496488f03ef7888879c4'.
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [error]   Directory /run/log/journal/503b9249a316496488f03ef7888879c4 added.
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [error]   Journal effective settings seal=no compress=no compress_threshold_bytes=8B
Feb 11 03:06:57 98adad2 resin-supervisor[684307]: [error]   File /run/log/journal/503b9249a316496488f03ef7888879c4/system.journal added.
[...]

PR #1592, opened 2021-02-11
Refactor journalctl monitor to only spawn new process on exit

20k-ultra:
See https://github.com/balena-io/balena-supervisor/issues/1591 for info. 

This PR solves the issue by only spawning more processes on exit of the original. In addition I have made the code continue to follow the previous logic of spawning another process on error message from stderr. This can be changed to just log the error but this PR is focused on just preventing an infinite loop spawning more processes.

Change-type: patch
Closes: #1591

### DIFF
--- a/src/logging/monitor.ts
+++ b/src/logging/monitor.ts
@@ -73,13 +73,18 @@ class LogMonitor {
 					this.handleRow(row);
 				}
 			},
-			async (data: Buffer) => {
-				log.error(
-					'Non-empty stderr stream from journalctl log fetching: ',
-					data.toString(),
-				);
+			(data: Buffer) => {
+				log.error('journalctl - balena.service stderr: ', data.toString());
+			},
+			() => {
+				// noop for closed
+			},
+			async () => {
+				log.debug('balena.service journalctl process exit.');
+				// On exit of process try to create another
 				await delay(JOURNALCTL_ERROR_RETRY_DELAY);
-				this.start();
+				log.debug('Spawning another process to watch balena.service logs.');
+				return this.start();
 			},
 		);
 	}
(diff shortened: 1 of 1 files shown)


20k-ultra:
I will be making a few changes to this PR since I do not like that a new process is spawned on any message from stderr.

I've spoken with @codewithcheese @alexgg and we all agree this can be improved.

20k-ultra:
I've made the PR not kill the process when a message is received on stderr. This is because if there is an error causing the process to exit then the process would exit. If it's a message that's not causing the process to exit but preventing logs then most likely spawning another process won't fix that error. This PR contains log improvements as well to indicate what is happening.

src/logging/monitor.ts
```diff
@@ -113,13 +118,17 @@ class LogMonitor {
 		onRow: (row: JournalRow) => void,
 		onError: (data: Buffer) => void,
 		onClose?: () => void,
```
pipex, line 120: I don't see much use to keeping the `onClose` argument with the new fix. The `streamLogsFromJournal` method is set to private so we know only the `LogMonitor` class calls it

pipex: Also, when is the the `close` event supposed to be emitted? I don't see it on the documentation https://nodejs.org/docs/latest-v12.x/api/process.html#process_process_events

pipex (commented):
I like the change, I would just remove the references to the 'close' event, but this looks like this solves the issue by ensuring a new process is started only if the previous one stopped.

src/logging/monitor.ts
20k-ultra: The onClose is used https://github.com/balena-io/balena-supervisor/blob/17b6b007780132898e7f6f6d9ee504e7ec6a1dbb/src/logging/monitor.ts#L142-L144 to set a follow value to true for a service. This is inside the backfill function which gets previous logs that were created while the Supervisor was not listening. This true value is used to indicate that the service can be followed now since it's been backfilled.

@codewithcheese are you able to confirm my understanding of what's going on here ?

The [onClose](https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_event_close) event is also from the Child process class not Process object.

I wanted to make the PR as least invasive so changing events other functions used was not something I wanted to do. If we wanted to we could discuss removing onClose to use onExit as I feel they achieve the same thing.

onClose -> stream has closed but process can still be running
onExit -> process has stopped which would also cause stream to close so in effect is the same

pipex: Thanks for clarifying that for me, I hadn't seen the backfill part of the code. It makes sense to keep the event then. I agree that it could probably work with `onExit` but is probably not super necessary

pipex (approved):
LGTM!

pipex:
@20k-ultra we should get this merged, there was another user hit by the bug today

Issue #1594, opened 2021-02-12
Incorrectly detecting services to stop that use Supervisor API


20k-ultra:
The following function incorrectly detects if we must stop a service because it's using the Supervisor API

https://github.com/balena-io/balena-supervisor/blob/17b6b007780132898e7f6f6d9ee504e7ec6a1dbb/src/compose/application-manager.ts#L389-L408

This function basically will only kills or adds a noop step. However, if you have a service that isn't using the supervisor API then a noop step is always added..

20k-ultra:
Created https://github.com/balena-io/balena-supervisor/issues/1595 to address changing checkTruthy to not return undefined. In this code snippet returning undefined has the same affect as returning false.

PR #1596, opened 2021-02-12
Patch killServicesUsingApi to not get stuck in noop loop

20k-ultra:
See issue for info. The original code would always attach noop step even if the container wasn't using the SupervisorAPI or running. This means it would wait for stopped containers to finish "stopping".

This PR resolves that issue by returning early if the service doesn't use the Supervisor API and if it does will only try to kill running services or wait for services that are currently in a stopping state. 

Change-type: patch
Closes: #1594

### DIFF
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -390,16 +390,18 @@ function killServicesUsingApi(current: InstancedAppState): CompositionStep[] {
 	const steps: CompositionStep[] = [];
 	_.each(current, (app) => {
 		_.each(app.services, (service) => {
-			if (
-				checkTruthy(
-					service.config.labels['io.balena.features.supervisor-api'],
-				) &&
-				service.status !== 'Stopping'
-			) {
+			const isUsingSupervisorAPI = checkTruthy(
+				service.config.labels['io.balena.features.supervisor-api'],
+			);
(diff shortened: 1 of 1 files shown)


src/compose/application-manager.ts
```diff
@@ -390,17 +390,21 @@ function killServicesUsingApi(current: InstancedAppState): CompositionStep[] {
...
+			if (service.status !== 'Stopping') {
+				// Stop this service
 				steps.push(generateStep('kill', { current: service }));
-			} else {
-				// We want to output a noop while waiting for a service to stop, as we don't want
-				// the state application loop to stop while this is ongoing
+			} else if (service.status === 'Stopping') {
+				// Wait for the service to finish stopping
 				steps.push(generateStep('noop', {}));
+			} else {
```
pipex, line 406: Is this necessary? I don't think the program can ever get here since the `if/else if` takes all cases?

pipex (changes requested):

src/compose/application-manager.ts
20k-ultra: no I added it for the comment but seems kinda silly now

20k-ultra: updated

pipex (approved):

Issue #1595, opened 2021-02-12
refactor checkTruthy to only return booleans


20k-ultra:
The [checkTruthy](https://github.com/balena-io/balena-supervisor/blob/17b6b007780132898e7f6f6d9ee504e7ec6a1dbb/src/lib/validation.ts#L64) function will return a boolean or undefined when it evaluates a value. The problem is this function should return true or false because the name itself says checkTruthy. For example, undefined and null is falsey so the function must return false but instead it returns undefined.

We can see that this function is missing the point of its existence because in places that it is used the code often does another null/undeinfed check on the value returned from checkTruthy. 

We can use this function to also create our own types of truthy/falsey values. For example 'off' is a truthy but we can make it be falsey.

20k-ultra:
I will have a PR for this soon.

PR #1598, opened 2021-02-15
Refactor checkTruthy to return more predictable values

20k-ultra:
Change-type: patch
Closes: #1595

### DIFF
--- a/src/config/types.ts
+++ b/src/config/types.ts
@@ -3,7 +3,7 @@ import * as t from 'io-ts';
 import * as _ from 'lodash';
 
 import { InternalInconsistencyError } from '../lib/errors';
-import { checkTruthy } from '../lib/validation';
+import { checkBooleanish, checkTruthy } from '../lib/validation';
 
 const permissiveValue = t.union([
 	t.boolean,
@@ -23,11 +23,10 @@ export const PermissiveBoolean = new t.Type<boolean, t.TypeOf<PermissiveType>>(
 				case 'string':
 				case 'boolean':
 				case 'number':
-					const val = checkTruthy(v);
-					if (val == null) {
+					if (!checkBooleanish(v)) {
 						return t.failure(v, c);
 					}
-					return t.success(val);
+					return t.success(checkTruthy(v));
 				case 'undefined':
 					return t.success(false);
 				case 'object':
--- a/src/device-config.ts
+++ b/src/device-config.ts
@@ -633,10 +633,8 @@ async function isVPNEnabled(): Promise<boolean> {
 	}
 }
 
-async function setVPNEnabled(value?: string | boolean) {
-	const v = checkTruthy(value || true);
-	const enable = v != null ? v : true;
-
+async function setVPNEnabled(value: string | boolean = true) {
+	const enable = checkTruthy(value);
 	if (enable) {
 		await dbus.startService(vpnServiceName);
 	} else {
--- a/src/lib/validation.ts
+++ b/src/lib/validation.ts
@@ -13,6 +13,8 @@ export interface CheckIntOptions {
 const ENV_VAR_KEY_REGEX = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
 const NUMERALS_REGEX = /^-?[0-9]+\.?0*$/; // Allows trailing 0 decimals
+const TRUTHY = ['1', 'true', true, 'on', 1];
+const FALSEY = ['0', 'false', false, 'off', 0];
 
 /**
  * checkInt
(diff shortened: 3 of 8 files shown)


src/config/types.ts
```diff
@@ -23,8 +23,9 @@ export const PermissiveBoolean = new t.Type<boolean, t.TypeOf<PermissiveType>>(
 				case 'string':
 				case 'boolean':
 				case 'number':
-					const val = checkTruthy(v);
-					if (val == null) {
+					const strict = false; // allow undefined
+					const val = checkBoolean(v, strict);
```
20k-ultra, line 27: The reason this code references checkBoolean instead of checkTruthy like all other places is because considering the effects of making values return only true or false means that if an invalid value is passed it will might turn settings off rather then throw an error like before. So this function exists to improve checkTruthy but still follow the previous logic.

20k-ultra (commented):
A bunch of places where I removed code is because now that checkTruthy returns only true or false we no longer have to check if a value is null/undefined.

src/network.ts
```diff
@@ -106,8 +106,7 @@ export const startConnectivityCheck = _.once(
 );
 
 export function enableConnectivityCheck(enable: boolean) {
-	const boolEnable = checkTruthy(enable);
-	enable = boolEnable != null ? boolEnable : true;
+	enable = checkTruthy(enable);
```
pipex, line 109: Should this be `!checkFalsey`? I'm wondering because of the previous 

```
enable = boolEnable != null ? boolEnable : true; 
```

20k-ultra: Considering that the previous code defaulted to true then using `!checkFalsey` would follow that philosophy because it will always be true unless the value explicitly matches a false value. I can make this change.

20k-ultra: @pipex added change.

src/lib/validation.ts
```diff
@@ -55,32 +57,48 @@ export function checkString(s: unknown): string | void {
...
+ */
+export function checkFalsey(v: unknown): boolean {
+	if (_.isString(v)) {
+		v = v.toLowerCase();
 	}
+	return (
+		FALSEY.filter((f) => {
+			return f === v;
+		}).length > 0
+	);
```
Page-, line 101:
```suggestion
	return FALSEY.includes(v);
```

src/lib/validation.ts
```diff
@@ -55,32 +57,48 @@ export function checkString(s: unknown): string | void {
...
+}
+
+/**
+ * checkFalsey
+ *
+ * Given an unknown value, determine if it evaluates to false.
+ *
+ */
+export function checkFalsey(v: unknown): boolean {
+	if (_.isString(v)) {
```
Page-, line 94: I generally prefer native versions to lodash when they're equivalent length/functionality, ie
```suggestion
	if (typeof v === 'string') {
```
here, but eg still things like `_.sum` that do things with no native equivalent. The reason being roughly that the native versions have the ability to be optimized better, are more standard nowadays, and are perfectly safe when you're not dealing with unknown types - lodash helps by making sure things like eg `.map` are always available (it's from the days when you'd often support browsers that didn't have such things natively available), and making sure it works consistently across null/arrays/objects (but with typescript we should be able to rely on having the right data types, and if not it's a bug we should fix rather than hide)

src/lib/validation.ts
```diff
@@ -55,32 +57,48 @@ export function checkString(s: unknown): string | void {
...
-		case 'off':
-		case 0:
-			return false;
-		default:
-			return;
+	return (
+		TRUTHY.filter((t) => {
+			return t === v;
+		}).length > 0
+	);
```
Page-, line 84:
```suggestion
	return TRUTHY.includes(v);
```

src/lib/validation.ts
```diff
@@ -55,32 +57,48 @@ export function checkString(s: unknown): string | void {
...
  * checkTruthy
  *
- * Given a value which can be a string, boolean or number, return a boolean
- * which represents if the input was truthy
+ * Given an unknown value, determine if it evaluates to true.
+ *
  */
-export function checkTruthy(v: unknown): boolean | undefined {
+export function checkTruthy(v: unknown): boolean {
 	if (_.isString(v)) {
```
Page-, line 77:
```suggestion
	if (typeof v === 'string') {
```

src/device-config.ts
```diff
@@ -634,9 +634,7 @@ async function isVPNEnabled(): Promise<boolean> {
 }
 
 async function setVPNEnabled(value?: string | boolean) {
-	const v = checkTruthy(value || true);
-	const enable = v != null ? v : true;
-
+	const enable = checkTruthy(value || true);
```
Page-, line 637: I assume this is meant to be only in the case that value is undefined? In which case it'd be better to use `??`, or even better use
```ts
async function setVPNEnabled(value?: string | boolean = true) {
```

src/network.ts
```diff
@@ -106,8 +106,8 @@ export const startConnectivityCheck = _.once(
 );
 
 export function enableConnectivityCheck(enable: boolean) {
-	const boolEnable = checkTruthy(enable);
-	enable = boolEnable != null ? boolEnable : true;
+	// Only disable if value explicilty matches falsey
```
Page-, line 109:
```suggestion
	// Only disable if value explicitly matches falsey
```

src/lib/validation.ts
20k-ultra: I agree but not sure how to work around the following issue when using includes.

> Argument of type 'unknown' is not assignable to parameter of type 'string | number | boolean'.
  Type 'unknown' is not assignable to type 'true'.ts(2345)

I avoid the use of as any but what do you think about `return TRUTHY.includes(v as any)` if we really wanted to avoid the filter.

src/lib/validation.ts
20k-ultra: I favor the std lib as well. Thanks for pointing this instance out.

src/device-config.ts
20k-ultra: I like the signature change better then having this inside the function.

src/lib/validation.ts
20k-ultra: I saw a good amount of instances use `as any` so I changed my mind.

20k-ultra:
@Page- added changes from your comments. thanks your feedback.

src/lib/validation.ts
Page-: In this specific case of validation I don't mind the cast, although you could also do it via
```ts
const FALSEY: unknown[] = ['0', 'false', false, 'off', 0];

Page- (commented):
LGTM just up to you whether you want to use `unknown[]` instead of `as any`

pipex (approved):
LGTM

PR #1601, opened 2021-02-18
Call restore for each healthcheck stub in v1 tests

cywang117:
In the after hook of the `41-device-api-v1.spec.ts` test suite, stubbed healthchecks aren't restored, causing all future tests that attempt to stub healthchecks to fail. 

Change-type: patch


20k-ultra (approved):
nice find.

cywang117:
> 
> 
> @cywang117 remember to use `git rebase` instead of `git merge` to update changes from master into your branch.

@pipex Thanks, will remember for next time!

Issue #1604, opened 2021-02-25
Properly handle legacy volumes


20k-ultra:
When we decoupled the docker components from application-manager.coffee into their own respective files one change made was that volumes were now given a label "io.balena.supervised: 1". This issue is that if a device upgraded from an old enough Supervisor versions then releases containing volumes without this label will make the apply target state calculation think a volume must be created but then when creating the volume the Supervisor will query for volumes by appId and it finds the existing volume then nothing happens and loops over and over.

For example we can see that the Supervisor wants to create a volume when one already exists and the only difference is the label:

```
Feb 25 02:35:49 e3a0a8f resin-supervisor[1961891]: [debug]   Attempting to create volume {"name":"ser2net-conf","appId":1503221,"config":{"driverOpts":{},"driver":"local","labels":{"io.balena.supervised":"true"}}}
Feb 25 02:35:49 e3a0a8f resin-supervisor[1961891]: [debug]   Found existing volume:  {"name":"ser2net-conf","appId":1503221,"config":{"labels":{},"driver":"local","driverOpts":{}}}
```

PR #1608, opened 2021-03-03
Infer legacy Volumes that do not have the supervised label

20k-ultra:
I spent a bit of time trying to figure out what the least impactful way to make the Supervisor detect Volumes without the supervised label. This fix solves that by using existing functions namely,  `deconstructDockerName` which tries to extract Volume info such as appId from a given volume ID. If the Supervisor cannot extract info from a Volume's inspect info which it created then it can assume that volume was not created by the Supervisor. Therefore, this patch allows the Supervisor to detect such legacy volumes and also logs whenever a Volume is not supervised to help with troubleshooting in the future if the Supervisor is ignoring a volume.

Change-type: patch
Closes: #1604

### DIFF
--- a/src/compose/volume-manager.ts
+++ b/src/compose/volume-manager.ts
@@ -3,11 +3,12 @@ import * as Path from 'path';
 import { VolumeInspectInfo } from 'dockerode';
 
 import constants = require('../lib/constants');
-import { NotFoundError } from '../lib/errors';
+import { NotFoundError, InternalInconsistencyError } from '../lib/errors';
 import { safeRename } from '../lib/fs-utils';
 import { docker } from '../lib/docker-utils';
 import * as LogTypes from '../lib/log-types';
 import { defaultLegacyVolume } from '../lib/migration';
+import log from '../lib/supervisor-console';
 import * as logger from '../logger';
 import { ResourceRecreationAttemptError } from './errors';
 import Volume, { VolumeConfig } from './volume';
@@ -24,8 +25,21 @@ export async function get({ name, appId }: VolumeNameOpts): Promise<Volume> {
 }
 
 export async function getAll(): Promise<Volume[]> {
-	const volumeInspect = await listWithBothLabels();
-	return volumeInspect.map((inspect) => Volume.fromDockerVolume(inspect));
+	const volumes = await list();
+	// Normalize inspect information to Volume types and filter any that fail
+	return volumes.reduce((volumesList, volumeInfo) => {
+		try {
+			const volume = Volume.fromDockerVolume(volumeInfo);
+			volumesList.push(volume);
+		} catch (err) {
+			if (err instanceof InternalInconsistencyError) {
+				log.debug(`Cannot parse Volume: ${volumeInfo.Name}`);
+			} else {
+				throw err;
+			}
+		}
+		return volumesList;
+	}, [] as Volume[]);
 }
 
 export async function getAllByAppId(appId: number): Promise<Volume[]> {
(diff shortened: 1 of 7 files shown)


20k-ultra:
working on tests but wanted to share patch to get the conversation going.

src/compose/volume-manager.ts
```diff
@@ -23,9 +24,29 @@ export async function get({ name, appId }: VolumeNameOpts): Promise<Volume> {
...
+	let volumes;
+	if (allowLegacy) {
+		// Returns all Volumes which might not have a supervised label
+		volumes = await list();
+	} else {
+		// Only get Volumes with supervised label
+		volumes = await listWithBothLabels();
+	}
+	// Normalize inspect information to Volume types and filter any that fail
+	return volumes.reduce((volumesList, volumeInfo) => {
```
20k-ultra, line 37: @Page- what are your thoughts on a cleaner way of doing this. I never check what options lodash has so maybe there is one. To be clear, I'm happy with this solution but I think you've pointed out smaller LOC implementations that achieve the same thing. I basically need a filter catch function..

src/compose/volume-manager.ts
```diff
@@ -23,9 +24,29 @@ export async function get({ name, appId }: VolumeNameOpts): Promise<Volume> {
...
+		// Only get Volumes with supervised label
+		volumes = await listByLabel();
+	}
+	// Normalize inspect information to Volume types and filter any that fail
+	return volumes.reduce((volumesList, volumeInfo) => {
+		try {
+			const volume = Volume.fromDockerVolume(volumeInfo);
+			volumesList.push(volume);
+		} catch (e) {
+			if (e.message.substring(0, 16) === 'Could not detect') {
```
pipex, line 42: I think creating a new exception type and throw that in `Volume.deconstructDockerName` might be  a better option here @20k-ultra

pipex: I think this doesn't work for correctly named volumes without the `io.balena.supervised` label, right? In that case `Volume.deconstructDockerName` doesn't throw and it is also not catched by `listByLabel`

pipex: Is there any drawback to always using `list()` instead of `listByLabel`? If there isn't I would get rid of the function parameter and just use `list()` by default

20k-ultra: I agree with creating a new exception type that we can use for better error handling of this type. I'll do this.

This error is only for failure to parse a volume name, there's no need for it to be aware of the `io.balena.supervised` label so your observations are by design.

I kept listByLabel (previously listWithBothLabels) even though no one is using it because when I'm patching an issue I want to avoid introducing new regressions at all costs so I write the patch in a way that fixes the issue without introducing new behavior. By using this toggle (allowLegacy) the codebase acts the same unless when this value is true. However, whenever I make a PR patch and the reviewer thinks there's optimizations we can squeeze in at the same time I'm cool to add them. i'll clean this up too.

pipex: Oh, I thought those volumes without the label were causing the issue. I see now.

For  the adding the `allowLegacy` arg, if you feel it might be safer to do it that way, then let's keep it. It is true that just relying on naming to identify volumes might be easy to break, so do what you believe it's best

20k-ultra:
All the files where I change values like '{}' to '[]' is because they are suppose to be arrays not objects. I learned this because I added listVolumes which some functions perform array functions such as reduce. This is when i saw that the default values of volumes, networks, containers, etc being objects was incorrect. Since I fixed that in mocked-dockerode I went to the test files to make sure they also had arrays.

pipex:
Nice catch on the '{}', my guess is that whoever wrote those tests got confused with the target state endpoint format (where they are objects)

pipex (approved):
LGTM

Issue #1602, opened 2021-02-24
Remove support for rsync deltas


dfunckt:
The Supervisor has support for both old rsync-based and new balena-engine deltas which is selectable at runtime via a config var. The old deltas offer no advantage over the new and there is no reason to enable them or have the code around in the Supervisor anymore.

cc @pipex

20k-ultra:
@cywang117 would you like to work on this one ? It'll be a change of scenery from the HTTP API :p

cywang117:
@20k-ultra I'll take a look!

cywang117:
Hey @dfunckt I'd like to get some context regarding rsync vs balenaEngine deltas if you don't mind :). Some observations first, please correct if wrong:
- [V2 deltas are based on rsync and V3 deltas are based on balenaEngine](https://github.com/balena-io/balena-supervisor/blob/master/src/lib/docker-utils.ts#L140). These are the only two versions supported currently.
- The [default delta version](https://github.com/balena-io/balena-supervisor/blob/master/src/device-config.ts#L168) is V2, rsync. 

With that in mind, is it possible that there are devices / running applications still using the V2 deltas? If the user upgrades to a proposed new Supervisor version without V2 delta support, would there be any breaking changes?

dfunckt:
Replied in that linked FD thread and on #1613, but for clarity:

> is it possible that there are devices / running applications still using the V2 deltas?

Yes, all devices on balenaOS <2.47 use rsync deltas.

> If the user upgrades to a proposed new Supervisor version without V2 delta support, would there be any breaking changes?

Yes, images produced by applying rsync deltas are fundamentally different to those produced by applying engine deltas. Rsync deltas squash the image layers and carry no metadata resulting in an image that has a single layer and appears as a 18KB image. Engine deltas maintain layers and metadata and the resulting images are identical to what a regular `pull` action would produce, and to do so require the image that the delta is being applied to, to be "well formed" (which rsync images are not) so the Supervisor performs a complete image pull the first time it switches from v2 deltas to v3. This "migration" is available in Supervisors included in balenaOS >=2.47 or thereabouts.

> If there would be breaking changes, my assumption is that we'd have to handle the delta version difference somehow within Supervisor, by converting V2 delta requests into V3. I see from the balena-delta repo that one difference lies in how images are referenced: V2 by image name, V3 by image ID, However I don't know what other differences between the two there are, without digging too deeply into the code.

[...]

PR #1613, opened 2021-03-05
Remove rsync delta (v2) support

cywang117:
As requested in https://github.com/balena-io/balena-supervisor/issues/1602, remove support for old rsync deltas. Removed all code related to Rsync deltas, and added some tests for previously-untested delta interfaces. Let me know if any questions!

See [related Flowdock thread] for context. If the potential incompatibility of V3 deltas with balenaOS <2.47 isn't a big issue, this PR might be okay for merge, otherwise it might be better as a draft PR until the relevant supervisor contracts are implemented.

Change-type: minor
Connects-to: #1602


cywang117:
@20k-ultra I'm only tagging you because you tagged me for this issue, feel free to ignore (or not) 😄

cywang117:
@balena-ci retest

dfunckt:
@cywang117, I think rsync (the binary and any dependencies it may have) can be removed from the Supervisor's image as well (which should slightly decrease its footprint on disk).

dfunckt:
Also, regarding the "old balenaOS, new Supervisor" issue you're describing, we're apparently going to be preventing older balenaOS devices from receiving the balenaOS 3 Supervisor, so that is a good cutoff point to include those changes without waiting for the contracts work (and it *also* makes somewhat more sense from a user's perspective -- no deltas v2 *is* a user-facing breaking change and we've long started the transition to v3). I've no idea about the logistics around this though and what it means for a PR to stay unmerged until the opportune time to include it in balenaOS 3.

cywang117:
@dfunckt Thanks for the additional context! I've removed the rsync binary. There will probably be conflicts when eventually rebasing + merging this PR, since OS3 isn't here yet, but we can deal with them when they arrive. The core code removal is done and should be sound though.

cywang117:
@dfunckt Thanks so much for the guidance :) I definitely missed a few things. I've removed the `docker-delta` and `resumable-request` dependencies, because they were both only used for v2 deltas. The isV2DeltaImage function has also been removed along with its accompanying test.  I'll go ahead and resolve the comments in your review.

Because of package.json and package-lock.json changes (removed 2 dependencies), there will probably be conflicts when attempting to merge this with v3 supervisors later on.

src/lib/docker-utils.ts
```diff
@@ -72,37 +64,18 @@ export async function fetchDeltaWithProgress(
...
-	// We need to make sure that we're not trying to apply a
-	// v3 delta on top of a v2 delta, as this will cause the
-	// update to fail, and we must fall back to a standard
-	// image pull
-	if (
-		deltaOpts.deltaVersion === 3 &&
-		(await isV2DeltaImage(deltaOpts.deltaSourceId))
-	) {
-		logFn(`Cannot create a delta from V2 to V3, falling back to regular pull`);
-		return await fetchImageWithProgress(imgDest, deltaOpts, onProgress);
```
dfunckt, line 101: I think we’ll need to keep this check.

The image on a device that has been updating using rsync deltas lacks the metadata needed to make it a “proper” docker image, which means trying to use it as a source to apply a v3 delta over fails. The only way forward in this case is to perform a regular pull — effectively performing a one-time migration by pulling the complete new image (and subsequently removing the old one).

The reason we still need this check even though the update path to balenaOS 3 will involve an update to the latest balenaOS versions first (hence recent Supervisors that know how to migrate rsync deltas) is that unless the user *also* pushes a new release in between, the local images may still be in that incompatible image format. We can technically only drop this check when all devices on the platform have switched away from rsync deltas.

Might worth expanding the code comment a bit to include this context.

cywang117: @dfunckt Apologies for dropping the ball on this. I've re-added `isV2DeltaImage` and amended the comment above `if (deltaVersion === 3 && isV2DeltaImage)` to reflect the context you've provided. Also re-added tests for this util and for the v3-on-v2-delta case,

dfunckt (commented):
Really good stuff @cywang117 !

PR #1621, opened 2021-03-12
Add nebra-hnt to raspberry pi variants

pipex:
Change-type: patch


pipex:
@balena-ci I self certify!

PR #1630, opened 2021-03-22
Enable strict options by default and only disable specific ones

Page-:
Change-type: patch

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from 'events';
-import { Transaction } from 'knex';
+import type { Transaction } from 'knex';
 import * as _ from 'lodash';
 import StrictEventEmitter from 'strict-event-emitter-types';
 import { inspect } from 'util';
@@ -54,7 +54,7 @@ export async function get<T extends SchemaTypeKey>(
 	key: T,
 	trx?: Transaction,
 ): Promise<SchemaReturn<T>> {
-	const $db = trx || db.models.bind(db);
+	const $db = trx || db.models;
 
 	if (Schema.schema.hasOwnProperty(key)) {
 		const schemaKey = key as Schema.SchemaKey;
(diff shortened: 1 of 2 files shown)


src/config/index.ts
```diff
@@ -235,7 +235,7 @@ export function valueIsValid<T extends SchemaTypeKey>(
 
 async function getSchema<T extends Schema.SchemaKey>(
 	key: T,
-	$db: Transaction,
+	$db: typeof db.models,
```
Page-, line 238: This change is necessary to pass `strictBindCallApply` - it used to work because `db.models.bind(db)` would convert to `any` and so would "pass" type checking

src/config/index.ts
```diff
@@ -183,7 +183,7 @@ export async function set<T extends SchemaTypeKey>(
 	if (trx != null) {
 		await setValuesInTransaction(trx);
 	} else {
-		await db.transaction((tx: Transaction) => setValuesInTransaction(tx));
```
Page-, line 186: I removed the unnecessary explicit type so we can automatically use the correct types implicitly

src/config/index.ts
```diff
@@ -54,7 +54,7 @@ export async function get<T extends SchemaTypeKey>(
 	key: T,
 	trx?: Transaction,
 ): Promise<SchemaReturn<T>> {
-	const $db = trx || db.models.bind(db);
```
Page-, line 57: I removed the `.bind` as it's not necessary - it's almost certainly a leftover from when `db` was a class before we converted it to a singleton

tsconfig.json
```diff
@@ -2,10 +2,11 @@
 	"compilerOptions": {
 		"target": "ES2019",
 		"module": "commonjs",
-		"noImplicitAny": true,
```
Page-, line 5: This is included under the `strict` banner

tsconfig.json
```diff
@@ -2,10 +2,11 @@
...
+		"strict": true,
+		"strictFunctionTypes": false,
+		"strictPropertyInitialization": false,
 		"preserveConstEnums": true,
 		"inlineSourceMap": true,
-		"strictNullChecks": true,
```
Page-, line 8: This is included under the `strict` banner

Page-:
@balena-ci rebase

Page-:
@balena-ci I self-certify!

PR #1635, opened 2021-03-29
Prevent unintended image removal when calling purge endpoints to remove volumes

cywang117:
[Issue](https://github.com/balena-io/balena-supervisor/issues/1611)

When clicking the "purge" dashboard action or querying /v1/purge or /v2/applications/:appId/purge, images would also be removed. This is unintended and undesired, since the images will have to be redownloaded, consuming bandwidth, and an error also appears on the logs. This patch sets a flag when purging to prevent images from being removed.

The error log mentioned above:
```shell
[event]   Event: Purge data error {"appId":1806459,"error":{"message":"Failed to apply state transition steps. Cannot read property 'match' of undefined Steps:[\"fetch\",\"fetch\",\"fetch\",\"createVolume\"]","stack":"Error: Failed to apply state transition steps. Cannot read property 'match' of undefined Steps:[\"fetch\",\"fetch\",\"fetch\",\"createVolume\"]\n    at /usr/src/app/src/device-state.ts:878:10"}}
[error]   Error on POST /v1/purge:  Error: Failed to apply state transition steps. Cannot read property 'match' of undefined Steps:["fetch","fetch","fetch","createVolume"]
[error]         at /usr/src/app/src/device-state.ts:878:10
[api]     POST /v1/purge 503 - 1274.285 ms
```

Purge events before and after proposed patch (top = first event):
<table>
    <thead>
        <tr>
            <th>Before</th>
            <th>After</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td colspan=2>Purge endpoint is called, lock updates / pause target state apply as needed</td>
        </tr>
        <tr>
            <td colspan=2>Get the current state and temporarily set services and volumes as empty, to be applied as intermediate state.</td>
        </tr>
        <tr>
            <td></td>
            <td width='50%'>Set the applicationManager-owned flag <code>isPurgingVolumes</code> to false (add <code>isPurgingVolumes</code> to the logic under <code>applicationManager.saveAndRemoveImages</code>)</td>
        </tr>
        <tr>
            <td colspan=2>Supervisor generates a series of steps based on current-intermediate state diffs. Steps below:</td>
        </tr>
        <tr>
            <td>
                <ol>
                    <li><code>kill</code></li>
                    <li><code>removeImage</code></li>
                    <li><code>removeVolume</code></li>
                    <li><code>fetch</code></li>
                    <li><code>createVolume / start</code></li>
                </ol>
            </td>
            <td>
                <ol>
                    <li><code>kill</code></li>
                    <li><code>removeVolume</code></li>
                    <li><code>createVolume / start</code></li>
                </ol>
            </td>
        </tr>
        <tr>
            <td colspan=2>Explicitly remove all appID-related volumes as returned by volumeManager.</td>
        </tr>
        <tr>
            <td></td>
            <td width='50%'>Set <code>isPurgingVolumes</code> to true (also set to true in event of thrown error)</td>
        </tr>
        <tr>
            <td colspan=2>Restore services and volumes and call applyIntermediateState again, then allow target state to continue applying.</td>
        </tr>
    </tbody>
</table>

Because of `isPurgingVolumes`, no `removeImage` step is ever generated, so there aren't any log errors and there doesn't need to be unnecessary bandwidth usage from delta/image fetches. However, there is a small issue which doesn't affect the state engine funnel, but it's a bit annoying. It also points to a potential logical inconsistency in `applicationManager.saveAndRemoveImages`. After volume(s) are removed and before start / createVolume steps are generated, there is an unexpected saveImage step, which errors because the image digest(name) is undefined:

```
Current Images: []

Target Images: [
{
  name: undefined,
  appId: ...
  serviceId: ...
}
]
```

Unsure why the name is `undefined`. It should be an image digest string.  Also , I don't think a saveImage step is needed. I am investigating this and the "temporary fix" is to filter out `saveImage` steps when `isPurgingVolumes` is true. (Without this filter, an error inevitably gets thrown and logged.)

I still need to write tests, but feel free to give feedback @20k-ultra @pipex. Thanks!

Change-type: patch


cywang117:
In any case, I've changed the logic to prevent an unnecessary saveImage step from being generated while purging, so the above won't happen anymore - it simply remains documented in case there's some issue related to this in the future.

src/compose/application-manager.ts
```diff
@@ -694,7 +702,7 @@ function saveAndRemoveImages(
 			const isNotSaved = !_.some(availableWithoutIds, (img) =>
 				_.isEqual(img, targetImage),
 			);
-			return isActuallyAvailable && isNotSaved;
+			return isActuallyAvailable && isNotSaved && !isPurgingVolumes;
```
pipex, line 705: I'm not sure I would modify this here @cywang117,  I would prefer to prevent the application manager from performing the [image cleanup step](https://github.com/balena-os/balena-supervisor/blob/master/src/compose/application-manager.ts#L209) on any intermediate state. The image database is going away and when that happens, the cleanup logic will be done through a `image prune` step, and it will still make sense to make it here.

Even if/when we move to actors, we will probably want to notify the application manager actor to skip cleanup on intermediate states, so that this logic will be maintained.

pipex (commented):
I think the general idea is OK, but we should do this in a more future compatible way. 

A question related to these changes. Could you check that the statement [here](https://github.com/balena-os/balena-supervisor/blob/master/src/device-api/common.js#L73-L79) is still true? From playing with balena compose last week I get the sense that this is no longer the case

```
// Now that we're not running anything, explicitly
// remove the volumes, we must do this here, as the
// application-manager will not remove any volumes
// which are part of an active application
```

Regarding the undefined `name` on the second applyIntermediate call, could you check that the apps state obtained in the call to [getCurrentState](https://github.com/balena-os/balena-supervisor/blob/7358db6af745f2f393e9bef80d629cb112c20ee1/src/device-api/common.js#L53) has the name? Maybe there is a bug there? I would expect the current service state to have the correct image name.

src/compose/application-manager.ts
cywang117: @pipex After checking out the old implementation, no cleanup step is ever generated. The doPurge steps are 100% as listed above: kill --> noop --> removeImage --> removeVolume --> fetch --> createVolume + start. We never enter the `if (cleanupNeeded)` block. This is because currently the cleanupNeeded boolean is determined by whether the docker image ID is in the image table, which will definitely change once the image table is removed.

From what I've investigated, the error is entirely in [`saveAndRemoveImages`](https://github.com/balena-os/balena-supervisor/blob/4f41901275df971e8769ab31f6d73234d72c1017/src/compose/application-manager.ts#L604), which generates the unnecessary removeImage/saveImage steps based on incomplete logic which hasn't taken volume purging status into account. I see the rationale behind not touching saveAndRemove though, since it'll be removed or changed significantly with image table removal.

If we don't want to change saveAndRemove, I think there'd be two ways to go about fixing this issue:
1. Wrap the [saveAndRemove call in getRequiredSteps](https://github.com/balena-os/balena-supervisor/blob/4f41901275df971e8769ab31f6d73234d72c1017/src/compose/application-manager.ts#L215) with a isPurging conditional, or isApplyingIntermediate conditional so it's skipped when purging.
2. Filter out removeImage / saveImage steps when isPurging / isApplyingIntermediate is true.

I'm more inclined to go with option 1 since it feels less hack-y.

Also, I'm actually more inclined to keep isPurging as a conditional over isApplyingIntermediate, since there could be a case where Supervisor is applying intermediate state but needs a remove or saveImage step (there aren't any of these cases right now, but there *could* be)

What do you think?

cywang117:
@pipex That comment is indeed still true, if I remove the [explicit volume removal call in doPurge](https://github.com/balena-os/balena-supervisor/blob/master/src/device-api/common.js#L77-L79) then volumes aren't removed. I haven't checked volume removal other situations but this sounds familiar, I remember talking about this before - that because of issues with mistaken volume removal, [volumes would no longer be removed without explicit action, even orphaned volumes](https://github.com/balena-os/balena-supervisor/issues/1026).

deviceState.getCurrentState queries balenaEngine directly through serviceManager, I think only an imageManager query would return the digest. I suppose that's why name is `undefined`. This [function](https://github.com/balena-os/balena-supervisor/blob/master/src/compose/images.ts#L95-L98) in `images.ts` has a comment which is revealing: it always expects the image name to not be undefined, because these image objects are being created from targetState, which always has the digest. However in the case of applyIntermediateTarget we're modifying a currentState object and passing it as target state, hence the error. I think this error would only appear in the applyIntermediate situation (no other methods try to pass currentState as targetState in some way that I know of), so it's probably not worth looking into unless we plan to incorporate applyIntermediateTarget into other places as well. Definitely worth making a note of though and good questions! Thoughts?

src/compose/application-manager.ts
pipex: > Wrap the saveAndRemove call in getRequiredSteps with a isPurging conditional, or isApplyingIntermediate conditional so it's skipped when purging.

This is what I meant, sorry. I would add the `isPurging` or `isIntermediate` to the `if` on line 209 and skip both the cleanup and the `saveAndRemoveImages` call altogether. When we remove the image database, the `saveAndRemoveImages` will go away and we will always perform a image prune as `cleanup` step.

I honestly prefer `isIntermediateTarget` or `isLocalTarget` or something, now we only do it for purging, but we also will need to change the `stop` and `restart` actions to use intermediate targets when we move the composer to its own library. If we do just `isPurging` that opens the floodgates to adding a bunch of other conditions.

pipex:
> That comment is indeed still true, if I remove the explicit volume removal call in doPurge then volumes aren't removed. I haven't checked volume removal other situations but this sounds familiar, I remember talking about this before - that because of issues with mistaken volume removal, volumes would no longer be removed without explicit action, even orphaned volumes.

Oh, good, it makes sense, I think what I saw on the balena-compose project was with networks and not volumes

> deviceState.getCurrentState queries balenaEngine/dockerode directly through serviceManager

You are right, I just checked [fromDockerContainer](https://github.com/balena-os/balena-supervisor/blob/7358db6af745f2f393e9bef80d629cb112c20ee1/src/compose/service.ts#L451) and it doesn't get the `imageName` from anywhere. We could try for now to get it either from the target state for that service, or querying the engine to get the repo digest for the service. The problem with the second option is that there can be more than one digest for the image, so that might cause problems. 

We should make a note to add the `imageName` to the service metadata in the future so we can get the full current state from the service itself.

src/compose/application-manager.ts
cywang117: @pipex Got it! That makes sense, I'll update the PR. I've also added tests so this PR can move out of draft status :)

cywang117:
I've opened an issue to track the `name: undefined` quirk [here](https://github.com/balena-os/balena-supervisor/issues/1649).

Otherwise, @20k-ultra @pipex I've added tests and this PR should be ready to go. common.js not being typed has become pretty inconvenient while working on this PR, I'll probably open an issue to get it converted to TS once and for all. I've added comments in the code where I've had to change implementations slightly because of type checks not passing, I'll comment in the code in GH as well to direct your guys' attention to them. As a result I'm not too pleased about how the implementation currently is so if you guys have any good ideas, feel free to drop them here. Thanks!

src/device-api/common.js
```diff
@@ -62,9 +62,55 @@ export async function doPurge(appId, force) {
...
+			 * for a more in-depth explanation of why aliases need to be filtered out.
+			 *
+			 * NOTE that we can't use safeStateClone which performs this filtering function because common.js
+			 * is not yet written in TypeScript, leading to a type error when trying to pass cloned state into
+			 * the 2nd applyIntermediate call.
+			 *
+			 * TODOs in order of achievability:
+			 * - Convert common.js to TypeScript & switch to using safeStateClone
+			 * - Prevent network aliases that are containerIds from being added to service config objects
+			 */
```
cywang117, line 106: @20k-ultra @pipex Here is the problem area I mentioned in my other comment. Hopefully I've documented in well enough here but lemme know if there are any questions. The `currentServices.forEach` block to filter out containerId aliases definitely feels like a bit of a hack, but in the absence of typed common.ts I'm not sure what to do to prevent `Networks do not match!` issue.

cywang117: Idea: don't merge this PR until `common.js` is converted? I could work on that next and the file isn't too big so it shouldn't take long (kappa).

cywang117: With the [latest changes](https://github.com/balena-os/balena-supervisor/commit/bdf8808c716b19b866c2c240e5dddf90918206f4) courtesy of @Page- , please ignore this comment about holding off merge til common.ts.

cywang117:
@balena-ci rebase

pipex (approved):

pipex:
Nice work @cywang117 !!

PR #1640, opened 2021-03-31
Disable docker layer caching on CircleCI

pipex:
Docker layer caching is a paid plan feature on Circle CI, it has started
being enforced by Circle. Disabling it solves this temporarily while we sort out
a better long term solution for supervisor CI

Change-type: patch

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,7 @@ defaults: &defaults
   steps:
     - setup_remote_docker:
         version: 18.09.3
-        docker_layer_caching: true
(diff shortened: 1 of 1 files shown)


cywang117 (approved):
LGTM!

cywang117:
Further context: this PR is because of the organization move from balena-io to balena-os. Docker Layer Caching is possible on balena-io but not balena-os. Thus a new goal is to remove our dependency on CircleCI and move to solutions with less friction.

PR #1642, opened 2021-03-31
Removed upgrades.md doc

20k-ultra:
Moved this doc to docs repo in https://github.com/balena-io/docs/pull/1801 because upgrade policy isn't enforced by the Supervisor repository and is more suitable as a general document for Balena.

Change-type: patch


20k-ultra:
@balena-ci I self-certify!

Issue #1636, opened 2021-03-29
12.4.6 Purging data: Cannot read property 'reduce' of null


20k-ultra:
https://github.com/balena-io/balena-supervisor/commit/183ea88a2ad3bbdf18ab96505920465f1a404df0 made some changes to how volumes are queried and I think a regression was introduced that causes null volumes to be returned for the purge data function which results in an exception being thrown

``` Purge data: 26.03.21 16:32:43 (+0100) Error purging data: TypeError: Cannot read property 'reduce' of null```

PR #1643, opened 2021-04-01
Patch list volumes to always return an array

20k-ultra:
I don't know why or when docker will return null when listing volumes but given the issue's log, when I write a test which passes null response to the volume manager list function, it throws the exact same error. There aren't many places in the codebase which use reduce so it was easy to find the code where `Cannot read property 'reduce' of null` could come from.

Additionally, when I ran my test with the original code before https://github.com/balena-os/balena-supervisor/commit/183ea88a2ad3bbdf18ab96505920465f1a404df0 it does handle null and returns an empty array! 

So, I'm confident this fixes the regression.

Change-type: patch
Closes: #1636

### DIFF
--- a/src/compose/volume-manager.ts
+++ b/src/compose/volume-manager.ts
@@ -154,6 +154,6 @@ export async function removeOrphanedVolumes(
 }
 
 async function list(): Promise<VolumeInspectInfo[]> {
-	const allVolumes = await docker.listVolumes();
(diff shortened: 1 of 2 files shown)


cywang117 (commented):
This LGTM! I think the null case was handled before because `_.unionBy` would return an empty array for non-array arguments.

20k-ultra:
@balena-ci I self-certify!

PR #1647, opened 2021-04-01
Default delta version to be 3

20k-ultra:
The API is sending `RESIN_SUPERVISOR_DELTA_VERSION: 3` anyways so this will just make the source code default to this version as well.

Change-type: patch


20k-ultra:
@dfunckt I saw that you implemented a lot of the delta code in the Supervisor back in 2017. Would you be able to give your input here. It's ok to make the Supervisor codebase default to delta version 3 correct ? Do you know if this could have some issue for openBalena devices ?

cywang117:
@20k-ultra, @dfunckt raised an [issue](https://github.com/balena-os/balena-supervisor/issues/1602) that might provide some context, I completed a draft [PR](https://github.com/balena-os/balena-supervisor/pull/1613) for that issue which is expected to be merged with OS v3. I'll summarize the relevant parts of the linked Flowdock thread from that issue for those that can't access FD:

- v3 deltas are preferred to v2 deltas and superior, so remove support rsync v2 deltas with OS v3.
- However, although OS v<2.47 supports v3 deltas, there are kernel bugs that prevent them from being used

Therefore, if supervisor versions from v12.5.4 (current latest at time of writing this comment) onward default to v3 deltas, and the user is running OS v<2.47 and doesn't want to update, it seems that the user would not be able to use deltas. 

In the draft PR I also made the default version changes that you did here, with the addition of removing rsync functions and dependencies. I'm wondering if it's ok to change the default version here, since you can still select a really old OS version when provisioning a new device on the dashboard: 

image

dfunckt:
> It's ok to make the Supervisor codebase default to delta version 3 correct?

@cywang117 nailed it -- it is okay to default to v3 as long as that Supervisor is guaranteed it won't run on balenaOS <2.47. balenaOS 3 makes for a natural cutoff point where this can be guaranteed by restricting it's Supervisor from appearing as an option to upgrade to on devices running balenaOS <3.0.

> Do you know if this could have some issue for openBalena devices?

None at all, openBalena has no way to generate deltas because it doesn't include the cloud builder or the delta server, so toggling deltas "on" on openBalena devices will fail anyway.

20k-ultra:
@cywang117 I'll close this one in favour of https://github.com/balena-os/balena-supervisor/pull/1613

Issue #1646, opened 2021-04-01
Document all configurations available


20k-ultra:
Add a configurations.md in the docs to show all variables available.

This will be used in the docs on a part page https://www.balena.io/docs/learn/manage/configuration/#variable-list which currently gets these variables from https://github.com/balena-io/docs/blob/381bfc296fb1c432e1c3543b5fddea7fce913ab1/pages/learn/manage/configuration.md.

PR #1648, opened 2021-04-02
Added configurations.md to document all configurable vars

20k-ultra:
I tried to find all instance of a configuration variable and verify when they were added from the changelog...hopefully I didn't miss any.

Change-type: patch
Closes: #1646

### DIFF
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -1,0 +1,81 @@
+# Configuration Variables
+
+Configuration variables allow you to provide runtime configuration to the host OS and Supervisor. These variables all begin with `BALENA_` or the legacy prefix `RESIN_`. Beginning with `Supervisor v7.0.0`, a number of them automatically will be sent to the cloud on provision.
+
+**Note:** Configuration variables defined in the dashboard will not apply to devices in local mode.
+
+## Variable list
+
+This list contains configuration variables that can be used with all balena devices, some of which will automatically appear for devices with Supervisor v7.0.0 and greater. While they may not automatically populate in the dashboard, most of these variables can still be used with older Supervisor versions, so be sure to check the _Valid from_ column:
+
+| Name | Type | Reboots | Default | Description | Valid from |
+| --- | --- | --- | --- | --- | --- |
+| BALENA_SUPERVISOR_OVERRIDE_LOCK | boolean | false | false | Override any existing locks preventing application changes. | v1.0.0 |
+| BALENA_SUPERVISOR_VPN_CONTROL | boolean | false | true | Enable / Disable VPN service on device. | v1.1.0 |
+| BALENA_SUPERVISOR_CONNECTIVITY_CHECK | boolean | false | true | Enable / Disable VPN connectivity check | v1.3.0 |
+| BALENA_SUPERVISOR_LOG_CONTROL | boolean | false | true | Enable / Disable logs being sent to the balena API | v1.3.0 |
+| BALENA_SUPERVISOR_POLL_INTERVAL | integer | false | 900000 | Define the balena API poll interval in milliseconds. This interval will only matter if the device is not connected to the VPN at the time an update is pushed, or if BALENA_SUPERVISOR_INSTANT_UPDATE_TRIGGER is set to false. Starting from Supervisor v9.13.0, the Supervisor will use a random time between 0.5 and 1.5 times this poll interval each time it checks the balenaCloud API. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v1.3.0 |
+| BALENA_SUPERVISOR_DELTA_REQUEST_TIMEOUT | integer | false | 30000 | Millisecond time within which an HTTP response containing delta metadata must be returned by the delta server. | v2.9.0 |
+| BALENA_SUPERVISOR_LOCAL_MODE | boolean | false | false | Enable / Disable local mode. | v4.0.0 |
+| BALENA_SUPERVISOR_DELTA_APPLY_TIMEOUT | integer | false | 0 | Millisecond time within which delta data will begin syncing to the local device. | v6.2.0 |
+| BALENA_SUPERVISOR_DELTA_RETRY_COUNT | integer | false | 30 | Number of retry attempts for downloading deltas. | v6.2.0 |
+| BALENA_SUPERVISOR_DELTA_RETRY_INTERVAL | integer | false | 10000 | Millisecond time before another delta download is attempted after failiure. | v6.2.0 |
+| BALENA_SUPERVISOR_DELTA_VERSION | integer | false | 3 | Specifies the delta version to use which has different APIs and download algorithms. | v7.9.0 |
+| BALENA_SUPERVISOR_PERSISTENT_LOGGING | boolean | true | false | Persists journal logs to disk so they will be available after a reboot event. | v7.15.0 |
+| BALENA_SUPERVISOR_INSTANT_UPDATE_TRIGGER | boolean | false | true | Enable / Disable instant triggering of updates when a new release is deployed. If set to false, the device will ignore the notification that is triggered when the device's target state has changed. In this case, the device will rely on polling to apply updates. Coupled with a large BALENA_SUPERVISOR_POLL_INTERVAL, this allows spreading out updates in large fleets to avoid overloading local networks when there is a large number of devices at one location. | v9.13.0 |
+| BALENA_HOST_FIREWALL_MODE | string | false | off | Toggle firewall modes between on, off, and auto. | v11.9.1 |
+| BALENA_HOST_DISCOVERABILITY | boolean | false | true | Enable / Disable Avahi to run which will allow the device to respond to requests such as network scans. | v11.9.2 |
+| BALENA_HOST_SPLASH_IMAGE | integer | true | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
+
+---
+
+In addition to these values, there may be some device-type specific configuration variables that can be set.
+
+## Config.txt
+
+> Supported boards: Any Raspberry Pi and balenaFin.
+
+This device type utilizes a [`config.txt` file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) of which we offer some values to be configured. The following values are not all the configurations found in the `config.txt` file and are just balena additions or edits to the OEM version.
+
+| Name | Type | Reboots | Default | Description | Valid from |
+| --- | --- | --- | --- | --- | --- |
+| BALENA_HOST_CONFIG_dtoverlay | string | true | "vc4-fkms-v3d" | Allows loading [custom device tree overlays](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/overlays/README). | v1.0.0 |
+| BALENA_HOST_CONFIG_device_tree_overlay | string | true | "vc4-fkms-v3d" | Allows loading [custom device tree overlays](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/overlays/README). | v1.0.0 |
+| BALENA_HOST_CONFIG_dtparam | string | true | "i2c_arm=on","spi=on","audio=on" | Allows setting parameters for the default overlay. | v1.0.0 |
+| BALENA_HOST_CONFIG_device_tree_param | string | true | "i2c_arm=on","spi=on","audio=on" | Allows setting parameters for the default overlay. | v1.0.0 |
+| BALENA_HOST_CONFIG_disable_splash | integer | true | 1 | Enable / Disable the splash screen to display image on boot. | v1.0.0 |
+| BALENA_HOST_CONFIG_gpu_mem | integer | true | 16 | Define device GPU memory in megabytes. | v1.0.0 |
+| BALENA_HOST_CONFIG_gpio | string | true | undefined | Allows GPIO pins to be set to specific modes and values at boot time. See [here](https://www.raspberrypi.org/documentation/configuration/config-txt/gpio.md) for more. | v1.0.0 |
+
+You can find more information on updating `config.txt` through configuration variables in our [Advanced Boot Configuration Guide](https://www.balena.io/docs/reference/os/advanced/).
+
+## ConfigFS
+
+> Supported boards: UP Board
+
+This device type utilizes a `configFS` [file](https://lwn.net/Articles/148973/) interface that allows definition of arbitrary functions and configurations.
+
+| Name | Type | Reboots | Default | Description | Valid from |
+| --- | --- | --- | --- | --- | --- |
+| BALENA_HOST_CONFIGFS_ssdt | string | true | undefined | Allows configuring a user defined SSDTs from userspace via the configfs interface. See [SSDT Overlays](https://www.kernel.org/doc/html/latest/admin-guide/acpi/ssdt-overlays.html#ssdt-overlays). | v10.9.2 |
+
+## Extra-uEnv
+
+> Supported boards: Nvidia Jetson Nano and TX2.
+
+This device type utilizes a file which we created to proxy configurations to [extlinux.conf](https://wiki.syslinux.org/wiki/index.php?title=Config).
+
+| Name | Type | Reboots | Default | Description | Valid from |
+| --- | --- | --- | --- | --- | --- |
+| BALENA_HOST_EXTLINUX_isolcpus | string | true | undefined | Allows to isolate CPU cores from kernel scheduler. See [kernel parameters](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html?highlight=isolcpu#the-kernel-s-command-line-parameters). | v7.10.0 |
+| BALENA_HOST_EXTLINUX_fdt | string | true | undefined | Specify a filename with extension from userspace to use as DTB. See [Device Tree Binary](https://elinux.org/Jetson/TX2_DTB). | v11.9.0 |
+
+## ODMDATA
+
+> Supported boards: Nvidia Jetson TX2.
+
+This device type utilizes a configurable hardware partition to set the mode for [UPHY lane assigment](https://docs.nvidia.com/jetson/l4t/index.html#page/Tegra%20Linux%20Driver%20Package%20Development%20Guide/adaptation_and_bringup_tx2.html#wwpID0E0NL0HA).
+
+| Name | Type | Reboots | Default | Description | Valid from |
+| --- | --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


20k-ultra:
@acostach can you verify my descriptions fir ODMDATA and ConfigFS SSDTs make sense.

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+This device type utilizes a `configFS` [file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) interface that allows definition of arbitrary functions and configurations.
+
+Name | Default | Description | Valid from
+--- | --- | --- | ---
+BALENA_HOST_CONFIGFS_ssdt | undefined | Allows configuring a user defined SSDTs from userspace via the configfs interface. See [SSDT Overlays](https://www.kernel.org/doc/html/latest/admin-guide/acpi/ssdt-overlays.html#ssdt-overlays). | v10.9.2
+
+## Extra-uEnv
+
+> Supported boards: Intel NUC, Nvidia Jetson Nano and TX2.
```
acostach, line 59: Only NVidia Jetson Nano and TX2 here, without Intel NUC

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+This device type utilizes a file which we created to proxy configurations to [extlinux.conf](https://wiki.syslinux.org/wiki/index.php?title=Config).
+
+Name | Default | Description | Valid from
+--- | --- | --- | ---
+BALENA_HOST_EXTLINUX_isolcpus | undefined | Allows to isolate CPU cores from kernel scheduler.  See [kernel parameters](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html?highlight=isolcpu#the-kernel-s-command-line-parameters). | v7.10.0
+BALENA_HOST_EXTLINUX_fdt | undefined | Allows configuring a user defined DTB from userspace. See [Device Tree Binary](https://elinux.org/Jetson/TX2_DTB). | v11.9.0
```
acostach, line 66: Just to be clear for our users, we should probably mention that the device tree should be specified as a file name including extension (i.e device-tree.dtb), and not as a full path.

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+## ODMDATA
+
+> Supported boards: Nvidia Jetson TX2.
+
+This device type utilizes a configurable hardware partition to set the mode for [UPHY lane assigment](https://docs.nvidia.com/jetson/l4t/index.html#page/Tegra%20Linux%20Driver%20Package%20Development%20Guide/adaptation_and_bringup_tx2.html#wwpID0E0NL0HA).
+
+Name | Default | Description | Valid from
+--- | --- | --- | ---
+BALENA_HOST_CONFIG_odmdata_configurations | 2 | Select with mode to configure UPGHY lanes. See [UPHY lane assignment](https://docs.nvidia.com/jetson/l4t/index.html#page/Tegra%20Linux%20Driver%20Package%20Development%20Guide/adaptation_and_bringup_tx2.html#wwpID0E0NL0HA). | v11.13.0
```
acostach, line 77: typo, should be UPHY

docs/configurations.md
20k-ultra: I pulled these values from the device types the Supervisor will actually match the backend with and Intel NUC is one of them https://github.com/balena-os/balena-supervisor/blob/0f906fd9d8f4f7dc2bfd9481e882eb47f2ebbba5/src/config/backends/extra-uEnv.ts#L61-L69

Should we change this ?

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+--- | --- | --- | ---
+BALENA_SUPERVISOR_OVERRIDE_LOCK | false | Override any existing locks preventing application changes. | v1.0.0
+BALENA_SUPERVISOR_VPN_CONTROL | true | Enable/ Disable VPN service on device. | v1.1.0
+BALENA_SUPERVISOR_CONNECTIVITY_CHECK | true | Enable / Disable VPN connectivity check | v1.3.0
+BALENA_SUPERVISOR_LOG_CONTROL | true | Enable / Disable logs being sent to the balena API | v1.3.0
+BALENA_SUPERVISOR_POLL_INTERVAL | 900000 | Define the balena API poll interval in milliseconds. This interval will only matter if the device is not connected to the VPN at the time an update is pushed, or if BALENA_SUPERVISOR_INSTANT_UPDATE_TRIGGER is set to false. Starting from Supervisor v9.13.0, the Supervisor will use a random time between 0.5 and 1.5 times this poll interval each time it checks the balenaCloud API. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v1.3.0
+BALENA_SUPERVISOR_DELTA_REQUEST_TIMEOUT | 30000 | Millisecond time within which an HTTP response containing delta metadata must be returned by the delta server. | v2.9.0
+BALENA_SUPERVISOR_LOCAL_MODE | false | Enable / Disable local-mode. | v4.0.0
```
cywang117, line 19: @20k-ultra I'll summarize the context surroundings discussions of some of these config vars from the FD thread:
`*_SUPERVISOR_OVERRIDE_LOCK` and `*_SUPERVISOR_LOCAL_MODE` are reported through the SDK and on the dashboard as either `"0"` or `"1"`. The format is a `string int`. There might be other vars but these are the main ones that I saw had the difference. However, these vars are documented here as being `true` or `false`. While the Supervisor has the function `checkTruthy` to convert `"0"` to `false` and so on, this documentation still differs from the dashboard / SDK, and both are user-facing resources. I'm wondering if it would be beneficial, at least, to note somewhere in this PR that `false` here might be listed as `"0"` on the SDK/dashboard?

cywang117: @20k-ultra yo! I know you're busy so if you tell me your decision on this, I can take this off your plate :D

docs/configurations.md
20k-ultra: messaging you to discuss if we should change the deviceTypes. Resolving this though as the devices mentioned are what the code is going to match this.

docs/configurations.md
20k-ultra: I understand what you're saying but don't think it's necessary. The point is that the documentation just has to convey what the default boolean value is. If we add 1 or 0 then we'll just have to keep changing our documentation to match what values we accept which as of right now are [10](https://github.com/balena-os/balena-supervisor/blob/14efec2c7df29107a7a24f2080d4868abf6ca528/src/lib/validation.ts#L16-L17) different values.

20k-ultra: I modified the tables to say what the data type is for each value. This was needed because for some configs I had undefined for the default but it wasn't clear what they should be set to.

cywang117: Oh geez. The 10 different values thing is a bit spooky, I feel like limiting allowed values to a simple true/false boolean would have been better. I see your point though.

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+## ConfigFS
+
+> Supported boards: UP Board
+
+This device type utilizes a `configFS` [file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) interface that allows definition of arbitrary functions and configurations.
+
+Name | Type | Default | Description | Valid from
+--- | --- | --- | --- | ---
+BALENA_HOST_CONFIGFS_ssdt | string | undefined | Allows configuring a user defined SSDTs from userspace via the configfs interface. See [SSDT Overlays](https://www.kernel.org/doc/html/latest/admin-guide/acpi/ssdt-overlays.html#ssdt-overlays). | v10.9.2
```
cywang117, line 55: With any configs that support either strings (`"value1"`) or comma-separated strings (`"value1","value2"`), it would be clearer to have a distinction if communicating data types. In the code I see that config vars that support this format are 
- ConfigFS: 
   - *_HOST_CONFIGFS_ssdt
- Config.txt:
   - *_HOST_CONFIG_dtparam
   - *_HOST_CONFIG_dtoverlay
   - *_HOST_CONFIG_device_tree_param
   - *_HOST_CONFIG_device_tree_overlay
   - *_HOST_CONFIG_gpio
- extlinux:
   - *_HOST_EXTLINUX_isolcpus

as opposed to config vars like *_HOST_EXTLINUX_fdt which only support `"value1"` and not `"value1","value2"`. Currently both these data types are listed as "string" in these tables. What do you think?

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+> Supported boards: Any Raspberry Pi and BalenaFin.
+
+This device type utilizes a `config.txt` [file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) which we offer some values to be configured.
+
+Name | Type | Default | Description | Valid from
+--- | --- | --- | --- | ---
+BALENA_HOST_CONFIG_disable_splash | integer | 1 | Enable / Disable the Raspberry Pi rainbow splash screen. | v1.0.0
+BALENA_HOST_CONFIG_dtparam | string | "i2c_arm=on","spi=on","audio=on" | Define DT parameters. | v1.0.0
+BALENA_HOST_CONFIG_gpu_mem | integer | 16 | Define device GPU memory in megabytes. | v1.0.0
```
cywang117, line 43: What about the following?
- *_HOST_CONFIG_dtoverlay
- *_HOST_CONFIG_device_tree_param
- *_HOST_CONFIG_device_tree_overlay
- *_HOST_CONFIG_gpio
- *_HOST_CONFIG_enable_uart

They're all listed (except `enable_uart`) in config-txt.ts, and `enable_uart` & `dtoverlay` at least are toggle-able options on the dashboard for RPis. Is there a limit to the kinds of config vars listed here?

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+BALENA_SUPERVISOR_VPN_CONTROL | boolean | true | Enable / Disable VPN service on device. | v1.1.0
+BALENA_SUPERVISOR_CONNECTIVITY_CHECK | boolean | true | Enable / Disable VPN connectivity check | v1.3.0
+BALENA_SUPERVISOR_LOG_CONTROL | boolean | true | Enable / Disable logs being sent to the balena API | v1.3.0
+BALENA_SUPERVISOR_POLL_INTERVAL | integer | 900000 | Define the balena API poll interval in milliseconds. This interval will only matter if the device is not connected to the VPN at the time an update is pushed, or if BALENA_SUPERVISOR_INSTANT_UPDATE_TRIGGER is set to false. Starting from Supervisor v9.13.0, the Supervisor will use a random time between 0.5 and 1.5 times this poll interval each time it checks the balenaCloud API. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v1.3.0
+BALENA_SUPERVISOR_DELTA_REQUEST_TIMEOUT | integer | 30000 | Millisecond time within which an HTTP response containing delta metadata must be returned by the delta server. | v2.9.0
+BALENA_SUPERVISOR_LOCAL_MODE | boolean | false | Enable / Disable local-mode. | v4.0.0
```
cywang117, line 19:
```suggestion
BALENA_SUPERVISOR_LOCAL_MODE | boolean | false | Enable / Disable local mode. | v4.0.0
```

cywang117 (commented):
Nice work adding data types - I just have a few questions :)

docs/configurations.md
acostach: Hi @20k-ultra, I see, so this backend matches the NUC too. However, this won't work because the NUC is x86 which uses grub, so it doesn't support device-trees and implicitly no extlinux file. The configuration file which can specify extra cmdline arguments for the NUC is named grub_extraenv (see internal discussion https://www.flowdock.com/app/rulemotion/r-supervisor/threads/Vs_mpEKwt6K_kvgQ-wGw6ZH9A930) and locations for it in the boot partition are https://github.com/balena-os/balena-intel/blob/315071b5e0d98d7dd776a8b662ee8c0f73595e18/layers/meta-balena-genericx86/recipes-core/images/resin-image.inc#L18

So, on the NUC the extra_os_cmdline can't work as it is because the configuration file differs in name and path, and custom device-tree can't work because there's no device tree support on this platform. Therefore, the NUC should not be mentioned in the docs for these 2 configuration options.

20k-ultra: In PR https://github.com/balena-os/balena-supervisor/pull/1421 I noted that extra_uEnv backend is supported because of https://github.com/balena-os/balena-intel/pull/311. Was there a change or was the Intel Nuc never suppose to be there ? I've prempively created https://github.com/balena-os/balena-supervisor/pull/1659 to approve once you confirm that this should be removed.

acostach: @20k-ultra just to clarify, does the supervisor use the 'grub_extraenv' file specifically for the Intel NUC somewhere in the sources, or does it use the same file extra_uEnv.txt for both jetson and NUC? It looks like it uses extra_uEnv.txt for the NUC too.

This PR https://github.com/balena-os/balena-intel/pull/311 mentions the grub environment file name "grub_extraenv"  and paths https://github.com/balena-os/balena-intel/pull/311/commits/5f8856bbef645b63cfa04593c5cad28f1586d467#diff-8ad673b0f5c6b06dca41dc93a32e63d4a66ff69ee5237718972a8687ae339084R20 , which aren't the same as for  ["extra_uEnv.txt"](https://github.com/balena-os/balena-supervisor/pull/1659/commits/e6eda0fca7b280aa5453972a26c36ab6895c227a#diff-7e32acc477451dde576c1cedba193ccc078e0f1068440be5b91571c059a3dc88L45).

I'm not aware of any change on the Intel NUC side, grub_extraenv should be the same as when it was added in [PR 311](https://github.com/balena-os/balena-intel/pull/311). So if we would want to have extra_os_cmdline="x,y,z" capability in SV for the intel NUC, https://github.com/balena-os/balena-supervisor/pull/1659/commits/e6eda0fca7b280aa5453972a26c36ab6895c227a#diff-7e32acc477451dde576c1cedba193ccc078e0f1068440be5b91571c059a3dc88L45 should mention these two paths specifically for the NUC:

```
<boot_partition>/grub/grub_extraenv \
<boot_partition>/EFI/BOOT/grub_extraenv \
```
 and leave untouched the extra_uEnv.txt file that's currently used for the jetson devices. 

extra_os_cmdline arguments would be the only option compatible with the NUC. odmdata and fdt are not supported on the x86 platform for the reasons I already mentioned.

20k-ultra: the extra_uenv backend does not use a different file for the intel nuc. I've removed it from this backend with https://github.com/balena-os/balena-supervisor/pull/1659 which is merged. Thanks for pointing this out. I guess I added it my some misunderstanding when I first wrote the feature.

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+This device type utilizes a `config.txt` [file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) which we offer some values to be configured.
+
+Name | Type | Default | Description | Valid from
+--- | --- | --- | --- | ---
+BALENA_HOST_CONFIG_disable_splash | integer | 1 | Enable / Disable the Raspberry Pi rainbow splash screen. | v1.0.0
+BALENA_HOST_CONFIG_dtparam | string | "i2c_arm=on","spi=on","audio=on" | Define DT parameters. | v1.0.0
+BALENA_HOST_CONFIG_gpu_mem | integer | 16 | Define device GPU memory in megabytes. | v1.0.0
+
+You can find more information on updating `config.txt` through configuration variables in our [Advanced Boot Configuration Guide][boot-config-guide].
```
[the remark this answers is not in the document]
20k-ultra: I left this because when the doc server compiles the markdown into html it will translate these variable into links. I see now though that this variable is not a global value and so will not get translated from the Supervisor docs.

docs/configurations.md
```diff
@@ -0,0 +1,77 @@
...
+
+## ConfigFS
+
+> Supported boards: UP Board
+
+This device type utilizes a `configFS` [file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) interface that allows definition of arbitrary functions and configurations.
```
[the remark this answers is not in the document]
20k-ultra: great catch!! The woes of copy pasta.

docs/configurations.md
20k-ultra: I see what you're saying but think strings is fine because "value1","value2" is still a string.

docs/configurations.md
20k-ultra: I've added the rest except enable_uart as it is not in the config.txt backend however it is a valid configuration otherwise just not one the Supervisor will apply https://github.com/balena-os/balena-supervisor/blob/2b30ce3b49f7286c74aa86556fad1f731223e783/src/config/backends/config-txt.ts#L32-L38

cywang117 (approved):
LGTM! :)

docs/configurations.md
```diff
@@ -0,0 +1,81 @@
...
+| BALENA_HOST_DISCOVERABILITY | boolean | true | Enable / Disable Avahi to run which will allow the device to respond to requests such as network scans. | v11.9.2 |
+| BALENA_HOST_SPLASH_IMAGE | integer | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
+
+---
+
+In addition to these values, there may be some device-type specific configuration variables that can be set.
+
+## Config.txt
+
+> Supported boards: Any Raspberry Pi and BalenaFin.
```
vipulgupta2048, line 36:
```suggestion
> Supported boards: Any Raspberry Pi and balenaFin.
```

docs/configurations.md
```diff
@@ -0,0 +1,81 @@
...
+
+This device type utilizes a `config.txt` [file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) of which we offer some values to be configured. The following values are not all the configurations found in the `config.txt` file and are just Balena additions or edits to the OEM version.
```
vipulgupta2048, line 38:
```suggestion
This device type utilizes a [`config.txt` file](https://www.raspberrypi.org/documentation/configuration/config-txt/README.md) of which we offer some values to be configured. The following values are not all the configurations found in the `config.txt` file and are just balena additions or edits to the OEM version.
```

Checkout the style guide: https://github.com/balena-io/docs#style-manual

vipulgupta2048 (commented):
Looking good.

20k-ultra:
we'll try to implement this in a contract https://github.com/balena-os/balena-supervisor/issues/1670

20k-ultra:
let's merge this @cywang117. It's too valuable to be lost as a closed PR.

20k-ultra:
I have added reboot required to the tables. All the backends are currently marked as needing a reboot because of this logic: https://github.com/balena-os/balena-supervisor/blob/d9223bd460fd4e3c7fba501e1313f7ed94bdfb25/src/device-config.ts#L606-L613

not sure if there are some optimization to make here so we don't always restart if it's not necessary.

cywang117:
@20k-ultra Definitely. I know that ideally, we should be shooting for the optimal long-term solution, but in this case, I agree that this "intermediary" step is too valuable to lose. Thanks for all your work on this :)

Issue #1609, opened 2021-03-04
Not detecting storage device plugged in with USB


20k-ultra:
It is possible to run devices such as a rpi on a storage device which is plugged in via a USB port. The Supervisor fails to detect these types of devices and the dashboard will show no storage drive.

The culprit is probably the regex match https://github.com/balena-io/balena-supervisor/blob/dc96cca5e5fddc277d5a466bcd17654ccce67995/src/lib/system-info.ts#L26

anujdeshpande:
hey @20k-ultra - let me know if I can do anything to help with this

20k-ultra:
Available in Supervisor v.12.5.8.

PR #1651, opened 2021-04-06
Improved mutable (/data) file system detection

20k-ultra:
Change-type: patch
Closes: #1609


20k-ultra:
there's some weird permission issue going on because I can't run my test locally anymore....I keep getting EACCESS on my /test/data files

src/lib/system-info.ts
```diff
@@ -1,17 +1,12 @@
...
 import { fs, child_process } from 'mz';
 
-export function getCpuUsage(): Promise<number> {
-	return new Promise((resolve) => {
-		osUtils.cpuUsage((percent) => {
-			resolve(Math.round(percent * 100));
-		});
-	});
+export async function getCpuUsage(): Promise<number> {
+	const cpuData = await systeminformation.cpuCurrentspeed();
```
pipex, line 6: Do `cpuCurrentSpeed` and `cpuUsage` return the same value?  `cpuCurrentSpeed` seems to return a clock frequency while the dashboard expects a load value?

pipex: There is a `getLoad` function on systeminformation which might help us get rid of the os-utils dependency https://github.com/sebhildebrandt/systeminformation/blob/master/lib/cpu.js#L1408

src/lib/system-info.ts
```diff
@@ -23,13 +18,8 @@ export async function getStorageInfo(): Promise<{
 	// First we find the block device which the data partition is part of
 	for (const partition of fsInfo) {
 		if (partition.mount === '/data') {
-			const match = partition.fs.match(blockDeviceRegex);
```
pipex, line 26: I think is the right decision to remove the match. This seems to restrict only to balena block devices, this was bound to bite us sooner or later.

src/lib/system-info.ts
20k-ultra: We're using systeminformation v4 so limited to https://systeminformation.io/v4/cpu.html. It's possible this was why we used another module to fetch cpu usage if v4 for systeminformation cannot do it. I had assumed the cpu speed was the same as usage.

20k-ultra: the speed here is not a calculation of utilization though so I need to find another way to get this value.

pipex: Why can't we update to system information v5?

20k-ultra: We could but I found a v4 function in the source code that gave us load. Their documentation is just missing stuff..

src/lib/system-info.ts
```diff
@@ -1,17 +1,12 @@
...
 
-export function getCpuUsage(): Promise<number> {
-	return new Promise((resolve) => {
-		osUtils.cpuUsage((percent) => {
-			resolve(Math.round(percent * 100));
-		});
-	});
+export async function getCpuUsage(): Promise<number> {
+	const cpuData = await systeminformation.currentLoad();
+	return Math.round(cpuData.avgLoad * 100);
```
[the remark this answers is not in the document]
pipex: Wait I think I figured it out on https://github.com/sebhildebrandt/systeminformation/blob/master/lib/cpu.js#L1412-L1413, avgLoad is calculated as 

```
      let loads = os.loadavg().map(function (x) { return x / util.cores(); });
      let avgLoad = parseFloat((Math.max.apply(Math, loads)).toFixed(2));
``
[os.loadavg()](https://nodejs.org/api/os.html#os_os_loadavg) returns the 1,5 and 15 minute averages and the loads divides that into the number of cores and and avgLoad is the max of that calculation. I don't really understand the logic there. 

Maybe we can look at both values and see how the values compare to the output of top or htop?

**Review by pipex — CHANGES_REQUESTED** [review]
Looks good but I'd like to make sure that we are returning the correct load value

File: src/lib/system-info.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (20k-ultra) [reply src/lib/system-info.ts:7]: avgLoad is over time and currentLoad is on that cycle. The interval of each cycle is probably < 2 seconds.

When I chose avgLoad I did watch what the Supervisor was saying the cpu load was while monitoring top and they matches ~1% which I think was just based on round implementation. So top would say 19 and SV said 18.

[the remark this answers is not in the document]
Comment (pipex) [reply src/lib/system-info.ts:7]: But if I understand correctly, if the device was under a high period of load, and then load drops, since `avgLoad` also considers the 15 minute load, then it would take some time to push the correct value to the cloud. Would you agree?

File: src/lib/system-info.ts [hunk 4]
```diff
@@ -1,17 +1,15 @@
...
 import * as _ from 'lodash';
 import { fs, child_process } from 'mz';
 
-export function getCpuUsage(): Promise<number> {
-	return new Promise((resolve) => {
-		osUtils.cpuUsage((percent) => {
-			resolve(Math.round(percent * 100));
-		});
-	});
+export async function getCpuUsage(): Promise<number> {
```
pipex, line 5: 🙌

pipex (approved):
LGTM

PR #1655, opened 2021-04-09
patch: Fix substring end parameter to get accurate CPU ID

vipulgupta2048:
The `end` parameter of the substring inside the `getCpuId()` function was trimming the last character of the CPU ID. This data was showing up in the API as inaccurate (presumably because maybe the Supervisor API is the one providing the data). 

Screenshot from 2021-04-06 15-25-37

From the screenshot below, one can observe the presence of a null byte after the CPU ID, when running the command `cat -A /proc/device-tree/serial-number`. I tried to `trim()` the null byte but it was no good. So was looking for better solutions than using `substring` to trim the last character. The only other solution described was the use of regex where the case was for multiple null bytes present in the string. 

Screenshot from 2021-04-09 19-51-53

Would like a solution better than substring though, recommendations welcome. 

cc: @Page- 
Closes: https://github.com/balena-io/open-balena-api/issues/631 (I was going to transfer but someone moved the repository to balenaOS 🙃)
Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>

### DIFF
--- a/src/lib/system-info.ts
+++ b/src/lib/system-info.ts
@@ -69,7 +69,7 @@ export async function getCpuId(): Promise<string | undefined> {
 	try {
 		const buffer = await fs.readFile('/proc/device-tree/serial-number');
 		// Remove the null byte at the end
-		return buffer.toString('utf-8').substr(0, buffer.length - 2);
+		return buffer.toString('utf-8').replace(/\0/g, '');
 	} catch {
 		return undefined;
 	}
--- a/test/38-sys-info.spec.ts
+++ b/test/38-sys-info.spec.ts
@@ -119,7 +119,7 @@ describe('System information', async () => {
 				]),
 			);
 			const cpuId = await sysInfo.getCpuId();
-			expect(cpuId).to.equal('1000000001b93f3');
+			expect(cpuId).to.equal('1000000001b93f3f');
 			fsStub.restore();
 		});
 	});


vipulgupta2048:
@balena-ci rebase

src/lib/system-info.ts
```diff
@@ -69,7 +69,7 @@ export async function getCpuId(): Promise<string | undefined> {
 	try {
 		const buffer = await fs.readFile('/proc/device-tree/serial-number');
 		// Remove the null byte at the end
-		return buffer.toString('utf-8').substr(0, buffer.length - 2);
+		return buffer.toString('utf-8').substr(0, buffer.length - 1);
```
20k-ultra, line 72:
```suggestion
		return buffer.toString('utf-8').replace(/\0/g, '')
```

20k-ultra (commented):
We also have to fix the test.. the expect CPU id is also missing the last char so should be `1000000001b93f3f`

https://github.com/balena-os/balena-supervisor/blob/6c0df43f1ad4f7d3852284039e3ca3d169e1d716/test/38-sys-info.spec.ts#L122

src/lib/system-info.ts
vipulgupta2048: LGTM, resolving.

20k-ultra (approved):

Issue #1618, opened 2021-03-09
Failure to validate IPAM network configuration


20k-ultra:
The function for [validating](https://github.com/balena-io/balena-supervisor/blob/master/src/compose/network.ts#L211-L224) IPAM network configurations is not working correctly and is leading to undefined behavior. In the case we had seen releases were failing to start containers because the Supervisor was getting the following error:

```
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [debug]   Attempting to create network {"name":"default","appId":1800203,"config":{"driver":"bridge","ipam":{"driver":"default","config":[{"subnet":"172.29.238.0/24"},{"subnet":"2001:3984:3989::/64"}],"options":{}},"enableIPv6":false,"internal":false,"labels":{},"options":{}}}
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [debug]   Creating network:  {"name":"default","appId":1800203,"config":{"driver":"bridge","ipam":{"driver":"default","config":[{"subnet":"172.29.238.0/24"},{"subnet":"2001:3984:3989::/64"}],"options":{}},"enableIPv6":false,"internal":false,"labels":{},"options":{}}}
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [error]   Scheduling another update attempt in 64000ms due to failure:  Error: Failed to apply state transition steps. network.create is not a function Steps:["start","createNetwork"]
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [error]         at /usr/src/app/src/device-state.ts:1594:10
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [error]       at processTicksAndRejections (internal/process/task_queues.js:97:5)
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [error]   Device state apply error Error: Failed to apply state transition steps. network.create is not a function Steps:["start","createNetwork"]
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [error]         at /usr/src/app/src/device-state.ts:1594:10
Mar 06 01:01:34 69f6f22 resin-supervisor[7887]: [error]       at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Upon further investigation we saw that the release had an malformed IMAP configuration for the docker-compose file however, since the Supervisor failed to catch this invalid config it cause a variable to not be initialized as the correct type and creating the above error.

20k-ultra:
[...]

PR #1656, opened 2021-04-09
Fix broken IPAM network validation

pipex:
Network validaton was failing to identify a bad IPAM network
configuration leading to supervisor failures (see #1618).

This PR also modifies the mocha configuration to perform recursive lookup of `*.spec.ts` files

Change-type: patch
Closes: #1618

### DIFF
--- a/test/compose/network.spec.ts
+++ b/test/compose/network.spec.ts
@@ -1,0 +1,112 @@
+import ChaiConfig = require('../lib/chai-config');
+const { expect } = ChaiConfig;
+
+import { Network } from '../../src/compose/network';
+
+describe('Network', () => {
+	describe('fromComposeObject', () => {
+		it('creates a default network configuration if no config is given', () => {
+			const network = Network.fromComposeObject('default', 12345, {});
+
+			expect(network.name).to.equal('default');
+			expect(network.appId).to.equal(12345);
+
+			// Default configuration options
+			expect(network.config.driver).to.equal('bridge');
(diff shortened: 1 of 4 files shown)


test/compose/network.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+						driver: 'default',
+						config: [
+							{
+								subnet: '172.20.0.0/16',
+							},
+						],
+						options: {},
+					},
+				}),
+			).to.throw();
```
20k-ultra, line 91: Can you modify the `.to.throw()` to be `.to.throw('Network IPAM config entries must have both a subnet and gateway')`. This will ensure that whatever is being thrown is useful and not some cryptic message that will make it hard for us to debug.

test/compose/network.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+						driver: 'default',
+						config: [
+							{
+								gateway: '172.20.0.1',
+							},
+						],
+						options: {},
+					},
+				}),
+			).to.throw();
```
20k-ultra, line 105: Remember this one too

20k-ultra:
self certify after if you'd like :+1:

test/compose/network.spec.ts
pipex: That's a good consideration, I'll do that

pipex:
@balena-ci I self certify!

Issue #1650, opened 2021-04-05
Convert /src/device-api/common.js to TypeScript


cywang117:
Most recently, [this PR](https://github.com/balena-os/balena-supervisor/pull/1635#discussion_r607009527) has encountered friction while dealing with common.js as a non-TypeScript module. I'm sure there have been other instances of friction in the past. It would benefit everyone greatly and reduce friction for future developers if we could just convert it once and for all.

20k-ultra:
agreed. This and https://github.com/balena-os/balena-supervisor/blob/master/src/proxyvisor.js are the last 2 js files.

cywang117:
Closing in favor of #1660.

PR #1657, opened 2021-04-13
Convert device-api/common.js to TypeScript

cywang117:
I've removed the `serviceAction` function - it doesn't seem to be used anywhere. Let me know if I should revert this removal.

Change-type: patch
Closes: #1650

### DIFF
--- a/src/device-api/common.ts
+++ b/src/device-api/common.ts
@@ -3,13 +3,19 @@ import * as _ from 'lodash';
 import { appNotFoundMessage } from '../lib/messages';
 import * as logger from '../logger';
 
+import { InstancedDeviceState } from '../types/state';
+import { App } from '../compose/app';
+
 import * as deviceState from '../device-state';
 import * as applicationManager from '../compose/application-manager';
 import * as serviceManager from '../compose/service-manager';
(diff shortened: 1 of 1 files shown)


20k-ultra:
Let's consider just removing this file entirely rather then converting to TS. If we can make the code that uses it to talk to the application manager directly that would be the best. It already knows how to restart a service we may just need to add the purge but there is a [cleanup] step but it looks to just remove images ? which probably removes their containers and volumes...

cywang117:
@20k-ultra See reference: https://github.com/balena-os/balena-supervisor/pull/1635#discussion_r605372192 a cleanup step never gets generated for `doPurge`. Also we want to keep images, just remove their containers & associated volumes, Perhaps if we want to combine `doPurge` into application manager we could create a new composition step, `purge`, that handles purge specifically?

20k-ultra:
Yup, creating a purge step is exactly what I meant. I was just mentioning that maybe the clean step may already do that but it didn't look like it since it was removing images.

cywang117:
Great idea! I'll close this PR so we can rethink / work on this idea - I'll also create & link an issue.

PR #1659, opened 2021-04-14
Refactor extra_uEnv to not match with intel nuc

20k-ultra:
From comment https://github.com/balena-os/balena-supervisor/pull/1648#discussion_r606252023 on another PR where I documented that the intel nuc will match with extra_uEnv @acostach said this is not true (reason in that comment). This PR makes the intel nuc not match with this backend.


20k-ultra:
@balena-ci I self-certify!

PR #1662, opened 2021-04-16
Change 'require' to 'import * from' for dependencies

pipex:
This refactors imports for all typescript compatible modules from
`require`, to `import * from`. This change has no effect in code
execution, it's only about code consistency.

This has the added benefit of allowing us to identify all non-typescript
external modules and look for possible alternatives.

Change-type: patch

### DIFF
--- a/src/compose/volume.ts
+++ b/src/compose/volume.ts
@@ -1,9 +1,6 @@
 import * as Docker from 'dockerode';
-import assign = require('lodash/assign');
-import isEqual = require('lodash/isEqual');
-import omitBy = require('lodash/omitBy');
-
-import constants = require('../lib/constants');
+import { assign, isEqual, omitBy } from 'lodash';
+import * as constants from '../lib/constants';
 import { docker } from '../lib/docker-utils';
 import { InternalInconsistencyError } from '../lib/errors';
 import * as LogTypes from '../lib/log-types';
--- a/src/lib/blink.ts
+++ b/src/lib/blink.ts
@@ -1,5 +1,6 @@
-import blinking = require('blinking');
+import blinking = require('blinking'); // blinking is a 'coffee' module and can only be imported this way
+import * as constants from './constants';
 
-import constants = require('./constants');
+const blink = blinking(constants.ledFile);
 
-export = blinking(constants.ledFile);
+export default blink;
--- a/src/lib/request.ts
+++ b/src/lib/request.ts
@@ -1,12 +1,12 @@
 import * as Bluebird from 'bluebird';
-import once = require('lodash/once');
+import * as _ from 'lodash';
 import * as requestLib from 'request';
 import * as resumableRequestLib from 'resumable-request';
 
 import * as constants from './constants';
 import * as osRelease from './os-release';
 
-import supervisorVersion = require('./supervisor-version');
+import supervisorVersion from './supervisor-version';
 
 export { requestLib };
 
@@ -40,7 +40,7 @@ type PromisifiedRequest = typeof requestLib & {
 	) => Bluebird<[requestLib.Response, any]>;
 };
 
-const getRequestInstances = once(async () => {
+const getRequestInstances = _.once(async () => {
 	// Generate the user agents with out versions
 	const osVersion = await osRelease.getOSVersion(constants.hostOSVersionPath);
 	const osVariant = await osRelease.getOSVariant(constants.hostOSVersionPath);
@@ -81,14 +81,14 @@ const getRequestInstances = once(async () => {
 	};
 });
 
-export const getRequestInstance = once(async () => {
+export const getRequestInstance = _.once(async () => {
 	return (await getRequestInstances()).request;
 });
 
-export const getRequestOptions = once(async () => {
+export const getRequestOptions = _.once(async () => {
 	return (await getRequestInstances()).requestOpts;
 });
 
-export const getResumableRequest = once(async () => {
+export const getResumableRequest = _.once(async () => {
 	return (await getRequestInstances()).resumable;
 });
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -16,7 +16,7 @@ import {
 } from './composition-steps';
 import * as targetStateCache from '../device-state/target-state-cache';
 import * as dockerUtils from '../lib/docker-utils';
-import constants = require('../lib/constants');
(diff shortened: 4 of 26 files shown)


src/lib/blink.ts
```diff
@@ -1,5 +1,6 @@
-import blinking = require('blinking');
+import blinking = require('blinking'); // blinking is a 'coffee' module and can only be imported this way
```
cywang117, line 1: This [blinking module](https://github.com/balena-io-modules/blinking) is actually a bit of a problem. It's an Javascript module with no typings first of all, and we also don't have type declaration file for it. It also hasn't been updated in 5 years. This [TODO comment](https://github.com/balena-os/balena-supervisor/blob/master/test/07-blink.spec.ts#L16) from two years ago has remained all this time. In my test refactor branch that I've been looking at, the blink test straight up doesn't pass TypeScript checks :|

Nothing wrong with your change here, but you were mentioning figuring out which external modules aren't TypeScript-safe and seeing if there were any alternatives, and this module is one I've experienced friction with in the past.

src/lib/request.ts
```diff
@@ -1,12 +1,12 @@
 import * as Bluebird from 'bluebird';
-import once = require('lodash/once');
+import * as _ from 'lodash';
```
cywang117, line 2: If we're looking for space improvements on the order of kilobytes, `import once from 'lodash/once` results in a smaller size than `import * as _ from 'lodash'`. [Here](https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark) is an interesting article which, while it's talking about bundle size on the frontend, can easily apply to the backend as well.

src/compose/volume.ts
```diff
@@ -1,16 +1,15 @@
 import * as Docker from 'dockerode';
-import assign = require('lodash/assign');
-import isEqual = require('lodash/isEqual');
-import omitBy = require('lodash/omitBy');
```
Page-, line 4: The usual reason to import lodash as `lodash/$x` is in order to tree shake and drop unused code for a smaller webpack output, using straight `lodash` requires importing the entirety of lodash which is quite large. I'm assuming that's the reason the supervisor had these as individual imports rather than importing everything

src/compose/volume.ts
```diff
@@ -1,16 +1,15 @@
...
+import * as _ from 'lodash';
+import * as constants from '../lib/constants';
 import { docker } from '../lib/docker-utils';
 import { InternalInconsistencyError } from '../lib/errors';
 import * as LogTypes from '../lib/log-types';
 import { LabelObject } from '../lib/types';
 import * as logger from '../logger';
 import * as ComposeUtils from './utils';
 
+const { assign, isEqual, omitBy } = _;
```
Page-, line 11: If you're going to import all of lodash/destructure it'd be much better to do
```suggestion
import { assign, isEqual, omitBy } from 'lodash';
```

src/lib/blink.ts
Page-: It hasn't been updated in so long because it hasn't needed to be :P  It's our own module as well (it was part of the supervisor before being separated out) so it's easy enough to convert to js and/or typescript and have it generate types from one of those

Page- (commented):
Btw just so you know there's no particular benefit atm to using `import = require` vs `import *` because typescript will transpile them both to the same thing

src/lib/blink.ts
Page-: https://github.com/balena-io-modules/blinking/pull/5

pipex: Nice one @Page-

For context, the only reason I left this comment is to leave a reminder to myself to remove the dependency or update the module to typescript. I will probably edit and make the same with other non-typescript dependencies

src/lib/request.ts
pipex: Yeah, I know this is probably more expensive, but the supervisor already imports lodash with `import * as _ from 'lodash'` in most places so changing it here wont have any effect.

src/compose/volume.ts
pipex: Shouldn't tree shake work either way? If I do

```
import module from 'mod';

mod.somefunc()
```

Shouldn't tree shaking get rid of everything except for `somefunc` and functions it requires?

src/compose/volume.ts
pipex: I tried to do that, but doesn't work without enabling `esModuleInterop`. Glad you mention it though, can you think of any drawbacks to enabling that option on tsconfig?

src/compose/volume.ts
pipex: > Btw just so you know there's no particular benefit atm to using import = require vs import * because typescript will transpile them both to the same thing

@Page- I know, the only reason I'm doing this is because I'm trying to generate a dependency tree diagram and these `require` imports show as `unresolved module` for some reason I haven't been able to figure out. Is a bonus that we get consistent code :)

See this thread for more context https://www.flowdock.com/app/rulemotion/r-supervisor/threads/ophCVmMO6NIY5axDZ4EkKe2focX

Page-: Tree-shaking only works on a per-export method for es-modules because it requires the semantic restrictions of the es-module syntax in order to prove the tree-shaking is safe, and as lodash isn't actually an es-module (ts just emulates it) it needs to include the entire module

src/compose/volume.ts
Page-: I don't think you need to enable `esModuleInterop` for it to work - I've seen it elsewhere without needing to enable `esModuleInterop`. Maybe you need to update lodash typings?

pipex: Oh, you are right, I think I got it confused with the `import assign from 'lodash/assign'` and its variants that do require esModuleInterop

src/lib/blink.ts
cywang117: @Page- Thanks for the speedy update Page!

src/compose/volume.ts
pipex: @Page- I found this article https://www.azavea.com/blog/2019/03/07/lessons-on-tree-shaking-lodash/ and it looks like lodash4 is exported as true esModule. 

I was also looking at our webpack config and I don't think we do any tree-shaking, only minimization. Maybe we could enable it?

Issue #1567, opened 2020-12-24
Remove dependency on mz package (?)


pipex:
The supervisor depends today on the [node modernize](https://github.com/balena-io/balena-supervisor/pull/1566) (mz) package, because of its promise support, but newer node versions have promise support builtin in the [fs package](https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_promises_api) and provide ways to promisify other operations such as `child_process.exec` (see the promisify example [here](https://nodejs.org/docs/latest-v12.x/api/child_process.html#child_process_child_process_exec_command_options_callback))

A code review to see what other functionality is used from that package might tell us the difficulty of this change.

cywang117:
@pipex Did a quick code review, current `mz` methods used can easily be replaced by a combination of native `fs.promises` and `util.promisify`. See table below:
mz | Node fs | Node child_process | is promise lib?
----|---------|---------------------|-------------
child_process.exec | | exec | N
child_process.spawn | | spawn | N
fs.exists | exists (DEPRECATED, should create a helper that has the same functionality using `fs.access` or `fs.stat`) | | N
fs.fsync | fsync | | N
fs.lstat | lstat | | Y
fs.mkdir | mkdir | | Y
fs.open | open | | Y
fs.readdir | readdir | | Y
fs.readFile | readFile | | Y
fs.rename | rename | | Y
fs.stat | stat | | Y
fs.unlink | unlink | | Y
fs.watch | watch | | N
fs.write | write | | N
fs.writeFile | writeFile | | Y

Looks like a straightforward fix to decrease Supervisor's binary size *very* slightly.

20k-ultra:
I love removing dependencies. Some of the standard lib functions also provide performance buffs!

PR #1663, opened 2021-04-16
Remove mz, mkdirp, body-parser dependencies

cywang117:
'mz' can be safely replaced with native `fs.promises` and native `util.promisify` for faster native methods.

'mkdirp' after Node v8 uses native `fs.(promises.)mkdir(path, { recursive: true})`, thus is redundant. 

'body-parser' is deprecated and [contained within express v4.x](http://expressjs.com/en/api.html).

Additionally, I've replaced instances of `fs.exists` with `fsUtils.existsAsync`, which is a refactor of `fsUtils.pathExistsOnHost`. This is because `fs.exists` is [deprecated](https://nodejs.org/docs/latest-v12.x/api/fs.html#fs_fs_exists_path_callback) in v12.

This reduces our binary size very slightly, and will likely speed up execution time slightly as well because of Node-native method usage. This will also help us on the path to better unit tests, as I've noticed that certain tests import 'fs' while others import 'mz' to be stubbed. By removing 'mz' we'll save time in this aspect.

Closes: #1567
Change-type: patch


src/device-state.ts
```diff
@@ -468,7 +467,7 @@ function emitAsync<T extends keyof DeviceStateEvents>(
 		: Array<DeviceStateEvents[T]>
 ) {
 	if (_.isArray(args)) {
-		return setImmediate(() => events.emit(ev as any, ...args));
+		return setImmediate(() => events.emit(ev as any, ...(args as any[])));
```
Page-, line 470: Why did this become necessary?

src/lib/fs-utils.ts
```diff
@@ -1,21 +1,25 @@
...
 import { ENOENT } from './errors';
 
+export const execAsync = promisify(exec);
+
 export function writeAndSyncFile(
 	path: string,
 	data: string | Buffer,
 ): Bluebird<void> {
-	return Bluebird.resolve(fs.open(path, 'w')).then((fd) => {
+	return Bluebird.resolve(fs.open(path, 'w')).then((fh) => {
```
Page-, line 16: Fwiw I normally see `fd` used as in file descriptor and so I'd stick with `fd` as it's what I've always seen to name a "file descriptor" and makes it easier for me when reading the code (I know it is technically a file handle, as in a data structure that contains a file descriptor, but I've never seen people bother to differentiate at the var name level before)

src/lib/fs-utils.ts
```diff
@@ -31,12 +35,10 @@ export function writeFileAtomic(
...
+		.tap((fh) => fh.sync())
+		.then((fh) => fh.close());
 }
 
-export function pathExistsOnHost(p: string): Bluebird<boolean> {
-	return Bluebird.resolve(fs.stat(Path.join(constants.rootMountPoint, p)))
-		.return(true)
-		.catchReturn(ENOENT, false);
+export function existsAsync(p: string): Bluebird<boolean> {
+	return Bluebird.resolve(fs.stat(p)).return(true).catchReturn(ENOENT, false);
```
Page-, line 43: `fs.access` is the recommended way of checking a file exists over fs.stat: https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fs_stat_path_options_callback
> To check if a file exists without manipulating it afterwards, fs.access() is recommended.

```suggestion
	return Bluebird.resolve(fs.access(p)).return(true).catchReturn(ENOENT, false);
```

src/lib/fs-utils.ts
```diff
@@ -31,12 +35,10 @@ export function writeFileAtomic(
 export function safeRename(src: string, dest: string): Bluebird<void> {
 	return Bluebird.resolve(fs.rename(src, dest))
 		.then(() => fs.open(Path.dirname(dest), 'r'))
-		.tap(fs.fsync)
-		.then(fs.close);
+		.tap((fh) => fh.sync())
+		.then((fh) => fh.close());
 }
 
-export function pathExistsOnHost(p: string): Bluebird<boolean> {
```
Page-, line 38: Having looked at the code I think it's worth having a helper function for `pathExistsOnHost` to make it clearer when a check is looking at the host fs and centralize the logic of how to check host fs vs container fs in a single place rather than copying that logic in multiple places where a mistake might get introduced

src/device-state.ts
cywang117: @Page it was popping up an error whenever I opened the file, this was for OCD purposes (it doesn't result in failed lint checks though)

src/lib/fs-utils.ts
cywang117: Makes sense.

src/lib/fs-utils.ts
cywang117: Sounds good, I'll change this

cywang117:
The failing test in the aarch64 suite is quite strange, since I can't repro locally. As we all know, the unit tests can definitely be improved, and this might be another instance - a test with a side effect?

src/lib/fs-utils.ts
cywang117: I would keep it as is normally, but this change was made to reduce confusion, as VSCode Intellisense was showing  `fd` as type fs.FileHandle`. So fh seemed more appropriate.

src/lib/fs-utils.ts
cywang117: I'll be working on this a bit more, will ping you again when I've pushed a proposed refactor if the refactor ends up being small enough

src/lib/fs-utils.ts
20k-ultra: I agree with Page but I suppose fh is the most correct so...I'm for `fh`

src/device-state.ts
20k-ultra: What was the linter/transpiler yelling exactly ? Seems odd that it's complaining now.

cywang117: @20k-ultra 
```
Type 'DeviceStateEvents[T] extends (...args: any) => void ? Parameters<DeviceStateEvents[T]> : DeviceStateEvents[T][]' must have a '[Symbol.iterator]()' method that returns an iterator. ts(2488)
```

It's one of those red squiggly line errors that don't fail lint checks or transpilation

20k-ultra: Try to see what you can find out when searching 'must have a '[Symbol.iterator]()' method that returns an iterator.". Maybe there's something we should do.

pipex: The function types are pretty strange, my guess is that they are left from an older version of typescript, maybe you xcan try something like 

```
// We keep compatibility with the StrictEventEmitter types
// from the outside, but within this function all hells
// breaks loose due to the liberal any casting
function emitAsync<T extends keyof DeviceStateEvents>(
	ev: T,
	...args: DeviceStateEvents[T] extends (...args: infer TArgs) => void
		? TArgs
		: Array<DeviceStateEvents[T]>
) {
	if (_.isArray(args)) {
		return setImmediate(() => events.emit(ev as any, ...args));
	} else {
		return setImmediate(() => events.emit(ev as any, args));
	}
}
```

src/lib/fs-utils.ts
pipex: I'm for `fd`. Whenever I see `fd` in a piece of code I know that code is dealing with a file (for me it comes from writing C), which makes it more readable at least for me.

src/device-state.ts
pipex: I'm not 100% sure, but I think the `isArray` check is not necessary, since `args` should always be an array given the `...`, even if the function is called with a single argument

src/lib/fs-utils.ts
cywang117: It seems we have a 50/50 split. With regards with `fh` vs `fd` being more familiar to devs, from my experience with Python, I am more familiar with `filehandle`. You can really make the argument for either one based on past experience, but we are writing a TypeScript module. Compromise: for *absolute* clarity how about I rename this variable to `file`?

src/device-state.ts
cywang117: @pipex Thanks! This works, and also you're right, the isArray check isn't needed.

src/lib/fs-utils.ts
cywang117: I've centralized the various fs calls within the codebase to use fs-utils. Examples include: 
- Replacing any `fs.stat` calls that are for the purpose of checking file existence with `existsAsync`
- Replacing multiple unlink calls for deleting files with a new fs-utils function, unlinkAll, which takes one or more filenames as arguments. This is because everywhere unlink is being called, it needed to have a .catch to handle when the file doesn't exist. `unlinkAll` does this in one place so we can deduplicate this functionality.
- Replace multiple calls to `mkdir(dirName, { recursive: true })` with an annotated fs-utils `mkdirRecur` to be clear about this functionality
- Refactor fs-utils to use `async`/`await` and native Promises instead of Bluebird

Misc:
- Remove an unused function from config/configJson.ts

Untouched fs functionality:
- pathOnHost() in config/configJson.ts, as it has extra context-specific calls that would make replacing it with a generic fs-util function unnecessarily convoluted.
- remountAndWriteAtomic() in config/backend.ts, as it's used only in the context of the device backends and probably won't see use anywhere else.

What I interpreted from your comment @Page- was the idea of centralizing various vague or multi-step fs functionalities, such as fs.unlink needing a catch whenever it's called. AFAIK there are no uses for `pathExistsOnHost` to deal with container FS's. I have added a comment under pathExistsOnHost in an attempt to clarify it.

cywang117:
Additionally I've removed the imports of the 'chai-config' conf file and the destructuring of `expect` from it, as it's not necessary. (I realize that this PR is creeping a bit in scope)

test/src-test/lib/fs-utils.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+	const fsUtilsRoot = path.resolve(rootMountPoint, 'fs');
+	const testFile1 = path.join(fsUtilsRoot, 'temp-fs-utils-file.1');
+	const testFile2 = path.join(fsUtilsRoot, 'temp-fs-utils-file.2');
+	const testFile3 = path.join(fsUtilsRoot, 'temp-fs-utils-file.3');
+
+	const createTestFiles = async () => {
+		// Check directory existence before running tests
+		try {
+			await fs.access(fsUtilsRoot);
+		} catch (_e) {
```
Page-, line 19:
```suggestion
		} catch {
```

test/src-test/lib/fs-utils.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+		});
+
+		it('should return whether a file exists', async () => {
+			expect(await fsUtils.existsAsync(testFile1)).to.be.true;
+			await fs.unlink(testFile1).catch(() => {
+				/* noop */
+			});
+			expect(await fsUtils.existsAsync(testFile1)).to.be.false;
+		});
+	});
```
cywang117, line 130: Probable reason for this failing is that fs.access might be flipping out because of invalid absolute directory

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
 }
 
-export function safeRename(src: string, dest: string): Bluebird<void> {
-	return Bluebird.resolve(fs.rename(src, dest))
-		.then(() => fs.open(Path.dirname(dest), 'r'))
-		.tap(fs.fsync)
-		.then(fs.close);
+export async function existsAsync(_path: string): Promise<boolean> {
+	try {
+		await fs.access(_path);
```
Page-, line 42:
```suggestion
export async function existsAsync(path: string): Promise<boolean> {
	try {
		await fs.access(path);
```

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
+		return true;
+	} catch (e) {
```
Page-, line 44:
```suggestion
	} catch {
```

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
+		return false;
+	}
+}
+
+/**
+ * Check if a path exists as a direct child of the device's root mountpoint,
+ * which is equal to constants.rootMountPoint (`/mnt/root`).
+ */
+export function pathExistsOnHost(_path: string): Promise<boolean> {
+	return existsAsync(path.join(constants.rootMountPoint, _path));
```
Page-, line 54:
```suggestion
export function pathExistsOnHost(path: string): Promise<boolean> {
	return existsAsync(path.join(constants.rootMountPoint, path));
```

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
+}
+
+/**
+ * Recursively create directories until input directory.
+ * Equivalent to mkdirp package, which uses this under the hood.
+ */
+export async function mkdirRecur(_path: string): Promise<void> {
+	await fs.mkdir(_path, { recursive: true });
```
Page-, line 62:
```suggestion
export async function mkdirRecur(path: string): Promise<void> {
	await fs.mkdir(path, { recursive: true });
```

test/src-test/lib/fs-utils.spec.ts
cywang117: It's not. We just don't remove the build folder properly it between test runs, so it's likely that an old file/script was interfering somehow. After doing `rm -rf build`, the tests passed.

src/lib/fs-utils.ts
cywang117: Results in `no-shadowed-variable  Shadowed name: 'path'`. I'll rename this param to `pathname` to be more explicit, since `_path` or `Path` are semantically not as correct.

test/src/lib/fs-utils.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+		});
+	});
+
+	/**
+	 * TODO: Un-skip this test after all fs tests that write to a test file system use
+	 * mock-fs instead. Hypothesis: exists isn't handling the relative directory it's
+	 * being passed well. When all unit tests use mock-fs, we can set process.env.ROOT_MOUNTPOINT
+	 * to `/mnt/root` so we can have an absolute path in all these tests.
+	 */
+	describe.skip('existsAsync', () => {
```
cywang117, line 93: Ignore what I mentioned before about this test suite finally working with the `npm run clean` script, it still works with `npm run test:fast` but not `npm run test`. After exhausting all other possibilities, the most likely hypothesis for me is that this is because of the directory level differences between build/ and test/, even though theoretically this feels like this is not the case. In another PR this week I want to refactor all fs-using tests to use mock-fs, then set process.env.ROOT_MOUNTPOINT in `00-init.spec.ts` to an absolute path instead of a relative one - I think this will solve the problem.

test/mocha.opts
```diff
@@ -1,3 +1,3 @@
 --exit
 --timeout 30000
-build/test/**/*.js
+build/test/**/*.spec.js
```
cywang117, line 3: Add `.spec` specificity to prevent any other files from running aside from those explicitly required by test suites, as some files might have side-effects immediately upon import.

pipex: I could've sworn i did this here https://github.com/balena-os/balena-supervisor/pull/1656/files#diff-7256b382c196e793feaa018e2a9d0d7a68fb0ede44d938a5c41663a21275140cL3

cywang117: @pipex  That change looks like it was `build/test/*.js` --> `build/test/**/*.js`, this one specifies further from that with `build/test/**/*.js` --> `build/test/**/*.spec.js`.

pipex: Yeah, I know, I meant to do on that PR what you did here

cywang117: Then it's fixed! :D

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
+ */
+export function pathExistsOnHost(pathname: string): Promise<boolean> {
+	return existsAsync(path.join(constants.rootMountPoint, pathname));
+}
+
+/**
+ * Recursively create directories until input directory.
+ * Equivalent to mkdirp package, which uses this under the hood.
+ */
+export async function mkdirRecur(pathname: string): Promise<void> {
```
pipex, line 61: Why not `mkdirp`?

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
+ * Check if a path exists as a direct child of the device's root mountpoint,
+ * which is equal to constants.rootMountPoint (`/mnt/root`).
+ */
+export function pathExistsOnHost(pathname: string): Promise<boolean> {
+	return existsAsync(path.join(constants.rootMountPoint, pathname));
+}
+
+/**
+ * Recursively create directories until input directory.
+ * Equivalent to mkdirp package, which uses this under the hood.
```
20k-ultra, line 59: Can remove this

src/network.ts
```diff
@@ -1,6 +1,6 @@
 import * as Bluebird from 'bluebird';
 import * as _ from 'lodash';
-import { fs } from 'mz';
+import * as fs from 'fs';
```
pipex, line 3:
```suggestion
import { promises as fs } from 'fs';
```
I like this better

pipex: Then you can still do `fs.lstat` below

src/lib/fs-utils.ts
```diff
@@ -1,42 +1,90 @@
...
-export function pathExistsOnHost(p: string): Bluebird<boolean> {
-	return Bluebird.resolve(fs.stat(Path.join(constants.rootMountPoint, p)))
-		.return(true)
-		.catchReturn(ENOENT, false);
+/**
+ * Get one or more paths as they exist in relation to host OS's root.
+ */
+export function getPathOnHost(p: string): string;
+export function getPathOnHost(p: string[]): string[];
+export function getPathOnHost(p: any): any {
```
20k-ultra, line 84:
```suggestion
export function getPathOnHost(path: string): string;
export function getPathOnHost(path: string[]): string[];
export function getPathOnHost(path: any): any {
```

20k-ultra (approved):

PR #1664, opened 2021-04-19
Update supervisor to typescript 4

pipex:
Change-type: patch

### DIFF
--- a/src/migrations/M00001.js
+++ b/src/migrations/M00001.js
@@ -21,7 +21,7 @@ exports.up = function (knex) {
 					console.log(
 						'Unable to locate config.json! Things may fail unexpectedly!',
 					);
-					resolve();
+					resolve(false);
 					return;
 				}
 				fs.readFile(configJsonPath, (err, data) => {
@@ -29,7 +29,7 @@ exports.up = function (knex) {
 						console.log(
 							'Failed to read config.json! Things may fail unexpectedly!',
 						);
-						resolve();
+						resolve(false);
 						return;
 					}
 					try {
@@ -44,13 +44,13 @@ exports.up = function (knex) {
 								);
 								return;
 							}
-							resolve();
+							resolve(false);
 						});
 					} catch (e) {
 						console.log(
 							'Failed to parse config.json! Things may fail unexpectedly!',
 						);
-						resolve();
+						resolve(false);
 					}
 				});
 			}).then(() => {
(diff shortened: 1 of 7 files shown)


pipex:
@Page- do you see any potential problems of upgrading to typescript 4? I wanted to update the blinking module to the typescript version you pushed and got some syntax error due to it being typescript 4 vs the supervisor being typescript 3.9, So i went ahead and updated the supervisor too :p

src/migrations/M00001.js
```diff
@@ -21,15 +21,15 @@ exports.up = function (knex) {
 					console.log(
 						'Unable to locate config.json! Things may fail unexpectedly!',
 					);
-					resolve();
+					resolve(false);
```
pipex, line 24: I added these because the typescript compiler was complaining with 

```
src/migrations/M00002.js:20:12 - error TS2794: Expected 1 arguments, but got 0. Did you forget to include 'void' in your type argument to 'Promise'?

20     return resolve();
              ~~~~~~~~~

  node_modules/typescript/lib/lib.es2015.promise.d.ts:33:34
    33     new <T>(executor: (resolve: (value: T | PromiseLike<T>) => void, reject: (reason?: any) => void) => void): Promise<T>;
                                        ~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'value' was not provided.
```

pipex:
@balena-ci retest

cywang117:
Weird that the checks passed on aarch64 but not i386 this time..

cywang117:
@pipex Looking at [TypeScript 4.0 breaking changes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#breaking-changes), the only one that might apply is [Operands for delete must be optional](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#operands-for-delete-must-be-optional)

pipex:
Nice find @cywang117. I didn't see get any compilation errors so I guess it should be fine

20k-ultra:
> Nice find @cywang117. I didn't see get any compilation errors so I guess it should be fine

Did you take a look at the changelog or just want the transpiler to catch issues ? I'm wondering if any other behaviour changes could be introduced too that weren't considered a breaking change and just a patch.

pipex:
Looking at the changelog i don’t anything about changes in transpilation and i fixed all compilation errors. I will do some end to end test just in case

pipex:
@balena-ci rebase

pipex:
I did some tests, I don't expect any surprises with this

20k-ultra (approved):

Issue #1645, opened 2021-04-01
Configurable var to stop sending hardware metrics


20k-ultra:
For users that are sensitive to bandwidth usage we can help by offering a configurable variable to toggle sending cpu/memory/temp/storage stats to the cloud. This feature can send data every 10 seconds at a minimum so it will definitely cause devices to make more noise.

We obtain the system information at https://github.com/balena-os/balena-supervisor/blob/16f981c478f26e7947a7f8bdc2a3d8333f3031e1/src/lib/system-info.ts#L96 which is reported in the reportCurrentState function https://github.com/balena-os/balena-supervisor/blob/16f981c478f26e7947a7f8bdc2a3d8333f3031e1/src/device-state/current-state.ts#L180.

We should step in at the reportCurrentState to not ask for system info if the device is configured not to send this. We just need to figure out what default value should be used. We can either set it to have null values OR don't even set the info key in the stateForReport.local object. 

When we do this we should add the configurable var to https://github.com/balena-os/balena-supervisor/issues/1646

20k-ultra:
The currently agreed solution is the send null data for each value. So, cpu null, mem null, etc. Not an empty object.

additionally, we found https://github.com/balena-os/balena-supervisor/issues/1669 would be another optimization.

cywang117:
Clarifying: send null once when flag is toggled on, update "last reported current state" as an in-memory object. Thus future reports which would contain null for sys info would filter it out based on state diff comparisons. 

@20k-ultra Does this PR depend on changes that should be made in #1669 to persist last reported current state?

cywang117:
Additionally, `is_undervolted` as a sysinfo value is not nullable in the API. I can only assume that we would not send an `is_undervolted` field with the null payload.

20k-ultra:
https://github.com/balena-os/balena-supervisor/issues/1669 is just an optimization for the Supervisor in general which we could add even without this metrics work. 

I would modify the `SystemInfo` type to allow null. There are some values which already allow undefined and I'd make those null instead of undefined as well. Then with `SystemInfo` now allowing null for all the values make a function which returns a disabled report that is used like:

```
const info = (await shouldReportMetrics())
		? await sysInfo.getSysInfoToReport()
		: sysInfo.disabledReport();
```

These are just my ideas of course, other solutions can be considered.

20k-ultra:
[...]

PR #1666, opened 2021-04-20
Add toggleable `SUPERVISOR_HARDWARE_METRICS` config

cywang117:
On devices with bandwidth sensitivity, this config var disables sending system information such as memory usage or cpu temp as current state.

Before we can merge this, we'll need to update open-balena-api to whitelist this config var, and think about how to convey that hardware metrics are greyed out on the dashboard. Currently, when SUPERVISOR_DISABLE_HARDWARE_METRICS is toggled off, the last known set of hardware metrics continues to be displayed on the dashboard summary page, which is misleading. We should discuss with balena-ui team about greying out the [DeviceMetrics.tsx](https://github.com/balena-io/balena-ui/blob/master/src/pages/dashboard/device/device-summary/DeviceMetrics.tsx) React component when DISABLE_HARDWARE_METRICS is false(y).

Also added test coverage for the current-state.ts module, as it's very important business logic that was previously untested,

Closes: #1645
Change-type: minor


pipex (commented):
Looks great @cywang117 

Just being nit-picky, any particular reason on making it `reportHardwareMetrics` instead of `disableHardwareMetrics`?

20k-ultra:
How the UI and API accept/display is out of scope for this PR. Simply making the toggle to stop sending data is enough. We don't have to wait for others to implement anything.

src/device-state/current-state.ts
```diff
@@ -177,7 +177,10 @@ const reportCurrentState = (): null => {
 		reportPending = true;
 		try {
 			const currentDeviceState = await deviceState.getStatus();
-			const info = await sysInfo.getSysInfoToReport();
+			const info = await sysInfo.getSysInfoToReport(
+				await config.get('reportHardwareMetrics'),
+			);
```
20k-ultra, line 182: I think adding the toggle to the report generate function adds necessary complexity to the signature when we can do:

```
	const info = (await config.get('reportHardwareMetrics'))
		? await sysInfo.getSysInfoToReport()
		: {};
```

My other concern with getting the config every time is if it adds another IO we can prevent. Especially since this happens every 10 seconds.. What if we do something like this to cache the value and update the cache as it changes:

```
let reportHardwareMetrics: boolean;

async function shouldReportMetrics() {
	if (!reportHardwareMetrics){
		reportHardwareMetrics =	await config.get('reportHardwareMetrics')
	}
	return reportHardwareMetrics
}

config.on('change', (changed) => {
	if (changed.reportHardwareMetrics != null) {
		// this value has changed
		reportHardwareMetrics = changed.reportHardwareMetrics
	}
});
```
then get info like:
```
	const info = (await shouldReportMetrics())
		? await sysInfo.getSysInfoToReport()
		: {};
```

src/lib/system-info.ts
```diff
@@ -75,7 +75,7 @@ export async function getCpuId(): Promise<string | undefined> {
 	}
 }
 
-const undervoltageRegex = /under.*voltage/;
+const undervoltageRegex = /[U|u]nder.*voltage/;
```
20k-ultra, line 78: what about this or some other flag to make the test case insensitive rather then making the pattern more complex.

```suggestion
const undervoltageRegex = new RegExp('under.*voltage', 'i');
```

test/38-sys-info.spec.ts
```diff
@@ -10,16 +10,34 @@ function toMb(bytes: number) {
...
+		fsSizeStub = stub(systeminformation, 'fsSize');
+		memStub = stub(systeminformation, 'mem').resolves(mockMemory);
+		currentLoadStub = stub(systeminformation, 'currentLoad').resolves(
+			mockCPU.load,
+		);
+		cpuTempStub = stub(systeminformation, 'cpuTemperature').resolves(
+			mockCPU.temp,
+		);
+		// @ts-ignore TS thinks we can't return a buffer...
+		fsStub = stub(fs, 'readFile').resolves(mockCPU.id);
```
20k-ultra, line 30: I'd prefer if the value was called `mockCPU.idBuffer` so it won't imply that Ids are always a buffer of bytes when really it's only a buffer because it was store in a file.

20k-ultra (changes requested):

20k-ultra:
Can you also delete https://github.com/balena-os/balena-supervisor/blob/master/test/data/serial-number in this PR. I added it by mistake before and the test changes are relevant to sys info.

cywang117:
@pipex 
> Just being nit-picky, any particular reason on making it `reportHardwareMetrics` instead of `disableHardwareMetrics`?

I hadn't thought about it, but it's an issue of semantics, opinion, and user readability I suppose. The user-facing config is <ins>true</ins> when Supervisor <ins>is</ins> sending hardware metrics, and <ins>false</ins> when Supervisor <ins>is not</ins>. I wouldn't mind `disableHardwareMetrics` though, if you feel strongly about it.

src/device-state/current-state.ts
cywang117: @20k-ultra This makes sense. Realistically SQLite3 should be able to handle many more queries per second than Supervisor is throwing at it, even with the reportCurrentState function making a SELECT query every 10 seconds, but we should also consider memory limitations on the device. While writing this PR I was thinking of whether to cache the reportHardware config value, so took a look at existing code to see if there were any existing examples. I'm down to make the change here, but if so, the caching change should also be made in the [actual report function](https://github.com/balena-os/balena-supervisor/blob/master/src/device-state/current-state.ts#L134-L140), which also fires every 10 seconds max.

cywang117: In fact we query for some of the same values [twice](https://github.com/balena-os/balena-supervisor/blob/master/src/device-state/current-state.ts#L71-L76) every 10 seconds :|

20k-ultra:
> @pipex
> 
> > Just being nit-picky, any particular reason on making it `reportHardwareMetrics` instead of `disableHardwareMetrics`?
> 
> I hadn't thought about it, but it's an issue of semantics, opinion, and user readability I suppose. The user-facing config is true when Supervisor is sending hardware metrics, and false when Supervisor is not. I wouldn't mind `disableHardwareMetrics` though, if you feel strongly about it.

I like `disableHardwareMetrics` as well just because it makes it more obvious that the metrics aren't queried and reported.

src/device-state/current-state.ts
20k-ultra: I think this makes it clear that we shouldn't implement the config retrieve cache in each module but just in the config component itself. Feel free to ignore the caching comment and we can do it in https://github.com/balena-os/balena-supervisor/issues/1667. 

Do you agree with not asking sysInfo to generate hte report if the feature is disabled though? I think you do.

cywang117: For sure, I wrote a comment in #1667 and we can look more into in the future.

> Do you agree with not asking sysInfo to generate hte report if the feature is disabled though? I think you do.

I agree and I've already made the changes locally, just haven't updated the remote.

20k-ultra:
Comments have been added to the original issue which specify we should send null for the metrics when it is disabled. There's a comment in the issue also describing how we should change the `SystemInfo` type to allow null values. https://github.com/balena-os/balena-supervisor/issues/1645#issuecomment-824989155

src/lib/system-info.ts
```diff
@@ -85,7 +85,13 @@ export async function undervoltageDetected(): Promise<boolean> {
 	}
 }
 
-export async function getSysInfoToReport() {
+export async function getSysInfoToReport(
```
20k-ultra, line 88: I would remove this function and just construct the info object in the current-state file which sends the data. This way we can do...

```js
	let info = {
		is_undervolted: await sysInfo.undervoltageDetected()
	}
	if (SEND_METRICS){
		const metrics = await deviceMetrics()
		info = {
			metrics
			...info
		}
	}

	async function deviceMetrics() {
		const mem = await sysInfo.getMemoryInformation()
		const storage = await sysInfo.getStorageInfo()
		return {
			cpu_usage: await sysInfo.getCpuUsage(),
			memory_usage: mem.used,
			memory_total: mem.total,
			storage_usage: storage.storageUsed,
			storage_total: storage.storageTotal,
			storage_block_device: storage.blockDevice,
			cpu_temp: await sysInfo.getCpuTemp(),
			cpu_id: await sysInfo.getCpuId(),
		}
	}
```

cywang117: Removed 👍

src/device-state/current-state.ts
```diff
@@ -177,7 +181,22 @@ const reportCurrentState = (): null => {
 		reportPending = true;
 		try {
 			const currentDeviceState = await deviceState.getStatus();
-			const info = await sysInfo.getSysInfoToReport();
+			const info = {
+				...((await config.get('disableHardwareMetrics'))
```
Page-, line 185: You could avoid needing to re-fetch this every time using a mechanism like the target-state where it only updates when it's changed: https://github.com/balena-os/balena-supervisor/blob/v12.6.4/src/device-state/target-state.ts#L43-L58

src/lib/system-info.ts
```diff
@@ -77,7 +77,7 @@ export async function getCpuId(): Promise<string | undefined> {
 	}
 }
 
-const undervoltageRegex = /under.*voltage/;
+const undervoltageRegex = new RegExp('under.*voltage', 'i');
```
Page-, line 80: Whilst the current version will work I normally see this as:
```suggestion
const undervoltageRegex =/under.*voltage/i;
```
With `new RegExp(...)` resolved for when the regexp is dynamically created, eg via string concatenation

20k-ultra:
@cywang117 can you ask me for another review after looking at Page's suggestions pleaes

src/device-state/current-state.ts
cywang117: Good idea, done.

cywang117:
Made changes per @Page- 's suggestions, ready for re-review

@pipex Added [test cases](https://github.com/balena-os/balena-supervisor/pull/1666/files#diff-515ffc947b31b73e1ec0e626bb17710db92281e29ff3e4c09f6376bd59317b4eR223-R286) for your merged PR #1692 

@20k-ultra I think you gotta resolve the "changes requested" yourself - I don't see an option:
image

docs/configurations.md
```diff
@@ -26,6 +26,7 @@ This list contains configuration variables that can be used with all balena devi
 | BALENA_HOST_FIREWALL_MODE | string | false | off | Toggle firewall modes between on, off, and auto. | v11.9.1 |
 | BALENA_HOST_DISCOVERABILITY | boolean | false | true | Enable / Disable Avahi to run which will allow the device to respond to requests such as network scans. | v11.9.2 |
 | BALENA_HOST_SPLASH_IMAGE | integer | true | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
+| BALENA_SUPERVISOR_DISABLE_HARDWARE_METRICS | boolean | false | false | Disables hardware metrics such as CPU temperature and memory usage from being reported to the backend and displayed in the dashboard. Useful for devices with bandwidth sensitivity. | TODO |
```
cywang117, line 29: I'll fix the TODO by the version after PR approval -- to prevent auto-merge from bulldozer, please comment your PR approval instead of explicitly approving it - I'll self-certify after fixing.

20k-ultra:
@cywang117 lgtm. not approving as per your request...I'm not sure how we'd overcome this type of issue since it's still prone to being wrong in the future but oh well..bigger fish to fry. AWESOME work on this.

cywang117:
@20k-ultra 
> If required reviews are enabled and a collaborator with write, admin, or owner access to the repository submits a review requesting changes, the pull request cannot be merged until the same collaborator submits another review approving the changes in the pull request.

^ I knew there was something else I should have mentioned! Anyways, tests should have passed by the time you see this, approve whenever you wanna work with the new changes 👍🏼

docs/configurations.md
```diff
@@ -26,6 +26,7 @@ This list contains configuration variables that can be used with all balena devi
 | BALENA_HOST_FIREWALL_MODE | string | false | off | Toggle firewall modes between on, off, and auto. | v11.9.1 |
 | BALENA_HOST_DISCOVERABILITY | boolean | false | true | Enable / Disable Avahi to run which will allow the device to respond to requests such as network scans. | v11.9.2 |
 | BALENA_HOST_SPLASH_IMAGE | integer | true | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
+| BALENA_SUPERVISOR_DISABLE_HARDWARE_METRICS | boolean | false | false | Disables hardware metrics such as CPU temperature and memory usage from being reported to the backend and displayed in the dashboard. Useful for devices with bandwidth sensitivity. | v12.8.0 |
```
cywang117, line 29: Added version (v12.8.0)

cywang117: Agreed that this might be a problem in the future, the PR might have to be added separately after the feature is merged to avoid this dev flow

src/device-state/current-state.ts
```diff
@@ -132,7 +132,7 @@ const getStateDiff = (): DeviceStatus => {
 
 const report = _.throttle(async () => {
 	const conf = await config.getMany([
-		'deviceId',
+		'deviceApiKey',
```
pipex, line 135: Have we always reported the `deviceApiKey` to the cloud? That's strange

src/device-state/current-state.ts
```diff
@@ -172,12 +171,41 @@ const report = _.throttle(async () => {
...
 }, constants.maxReportFrequency);
 
+/**
+ * Listen for config changes to disableHardwareMetrics
+ * to avoid a db query every 10 seconds
+ */
+let disableHardwareMetrics: boolean;
+(async () => {
+	await config.initialized;
+	config.on('change', (conf) => {
```
pipex, line 181: I don't like this, we are already querying the database every 10 seconds for other configurations such as local mode, api timeout, etc. I think we should step away from using the config module this way as it makes code more complex and harder to read.

test/src/device-state/current-state.spec.ts
```diff
@@ -0,0 +1,580 @@
...
+			lastReportedState.local = {
+				config: { ENV_VAR_3: '1' },
+			};
+			stateForReport.local = {
+				config: { ENV_VAR_3: '0' },
+			};
+			expect(getStateDiff()).to.deep.equal({
+				local: { config: { ENV_VAR_3: '0' } },
+			});
+			resetGlobalStateObjects();
```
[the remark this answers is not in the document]
pipex: What do you mean? Even if the config has been changed in the target state, the device still needs to report it has received and applied the configuration as the current state. Am I understanding the question correctly?

pipex (changes requested):
I think we should remove the event listener for config changes in hardware metrics. I think it only adds complexity

src/device-state/current-state.ts
cywang117: Yeah I believe it's [needed](https://github.com/balena-os/balena-supervisor/blob/4eb571e992907512521f4e8e584433d44a34c67f/src/device-state/current-state.ts#L81) for the patch request to the backend, it was here before this PR too

pipex: I understand is necessary as a header, but in the body? In any case I agree that is not related to this PR

test/src/device-state/current-state.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+		const report = currentState.__get__('report');
+		let localDisableHardwareMetrics: any;
+		const localAppUpdatePollInterval = 1000;
+
+		const unhandledRejectionHandler = () => {
+			/* noop */
+		};
+
+		before(() => {
+			stub(deviceState, 'getStatus').resolves({});
```
pipex, line 472: We should make a note to improve this module. The fact that so many stubs are necessary and stuff like the sleep and wait for the function to be called tell us that there is something wrong with the architecture

test/src/device-state/current-state.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+
+			expect(stateForReport).to.deep.equal({
+				local: { is_undervolted: false, cpu_usage: 20 },
+				dependent: {},
+			});
+
+			process.removeListener('unhandledRejection', unhandledRejectionHandler);
+		});
+	});
+});
```
pipex, line 585: Nice work with the tests

20k-ultra:
> I think we should remove the event listener for config changes in hardware metrics. I think it only adds complexity

I like the optimization but haven't read how we did it. I'll check in a bit.

docs/configurations.md
20k-ultra, line 29:
```suggestion
| BALENA_SUPERVISOR_HARDWARE_METRICS | boolean | false | true | Periodically queries the device for hardware metrics such as CPU utilization, CPU temperatue, memory usage, and disk space. This info is then reported back to the cloud quite frequently (10 second intervals if there are changes) | v12.8.0 |
```

I am glad @pipex mentioned that this var name isn't like our existing ones. We usually have configurations for a feature which are default to true and can be explicitly set to true or false (see local mode, deltas, persistent logging...they don't say enable/disable they just say the name of the feature).

I'll mentioned we do have `BALENA_HOST_CONFIG_disable_splash` but that is a host config not Supervisor config so it might make sense for those but also, it might have been a mistake adding disable_splash.

20k-ultra (changes requested):

src/device-state/current-state.ts
cywang117: What do you mean? The only place deviceApiKey is used is in the [header](https://github.com/balena-os/balena-supervisor/blob/4eb571e992907512521f4e8e584433d44a34c67f/src/device-state/current-state.ts#L84)

test/src/device-state/current-state.spec.ts
cywang117: Agreed. The stubs for fine and expected for unit testing, but having to `sleep(200)` for a function call to complete because it's the only way to test its effects isn't great.

docs/configurations.md
cywang117: I went with `Toggle hardware metrics reporting to the cloud, which occurs every 10 seconds when there are changes. Metrics include CPU utilization, CPU temperature, memory usage, and disk space. Useful for devices with bandwith sensitivity.` as a description to combine your suggested change's intent and my original description's intent, since I'm not completely sold on your suggestion's wording. Lmk what you think

cywang117:
Ready for review: did some final integration tests by modifying incoming target state. *_SUPERVISOR_HARDWARE_METRICS gets stored in config & deviceConfig db tables as follows:
image
image

Toggling HARDWARE_METRICS between '0' and '1' results in db & current state reporting changes. As of now, when a null patch is sent, the device metrics UI component is removed entirely (There is a related UI issue here: https://github.com/balena-io/balena-ui/issues/4726). There is also no HARDWARE_METRICS in the API but I've made a PR here: https://github.com/balena-io/open-balena-api/pull/671 that's ready for review.

src/device-state/current-state.ts
pipex: You are right, I was misreading the code

pipex (approved):
LGTM

cywang117:
@20k-ultra Ready for your approval 👍🏼

20k-ultra (approved):

PR #1674, opened 2021-04-26
Support app uuids and release version on target state

pipex:
We store the uuid and releaseVersion on the individual docker components so that when
building a current state, we can match them with the target state.

This only accepts uuids from the target state but appId and serviceId still
remain the main method for comparing current and target state

This is a breaking change that requires a new state endpoint version

See balena-io/open-balena-api#504

Change-type: major
Depends-on: balena-io/open-balena-api#504
Depends-on: #1676

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -811,11 +811,29 @@ export class App {
 			...opts,
 		};
 
+		const isService = (svc: ServiceComposeConfig) =>
+			!svc.labels ||
+			!svc.labels['io.balena.image.class'] ||
+			svc.labels['io.balena.image.class'] === 'service';
+
+		const isDataStore = (svc: ServiceComposeConfig) =>
+			!svc.labels ||
+			!svc.labels['io.balena.image.store'] ||
+			svc.labels['io.balena.image.store'] === 'data';
+
 		// In the db, the services are an array, but here we switch them to an
 		// object so that they are consistent
 		const services: Service[] = await Promise.all(
-			(JSON.parse(app.services) ?? []).map(
-				async (svc: ServiceComposeConfig) => {
+			(JSON.parse(app.services) ?? [])
+				.filter(
+					// For the host app, `io.balena.image.*` labels indicate special way
+					// to install the service image, so we ignore those we don't know how to
+					// handle yet. If a user app adds the labels, we treat those services
+					// just as any other
+					(svc: ServiceComposeConfig) =>
+						!app.isHost || (isService(svc) && isDataStore(svc)),
+				)
+				.map(async (svc: ServiceComposeConfig) => {
 					// Try to fill the image id if the image is downloaded
 					let imageInfo: ImageInspectInfo | undefined;
 					try {
@@ -837,8 +855,7 @@ export class App {
 						svc,
 						(thisSvcOpts as unknown) as DeviceMetadata,
 					);
-				},
-			),
+				}),
 		);
 
 		return new App(
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -31,9 +31,11 @@ export interface AppConstructOpts {
 	appId: number;
 	appName?: string;
 	commit?: string;
+	isHost?: boolean;
 	releaseId?: number;
+	releaseVersion?: string;
 	source?: string;
-
+	uuid?: string;
 	services: Service[];
 	volumes: Dictionary<Volume>;
 	networks: Dictionary<Network>;
@@ -56,8 +58,11 @@ export class App {
 	// When setting up an application from current state, these values are not available
 	public appName?: string;
 	public commit?: string;
+	public isHost?: boolean;
 	public releaseId?: number;
+	public releaseVersion?: string;
 	public source?: string;
+	public uuid?: string;
 
 	// Services are stored as an array, as at any one time we could have more than one
 	// service for a single service ID running (for example handover)
@@ -69,17 +74,22 @@ export class App {
 		this.appId = opts.appId;
 		this.appName = opts.appName;
 		this.commit = opts.commit;
+		this.isHost = opts.isHost;
 		this.releaseId = opts.releaseId;
+		this.releaseVersion = opts.releaseVersion;
 		this.source = opts.source;
 		this.services = opts.services;
 		this.volumes = opts.volumes;
 		this.networks = opts.networks;
+		this.uuid = opts.uuid;
 
 		if (this.networks.default == null && isTargetState) {
 			// We always want a default network
 			this.networks.default = Network.fromComposeObject(
 				'default',
 				opts.appId,
+				// Target state always has a uuid set
+				opts.uuid!,
 				{},
 			);
 		}
@@ -162,12 +172,17 @@ export class App {
 			),
 		);
 
+		// Do this only for the user app as a way to identify it
+		// from other apps. For now the state endpoint is still
+		// expecting a single commit for the device. Once we move to full multi-app
+		// we can remove the commitStore table altogether as information about
+		// the release is already on the container
 		if (
 			steps.length === 0 &&
 			target.commit != null &&
-			this.commit !== target.commit
+			this.commit !== target.commit &&
+			!target.isHost
 		) {
-			// TODO: The next PR should change this to support multiapp commit values
 			steps.push(
 				generateStep('updateCommit', {
 					target: target.commit,
@@ -266,6 +281,10 @@ export class App {
 		removePairs: Array<ChangingPair<Service>>;
 		updatePairs: Array<ChangingPair<Service>>;
 	} {
+		// TODO: This is comparing by service id, so changing environment will forcibly
+		// trigger a reinstall of the services, an since image names are also
+		// different it will also trigger a fetch
+		// comparison should probably be done using serviceName as with docker compose
 		const currentByServiceId = _.keyBy(current, 'serviceId');
 		const targetByServiceId = _.keyBy(target, 'serviceId');
 
@@ -751,13 +770,13 @@ export class App {
 			if (conf.labels == null) {
 				conf.labels = {};
 			}
-			return Volume.fromComposeObject(name, app.appId, conf);
+			return Volume.fromComposeObject(name, app.appId, app.uuid, conf);
 		});
 
 		const networks = _.mapValues(
 			JSON.parse(app.networks) ?? {},
 			(conf, name) => {
-				return Network.fromComposeObject(name, app.appId, conf ?? {});
+				return Network.fromComposeObject(name, app.appId, app.uuid, conf ?? {});
 			},
 		);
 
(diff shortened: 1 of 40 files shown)


pipex:
This should be ready for reviewing but still needs integration testing with the proper API state endpoint so left it as draft to prevent merging

src/compose/application-manager.ts
```diff
@@ -234,6 +236,7 @@ export async function getRequiredSteps(
 
 		// We want to remove images before moving on to anything else
 		if (steps.length === 0) {
+			// TODO: this should be comparing by uuid if available
```
cywang117, line 239: If we're completely supporting UUIDs, the keys for the apps object in the target state from v3 target state endpoint would be UUID, so we could step in [here](https://github.com/balena-os/balena-supervisor/blob/master/src/compose/application-manager.ts#L200-L201) and check the key type with regex, like so:
```
const appIdRegex = /^[0-9]+$/;
const currentAppIdsOrUuids = Object.keys(currentApps).map((i) => i.match(appIdRegex) ? parseInt(i, 10) : i);
const targetAppIdsOrUuids = Object.keys(targetApps).map((i) => i.match(appIdRegex) ? parseInt(i, 10) : i);
```
This change would satisfy the TODO comment you made here.

cywang117: Or just check key length, if UUID length is consistent and appIds will never reach the same length as UUIDs. Regex feels safer though

src/compose/types/service.ts
```diff
@@ -186,7 +186,7 @@ export interface DeviceMetadata {
 	imageInfo?: Dockerode.ImageInspectInfo;
 	uuid: string | null;
 	appName: string;
-	version: string;
+	version: string; // TODO: this is not used anywhere
```
cywang117, line 189: What's the resistance behind removing this?

src/device-state.ts
```diff
@@ -578,10 +578,17 @@ export async function getStatus(): Promise<DeviceStatus> {
 	// at that point we can filter non-system apps leaving a single user app.
 	// After this, for true multi-app, we will need to report our status back in a
 	// different way, meaning this function will no longer be needed
+
+	// TODO: in tri-app, we need to filter first the user app from the host and supervisor apps
+	// on the to get the correct user app id
```
cywang117, line 583:
```suggestion
	// on the state to get the correct user app id
```

Can also remove the comment block above it, they appear to be saying the same thing

src/device-state/target-state.ts
```diff
@@ -135,7 +135,7 @@ export const update = async (
 			);
 		}
 
-		const endpoint = url.resolve(apiEndpoint, `/device/v2/${uuid}/state`);
+		const endpoint = url.resolve(apiEndpoint, `/device/v3/${uuid}/state`);
```
cywang117, line 138: This v3 endpoint will be live this week?

src/compose/service-manager.ts
```diff
@@ -59,12 +59,30 @@ const volatileState: Dictionary<Partial<Service>> = {};
 export const getAll = async (
 	extraLabelFilters: string | string[] = [],
 ): Promise<Service[]> => {
-	const filterLabels = ['supervised'].concat(extraLabelFilters);
-	const containers = await listWithBothLabels(filterLabels);
+	// From balenaOS v3 containers will be created from multiple locations
+	// (preloaded in the OS, start-resin-supervisor script). All of these
+	// containers will be required to provide an app-uuid so is a good time to
+	// phase out the 'supervised' label. On the next major release after the
+	// v3 supervisor, this code should be removed
```
20k-ultra, line 66: Images have just started adding the supervised label on creation. I say just started because I had to patch a bug from a device having old images without this label https://github.com/balena-os/balena-supervisor/pull/1608. Do we really want to start removing the supervised label from services ?

src/compose/volume-manager.ts
```diff
@@ -102,7 +102,15 @@ export async function createFromPath(
...
 	oldPath: string,
 ): Promise<Volume> {
-	const volume = Volume.fromComposeObject(name, appId, config);
+	const volume = Volume.fromComposeObject(
+		name,
+		appId,
+		// We don't have a uuid available here, but we need one to create a volume
+		// from a compose object. We pass uuid as undefined here so that we will
+		// fallback to id comparison for apps
+		undefined as any,
```
20k-ultra, line 111: Do we want to hack around the transpiler to do this or should we be passing in another value like empty string ? I'm not saying that's the correct thing to do but doing undefined as any seems like we're trying to do something that the transpiler knows is wrong.

src/compose/volume.ts
```diff
@@ -57,9 +60,14 @@ export class Volume {
 
 		// We only need to assign the labels here, as when we
 		// get it from the daemon, they should already be there
+		filledConfig.labels = {
+			...filledConfig.labels,
+			...constants.defaultVolumeLabels,
+			'io.balena.app-uuid': uuid,
+		};
```
20k-ultra, line 67: Why not do this in the initialization of filledConfig above like

```
		const filledConfig: VolumeConfig = {
			driverOpts: config.driver_opts || {},
			driver: config.driver || 'local',
			labels: {
				'io.balena.app-uuid': uuid,
				...ComposeUtils.normalizeLabels(config.labels || {}),
				...constants.defaultVolumeLabels,
			},
		};
```

src/compose/application-manager.ts
cywang117: Update: targetApps parameter is always [keyed by appId](https://github.com/balena-os/balena-supervisor/blob/master/src/device-state/db-format.ts#L31-L34), so this change is unnecessary.

src/lib/validation.ts
```diff
@@ -377,11 +377,11 @@ export function isValidAppsObject(obj: any): boolean {
...
 	}
 
-	return _.every(obj, (v, appId) => {
-		if (!isValidShortText(appId) || !checkInt(appId)) {
+	return _.every(obj, (v, uuid) => {
+		if (!isValidShortText(uuid)) {
 			log.debug(
-				'Invalid appId passed to validation.isValidAppsObject\nApp ID:',
-				inspect(appId),
+				'Invalid uuid passed to validation.isValidAppsObject\nUUID:',
```
20k-ultra, line 383:
```suggestion
				'Invalid UUID passed to validation.isValidAppsObject\nUUID:',
```

20k-ultra: Makes sense in code to follow naming conventions but in log messages UUID can be uppercase as acronyms usually are I believe.

test/34-firewall.spec.ts
```diff
@@ -40,6 +41,9 @@ describe('Host Firewall', function () {
 			},
 		} as Docker.Image);
 
+		await db.initialized;
+		await db.models('app').del();
```
20k-ultra, line 45: does this affect other test suites which depend on this table ?

test/40-app-uuids.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+		});
+	});
+
+	describe('Target state', () => {
+		before(async () => {
+			await db.models('app').del();
+		});
+		it('should correctly validate apps keyed by UUID', async () => {
+			await deviceState.setTarget(uuidTargetState);
+		});
```
20k-ultra, line 58: there's nothing being tested here because there's no expect ?

src/compose/volume-manager.ts
20k-ultra: Should we modify the function params to have a default which is undefined ?

src/compose/app.ts
```diff
@@ -162,12 +169,12 @@ export class App {
 			),
 		);
 
+		// TODO: remove this when the updateCommit table disappears
 		if (
 			steps.length === 0 &&
 			target.commit != null &&
 			this.commit !== target.commit
 		) {
-			// TODO: The next PR should change this to support multiapp commit values
```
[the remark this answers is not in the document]
pipex: Because the commit table is already multiapp. This used to be a column in the config table and I think they just didn't remove the comment

src/compose/app.ts
```diff
@@ -821,13 +832,32 @@ export class App {
...
+			.filter(
+				(svc) =>
+					!svc.config.labels['io.balena.image.class'] ||
+					svc.config.labels['io.balena.image.class'] === 'service',
+			)
+			.filter(
+				(svc) =>
+					!svc.config.labels['io.balena.image.store'] ||
+					svc.config.labels['io.balena.image.store'] === 'data',
+			);
```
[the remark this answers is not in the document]
pipex: Because the other way looks prettier :), I'll make the change

src/compose/app.ts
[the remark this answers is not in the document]
20k-ultra: gotcha. just wanted to confirm

src/compose/app.ts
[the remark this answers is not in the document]
20k-ultra: yeah I agree it's pushing it for readability

src/compose/application-manager.ts
```diff
@@ -368,35 +372,148 @@ export async function getCurrentAppsForReport(): Promise<
...
+		const components = componentGroups[appId];
+
+		// fetch the correct uuid from any component within the appId
+		const uuid = [
+			components.services[0]?.uuid,
+			components.volumes[0]?.uuid,
+			components.networks[0]?.uuid,
+		]
+			.filter((u) => u != null)
+			.shift()!;
```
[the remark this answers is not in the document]
pipex: My guess is that javascript does a pointer update to implement this method, if there are any performance gains I'm not sure they justify sacrificing readability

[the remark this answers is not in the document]
20k-ultra: agreed then.

src/compose/app.ts
```diff
@@ -821,13 +832,32 @@ export class App {
 				},
 			),
 		);
+
+		// Ignore any image class that is not a service to prevent
+		// the supervisor trying to install images it doesn't understand
+		const services = allServices
```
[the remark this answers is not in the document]
pipex: Yeah, the problem is that there is an overloading of the word service. On the compose they are defined as services but they can have a different image class, and one of the image classes is 'service'. So this gets all services and then filters out all those that are not of image class service

test/34-firewall.spec.ts
cywang117: I don't think it would, because we don't want database entry persistence between test suites anyway, and this call only deletes the current rows in `app` table. 

However the fact that we have to do this manually is pretty scuffed. With mocha v8 global hooks, we can reset the database to default values before each root `describe` block. I will be (and am kinda in the process of) looking into this .

src/compose/volume-manager.ts
pipex: That's a good question, I think the function needs to assume that uuid is given, since from now on it should always come as part of the target state. The `createFromPath` function is only called for legacy volumes, which is why this workaround is here. I didn't write this part of the code but I think I'm ok with doing it this way

src/compose/types/service.ts
pipex: I used VSCode get references for this, I'm not sure if there is a case that may not have been caught as part of a type reassignment. Searching the code for the string `"version"`, I'm pretty convinced it's not used. I'll remove it and hopefully we'll catch during testing if I missed something

pipex:
Changes in this PR have been merged in #1765

pipex:
This was superseded by #1765

PR #1676, opened 2021-04-26
Bump dockerode types to 2.5.34

pipex:
This PR updates dockerode types to the latest 2.x version, removing the need
for custom composer types for network. It also modifies network tests to use the new types

The also introduces a fix to a mock-dockerode bug that was resulting in test mocks not being properly cleaned up between tests when using `withTestData`

Change-type: minor


src/compose/network.ts
```diff
@@ -28,50 +23,45 @@ export class Network {
...
+		network: dockerode.NetworkInspectInfo,
+	): Network {
 		const ret = new Network();
 
 		const match = network.Name.match(/^([0-9]+)_(.+)$/);
 		if (match == null) {
 			throw new InvalidNetworkNameError(network.Name);
 		}
+
+		// TODO: this seems uncessary with the regex above
```
cywang117, line 36: Agreed. Looking at the regex, the `|| null` bit seems unnecessary. None of the following examples match the regex except for the all-numbers one:
image

The `checkInt` call might be required to convert string to int though, since the regex would match and return a string from the network string. checkInt will return `undefined` for non-integer parameters, so the `|| null` is doubly unnecessary.

src/compose/network.ts
```diff
@@ -90,19 +80,26 @@ export class Network {
...
+			ipam: {
+				driver,
+				config: config.map((conf) => ({
+					...(conf.subnet && { subnet: conf.subnet }),
+					...(conf.gateway && { gateway: conf.gateway }),
+					...(conf.ip_range && { ipRange: conf.ip_range }),
+					// TODO: compose defines aux_addresses as a dict but dockerode and the
+					// engine accepts a single AuxAddress. What happens when multiple addresses
+					// are given?
+					...(conf.aux_addresses && { auxAddress: conf.aux_addresses }),
```
cywang117, line 99: The engine should [accept](https://docs.docker.com/engine/reference/commandline/network_create/#specify-advanced-options) multiple aux addresses, so I assume Dockerode does too - do the Dockerode typings not reflect this?

pipex: The engine should, but it looks like the dockerode typings go by what is defined on the API https://docs.docker.com/engine/api/v1.37/#operation/NetworkCreate

src/compose/network.ts
pipex: So if, the network name doesn't match the regex, then the name doesn't contain an appId, and the first `if` will catch it. Even doing `checkInt` is unnecessary because the first part of the match should be a number at that point

20k-ultra: `checkInt` converts the value into a number type which the regex test won't btw.

20k-ultra: I'd leave it since we only gain 4 LOC

test/41-device-api-v1.spec.ts
```diff
@@ -298,11 +304,13 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 	describe('POST /v1/apps/:appId/start', () => {
 		it('does not allow starting an application when there is more than 1 container', async () => {
 			// Every test case in this suite has a 3 service release mocked so just make the request
```
20k-ultra, line 306: We can remove all instances of this comment since we now specify the containers (via testWithData) for all requests.

20k-ultra:
lgtm, self-certify after making the above changes mention

test/41-device-api-v1.spec.ts
20k-ultra: nvm I see the tests which still have this comment aren't running with `testWithData`

PR #1677, opened 2021-04-26
Add support for host overlays (WIP)

pipex:
This commit adds support for host extensions on the target state
endpoint. When host extensions configuration changes are detected
the supervisor will pass the target entirely to the OS `update-data-store`
script.

Change-type: minor
Depends-on: #1674


sync/device.ts
```diff
@@ -24,7 +24,7 @@ export async function getSupervisorContainer(
 ): Promise<Docker.ContainerInfo> {
 	// First get the supervisor container id
 	const containers = await docker.listContainers({
-		filters: { name: ['resin_supervisor'] },
+		filters: { name: ['balena-supervisor'] },
```
alexgg, line 27: @pipex we agreed to go for `balena_supervisor` as there was already an open PR https://github.com/balena-os/meta-balena/pull/2171

pipex: Yes, I only added this for the benefit of testing. I'll change it before we merge

Issue #1116, opened 2019-10-21
/v2/applications/:appId/restart-service recreates the service


hedss:
Whilst the `start` endpoint starts a service without recreating it, the `restart` endpoint actually recreates the container. This isn't consistent, and potentially constitutes a problem.

pdcastro:
For what it is worth (maybe not much), based on my playing with the web dashboard and ssh to a NUC device and tests on my laptop (but not having looked at supervisor source code), I gather that:
* `docker restart` = `docker stop` + `docker start`
* Dashboard / supervisor's stop = `docker stop`
* Dashboard / supervisor's start = `docker start`
* Dashboard / supervisor's restart = `docker rm -f` + `docker run`

cywang117:
I am not sure about the design rationale behind removing + recreating containers on restart, but the docker restart command does not do this, so it would be better if we emulated this. The change would theoretically just be to add a `removeContainer` kill option to [this line](https://github.com/balena-os/balena-supervisor/blob/master/src/device-api/common.js#L33) as follows:
```
await serviceManager.kill(service, {wait: true, removeContainer: false});
```

cywang117:
To clarify further, as I've just investigated all possible places where restart happens:
1. Clicking the `Restart` button on the summary page AND the action page recreate ALL containers.
2. Clicking the `Restart` button by a single service in the services table on the summary page does NOT recreate that container.

Per the May 4th balena-io product call, we should unify this behavior on the Supervisor, so that all `restart` actions remove + recreate containers.

hedss:
Been a while since I raised this, but I'm also considering using balena for a project, and I'd like to know the rationale behind recreating the containers. Certainly when @CameronDiver and I originally discussed this, we agreed this was not ideal and was probably originally in error, because:

1. This is not how Docker behaves, and this actually makes a huge difference in development.
2. You could potentially end up in a situation where you blow away overlays you might otherwise expect to be populated.

Thanks!

cywang117:
Hi @hedss, thanks for bringing this point up. I realize that in my previous comment from May 5, I wasn't very clear about balena's rationale behind implementing restart as a recreate action. Let me try to explain, as I realize that the internal call documentation I linked may not be accessible to everyone:

[...]

PR #1682, opened 2021-04-28
Patch restart to not remove & recreate containers

cywang117:
Closes: #1116
Change-type: patch

### DIFF
--- a/src/device-api/common.js
+++ b/src/device-api/common.js
@@ -30,7 +30,10 @@ export async function doRestart(appId, force) {
 
 			return deviceState.pausingApply(async () => {
 				return Bluebird.each(app.services, async (service) => {
-					await serviceManager.kill(service, { wait: true });
+					await serviceManager.kill(service, {
+						wait: true,
+						removeContainer: false,
+					});
 					await serviceManager.start(service);
 				});
 			});


src/device-api/common.js
```diff
@@ -30,7 +30,10 @@ export async function doRestart(appId, force) {
 
 			return deviceState.pausingApply(async () => {
 				return Bluebird.each(app.services, async (service) => {
-					await serviceManager.kill(service, { wait: true });
+					await serviceManager.kill(service, {
```
pipex, line 33: I'm not sure about this, I don't think the API should be making calls to `serviceManager` in the first place. I think we need to figure out a way to do this through providing target state

20k-ultra: I don't think that's within the scope of this PR tbh. We can resolve the issue with this simple addition.

pipex: Maybe, but a one line change also validates this code and adds to technical debt, so we need to at least to make a priority issue about refactoring the API code

cywang117: @pipex I believe this was done before using intermediateState but there was an image removal bug, see: https://github.com/balena-os/balena-supervisor/pull/1524

cywang117: I fixed this image removal bug for purge in #1635, we could apply the change here as well.

src/device-api/common.js
```diff
@@ -30,7 +30,10 @@ export async function doRestart(appId, force) {
...
+						wait: true,
+						removeContainer: false,
```
pdcastro, line 35: Adding my 2 cents... Is this change really desirable though? And is it effectively a breaking change, when the previous behavior has been around for at least 1.5 years, and maybe since ever? Not that breaking changes cannot be made, but they perhaps warrant a bit more discussion.

On one hand, one could argue that we should try to emulate `docker restart` and that restart should mean stop + start because "that's what it usually means". On the other hand, isn't stop + start "less useful" than re-creating the container? If an app service got stuck in a bad state, recreating the container is more useful and powerful than stop + start, which might simply "resume being stuck". Perhaps if the supervisor / dashboard offered an additional "reset" action as suggested by Joe in the support thread, then this change to restart could be introduced at the same time, and users would still have a way of recreating containers other than using restart.

One could also make the case that the only thing we need is to document that restart means recreating the container, and not changing what it currently does. Check this comment about the dashboard having a tooltip to clarify to users that data saved to the container will be lost:
https://github.com/balena-io/balena-ui/issues/3099#issuecomment-828532247

Don't let this comment of mine be a blocker to anything, but take it into account before making your decision. :-)

src/device-api/common.js
cywang117: I'll implement the change to use intermediateState instead of serviceManager in a new PR, closing this one as it's no longer relevant though.

src/device-api/common.js
cywang117: Hey Paulo, thanks for your comment, After discussion in the May 4th product call, we've agree on keeping the restart behavior as it is, However, the restart behavior throughout the Supervisor is not consistent, so I'll be working on that in another PR. An additional change we're making, which I think you suggested as well, is to clarify in the docs (and on the dashboard actions tab) that restart will remove + recreate containers.

cywang117:
Closing this PR, will be working on another one that unifies restart action behavior on the Supervisor. See my latest comment in #1116 for details. I'll also be referencing that issue in the new PR.

PR #1683, opened 2021-04-29
Update balena-register-device and send extra info at provision time

Page-:
This extra info will mean the API is able to immediately set default
config vars based on the os/supervisor version so that they are
available on the first target state fetch rather than having a delay
whilst waiting for the supervisor to report them as part of a state
patch

Update balena-register-device from 6.1.6 to 7.2.0

Change-type: patch


Page-:
@balena-ci rebase

cywang117:
@Page- Integration testing on a device works, confirming that `SUPERVISOR_DELTA`, `SUPERVISOR_DELTA_VERSION`, and `SUPERVISOR_POLL_INTERVAL` are set on the first target state poll after provision using `balena-device-register`. Since configs are set before app updates happen in the state funnel, there are no more race conditions on this front 🎉

src/config/schema-type.ts
```diff
@@ -211,6 +211,10 @@ export const schemaTypes = {
 			apiTimeout: PermissiveNumber,
 			registered_at: t.union([PermissiveNumber, NullOrUndefined]),
 			deviceId: t.union([PermissiveNumber, NullOrUndefined]),
+			supervisorVersion: t.union([t.string, t.undefined]),
+			osVersion: t.union([t.string, t.undefined]),
+			osVariant: t.union([t.string, t.undefined]),
+			macAddress: t.union([t.string, t.undefined]),
```
cywang117, line 217:
```suggestion
			supervisorVersion: t.union([t.string, NullOrUndefined]),
			osVersion: t.union([t.string, NullOrUndefined]),
			osVariant: t.union([t.string, NullOrUndefined]),
			macAddress: t.union([t.string, NullOrUndefined]),
```

Or does the balena API not accept null for these fields?

Page-: balena-device-register types don't accept null for them because it makes no sense to send null as they'll default to null, but the api would happily accept null as a reaffirmation of the default, it'd just be a small waste of bandwidth

Page-:
@balena-ci rebase

src/config/schema-type.ts
cywang117: I see, that makes sense. Thanks for clarifying!

cywang117 (approved):

PR #1686, opened 2021-04-30
Add tilde as a shortcut for project root, allowing shorter imports.

cywang117:
Imports may be made with file paths like so: `~/src/lib/constants.ts`, where `~` is the project root.

Change-type: patch


20k-ultra:
I like this. Will clean up imports to not need ../../../src etc

Issue #1689, opened 2021-05-01
Fix network driver_opts from compose to dockerconfig


quentingllmt:
I'm trying to use the balena supervisor with a custom balena os image with macvlan support. Macvlan needs the `drivers_opts` but it seems in the code that they are not passed to the dockerconfig object when creating it (`compose/network.ts`) (correctly parsed but not passed to *dockerode.NetworkCreateOptions*) in `toDockerConfig()`.

Why is that ? Is this just a missing property assignment ? Anyway, I think it should be implemented unless there something I don't know about the mechanism (never worked with dockerode or on this repository, I'm just trying to find the cause of my issue).

Fix: Add `Options: this.config.options,` line 133 might works if I understand the code correctly.

Thanks

cywang117:
Hey @quentingllmt, thanks for taking the initiative and opening a PR. We'll take a look at this and get back to you :)

PR #1691, opened 2021-05-01
Fix passing driver_opts from compose to docker network creation

quentingllmt:
Connects-to: https://github.com/balena-os/balena-supervisor/issues/1689
Change-Type: patch
Signed-off-by: quentinGllmt <quentin@quentingllmt.fr>

### DIFF
--- a/src/compose/network.ts
+++ b/src/compose/network.ts
@@ -130,6 +130,7 @@ export class Network {
 			Name: Network.generateDockerName(this.appId, this.name),
 			Driver: this.config.driver,
 			CheckDuplicate: true,
+			Options: this.config.options,
 			IPAM: {
 				Driver: this.config.ipam.driver,
 				Config: this.config.ipam.config.map((conf) => {
--- a/test/src/compose/network.spec.ts
+++ b/test/src/compose/network.spec.ts
@@ -110,6 +110,62 @@ describe('compose/network', () => {
 				'Network IPAM config entries must have both a subnet and gateway',
 			);
 		});
+
+		it('parses values from a compose object', () => {
+			const network1 = Network.fromComposeObject('default', 12345, {
+				driver: 'bridge',
+				enable_ipv6: true,
+				internal: false,
+				ipam: {
+					driver: 'default',
+					options: {
+						'com.docker.ipam-option': 'abcd',
+					},
+					config: [
+						{
+							subnet: '172.18.0.0/16',
+							gateway: '172.18.0.1',
+						},
+					],
+				},
+				driver_opts: {
+					'com.docker.network-option': 'abcd',
+				},
+				labels: {
+					'com.docker.some-label': 'yes',
+				},
+			});
+
+			const dockerConfig = network1.toDockerConfig();
+
+			expect(dockerConfig.Driver).to.equal('bridge');
+			// Check duplicate forced to be true
+			expect(dockerConfig.CheckDuplicate).to.equal(true);
+			expect(dockerConfig.Internal).to.equal(false);
+			expect(dockerConfig.EnableIPv6).to.equal(true);
+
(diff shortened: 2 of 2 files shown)


pipex:
Hi Quentin, thanks for the PR! 

Would you mind adding some sample configurations on the [tests for this module](https://github.com/balena-os/balena-supervisor/blob/master/test/src/compose/network.spec.ts#L19)? That will allow us ensure that that config is being interpreted properly.

Thanks again!

quentingllmt:
Hi @pipex,

I'm not really used to writing tests (furthermore for an app I know nothing of internally) but I did try anyway if it can help for a fast merge. I just hope those tests are what you are expecting. 
It seems to me you were not testing the `network.toDockerConfig()` output at all and that's where I made the modifications so I have added a new test case for that export which pretty much test everything. Hope that's OK !

pipex:
That's a great catch @quentingllmt, nice work with the tests. I don't have any additional comments, I'll wait for my colleagues to take a second look and will approve

pipex:
@balena-ci rebase

20k-ultra (changes requested):
Just to establish some terminology, the test suite is the **describe** block and a test case is the **it** blocks inside a suite. 

In your tests you've added a suite with a test which are great but my only request is that you move your test case "creates a valid network create options from the compose config" into the "creating a network from a compose object" test suite found on line 7. Then delete your test suite since it will be empty.

After that I think changing the test case name to "parses values from a compose object" is more clear for the case as well.

quentingllmt:
@20k-ultra You're right. Done
When do you think this PR could be approved ?

20k-ultra (approved):

20k-ultra:
Squash (fixup) all your commits into the first one then rebase master and force push with lease. I'm realizing our contribution docs do not contain these instructions so I've created a PR to help make this easier next time.

Check out exactly what I mean about squashing your history and rebasing master https://github.com/balena-os/balena-supervisor/pull/1697/commits/d3ddb4e2530fcf11b95dbdd4505ee95a0713c70d?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

quentingllmt:
@20k-ultra Done but versionbot failed.. Why ?

cywang117:
@quentingllmt I believe this is because the `Change-type` footer is mis-capitalized as `Change-Type` instead. It's not as intuitive as we'd like, so I'll fix it, no worries. Thanks for this contribution!

cywang117:
@balena-ci rebase

cywang117:
@quentingllmt Looks like our ci is having some issues, would you mind:
- Changing any commit footers that use `Change-Type` to `Change-type`
- Rebasing on top of master
?

Thanks again!

20k-ultra:
@quentingllmt you have done a great job with our feedback and I'm so sorry for all the friction. Versionist is getting tripped up because your commit message contains the additional footers so it cannot find the Change-type.

Right now your commit is:

```
Fix parsing driver_opts from compose to docker network creation

Change-Type: patch
Signed-off-by: quentinGllmt <quentin@quentingllmt.fr>

network create config tests

test suite refactoring
```

But if you make it:

```
Fix parsing driver_opts from compose to docker network creation

Change-Type: patch
Signed-off-by: quentinGllmt <quentin@quentingllmt.fr>
```

then your PR will pass and merge

quentingllmt:
@20k-ultra  Yes that's what I thought and I already did that in the last commit. It should be fine now!

20k-ultra:
thanks @quentingllmt! You can get this change running a device by upgraded to Supervisor v12.6.8 from the self service upgrade section of the dashboard. Test this on a development device first as these releases happen often and only rely on unit testing.

Issue #1680, opened 2021-04-27
Log better messages for cloud API error responses


20k-ultra:
A lot of times the API status responses will just log the status code and nothing else. We can provide some more information based on the status code and also dump the headers in case they include useful info.

For example you will see something like:

```
f7402c2  [error]   Error from the API: 401
f7402c2  [error]   Non-200 response from the API! Status code: 401 - message: Error
f7402c2  [error]         at /usr/src/app/dist/app.js:22:554770
f7402c2  [error]       at runMicrotasks (<anonymous>)
f7402c2  [error]       at processTicksAndRejections (internal/process/task_queues.js:97:5)
f7402c2  [error]       at async /usr/src/app/dist/app.js:22:554078
f7402c2  [error]       at async /usr/src/app/dist/app.js:22:555600
```

which is not useful.

20k-ultra:
We can mask the message to say there is an authentication issue and to obtain new API KEYS for the dashboard for example. However, this is a rare case so providing instructions on how to do that isn't necessary yet since I've only seen it happen once.

pipex:
I closed the issue because of the PR but the PR doesn't really cover all cases. So reopened

20k-ultra:
I think we missed an opportunity on https://github.com/balena-os/balena-supervisor/pull/1692https://github.com/balena-os/balena-supervisor/pull/1692. We should've made https://github.com/balena-os/balena-supervisor/blob/853acae4a8e5b4ef8f67a16fedd1d550611477bb/src/device-state/current-state.ts#L94 log that there was an error updating the current device state. In addition to logging what is in the response's body we should explain what the failure means. For example, a lot of times people think the 503 error is causing their application to not work but that's just the device cannot update the dashboard.

20k-ultra:
related: https://github.com/balena-os/balena-supervisor/issues/1685

20k-ultra:
We have merged in https://github.com/balena-os/balena-supervisor/pull/1692 which has added some improvements but I think we can further improve API error messages because the following example is pretty verbose with a lot of unnecessary info

```
Jun 02 14:40:07 203db23 resin-supervisor[4770]: [error]   Error from the API: 503
Jun 02 14:40:07 203db23 resin-supervisor[4770]: [error]   Non-200 response from the API! Status code: 503 - message: Error: "The API cannot handle your re>
[...]

PR #1692, opened 2021-05-03
Log error responses from API when reporting state

pipex:
Change-type: patch
Relates-to: #1680

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -34,8 +34,8 @@ const stateForReport: DeviceStatus = {
 let reportPending = false;
 
 class StatusError extends Error {
-	constructor(public statusCode: number) {
-		super();
+	constructor(public statusCode: number, public statusMessage?: string) {
+		super(statusMessage);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pipex:
Drafting for the moment as we want to see if any filtering is required from API messages

pipex:
Ok, taking of draft. Changed commit message to make it more specific, this is only when reporting state and doesn't include possible errors when getting target state or for streaming logs.

cywang117:
Looks good. I have a thought -- if #1666 is merged before this, you could edit the test [here](https://github.com/balena-os/balena-supervisor/pull/1666/files#diff-515ffc947b31b73e1ec0e626bb17710db92281e29ff3e4c09f6376bd59317b4eR391-R412) to display statusMessages from the cloud if we wanna cover all our bases. Otherwise if this merges first (more likely), I'll rebase & add the test. Both are ok with me.

pipex:
@balena-ci rebase

pipex:
Can we get an approval here @cywang117 @20k-ultra 🙏🏻

20k-ultra (approved):

Issue #1693, opened 2021-05-04
New target state prevented from being applied, when network without both subnet and gateway created


anujdeshpande:

pipex:
Here is the replication

Create the following docker-compose and perform a push in local mode. You'll get a looping exception with the message "Network IPAM config entries must have both a subnet and gateway", and will not be able to push a correct target state after that.

```
version: '2.1'

services:
  hello:
    build: ./hello-world
    networks:
      my_net:
        ipv4_address: 172.16.238.10

networks:
  my_net:
    driver: bridge
    ipam:
      driver: default
      config: 
        - gateway: 172.16.238.1
```

pipex:
The PR #1656 fixed validation for network ipam config, checking that both network and subnet are defined for each ipam config entry (as described in the [docker documentation](https://docs.docker.com/compose/compose-file/compose-file-v2/#networks)) 

The old was querying the wrong parts of the configuration making the validation never to happen [here](https://github.com/balena-os/balena-supervisor/pull/1656/files#diff-f9a66d4ba0463ef433190b8c8c5b5dd77886557585cd0c6d01f27f0398ad1f63R217).

After that PR, the validations throws an exception if the network target state is incorrect, but this turns out to be the wrong approach, because that exception is also triggered when querying target state. 

This isn't a problem in normal operation, but it is in local mode, because local mode queries the old target state before sending a new one. Since the query fails, the CLI can never push the new target state.

I think the solution is just to show a warning on the supervisor logs, since a poorly configured network will not cause the engine to misbehave, it will only render the container unreachable through the provided address.

20k-ultra:
> I think the solution is just to show a warning on the supervisor logs, since a poorly configured network will not cause the engine to misbehave, it will only render the container unreachable through the provided address.

I think this is a product question @pipex. We totally should prevent the Supervisor from becoming unresponsive as the original reporter in the jellyfish link mentioned but don't know if logging a misconfigured network but continuing to apply target state is better then throwing the exception and not continuing to apply the target state.

pipex:
@20k-ultra  I totally agree that the supervisor should be able to reject a target state, which [it does](https://github.com/balena-os/balena-supervisor/blob/master/src/device-state.ts#L51-L53) and we should work on improving that part of the codebase (better validation, surfacing errors to the dashboard, etc.).

But I also think is clear that the validation cannot happen in the way is being done now, as an exception when building the object, because that will break other parts of the codebase. 

[...]

PR #1694, opened 2021-05-04
Show warning instead of exception for invalid network config

pipex:
A previous PR (#1656) fixed validation for network ipam config,
checking that both network and subnet are defined for each ipam config entry
(as described in the docker documentation).

After that PR, the validations throws an exception if the network target state is incorrect,
but this turns out to be the wrong approach, because that exception is also triggered
when querying target state.

This isn't a problem in normal operation, but it is in local mode, because local
mode queries the old target state before sending a new one. Since the query fails,
the CLI can never push the new target state.

This PR replaces the exception with a warning on the logs, since a
misconfigured network won't cause any engine failures, it will just
prevent containers to communicate through the provided network.

A future improvement should move this validation to an earlier point in the process,
so the target state can get rejected before it even gets to a point it
can be used.

Relates-to: #1693
Change-type: patch

### DIFF
--- a/src/compose/network.ts
+++ b/src/compose/network.ts
@@ -5,14 +5,12 @@ import * as dockerode from 'dockerode';
 import { docker } from '../lib/docker-utils';
 import logTypes = require('../lib/log-types');
 import * as logger from '../logger';
+import log from '../lib/supervisor-console';
 import * as ComposeUtils from './utils';
 
 import { ComposeNetworkConfig, NetworkConfig } from './types/network';
 
-import {
-	InvalidNetworkConfigurationError,
-	InvalidNetworkNameError,
-} from './errors';
+import { InvalidNetworkNameError } from './errors';
 
 export class Network {
 	public appId: number;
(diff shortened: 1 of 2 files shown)


pipex:
@balena-ci retest

src/compose/network.ts
```diff
@@ -200,8 +198,9 @@ export class Network {
 		// Check if every ipam config entry has both a subnet and a gateway
 		_.each(_.get(config, 'ipam.config', []), ({ subnet, gateway }) => {
 			if (!subnet || !gateway) {
-				throw new InvalidNetworkConfigurationError(
-					'Network IPAM config entries must have both a subnet and gateway',
+				log.warn(
+					'Network IPAM config entries must have both a subnet and gateway defined.' +
+						' Your app containers not be able to reach your container through the defined network address otherwise.',
```
cywang117, line 203:
```suggestion
						' Your app containers might not be able to reach your container through the defined network address otherwise.',
```

cywang117: Just the typo, otherwise LGTM!

pipex: Nice catch! Fixed!

src/compose/network.ts
```diff
@@ -200,8 +198,9 @@ export class Network {
 		// Check if every ipam config entry has both a subnet and a gateway
 		_.each(_.get(config, 'ipam.config', []), ({ subnet, gateway }) => {
 			if (!subnet || !gateway) {
-				throw new InvalidNetworkConfigurationError(
-					'Network IPAM config entries must have both a subnet and gateway',
+				log.warn(
+					'Network IPAM config entries must have both a subnet and gateway defined.' +
+						' Your app containers might not be able to reach your container through the defined network address otherwise.',
```
20k-ultra, line 203: let's discuss if we want to log a warning since this might have unexpected behaviour when the SV skips over creating this network and other parts of the application need it.

20k-ultra (changes requested):

20k-ultra (approved):
We'll work on https://github.com/balena-os/balena-supervisor/issues/1706 to improve the bigger issue here but this change will be good for the mean time.

PR #1696, opened 2021-05-05
Improve testing for supervisor composition modules

pipex:
This PR cleans up testing for supervisor compose modules. It also fixes broken tests for application manager and removes a lot of dependencies for those tests on DB and other unnecessary mocks. There are probably a lot of cases that tests are missing but this should make writing new tests a lot easier.


20k-ultra:
let's make add mockerode.ts to https://github.com/balena-io-modules/mockerode :)

20k-ultra:
looking forward to this :)

src/compose/application-manager.ts
```diff
@@ -179,21 +179,38 @@ export async function getRequiredSteps(
...
+		availableImages = [] as Image[],
+		containerIdsByAppId = {} as { [appId: number]: Dictionary<string> },
+	} = {},
+) {
+	// get some required data
+	const [{ localMode, delta }, cleanupNeeded] = await Promise.all([
+		config.getMany(['localMode', 'delta']),
+		imageManager.isCleanupNeeded(),
+	]);
+
```
pipex, line 213: These are the only changes to a module outside of test. Without this, testing the `getRequiredSteps` function was way too difficult

pipex:
There are a lot of commits on this PR but I'll squash everything after review

test/lib/mockerode.ts
```diff
@@ -35,6 +35,26 @@ export type PartialImageInspectInfo = DeepPartial<
...
+				...res,
+				[fn]: () => {
+					throw Error(
+						`Fake method not implemented: ${prototype.constructor.name}.${fn}()`,
+					);
+				},
+			}),
+			{} as Methods<Prototype>,
+		);
+}
```
cywang117, line 56: Good idea! I checked real quick and the test fails with thrown error when a not-yet-implemented method is called in a suite.

cywang117: Never mind, I just saw your comment below in this commit (7db7ce74f1e429ede495ab5d72736c4af04339df) about avoiding silent fails

test/src/compose/volume-manager.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+					Name: Volume.generateDockerName(111, 'main'),
+					Labels: {
+						'io.balena.supervised': '1',
+					},
+				}),
+			];
+			// Perform test
+			await mockerode.withMockerode(
+				async (dockerode) => {
+					// Volume to try again create
```
cywang117, line 179: I understand this comment's meaning when looking at the whole test suite, but it's a bit confusing by itself. Maybe something like `// Generate compose data for volume which was already set up in mock engine`

test/lib/db-helper.ts
```diff
@@ -0,0 +1,82 @@
...
+			(db.models as sinon.SinonStub).resetHistory();
+			(db.upsertModel as sinon.SinonStub).resetHistory();
+		},
+
+		// Destroys the in-memory database and resets environment
+		async destroy() {
+			await knex.destroy();
+
+			// Restore the old datbase path
+			process.env.DATABASE_PATH = oldDatabasePath;
```
cywang117, line 70: I suppose this is necessary because there are still tests that still use a file system database such as 05-db.spec.ts. We should look at removing all physical db tests next and replace them with in-memory tests. We can setup a default in-memory database per test file with a root-level `beforeEach` mocha hook

test/lib/db-helper.ts
```diff
@@ -0,0 +1,82 @@
...
+					// We don't run the migrations again
+					if (r.name !== 'knex_migrations') {
+						await trx.raw(`DELETE FROM ${r.name}`);
+					}
+				}
+
+				// The supervisor expects this value to already have
+				// been pre-populated
+				await trx('deviceConfig').insert({ targetValues: '{}' });
+			});
```
cywang117, line 58: Looks like a bit of duplicated work with test/lib/prepare.ts: https://github.com/balena-os/balena-supervisor/blob/ff4a31a0e6269b823a04eeda2f433665617f583c/test/lib/prepare.ts#L10-L23. Is the plan to de-duplicate when all tests that require a database use an in-memory database?

test/lib/mockerode.ts
```diff
@@ -641,25 +722,114 @@ export class MockEngine {
...
+	// both the image and the tag are removed.
+	// Does not remove images from a registry.
+	// You cannot remove an image of a running container unless you use the
+	// -f option. To see all images on a host use the docker image ls command.
+	//
+	// You can remove an image using its short or long ID, its tag, or its digest.
+	// If an image has one or more tags referencing it, you must remove all of them
+	// before the image is removed. Digest references are removed automatically when
+	// an image is removed by tag.
+	removeImage(name: string, { force } = { force: false }) {
```
cywang117, line 779: Is this a convenience method for for Dockerode's `docker.getImage(img.dockerImageId).remove({ force: true })` to make tests simpler?

cywang117: Never mind, I understand why now.

test/src/compose/images.spec.ts
```diff
@@ -0,0 +1,481 @@
...
+			async (mockerode) => {
+				// Check that the image exists
+				// this is really checking that mockerode works ok
+				await expect(
+					mockerode.getImage(imageToRemove.name).inspect(),
+					'image exists on the engine before test',
+				).to.not.be.rejected;
+
+				// Check that only one image with this dockerImageId exists in the db
+				// in memory db is a bit flaky sometimes, this checks for issues
```
cywang117, line 251: > in memory db is a bit flaky

How so?

test/lib/application-state-mock.ts
```diff
@@ -31,7 +31,7 @@ export function mockManagers(svcs: Service[], vols: Volume[], nets: Network[]) {
 			}
 			return s;
 		});
-		console.log('Calling the mock', svcs);
+		// console.log('Calling the mock', svcs);
```
cywang117, line 34: Can we just delete this line?

cywang117:
Excellent work with mockerode! Nice module. Is moving it to https://github.com/balena-io-modules/mockerode an option?

test/src/compose/volume-manager.spec.ts
pipex: hahaha, don't know what I was thinking when I wrote that :p

test/lib/db-helper.ts
pipex: A copy paste really, I want the prepare file to go away eventually

test/src/compose/images.spec.ts
pipex: I'm not sure, that module failed one of the tests randomly because of the database entry not being there, which doesn't makes sense because the entry is being inserted a couple of lines before.

test/lib/application-state-mock.ts
pipex: yup, will do

pipex:
I don't think so @cywang117, is not really feature complete and I'm still cautious on how to use it, I don't want people to rely blindly on it working correctly since its behavior may diverge from docker behavior. That's why on tests I check that the function was called with the right arguments instead of checking the state of the engine afterwards (see [here](https://github.com/balena-os/balena-supervisor/pull/1696/files#diff-8ac4a7790ec47922c45f35fa931901531dbe09f524171fe57ab45b8a93adcb70R493-R495) for instance)

> Excellent work with mockerode! Nice module. Is moving it to https://github.com/balena-io-modules/mockerode an option?

test/lib/db-helper.ts
pipex: I'm not totally sure that's necessary, but since the prepare.ts db setup changes the env var I don't want to affect those tests. 

I'm not sure about creating a database file with a mocha hook, I think using the helper on just the tests that use the db is enough (there should not be that many)

20k-ultra:
are you constantly rebasing master so when this is ready you don't have a bunch of conflicts to fix ?

pipex:
@20k-ultra this is ready now to review. Yes I’ve been rebasing buy we don’t change a lot of tests so it hasn’t been too much

pipex:
@balena-ci rebase

test/lib/db-helper.ts
```diff
@@ -0,0 +1,82 @@
+import * as constants from '../../src/lib/constants';
+import * as db from '../../src/db';
+import * as sinon from 'sinon';
+
+import rewire = require('rewire');
+
+// Creates a test database and returns an
```
20k-ultra, line 7: Returns a query builder ?

test/src/compose/service.spec.ts
```diff
@@ -0,0 +1,1034 @@
...
+				'2000/tcp': {},
+				'2001/tcp': {},
+				'2002/tcp': {},
+				'2003/tcp': {},
+				'243/udp': {},
+			});
+		});
+
+		it('should correctly handle large port ranges', async function () {
+			this.timeout(60000);
```
20k-ultra, line 231: what's this for ?

20k-ultra:
self certify when you like.

test/src/compose/service.spec.ts
pipex: No idea, that was there before probably. I guess is a timeout for the test? Not sure why it's needed though

test/lib/db-helper.ts
pipex: right, will fix right away

pipex:
@balena-ci I self certify!

PR #1697, opened 2021-05-05
Refactor README to include more specific PR documentation

20k-ultra:
Change-type: patch

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,13 +31,31 @@ decision making processes!
 
 ### Pull requests
 
-Here's a few guidelines to make the process easier for everyone involved.
+Your PR will require the following to pass:
 
-- Every PR _should_ have an associated issue, and the PR's opening comment should say "Fixes #issue" or "Closes #issue".
-- We use [Versionist](https://github.com/resin-io/versionist) to manage versioning (and in particular, [semantic versioning](https://semver.org)) and generate the changelog for this project.
-- At least one commit in a PR should have a `Change-Type: type` footer, where `type` can be `patch`, `minor` or `major`. The subject of this commit will be added to the changelog.
-- Commits should be squashed as much as makes sense.
-- Commits should be signed-off (`git commit -s`)
+ - Each commit be signed-off
+ - At least one commit in the PR contains a `Change-type` value.
+
+See below for more information.
+
+#### Commit signatures
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,13 +31,31 @@ decision making processes!
...
-- Commits should be squashed as much as makes sense.
-- Commits should be signed-off (`git commit -s`)
+ - Each commit be signed-off
+ - At least one commit in the PR contains a `Change-type` value.
+
+See below for more information.
+
+#### Commit signatures
+
+Commits should be signed-off (`git commit -s`) and at least one commit contains a `Change-type` so that [Versionist](https://github.com/resin-io/versionist) can correctly update the application's [semantic versioning](https://semver.org)). Change-type's value is either patch, minor, or major.
```
pipex, line 43: Should we keep requiring the signature? The commit already has an author and the PR is reviewed by us.

README.md
```diff
@@ -31,13 +31,31 @@ decision making processes!
...
+
+#### Clean history
+
+Fixup (squash) as many commits as you can without bundling too many changes into a single commit. However, also don't bundle too many changes as it makes reviewing and regression debugging harder.
```
pipex, line 47: This is a bit confusing imho. Bundle your changes but don't bundle too much? We could say something like 'be conscious about the way your bundle your commits in order to make reviewing easier. This will improve the chances of your PR being reviewed'

Another route could be add some examples of good and bad commits, and linking to https://chris.beams.io/posts/git-commit/

cywang117, line 47:
```suggestion
Fixup (squash) as many commits as you can without bundling too many changes into a single commit. However, don't bundle too many changes as it makes reviewing and regression debugging harder. A good rule of thumb to follow when trying to determine which commits to squash together is to group the ones that relate to one atomic change, including the atomic change's tests if any.
```

README.md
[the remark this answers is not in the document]
cywang117: After the most recent 3rd party PR, I think it's worth mentioning that the footer should be the last bit of the PR, to avoid the pain the user went through figuring this out.

20k-ultra:
I'm in favour of making a general balena contribution doc rather then adding another one to this repo. Will make a PR linking to the balena org contribution doc when it's made.

PR #1698, opened 2021-05-06
Optimize current state filtering

Page-:


src/types/state.ts
```diff
@@ -40,7 +40,9 @@ export interface DeviceStatus {
 		};
 	} & DeviceReportFields;
 	// TODO: Type the dependent entry correctly
-	dependent?: any;
+	dependent?: {
+		[key: string]: any;
+	};
```
Page-, line 45: This is very slightly better and was necessary for `keyof DeviceStatus['dependent']` above to actually return a reasonable type

src/device-state/current-state.ts
```diff
@@ -108,25 +108,25 @@ const getStateDiff = (): DeviceStatus => {
...
-			.value(),
-		dependent: _(stateForReport.dependent)
-			.omitBy((val, key: keyof DeviceStatus['dependent']) =>
-				_.isEqual(lastReportedDependent[key], val),
-			)
-			.omit(INTERNAL_STATE_KEYS)
-			.value(),
+		local: _.omitBy(
+			stateForReport.local,
+			(val, key: keyof NonNullable<DeviceStatus['local']>) =>
```
Page-, line 113: Without the `NonNullable` this was returning a type of `never` since undefined cannot have a `keyof` - it was working because apparently indexing with `never` just works?

src/device-state/current-state.ts
```diff
@@ -108,25 +108,25 @@ const getStateDiff = (): DeviceStatus => {
...
-			.omitBy((val, key: keyof DeviceStatus['local']) =>
-				_.isEqual(lastReportedLocal[key], val),
-			)
-			.omit(INTERNAL_STATE_KEYS)
-			.value(),
-		dependent: _(stateForReport.dependent)
-			.omitBy((val, key: keyof DeviceStatus['dependent']) =>
-				_.isEqual(lastReportedDependent[key], val),
-			)
-			.omit(INTERNAL_STATE_KEYS)
```
Page-, line 121: Combining this `omit` into the one above avoids having to loop through the object a second time and create an extra result object which should reduce cpu usage both for the filtering and for the gc

src/device-state/current-state.ts
```diff
@@ -108,25 +108,25 @@ const getStateDiff = (): DeviceStatus => {
...
+				_.isEqual(lastReportedDependent[key], val) ||
+				INTERNAL_STATE_KEYS.includes(key),
+		),
 	};
 
-	const toOmit: string[] = sysInfo.filterNonSignificantChanges(
-		lastReportedLocal as sysInfo.SystemInfo,
-		stateForReport.local as sysInfo.SystemInfo,
-	);
-	diff.local = _.omit(diff.local, toOmit);
```
Page-, line 129: Same story, combining this in reduces the work/extra object and the corresponding cpu savings

src/lib/system-info.ts
```diff
@@ -120,28 +120,22 @@ const significantChange: { [key in keyof SystemInfo]?: number } = {
...
-				return true;
-			}
-
-			return (
-				Math.floor((value as number) / bucketSize) !==
-				Math.floor((past[key] as number) / bucketSize)
-			);
-		}),
-	) as Array<keyof SystemInfo>;
+export function isSignificantChange(
```
Page-, line 123: I changed to a single function that just returns whether the change is significant or not, it's more versatile and imo it's also simpler due to avoiding the weird keys/omitBy combo, plus the previous function name was a bit misleading as it did not filter but instead returned a list of keys

src/device-state/current-state.ts
```diff
@@ -108,25 +108,25 @@ const getStateDiff = (): DeviceStatus => {
...
-				_.isEqual(lastReportedLocal[key], val),
-			)
-			.omit(INTERNAL_STATE_KEYS)
-			.value(),
-		dependent: _(stateForReport.dependent)
-			.omitBy((val, key: keyof DeviceStatus['dependent']) =>
+		local: _.omitBy(
+			stateForReport.local,
+			(val, key: keyof NonNullable<DeviceStatus['local']>) =>
+				INTERNAL_STATE_KEYS.includes(key) ||
```
Page-, line 114: Moving this before the `_.isEqual` avoids the need to run `_.isEqual` when we're just going to discard it anyway, I think the keys we discard should be fairly cheap to compare so it probably won't matter much overall but it makes sense as a best practice to put the cheaper check first

Page-:
@balena-ci rebase

cywang117:
@balena-ci rebase

cywang117 (approved):

PR #1701, opened 2021-05-06
Patch awaiting response when checking if supervisor0 network exists

20k-ultra:
Our changes to move to using standard lib fs functions added a regression where checking if the supervisor0 network exists is not awaited for the response. This fixes that. https://github.com/balena-os/balena-supervisor/commit/4a2ac557ef11d334154bafa0cab4e7272cbed749#diff-bb7547de5afc63925925900846ce189e9b34cc4b8d1916e70928fa44f98a8fd0L73

### DIFF
--- a/src/compose/network-manager.ts
+++ b/src/compose/network-manager.ts
@@ -70,7 +70,7 @@ export async function remove(network: Network) {
 }
 
 export async function supervisorNetworkReady(): Promise<boolean> {
-	const networkExists = exists(
+	const networkExists = await exists(
 		`/sys/class/net/${constants.supervisorNetworkInterface}`,
 	);
 	if (!networkExists) {


20k-ultra:
@balena-ci I self-certify!

pipex:
@balena-ci retest

pipex:
@balena-ci rebase

src/compose/network-manager.ts
```diff
@@ -70,7 +70,7 @@ export async function remove(network: Network) {
 }
 
 export async function supervisorNetworkReady(): Promise<boolean> {
-	const networkExists = exists(
+	const networkExists = await exists(
```
Page-, line 73: This will also get caught automatically in typescript 4.3: https://devblogs.microsoft.com/typescript/announcing-typescript-4-3-beta/#errors-on-always-truthy-promise-checks 🎉

20k-ultra: Awesome! I was sure there is a way to detect this sort of issue..

20k-ultra: @cywang117 look

PR #1702, opened 2021-05-06
Testing versionist

20k-ultra:


cywang117:
@20k-ultra Change-Type doesn't work? Oh bother..

cywang117:
Disregard the above, I saw your comment in #1691

20k-ultra:
You can see from the commits which passed and failed that the issue was the commit message footers were not considered footers because they weren't at the bottom.

PR #1707, opened 2021-05-07
Upgrade mocha to v8, use mocha fixtures for chai plugins

cywang117:
Chai plugins are hooked in at the start of the test suite and globally available without needing to import them throughout the tests. This way we can remove test/lib/chai-config. I've also removed mocked-event-tracker since it just spies on event-tracker, and we can do that more atomically whenever it's needed instead of having it globally, which would just end up being confusing.

I still need to remove the imports from 00-init.spec.ts, that will have to wait for a future PR.

Change-type: patch


cywang117:
Confirmed that `npm run test` and `npm run test:fast` still both work for the full test suite.

cywang117:
@balena-ci rebase

test/00-init.spec.ts
```diff
@@ -33,4 +34,3 @@ fs.writeFileSync(
 import './lib/mocked-dbus';
 import './lib/mocked-dockerode';
 import './lib/mocked-iptables';
-import './lib/mocked-event-tracker';
```
cywang117, line 36: This file only contains a `sinon.spy` call to the event-tracker module. We should make sinon calls when needed within individual suites - having a global sinon stub/spy is not intuitive since you need to remember to import that stub/spy in tests when you need it, or risk a `Attempted to wrap ${FN} which is already wrapped` error.

test/07-blink.spec.ts
```diff
@@ -13,7 +13,6 @@ describe('blink', () => {
 	});
 
 	it('writes to a file that represents the LED, and writes a 0 at the end to turn the LED off', async () => {
-		// TODO: Fix the typings for blink
```
cywang117, line 16: `blinking` was updated to v0.0.4 with TypeScript support recently, which resolves this TODO.

test/config/fixtures.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Also, avoid setting global mutable variables here.
+ */
+export const mochaGlobalSetup = function () {
+	console.log('Setting up global fixtures for tests...');
+
+	/* Setup chai assertion plugins */
+	chai.use(chaiAsPromised);
+	chai.use(sinonChai);
+	// tslint:disable-next-line
+	chai.use(require('chai-events'));
```
pipex, line 24: Can't you do `import * as chaiEvents` above and then do `chai.use(chaiEvents)` here?

Did you try removing that dependency? I see a couple of test suites that import it but I don't see any of the assertions related to the package https://www.npmjs.com/package/chai-events

pipex:
This is awesome. I just have the question about `chaiEvents` but otherwise I say let's merge it

test/config/fixtures.ts
cywang117: Sounds good @pipex, looks like it's being imported in a number of places but the assertions are not used.

cywang117: I've scrubbed all mentions of chai-events from the codebase :)

cywang117:
@balena-ci I self-certify!

PR #1711, opened 2021-05-18
WIP: Test mock setup improvements

cywang117:
Various testing setup and mocking improvements in an attempt to make unit tests more atomic and less reliant on third parties. Changes made with the goal of removing imports with side effects for tests. Changes include:

- Better mocked-dbus.ts module
- Encapsulated lib/dbus.ts module to prevent import with side effect. (Specifically, encapsulate `dbus.getBus` call. Previously it breaks during tests, as it's called first when it's imported before `mochaGlobalSetup` can take place.)
- Remove physical LED_FILE for blink test in favor of mock-fs
- Suppress `log.api` logs in console during tests
- fs-utils tests: Remove duplicate hooks, fix `exists` test suite which was previously skipped

WIP; more to come


20k-ultra:
let's keep https://github.com/balena-os/balena-supervisor/pull/1696 in mind and not break anything I guess ?

cywang117:
Closing because of obsolete changes.

PR #1712, opened 2021-05-25
Revert multiline bullet points in notes

cywang117:
Docs engine cannot support > - markdown pattern which was added by me in #1708.

Should we merge this though, or revert the previous merge? 

Change-type: patch

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -208,9 +208,7 @@ $ curl -X POST --header "Content-Type:application/json" \
 
 ### POST /v1/purge
 
-> **Note:** 
-> - On devices with supervisor version lower than 7.22.0, replace all `BALENA_` variables with `RESIN_`, e.g. `RESIN_SUPERVISOR_ADDRESS` instead of `BALENA_SUPERVISOR_ADDRESS`.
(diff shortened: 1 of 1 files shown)


cywang117:
image

^ This time I checked with a local docs server just to be sure. This is what it looks like with this PR.

The current doc looks like this for the bullet-pointed entries: https://www.balena.io/docs/reference/supervisor/supervisor-api/#post-v1restart. 

I'll self-certify on Monday if no one comments, since I would expect this to be acceptable.

20k-ultra (approved):
I don't expect the docs to ever allow bullet points in a notes section tbh.

cywang117:
> I don't expect the docs to ever allow bullet points in a notes section tbh.

big sad, I love GitHub markdown

20k-ultra:
I think it's just markdown not github specifically markdown but I do know they have special features like checkboxes. 

I'm not sure what is required to make the engine render bullet points in the note section but I said that because I personally thought bullet points in the note section would be strange. To me it's like seeing a photo with a caption below it describing what's in the photo and then having bullet points with more info. I'm saying this because you can tell me i'm wrong and extend the engine to render bullet points in the notes section :p

Issue #1673, opened 2021-04-23
New CPU usage calculation causes cpu spikes only engine sees


20k-ultra:
See https://github.com/balena-os/balena-supervisor/issues/1673#issuecomment-850010197 for explanation of what exactly is happening.

When running v11.4.10 the usage reminds below 5% for over 600 samples but if you run 12.5.10 it goes up to 100%. When you check what process within the resin_supervisor container may be using up so much process, the only thing running is the supervisor node process which never goes above 10% either.

`balena stats resin_supervisor` for sampling cpu usage
`balena exec resin_supervisor top` for seeing processes running

[supervisor_v11-4-10.log](https://github.com/balena-os/balena-supervisor/files/6367677/supervisor_v11-4-10.log) 
[supervisor_v12-0-0.log](https://github.com/balena-os/balena-supervisor/files/6367680/supervisor_v12-0-0.log) 
[supervisor_v12-5-0.log](https://github.com/balena-os/balena-supervisor/files/6367564/supervisor_v12-5-0.log) 
[supervisor_v12-5-7.log](https://github.com/balena-os/balena-supervisor/files/6367692/supervisor_v12-5-7.log) 
[supervisor_v12-5-8.log](https://github.com/balena-os/balena-supervisor/files/6367625/supervisor_v12-5-8.log)

From the above logs sampling cpu usage over 600 times shows that the Supervisor is doing something new which causes large cpu spikes in v12.5.8. In that change we modified how the cpu usage is calculate for the dashboard (not the same as engine stats) which is probably the culprit https://github.com/balena-os/balena-supervisor/commit/204475d3dcbbb8a69b681d0348b39b5b486b628b.

20k-ultra:
I have observed that calculating the cpu information is executed several times (maybe 10) in quick succession for some reason which is where the cpu spike comes from. I'm not sure if it's the calculation of getting the usage or the fact it is performed so many times quickly. Still investigating.

20k-ultra:
It's not the new cpu usage algorithm that causes the spikes because when I return a static value such as 5 I still see the CPU spikes. I have observed that the Supervisor just asks for the cpu usage in quick succession once and awhile. I just have to find why it's asking more then before.

20k-ultra:
[...]

PR #1715, opened 2021-05-28
Prevent a recursive loop when reporting current state

20k-ultra:
The code previously calculated the current state until the state it just calculates is the same as the last calculated state. This is an issue with device metrics because a single CPU temp/usage, bit in memory usage, etc would cause a loop making the device calculate the current state over and over until it breaks out. Theoretically it could get stuck into this loop forever. 

After applying this fix I saw that calculating the current state was only performed every 10 seconds which matches the maxReportFrequency is value. There are more optimizations to be done in this module with regards to calculating current state diff because as we increase the size of the current state that function will be performing more and more operations. This is evident in CPU usage as I documented in #1673 which shows CPU spikes growing as the Supervisor continues to be developed. We should optimize the state diff check function next to try and reduce the 10-50% CPU spikes I was seeing even with this change however it was usually a 18% cpu spike every 10 seconds when the current state diff was ran. Only once was it greater then 18% over 5 minutes.

Closes: #1673
Change-type: patch

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -203,14 +203,10 @@ const reportCurrentState = (): null => {
 				...currentDeviceState.dependent,
 			};
 
-			const stateDiff = getStateDiff();
-			if (_.size(stateDiff) === 0) {
-				reportPending = false;
-				return null;
-			}
-
+			// Report current state
 			await report();
-			reportCurrentState();
+			// Finishing pending report
+			reportPending = false;
 		} catch (e) {
 			eventTracker.track('Device state report failure', { error: e });
 			// We use the poll interval as the upper limit of
@@ -241,5 +237,6 @@ export const startReporting = () => {
 	// But check once every max report frequency to ensure that changes in system
 	// info are picked up (CPU temp etc)
 	setInterval(doReport, constants.maxReportFrequency);
-	return reportCurrentState();
+	// Try to perform a report right away
+	return doReport();
 };
--- a/test/src/device-state/current-state.spec.ts
+++ b/test/src/device-state/current-state.spec.ts
@@ -505,25 +505,6 @@ describe('device-state/current-state', () => {
 			report.cancel();
 		});
 
-		it('does not report if current state has not changed', async () => {
-			// Use a temporary unhandledRejectionHandler to catch the promise
-			// rejection from the Bluebird.delay stub
-			process.on('unhandledRejection', unhandledRejectionHandler);
-			spy(_, 'size');
-
-			reportCurrentState();
-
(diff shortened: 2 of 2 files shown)


20k-ultra:
The report function does a state diff so we don't have to do this state diff when generating the report.

test/src/device-state/current-state.spec.ts
```diff
@@ -505,25 +505,6 @@ describe('device-state/current-state', () => {
...
-			// TODO: is there a better way to test this? Possible race condition
-			await sleep(200);
-
-			expect(stateForReport).to.deep.equal({ local: {}, dependent: {} });
-			expect(_.size as SinonSpy).to.have.returned(0);
-
-			(_.size as SinonSpy).restore();
-			process.removeListener('unhandledRejection', unhandledRejectionHandler);
-		});
-
```
20k-ultra, line 526: I removed this case because it's incorrect for the reportCurrentState to decide if a state should be reported. That is implemented in the report function. Having both of these functions check if there is a change was redundant. This report function has the same test to make sure it's not reporting the same state https://github.com/balena-os/balena-supervisor/blob/a4be293ec48755f7c3c75a0db1db6bacb3f628d8/test/src/device-state/current-state.spec.ts#L407-L411

src/device-state/current-state.ts
```diff
@@ -203,14 +203,10 @@ const reportCurrentState = (): null => {
...
 
-			const stateDiff = getStateDiff();
-			if (_.size(stateDiff) === 0) {
-				reportPending = false;
-				return null;
-			}
-
+			// Report current state
 			await report();
-			reportCurrentState();
```
pipex, line 213: So before this, the function would fall into this infinite loop and the interval call on the `startReporting` would always fail. Weird we never caught it before.

test/src/device-state/current-state.spec.ts
pipex: I agree

pipex:
LGTM

src/device-state/current-state.ts
20k-ultra: I had a conversation with Page about this. The system made sense before because changes were pushed rather then queried but with the addition of device metrics we now had a need to query for metrics on an interval. That's why the startReporting function is odd to have onChange and on interval.

https://github.com/balena-os/balena-supervisor/commit/eeda7e3ab17dc68d592acc2e4f58f20dad9a677a#diff-4201a9052aedc0dfdf3d8eb0566ecc3ac822602e1a872e858cddecb2ac772b8cR204-R214

20k-ultra:
@balena-ci I self-certify!

PR #1720, opened 2021-06-08
Improve target state typings

Page-:
Change-type: patch

### DIFF
--- a/src/types/state.ts
+++ b/src/types/state.ts
@@ -55,8 +55,8 @@ export interface TargetState {
 		apps: {
 			[appId: string]: {
 				name: string;
-				commit: string;
-				releaseId: number;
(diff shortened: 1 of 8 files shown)


src/types/state.ts
```diff
@@ -55,8 +55,8 @@ export interface TargetState {
 		apps: {
 			[appId: string]: {
 				name: string;
-				commit: string;
-				releaseId: number;
+				commit?: string;
```
pipex, line 58: Are these fields really optional for v2?

pipex (approved):
Looks good. LGTM

src/types/state.ts
Page-: Yes, apps start out without any release at which point this will be null

Issue #1525, opened 2020-11-19
Document the supervisor firewall rules behaviour and use cases


alexgg:
Customers are confused about this feature which has not been documented anywhere.
See https://jel.ly.fish/22ab7f29-f937-4a63-a3e7-35a79579db7d

20k-ultra:
Thanks Alex, we'll talk about what we should do since this feature isn't meant to be advertised yet but it is causing problems for people that are modifying their firewall.

For anyone that stumbles on this issue to have your iptable rules persist even when you disable the firewall mode in the Supervisor which still wipes your changes, place them into the raw table as the Supervisor only clears the filter table.

cywang117:
>  since this feature isn't meant to be advertised yet
@20k-ultra What do you mean? And/or is this state of non-advertisement for FIREWALL_MODE still valid in May 2021?

20k-ultra:
The firewall feature was an experimental feature that should not have affected anyone but the users which configured a mode. However since https://github.com/balena-os/balena-supervisor/issues/1482 is a thing then the firewall is impacting users who aren't even using it which should be resolved as soon as we can.

Issue #1482, opened 2020-10-08
Modifications to firewall (iptables) always removed by Supervisor


20k-ultra:
The Supervisor has added firewall configuration functionality in v11.9.1 which applies rules based on a mode. If a user wants to apply their own iptable rules they would use `BALENA_HOST_FIREWALL_MODE=off` configuration variable however it has been reported and reproduced that this is not the behaviour.

If you set rules in iptable with your own service and have firewall mode set to off the Supervisor will override and remove your settings. We have created https://github.com/balena-io-playground/iptables-test which easily reproduces this behaviour.

20k-ultra:
For anyone that stumbles on this issue to have your iptable rules persist even when you disable the firewall mode in the Supervisor which still wipes your changes, place them into the raw table as the Supervisor only clears the filter table.

20k-ultra:
@pipex I know we haven't resolved this because we wanted to figure out what to do with the firewall. I think an acceptable stop gap is to at-least, make the Supervisor not clear ALL rules in iptables when the firewall mode is set to off.

pipex:
@20k-ultra I think that is part of the solution, but I don't know how easy is to tell iptables-restore to do that. Otherwise doing that might mean a full reimplementation of the feature

mycloudllc:
With this being a current issues is there anyway around having to set iptable rules to the raw table? I have a few mangle rules I need  to add, but upon reboot those are wiped as well. Any ideas?

cywang117:
This fix is available with v14.9.2+ Supervisors.

PR #1721, opened 2021-06-09
Document the current state of HOST_FIREWALL_MODE

cywang117:
This set of docs can provide info for support agents until
issue 1482 is resolved. Expect this document to change as firewall
implementation is improved.

Connects-to: #1482
Connects-to: #1525
Change-type: patch


cywang117:
Going to merge this since it's useful as a reference doc, we can update it once we revisit firewall implementation

cywang117:
@balena-ci I self-certify!

Issue #1459, opened 2020-09-14
supervisor should not start after a HUP without checking the breadcrumbs (this ensure the hup rollback services have run and all is ok))


floion:
After a HUP the device is rebooted. When the boards boots, there are a few systemd rollback services that run to ensure the HUP was ok. These services are `rollback-health` and `rollback-altboot`. The supervisor should only start after these rollback services have finished. If the supervisor starts while these services have not finished then the supervisor may reboot the board (for example if the target state is different from the current state, the supervisor will do the changes to apply the target state and reboot the board) and if the rollback services have not done their part (clear upgrade_available, reset boot counter etc) then at next boot the board will boot from wrong rootfs as well as other bad things will happen (especially on nvidia based boards)

20k-ultra:
hey @xginn8, there's been a lot of work done on HUP so just checking this is something the Supervisor will still need to implement.

xginn8:
@20k-ultra as the supervisor absorbs some of the functionality of HUP as part of tri-app, we'll need to figure this out in the supervisor itself.

20k-ultra:
Is this something the Supervisor should start implementing or do we need to figure out / implement some other bits ?

acostach:
@20k-ultra @xginn8 do you have any updates on this?

xginn8:
i think this can be implemented in the supervisor. basically, until rollbacks have completed, no changes should be applied.

20k-ultra:
Is the Supervisor just checking the status of rollback-health and rollback-altboot to be exited ? That seems pretty trivial and we just have to discuss what should the Supervisor do if there are breadcrumbs. 

I like the idea of the Supervisor doing this check tbh. It seems like it will prevent a severe class of issues and absorbs HUP responsibilities since we'd like all applications (Supervisor is one) to be managed the same. What they do when running is up to them.

20k-ultra:
we can created a "reboot lock" much like an update lock. When the reboot lock exists the Supervisor will never try to reboot the device.

The first implementation of reboot locks doesn't have to look for a reboot.lock file, instead we can just check if the "breadcrumbs" which the rollback services themselves check to see if they should rollback. If these state files exist (unsure what they are exactly at this time) then convey that there is a reboot lock.

cywang117:
[...]

PR #1723, opened 2021-06-14
Prevent updates/reboots with locks when HUP breadcrumbs present

cywang117:
On HUP rollback, some services need to complete before
it's safe for the Supervisor to reboot the device when
applying state changes. rollback-{health|altboot}-breadcrumb
are the two files that Supervisor locks the device on when
present in this patch.

Not closing issue 1459 because there is a possible case where,
on altboot rollback, the breadcrumbs are not present. 1459
may be closed when this edge case is investigated.

Added some update-lock.ts tests.

Needs some more integration testing, but adding the rollback files manually in 
`/mnt/state` causes the update locks to behave as expected. Will integration test
more tomorrow.

Change-type: patch
Connects-to: #1459

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -638,6 +638,7 @@ export function reportCurrentState(
 }
 
 export async function reboot(force?: boolean, skipLock?: boolean) {
+	await updateLock.ensureNoHUPBreadcrumbsOnHost();
 	await applicationManager.stopAll({ force, skipLock });
 	logger.logSystemMessage('Rebooting', {}, 'Reboot');
 	const $reboot = await dbus.reboot();
@@ -647,6 +648,7 @@ export async function reboot(force?: boolean, skipLock?: boolean) {
 }
 
 export async function shutdown(force?: boolean, skipLock?: boolean) {
+	await updateLock.ensureNoHUPBreadcrumbsOnHost();
 	await applicationManager.stopAll({ force, skipLock });
 	logger.logSystemMessage('Shutting down', {}, 'Shutdown');
 	const $shutdown = await dbus.shutdown();
@@ -757,7 +759,7 @@ function applyError(
 		if (err instanceof UpdatesLockedError) {
 			const message = `Updates are locked, retrying in ${prettyMs(delay, {
 				compact: true,
-			})}...`;
+			})}. Reason: ${err.message}`;
 			logger.logSystemMessage(message, {}, 'updateLocked', false);
 			log.info(message);
 		} else {
--- a/src/lib/update-lock.ts
+++ b/src/lib/update-lock.ts
@@ -5,8 +5,9 @@ import { promises as fs } from 'fs';
 import * as path from 'path';
 import * as Lock from 'rwlock';
 
-import { ENOENT, UpdatesLockedError } from './errors';
-import { getPathOnHost } from './fs-utils';
+import * as constants from './constants';
+import { ENOENT, EEXIST, UpdatesLockedError } from './errors';
+import { getPathOnHost, pathExistsOnHost } from './fs-utils';
 
 type asyncLockFile = typeof lockFileLib & {
 	unlockAsync(path: string): Bluebird<void>;
@@ -33,6 +34,29 @@ function lockFilesOnHost(appId: number, serviceName: string): string[] {
 			path.join(lockPath(appId), serviceName, filename),
 		),
 	);
+}
+
+/**
+ * Check for rollback-{health|altboot}-breadcrumb, two files that exist while
+ * rollback-{health|altboot}.service have not exited. If these files exist,
+ * prevent reboot. If the Supervisor reboots while those services are still running,
+ * the device may become stuck in an invalid state during HUP.
+ */
+export function ensureNoHUPBreadcrumbsOnHost(): Promise<boolean | never> {
+	return Promise.all(
+		[
+			'rollback-health-breadcrumb',
+			'rollback-altboot-breadcrumb',
+		].map((filename) =>
+			pathExistsOnHost(path.join(constants.stateMountPoint, filename)),
+		),
+	).then((existsArray) => {
+		const anyExists = existsArray.some((exists) => exists);
+		if (anyExists) {
+			throw new UpdatesLockedError('Waiting for Host OS update to finish');
+		}
+		return anyExists;
+	});
 }
 
 const locksTaken: { [lockName: string]: boolean } = {};
@@ -65,6 +89,24 @@ function dispose(release: () => void): Bluebird<void> {
 		.finally(release)
 		.return();
 }
+
+const lockExistsErrHandler = (err: Error, release: () => void) => {
+	let errMsg = err.message;
+	if (EEXIST(err)) {
+		// Extract appId|appUuid and serviceName from lockfile path for log message
+		// appId: [0-9]{7}, appUuid: [0-9a-w]{32}, short appUuid: [0-9a-w]{7}
+		const pathMatch = err.message.match(
+			/\/([0-9]{7}|[0-9a-w]{32}|[0-9a-w]{7})\/(.*)\/(?:resin-)?updates.lock/,
+		);
+		if (pathMatch && pathMatch.length === 3) {
+			errMsg = `Lockfile exists for ${JSON.stringify({
+				serviceName: pathMatch[2],
+				[/^[0-9]{7}$/.test(pathMatch[1]) ? 'appId' : 'appUuid']: pathMatch[1],
+			})}`;
+		}
+	}
+	return dispose(release).throw(new UpdatesLockedError(errMsg));
+};
 
 /**
  * Try to take the locks for an application. If force is set, it will remove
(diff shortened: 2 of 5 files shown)


test/src/lib/update-lock.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+		beforeEach(() => {
+			// @ts-ignore
+			unlockSyncStub = stub(lockFile, 'unlockSync').callsFake((lockPath) => {
+				// Throw error on process.exit for one of the two lockpaths
+				if (lockPath === testLockPaths[1]) {
+					throw new Error(
+						'handled unlockSync error which should not crash test process',
+					);
+				}
+			});
```
cywang117, line 115: This tests both error thrown + no error thrown case when attempting to cleanup lockfiles on Node process exit. The error thrown case should be handled silently by the code currently.

test/src/lib/update-lock.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+				constants.rootMountPoint,
+				updateLock.lockPath(appId),
+				service,
+			)]: {
+				'updates.lock': '',
+			},
+		});
+	};
+
+	// TODO: Remove these hooks when we don't need './test/data' as test process's rootMountPoint
```
cywang117, line 42: These `before`/`after` hooks can be removed when every test that interacts with the fs in some way is switched over to mock-fs instead of using the physical file system.

20k-ultra:
@cywang117 can you ping for review again once the integration tests are done. I'm worried if significant changes will be made then we have to review again.

20k-ultra:
How is the integration testing going ?

cywang117:
Commented in Flowdock, I'll paste my comment here for reference:

Quick update regarding the HUP breadcrumbs reboot lock PR. I've figured out why I've been having trouble testing it on the TX2 now! I'd previously added breadcrumb checks in `updateLock.lock` which is fine, and this function gets called by `applicationManager.lockingIfNecessary`, but when you follow the code flow upwards to `deviceState.reboot`, the lock function is only reached when there are services and an application present on the device, which the TX2 did not have. This is why I couldn't see the behavior no matter what I tried. It's pretty obvious in hindsight and I'm shaking my head at myself slightly (haha 😢), but now I know how to modify the code to separate the HUP breadcrumb checking functionality from the service-level lockfile checking functionality. 

I'll be making updates to the PR early next week and integration testing it locally then on Ross's TX2, and I should have a working PR ready for review by late next week provided there are no hiccups. (Fingers crossed!)

As far as unit tests go, I think I just need to add a case to test for prevention of deviceState.reboot or deviceState.shutdown when breadcrumbs are present. All else is actual device testing (I've written up a how-to and a convenience script for testing this and I'll share those when I'm done with everything)

cywang117:
@20k-ultra @pipex This PR is ready for review! Integration tests completed successfully. [Here](https://www.notion.so/HUP-rollback-tests-6827d84a7f6b4d798ac0ceccb88582a4) is the process I followed.

image

src/device-state.ts
```diff
@@ -647,6 +648,7 @@ export async function reboot(force?: boolean, skipLock?: boolean) {
 }
 
 export async function shutdown(force?: boolean, skipLock?: boolean) {
+	await updateLock.ensureNoHUPBreadcrumbsOnHost();
```
cywang117, line 651: Adding the HUP checks here instead of in updateLock.lock. That function is currently only entered when there is an application running on the device, because `lockingIfNecessary` is part of `applicationManager`. However when no app / service is present, updateLock.lock never gets called, hence why I was having trouble with integration tests before.

test/05-device-state.spec.ts
```diff
@@ -355,4 +356,26 @@ describe('deviceState', () => {
...
+			.then(() => expect.fail())
+			.catch((err: any) => {
+				expect(err.message).to.equal(testErrMsg);
+			});
+		await deviceState
+			.shutdown()
+			.then(() => expect.fail())
+			.catch((err: any) => {
+				expect(err.message).to.equal(testErrMsg);
+			});
```
20k-ultra, line 377: I think this will fail if the promise does not throw but also passes when it does throw and matches our error (also matches on the error type which is nice since some logic does err instanceof)

`
await expect(deviceState.reboot()).to.eventually.throw(new UpdatesLockedError(testErrMsg));
await expect(deviceState.shutdown()).to.eventually.throw(new UpdatesLockedError(testErrMsg));
`

20k-ultra: feel free to see if this cleans up any other test case code such as the update-lock spec

test/src/lib/update-lock.spec.ts
```diff
@@ -0,0 +1,273 @@
...
+import * as Bluebird from 'bluebird';
+
+import rewire = require('rewire');
+import mockFs = require('mock-fs');
+
+import * as constants from '../../../src/lib/constants';
+import { UpdatesLockedError } from '../../../src/lib/errors';
+
+describe('lib/update-lock', () => {
+	const updateLock = rewire('../../../src/lib/update-lock');
```
20k-ultra, line 13: glad to see rewire being useful for others as well. I've really enjoyed it.

src/lib/update-lock.ts
```diff
@@ -35,6 +36,29 @@ function lockFilesOnHost(appId: number, serviceName: string): string[] {
...
+ * rollback-{health|altboot}.service have not exited. If these files exist,
+ * prevent reboot. If the Supervisor reboots while those services are still running,
+ * the device may become stuck in an invalid state during HUP.
+ */
+export function ensureNoHUPBreadcrumbsOnHost(): Bluebird<boolean | never> {
+	return Bluebird.all(
+		[
+			'rollback-health-breadcrumb',
+			'rollback-altboot-breadcrumb',
+		].map((filename) =>
```
20k-ultra, line 50: can we replace this bluebird call is [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) ?

20k-ultra: I think you did it since the file had bluebird all over it anyways. I don't want to push to scope of this PR but we should refactor this file to remove bluebird. Especially since the lock function is callback hell.

20k-ultra (commented):
killed it! Do you know what the Supevisor API response is when you try to hit shutdown or reboot from the dashboard while these breadcrumbs exist ? Check that out then you can self certify.

I'd say let's hold off on the refactor for another PR to prevent scope creep here. This is already a substantial change so we don't want to introduce too much at once. Feel free to refactor bluebird out and remove the callback hell from lock() if you want after this.

20k-ultra:
@balena-ci rebase

src/lib/update-lock.ts
cywang117: I totally forgot to switch out Bluebird for Promise here. Explanation, this function was just used inside updateLock.lock before, which uses Bluebird promises because of `.catchReturn` and `.disposer` which native Promises don't implement. I've since removed ensureNoHUP... from updateLock so Bluebird promises aren';t needed anymore. Thanks for the reminder!

test/src/lib/update-lock.spec.ts
cywang117: Same! It's been invaluable in tests.

test/05-device-state.spec.ts
cywang117: `await expect(deviceState.shutdown()).to.eventually.throw(new UpdatesLockedError(testErrMsg));` doesn't work but removing the async/await pattern from the test case does:
```
expect(deviceState.reboot()).to.eventually.throw(new UpdatesLockedError(testErrMsg));
expect(deviceState.shutdown()).to.eventually.throw(new UpdatesLockedError(testErrMsg));
```

Verified that the stub restore call after these 2 expects is called as well. Chai-as-promised handles the underlying promise so no need for async/await I guess.

cywang117: That's weird, because some other test cases use async/await no problem for a setup similar to this..

cywang117:
```
await expect(deviceState.reboot()).to.be.rejectedWith(testErrMsg);
await expect(deviceState.shutdown()).to.be.rejectedWith(testErrMsg);
```
works. Gotta remember to use `rejectedWith` instead of `throw` (note that `rejectedWith(new UpdatesLockedError(testErrMsg)` will not pass). I'm almost certain I've dealt with this exact scenario before with chai-as-promised while writing tests, which means I almost prefer the non chai-as-promised method I had originally:
```
await promise()
.then(() => expect.fail())
.catch(err => {
  expect(err).to.be.instanceOf(UpdatesLockedError);
  expect(err.message).to.equal(testErrMsg);
})
```

cywang117: Already committed to using chai-as-promised though because I spent time on it, so I'll change it 👍🏼

cywang117:
@20k-ultra I removed Bluebird from updateLock.ensureNoHUPBreadcrumbsExist but not the updateLock.lock function, since we still need Bluebird there because of the .disposer util which native Promises doesn't have. I'd love to remove this promise hell (and looked into it while coding this PR) but we'd need a good native alternative.

cywang117:
> Do you know what the Supevisor API response is when you try to hit shutdown or reboot from the dashboard while these breadcrumbs exist?
```
root@92f8b89:/mnt/state# curl -X POST --header "Content-Type:application/json" "10.114.104.1:48484/v1/reboot?apikey=dc28
27d153e94e92a9d2f5ab7a2cda5b" --data '{"force": true}'
{"Data":"","Error":"Waiting for Host OS update to finish"}
```
Looks good to me :)

cywang117:
@balena-ci I self-certify!

PR #1725, opened 2021-06-17
Add revpi-connect, revpi-core-3 to Raspberry Pi variants

floion:
We need the supervisor to be able to manage config.txt changes for these
Revolution Pi boards too.

Change-type: patch
Signed-off-by: Florin Sarbu <florin@balena.io>


cywang117 (approved):
LGTM

PR #1728, opened 2021-06-24
Add download progress LED blink

tmigone:
On devices that have ACT LED, blink it while a new release
is being downloaded.

Change-type: minor
Signed-off-by: Tomás Migone <tomas@balena.io>

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -14,6 +14,7 @@ import * as config from '../config';
 import { SchemaTypeKey, SchemaReturn } from '../config/schema-type';
 import * as eventTracker from '../event-tracker';
 import * as deviceState from '../device-state';
+import { downloadProgress } from '../lib/download-progress';
 
 // The exponential backoff starts at 15s
 const MINIMUM_BACKOFF_DELAY = 15000;
@@ -234,6 +235,7 @@ export const startReporting = () => {
 
 	// If the state changes, report it
 	deviceState.on('change', doReport);
+	deviceState.on('change', downloadProgress);
 	// But check once every max report frequency to ensure that changes in system
 	// info are picked up (CPU temp etc)
 	setInterval(doReport, constants.maxReportFrequency);
--- a/src/lib/download-progress.ts
+++ b/src/lib/download-progress.ts
@@ -1,0 +1,88 @@
+import * as _ from 'lodash';
+import blink = require('./blink');
+import { DeviceStatus, TargetState } from '../types/state';
+import * as deviceState from '../device-state';
+import * as targetState from '../device-state/target-state';
+
+interface ServiceState {
+	id: string;
+	status: string;
+	download_progress: number | null;
+}
+
+const DOWNLOAD_BLINK_THROTTLE = 10000;
+const DOWNLOAD_BLINK_PERIOD_BASE = 100;
+const DOWNLOAD_BLINK_PERIOD_STEP = 10;
+
+const throttledBlink = _.throttle((progress: number) => {
+	blink.pattern.stop();
+	if (progress < 100) {
+		blink.pattern.start({
+			blinks: 1,
+			pause: 0,
+			onDuration:
+				DOWNLOAD_BLINK_PERIOD_BASE + DOWNLOAD_BLINK_PERIOD_STEP * progress,
+			offDuration:
+				DOWNLOAD_BLINK_PERIOD_BASE + DOWNLOAD_BLINK_PERIOD_STEP * progress,
+		});
+	}
+}, DOWNLOAD_BLINK_THROTTLE);
+
+export async function downloadProgress() {
+	const currentDeviceState = await deviceState.getStatus();
+
+	if (currentDeviceState.local?.update_pending) {
+		// Get target services and current state
+		const targetDeviceState = await targetState.get();
+		const targetServicesIDs = getTargetServicesIds(targetDeviceState);
+		const currentServicesStates = getCurrentServicesStates(currentDeviceState);
+
+		// Calculate overall progress
+		let progress = 0;
+		for (const id of targetServicesIDs) {
+			const service = currentServicesStates.find((s) => s.id === id);
+			if (service) {
+				if (['Downloaded', 'Installing', 'Running'].includes(service.status)) {
+					// States after 'Downloading' report 'null' download_progress. Patch those to 100%
+					progress += 100;
+				} else {
+					progress += service.download_progress ?? 0;
+				}
+			}
+		}
+		progress /= targetServicesIDs.length;
+
+		// Blink!
+		throttledBlink(progress);
+	} else {
+		blink.pattern.stop();
+	}
+}
+
+function getTargetServicesIds(target: TargetState): string[] {
+	const servicesStatus: string[] = [];
+	_.toPairs(target.local?.apps).map(([, app]) => {
+		_.toPairs(app.services).map(([, service]) => {
+			servicesStatus.push(service.imageId.toString());
+		});
+	});
+
+	return servicesStatus;
+}
+
+function getCurrentServicesStates(current: DeviceStatus): ServiceState[] {
+	const serviceStatus: ServiceState[] = [];
+	_.toPairs(current.local?.apps).map(([, app]) => {
+		_.toPairs(app).map(([, service]) => {
+			_.toPairs(service).map(([id, serviceDetails]) => {
+				serviceStatus.push({
+					id,
+					status: serviceDetails.status,
+					download_progress: serviceDetails.download_progress,
+				});
+			});
+		});
+	});
+
+	return serviceStatus;
+}


src/device-state/current-state.ts
```diff
@@ -234,6 +235,7 @@ export const startReporting = () => {
 
 	// If the state changes, report it
 	deviceState.on('change', doReport);
+	deviceState.on('change', downloadProgress);
```
20k-ultra, line 238: The startReporting function is for sending the current state to the cloud so flashing the LED given download progress is not relevant. I think moving it outside the function to the top of the file with a comment saying "This will indicate download progress via the device LED" is a good compromise.

20k-ultra (changes requested):
download-progress looks great but can you add some testing. Create a file in /test/src/lib called download-progress.spec.ts. Refer to other test files for how we write tests.

Implementing cases that cover just the exported functions is enough but you can use rewire as shown in this [test suite](https://github.com/balena-os/balena-supervisor/blob/master/test/src/lib/update-lock.spec.ts) on how we've tested private function as well as exported.

cywang117:
I was about to say as well, some tests would be much appreciated :), in test/src/lib/download-progress.spec.ts as Miguel said.

Please ignore the test files that start with numbers -- we're slowly but surely moving away from numbered-file based testing to one-test-file-per-module based testing.

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+	});
+
+	return servicesStatus;
+}
+
+function getCurrentServicesStates(current: DeviceStatus): ServiceState[] {
+	const serviceStatus: ServiceState[] = [];
+	_.toPairs(current.local?.apps).map(([, app]) => {
+		_.toPairs(app).map(([, service]) => {
+			_.toPairs(service).map(([id, serviceDetails]) => {
```
cywang117, line 77: If we're not using the mapped array (from `.map`) anywhere, wouldn't a `.forEach` be better?

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+		// Blink!
+		throttledBlink(progress);
+	} else {
+		blink.pattern.stop();
+	}
+}
+
+function getTargetServicesIds(target: TargetState): string[] {
+	const servicesStatus: string[] = [];
+	_.toPairs(target.local?.apps).map(([, app]) => {
```
cywang117, line 64: Same as my other `.map` -> `.forEach` comment

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+			if (service) {
+				if (['Downloaded', 'Installing', 'Running'].includes(service.status)) {
+					// States after 'Downloading' report 'null' download_progress. Patch those to 100%
+					progress += 100;
+				} else {
+					progress += service.download_progress ?? 0;
+				}
+			}
+		}
+		progress /= targetServicesIDs.length;
```
cywang117, line 53: Looking at the code it seems pretty certain that `targetServicesIDs.length` will never be 0 if there's an application running. What happens if there aren't any apps/services running (Device is "Factory build"), and you trigger a config change?

cywang117: I suppose this thought is only relevant if you keep `deviceState.on('change', downloadProgress)` since 'change' events are also emitted on config changes. If you refactor so that downloadProgress is only called on application changes though, this shouldn't be an issue.

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+		blink.pattern.stop();
+	}
+}
+
+function getTargetServicesIds(target: TargetState): string[] {
+	const servicesStatus: string[] = [];
+	_.toPairs(target.local?.apps).map(([, app]) => {
+		_.toPairs(app.services).map(([, service]) => {
+			servicesStatus.push(service.imageId.toString());
+		});
```
pipex, line 67: This seems a bit convoluted for what you are trying to achieve. Could something like this work? 

```suggestion
       Object.values(target.local?.apps)
           .reduce((services, app) => services.concat(Object.values(app.services).map(s => s.imageId.toString())), [])
```

pipex: It may be that I'm not entirely accustomed to lodash methods though

src/device-state/current-state.ts
pipex: I think it depends on the definition of reporting. I'm personally OK with it being set here, I ultimately would like to get away from this type of event listener or at least I would like to have it being defined within a better API for device state.

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+	download_progress: number | null;
+}
+
+const DOWNLOAD_BLINK_THROTTLE = 10000;
+const DOWNLOAD_BLINK_PERIOD_BASE = 100;
+const DOWNLOAD_BLINK_PERIOD_STEP = 10;
+
+const throttledBlink = _.throttle((progress: number) => {
+	blink.pattern.stop();
+	if (progress < 100) {
```
pipex, line 19: This check seems unnecessary if the downloadProgress function is tested

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+		});
+	}
+}, DOWNLOAD_BLINK_THROTTLE);
+
+export async function downloadProgress() {
+	const currentDeviceState = await deviceState.getStatus();
+
+	if (currentDeviceState.local?.update_pending) {
+		// Get target services and current state
+		const targetDeviceState = await targetState.get();
```
pipex, line 36: I think you might be able to use `deviceState.getTarget` and skip an import

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+const DOWNLOAD_BLINK_PERIOD_STEP = 10;
+
+const throttledBlink = _.throttle((progress: number) => {
+	blink.pattern.stop();
+	if (progress < 100) {
+		blink.pattern.start({
+			blinks: 1,
+			pause: 0,
+			onDuration:
+				DOWNLOAD_BLINK_PERIOD_BASE + DOWNLOAD_BLINK_PERIOD_STEP * progress,
```
pipex, line 24:
```suggestion
                                // Slow down blink frequency as download reaches 100%
				DOWNLOAD_BLINK_PERIOD_BASE + DOWNLOAD_BLINK_PERIOD_STEP * progress,
```

src/lib/download-progress.ts
pipex: I misunderstood, ignore this

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+import * as deviceState from '../device-state';
+import * as targetState from '../device-state/target-state';
+
+interface ServiceState {
+	id: string;
+	status: string;
+	download_progress: number | null;
+}
+
+const DOWNLOAD_BLINK_THROTTLE = 10000;
```
pipex, line 13: Any reason for this particular combination of timings? Maybe you could add a comment for future context?

src/lib/download-progress.ts
```diff
@@ -0,0 +1,88 @@
...
+const DOWNLOAD_BLINK_PERIOD_BASE = 100;
+const DOWNLOAD_BLINK_PERIOD_STEP = 10;
+
+const throttledBlink = _.throttle((progress: number) => {
```
pipex, line 17: I wonder if because of the throttle might get to a case where the blinking gets stuck on a pattern. For instance, let's say the function gets called with `downloadProgress(95)`, and downloading the last 5% takes less than 10 seconds. The function will miss the `downloadProgress(100)` just as the download is finished and it will have to wait until the next time current state changes, which may be an undetermined span of time. 

It's probably not that bad, but it may lead to ambiguous signaling.

src/lib/download-progress.ts
20k-ultra: I'm in favour of standard lib functions :D with node upgrades we can get performance buffs for free

tmigone:
Thanks guys for the review. I'll add testing and address your comments!

src/device-state/current-state.ts
tmigone: Yeah my thinking was that blinking the LED is also reporting information in a way. Added a comment to clarify what's doing but I'll leave it here unless you feel very strongly about it.

src/lib/download-progress.ts
tmigone: No reason other than it looked good. I tried some other combinations and settled on this one eventually as it felt easy to spot. Noted on the comments.

src/lib/download-progress.ts
tmigone: Yes that's certainly a possibility and it's the reason the throttle is 10s and not more. From my (limited) experience even when reaching 100% the supervisor takes way longer to apply the update. During that time `update_pending` is still set to `true` so the throttled blink function will still get invoked and turn off the LED. 

As an extra precaution I also added [this](https://github.com/balena-os/balena-supervisor/blob/aa2370689491abcffa25a1e1c7a2c9f0cbd91241/src/lib/download-progress.ts#L58) which should always be triggered when `update_pending` goes from `true` to `false` (according to my understanding).

So my conclusion is that yes you could be reporting ambiguous signalling but most probably for at most 10 seconds.

src/lib/download-progress.ts
tmigone: Ah good call, i was not aware of that method.

src/lib/download-progress.ts
tmigone: Good catch. I guess you are right, if the device is on "Factory build" there are no services on the target state so length would be 0. So triggering a config change could end the universe by dividing by 0. I didn't consider this scenario! 

It could inadvertedly be solved at the moment if the config change event does not set `update_pending` to `true`, do you know if that's the case? In any case this seems like a more solid approach :P :

> If you refactor so that downloadProgress is only called on application changes though, this shouldn't be an issue.

How would you tell a config change from an app change? Just test the `config` and `apps` keys on `state.local`?

src/lib/download-progress.ts
tmigone: Yes that's a lot clearer and actually pretty close to what my first attempt looked like. I then noticed liberal usage of lodash and `toPairs` in the codebase for similar problems and (wrongly?) assumed that was the norm. 

I'm 100% on board the `Object.values()` train so going to re write this.

src/lib/download-progress.ts
cywang117: > It could inadvertedly be solved at the moment if the config change event does not set update_pending to true, do you know if that's the case?

Looks like a config change definitely [sets update_pending to true](https://github.com/balena-os/balena-supervisor/blob/0cc6920602468b824b08fddaa365c45ba1ff27db/src/device-state.ts#L843-L845).

> How would you tell a config change from an app change? Just test the config and apps keys on state.local?

If image fetch is the only time when you'd want blinking progress, I wonder if you can add the downloadProgress as a callback in only the image fetching step, since that's the only one where you need to download something. Maybe in this [onProgress callback](https://github.com/balena-os/balena-supervisor/blob/fd06c06092b61d349e8b40f66f847065368810e3/src/compose/images.ts#L132-L139) for images.triggerFetch? You can even pass in an `onFinish` callback to stop the blink pattern per what Felipe mentioned below about the `downloadProgress(95)` as an extra safeguard if you'd like.

cywang117: cc @pipex Do you think this would be a good idea to prevent what you mentioned [below](https://github.com/balena-os/balena-supervisor/pull/1728#discussion_r661578710)?

src/lib/download-progress.ts
pipex: Yeah, the supervisor is an old project, so it has a mix of different code styles. We are trying to get away from lodash where we can, so implementing using standard lib is ideal.

src/lib/download-progress.ts
pipex: The problem with the `onProgress` callback is that is per image, while on this case we need the download progress for the application release. I think is fine if we move the `targetServiceIDs` calculation outside of the if and add a `targetServiceIDs.length > 0` to the condition

src/lib/download-progress.ts
tmigone: Awesome, thanks for the input.

src/lib/download-progress.ts
tmigone: Rewriting this to use `Object.values()`, see other comment from Felipe.

src/lib/download-progress.ts
tmigone: Rewriting this to use Object.values(), see other comment from Felipe.

src/lib/download-progress.ts
cywang117: That's a good point ^ ignore my suggestion then :)

tmigone: That sounds good!

20k-ultra:
@tmigone I hope we can still get this merged since you put so much work into it so far

PR #1738, opened 2021-07-02
Allow users to override HUP lock if device is stuck in invalid state

cywang117:
This functionality is needed when breadcrumbs aren't deleted after a HUP
rollback for whatever reason.

Change-type: patch
Connects-to: #1459

### DIFF
--- a/src/lib/update-lock.ts
+++ b/src/lib/update-lock.ts
@@ -42,7 +42,11 @@ function lockFilesOnHost(appId: number, serviceName: string): string[] {
  * prevent reboot. If the Supervisor reboots while those services are still running,
  * the device may become stuck in an invalid state during HUP.
  */
-export function ensureNoHUPBreadcrumbsOnHost(): Promise<boolean | never> {
+export function abortIfHUPInProgress({
+	force = false,
+}: {
+	force: boolean | undefined;
+}): Promise<boolean | never> {
 	return Promise.all(
 		[
 			'rollback-health-breadcrumb',
@@ -52,7 +56,7 @@ export function ensureNoHUPBreadcrumbsOnHost(): Promise<boolean | never> {
 		),
 	).then((existsArray) => {
 		const anyExists = existsArray.some((exists) => exists);
-		if (anyExists) {
+		if (anyExists && !force) {
 			throw new UpdatesLockedError('Waiting for Host OS update to finish');
 		}
 		return anyExists;
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -638,7 +638,7 @@ export function reportCurrentState(
 }
 
 export async function reboot(force?: boolean, skipLock?: boolean) {
-	await updateLock.ensureNoHUPBreadcrumbsOnHost();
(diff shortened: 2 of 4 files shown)


src/lib/update-lock.ts
```diff
@@ -42,7 +42,11 @@ function lockFilesOnHost(appId: number, serviceName: string): string[] {
  * prevent reboot. If the Supervisor reboots while those services are still running,
  * the device may become stuck in an invalid state during HUP.
  */
-export function ensureNoHUPBreadcrumbsOnHost(): Promise<boolean | never> {
+export function ensureNoHUPBreadcrumbsOnHost({
```
klutchell, line 45: Can we rename this function to something like `foundHUPBreadcrumbsOnHost` to avoid any confusion around the `ensureNo` prefix?

pipex: I also don't love the name. I don't think it explains well what the function does, particularly since you cannot infer from the name and reading the code that the function will throw if the condition is not met. 

I think it would make more sense to make the signature more similar to that of the `lock` function (or the `lockIfNecessary` function), where the callback doesn't get executed if the condition is not met. I'm OK with not doing that if the name reflects the behavior though.

pipex (changes requested):
Everything looks good, except for the naming of the function which doesn't really reflect the functionality as I mentioned in the comment

src/lib/update-lock.ts
cywang117: How about `lockDuringHUP`?

cywang117: @pipex You said that it'd be better if this lock functionality had a signature similar to updateLock.lock. This function was a part of updateLock.lock originally, but because the lock doesn't work when there's no application on the device, I separated this functionality out. If I wanted to keep HUP breadcrumb locking functionality within updateLock.lock, I could do so by allowing the appId paramter to be optional, and accounting for different types of locking conditions in updateLock,lock. This would be a bit more of a refactor than than the HUP breadcrumbs PR that was already merged, so I decided against it.

That being said, I can take a look at a refactor if you and @20k-ultra think it's something that would simplify the code.

pipex: I think the refactor is outside scope for the PR, but eventually we should cleanup locking behavior on the codebase. For the function name. Maybe something like `abortIfHUPInProgress`? I want it to make clear that the operation will be interrupted (through an exception) if there is HUP.

20k-ultra:
@cywang117 can you fix this and self-certify so then I can update https://github.com/balena-os/meta-balena/pull/2236 to include this change as well as it seems important to send together

src/lib/update-lock.ts
cywang117: Sounds good. I'll just change the name then, but we can make a note to look at simplifying the lock code at a future time

cywang117:
@20k-ultra I've made the function rename change, but I think @pipex needs to approve because of the previous request changes. I'll try just in case though:

cywang117:
@balena-ci I self-certify!

pipex (approved):
LGTM

PR #1741, opened 2021-07-05
Clarify that /v2/applications/:appId/stop-service is temporary

20k-ultra:
Text is copied from the /v1/apps/:appId/stop description

Change-type: patch


20k-ultra:
@balena-ci self-certify

20k-ultra:
@balena-ci I self-certify!

PR #1742, opened 2021-07-06
patch: Live push starts node with `--inspect`

codewithcheese:
Not sure if you already have a method for attaching a debugger. If not, here is what I use.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "attach",
+            "name": "Attach to remote",
+            "address": "127.0.0.1",
+            "port": 9229,
+            "cwd": "${workspaceRoot}",
+            "program": "${workspaceRoot}/src/app.ts",
+            "outFiles": [
+              "${workspaceRoot}/dist/bundle.js"
+            ],
+            "sourceMaps": true,
+            "localRoot": "${workspaceFolder}",
+            "remoteRoot": "/usr/src/app"
+        }
+    ]
+} 
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,14 +1,10 @@
 ARG ARCH=%%BALENA_ARCH%%
 ARG NODE_VERSION=12.16.2
 
-FROM balenalib/$ARCH-alpine-supervisor-base:3.11 as BUILD
+FROM balenalib/$ARCH-alpine-node:$NODE_VERSION as BUILD
 
 ARG ARCH
 ARG NODE_VERSION
-ARG NODE_ARCHIVE="node-no-intl-v${NODE_VERSION}-linux-alpine-${ARCH}.tar.gz"
-ARG S3_BASE="https://resin-packages.s3.amazonaws.com"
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -50,7 +50,9 @@ RUN npm ci --build-from-source --sqlite=/usr/lib
 
 # We only run these commands when executing through
 # livepush, so they are presented as livepush directives
-#dev-run=apk add --no-cache ip6tables iptables
+# overwrite pre-built node with package version so --inspect can be used
+#dev-run=apk add --no-cache ip6tables iptables nghttp2-dev npm nodejs
```
20k-ultra, line 54: I spoke with Tom and this code is unavoidable because the node version which ships with the Supervisor does not have the inspect components. The issue is that the version installed might differ slightly (patches) then the version being ran outside of livepush. I believe this isn't an issue in reality since test suites do not run with this livepush version so any issues would be found.

20k-ultra (commented):
I'll wait to hear others thoughts but I'm cool with this being merged and looking forward to using it more. Thanks for adding.

Dockerfile
pipex: Is the alpine nodejs package just a different patch version or can it be another version entirely? Even though tests run on the proper environment. If node behavior is different this might reduce usefulness of the inspect option

20k-ultra: well now I want to see the actual size diff when we compile node with system-icu or small-icu. I haven't been successful testing yet....

20k-ultra: if we can just ship some ICU to satisfy the interpreter requirements then we won't need this line

codewithcheese: @pipex there appears to be a single `nodejs` version available for each version of alpine. supervisor is using alpine version 3.11 which has node version 12.22.1-r0. the prebuilt version is 12.16.2.

see https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.11

20k-ultra:
I want to get this in but want to see if we should just ship the required pieces to run the interpreter in the SV base image rather then reinstalling node in livepush. I just haven't had luck building the base image when copying the actual build script which removes the data required for the interpreter https://github.com/product-os/node-arm/blob/d0cfa75c5d8a3200ae4324b04ffc62b4667f856c/alpine/amd64/build.sh#L29

20k-ultra:
We found out that small-icu only adds ~3 MB so now we just have to see if that satisfies the requirements for using `--inspect`. Further discussion on https://www.flowdock.com/app/rulemotion/r-supervisor/threads/wWiPbOHT72Xapptl1m_3UN2zWXd

20k-ultra:
@balena-ci rebase

20k-ultra:
Well, I'm thinking we can just merge this to get the workflow enabled for development environments and then once we confirm that it's valuable we can move towards getting it enabled on production built Supervisor releases as well.

Shall we merge @pipex @codewithcheese ?

codewithcheese:
@20k-ultra I'm back from leave this week (had to extend it last week), I will give it a test with small itc and try work out what we need to ensure its built

20k-ultra:
I'm continuing the PR on this work. I've pushed a few more changes and it's working I just have to figure out what the cost is for this (space mostly).

pipex:
This is unnecessary. Since moving to alpine Node, debugging can be enabled by sending SIGUSR1 to the process

PR #1745, opened 2021-07-08
Fix db-helper module for tests

pipex:
The previous module was using `rewire` to get the knex instance from the
db module but that was leading to issues when running tests using `test:fast`.
This provides a fix for the test module that just removes the destroy
call entirely (it turns out it is not necessary).

Change-type: patch


pipex:
@balena-ci I self certify!

PR #1746, opened 2021-07-09
Files needed to build the Supervisor on balenaCloud

20k-ultra:
docker-compose sets correct services name for supervisor container and the Dockerfile.template allows it to be built on multiple architectures

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,0 +1,132 @@
+ARG ARCH=%%BALENA_ARCH%%
+ARG NODE_VERSION=12.16.2
+
+FROM balenalib/$ARCH-alpine-supervisor-base:3.11 as BUILD
+
+ARG ARCH
+ARG NODE_VERSION
+ARG NODE_ARCHIVE="node-no-intl-v${NODE_VERSION}-linux-alpine-${ARCH}.tar.gz"
+ARG S3_BASE="https://resin-packages.s3.amazonaws.com"
+ARG NODE_LOCATION="${S3_BASE}/node/v${NODE_VERSION}/${NODE_ARCHIVE}"
+
+# DO NOT REMOVE THE cross-build-* COMMANDS
+# The following commands are absolutely needed. When we
+# build for ARM architectures, we run this Dockerfile
+# through sed, which uncomments these lines. There were
+# other options for achieving the same setup, but this seems
+# to be the least intrusive. The commands start commented
+# out because the default build for balenaCI is amd64 (and
+# we can't run any sed preprocessing on it there)
+# RUN ["cross-build-start"]
+
+WORKDIR /usr/src/app
+
+RUN apk add --no-cache \
+	g++ \
+	git \
+	make \
+	python \
+	curl \
+	binutils \
+	libgcc \
+	libstdc++ \
+	libuv \
+	sqlite-libs \
+	sqlite-dev \
+	dbus-dev
+
+COPY build-utils/node-sums.txt .
+
+# Install node from balena's prebuilt cache
+RUN curl -SLO "${NODE_LOCATION}" \
+	&& grep "${NODE_ARCHIVE}" node-sums.txt | sha256sum -c - \
+	&& tar -xzf "${NODE_ARCHIVE}" -C /usr/local --strip-components=1 \
+	&& rm -f "${NODE_ARCHIVE}" \
+	&& strip /usr/local/bin/node
+
+COPY package*.json ./
+
+RUN npm ci --build-from-source --sqlite=/usr/lib
+
+# We only run these commands when executing through
+# livepush, so they are presented as livepush directives
+#dev-run=apk add --no-cache ip6tables iptables
+#dev-copy=entry.sh .
+#dev-cmd-live=LIVEPUSH=1 ./entry.sh
+
+COPY build-utils ./build-utils
+COPY webpack.config.js tsconfig.json tsconfig.release.json ./
+COPY src ./src
+COPY test ./test
+COPY typings ./typings
+
+RUN npm run test-nolint \
+	&& npm run build
+
+# Run the production install here, to avoid the npm dependency on
+# the later stage
+RUN npm ci --production --no-optional --unsafe-perm --build-from-source --sqlite=/usr/lib \
+	&& npm cache clean --force \
+	# For some reason this doesn't get cleared with the other
+	# cache
+	&& rm -rf node_modules/.cache \
+	# Remove various uneeded filetypes in order to reduce space
+	# We also remove the spurious node.dtps, see https://github.com/mapbox/node-sqlite3/issues/861
+	&& find . -path '*/coverage/*' -o -path '*/test/*' -o -path '*/.nyc_output/*' \
+		-o -name '*.tar.*'      -o -name '*.in'     -o -name '*.cc' \
+		-o -name '*.c'          -o -name "*.ts"     -o -name '*.eslintrc' \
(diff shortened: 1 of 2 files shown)


20k-ultra:
Dockerfile needs some more work. Build logs show

```
Step 1/38 : ARG ARCH=raspberrypi4-64
Step 2/38 : ARG NODE_VERSION=12.16.2
Step 3/38 : FROM balenalib/$ARCH-alpine-supervisor-base:3.11 as BUILD
pull access denied for balenalib/raspberrypi4-64-alpine-supervisor-base, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

Dockerfile.template
```diff
@@ -0,0 +1,132 @@
+ARG ARCH=%%BALENA_ARCH%%
```
pipex, line 1: Should we wait to add this dockerfile when we have the actions ready? Otherwise we'll have two dockerfiles for a while

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+  balena-supervisor:
+    build: ./
+    privileged: true
+    tty: true
+    restart: always
+    network_mode: host
+    environment:
+      - DOCKER_ROOT=/mnt/root/var/lib/docker
+      - DOCKER_SOCKET=/var/run/balena-engine.sock
+#       - "BOOT_MOUNTPOINT=$BOOT_MOUNTPOINT"                       <--- Must be set as service variable
```
pipex, line 14: What do you mean by must be set as service variable? A few of these could be set here, e.g. `DELTA_ENDPOINT`, `LISTEN_PORT`

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+#       - "DELTA_ENDPOINT=$DELTA_ENDPOINT"                         <--- Must be set as a service variable
+#       - "LED_FILE=${LED_FILE}"                                   <--- Must be set as a service variable
+#       - "LISTEN_PORT=$LISTEN_PORT"                               <--- Must be set as a service variable
+#       - "NODE_EXTRA_CA_CERTS=${NODE_EXTRA_CA_CERTS}"             <--- Must be set as a service variable
+#       - "SUPERVISOR_IMAGE=${SUPERVISOR_IMAGE}:${SUPERVISOR_TAG}" <--- Must be set as aservice variable
+    volumes:
+      - 'resin-data:/data'
+#       - type: bind                                               <--- Bind mounts are not allowed
```
pipex, line 22: Not sure if it's useful to add the bind mounts as comments

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+#      - type: bind                                                <--- Bind mounts are not allowed
+#         source: /usr/share/ca-certificates
+#         target: /usr/share/ca-certificates
+#         read_only: true
+#       - type: bind                                               <--- Bind mounts are not allowed
+#         source: / 
+#         target: /mnt/root 
+    labels:
+      io.balena.features.balena-api: '1'
+#     io.balena.features.balena-socket: '1'                        <--- TODO: Make Supervisor use DOCKER_SOCKET instead of DOCKER_HOST
```
pipex, line 57: That's a good idea, maybe just uncomment this line. Labels will only have an impact once the supervisor can update itself so they are now just a placeholder. 

You could add dbus here as well

Dockerfile.template
20k-ultra: Don't we want the compose file so we can build the Supervisor on balenaCloud ? I don't understand how we'll need just the compose but not the dockerfile to build a release since we're adding the compose to correctly build the Supervisor (use the right service name).

docker-compose.yml
20k-ultra: I wrote those because from the [startup](https://github.com/balena-os/meta-balena/blob/c681bc3f4759e8f337ace3e5483bfa84d2ed268d/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor/start-balena-supervisor#L60-L85) script it uses some env vars which means the Supervisor's runtime uses some dynamically set variables that we won't know at build time.

...I just thought of an issue regarding this and will get back to you on FD

docker-compose.yml
20k-ultra: I added them since those are needed to run the Supervisor and I've already gone through the work of adding what we need to actually run the Supervisor with this docker-compose file we just have to do whatever is required now to allow these requirements. Without them we'd have to figure out again what we need.

20k-ultra: We could remove them if you wanted to. I thought they were good documentation for the future with no harm.

Dockerfile.template
pipex: We are not building the supervisor on balenaCloud yet, not until we can have the CI do it. We can still deploy in the same way, but now we need to make sure the docker-compose is on the local folder when we do manual deploy, as described [here](https://www.balena.io/docs/reference/balena-cli/#deploy-appname-image)

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
+version: '2.1'
+volumes:
+  resin-data: {}
+services:
+  balena-supervisor:
```
cywang117, line 5: Would this become the name of the container?

20k-ultra: yeah this is the most important part of this whole thing (the reason why we're adding it actually). Without this Supervisor builds create the container in the api with the default "main" name so we have to set it here.

Are you thinking that the current name is `balena_supervisor` not `balena-supervisor`. The sysd service uses a hyphen so I'd like to match them. What do you think @cywang117 @pipex ? do we use undercore or hypen ?

docker-compose.yml
20k-ultra: added dbus and removed comment on this label since it would have to be removed again in the future anyways.

docker-compose.yml
klutchell: Note that if deployed via balena as an app, the container name would actually be `<app_id>_<service_name>` so something like `123456_balena-supervisor`.

If building or running with normal docker-compose the container name would be `<project_name>_<service_name>` where project name is the name of the parent folder by default.

The systemd service is going to need to be updated either way because currently we directly start/stop the container by name.

Dockerfile.template
```diff
@@ -0,0 +1,132 @@
...
+	DEFAULT_MIXPANEL_TOKEN=$DEFAULT_MIXPANEL_TOKEN
+COPY avahi-daemon.conf /etc/avahi/avahi-daemon.conf
+
+VOLUME /data
+HEALTHCHECK --interval=5m --start-period=1m --timeout=30s --retries=3 \
+	CMD wget http://127.0.0.1:${LISTEN_PORT:-48484}/v1/healthy -O - -q
+
+# RUN ["cross-build-end"]
+
+CMD ["/usr/src/app/entry.sh"]
```
klutchell, line 132: @pipex  should we be adding our supervisor label in here?
```Dockerfile
LABEL io.balena.supervisor=1
```

docker-compose.yml
20k-ultra: good points @klutchell. I'm slowly thinking that I don't have enough context for this PR and just did it because @pipex asked me to.. maybe someone else should take over ?

Dockerfile.template
pipex: @klutchell let's split that into a different PR along with the cleanup changes since it a bit more urgent. 

I think I want to hold of on adding a docker-compose to the supervisor since we don't have good automation for this yet.

Issue #1616, opened 2021-03-08
Images from old release not cleaned up


20k-ultra:
A device has 5 releases and all are running fine. After deploying another release with only 2 services now, the other 3 are removed but their images remained on the device. Confirmed in the docker-compose file of the release that the services should be removed which they are but the Images should also be removed...

The Supervisor has some cleanup functionality for images but not sure why it isn't triggering for this device. If you restart the Supervisor you can see applying target state follow by a finished applying target state so the Supervisor thinks there is nothing to clean up. (This is where the cleanup code exists)

https://github.com/balena-io/balena-supervisor/blob/dc96cca5e5fddc277d5a466bcd17654ccce67995/src/compose/application-manager.ts#L210-L212

20k-ultra:
The issue with this instance was that docker images were cleaned up but the database still have entries for these images. We should audit the image cleanup to also cleanup the database entries.

20k-ultra:
Figuring out how to remove the image table from the database file would also solve this however another issue will be created if we go that route.

20k-ultra:
Can we remove entries to images from other applications ? I found that image records from another application were causing the Supervisor to behave weird.

pipex:
Here are some steps to replicate one border case that caused this by Andrea

> Steps to replicate the problem:
> BUILD AND DEPLOY THE SAME IMAGE IN TWO DIFFERENT APPS:
> 1) balena-staging build (in my case it was for rpi4)
> 2) balena-staging deploy support_test balena-console-logs_main (application with one device)
> 3) balena-staging deploy support_test_2 balena-console-logs_main (application without deivces)
> 
> MOVE THE DEVICE
> 1) go to the dashboard and move the device from one application to another.
> 
> OS: balenaOS 2.67.3+rev4
> Supervisor: 12.3.0

cywang117:
Here is a possible way to easily reproduce the issue:

1. `balena push` [node-hello-world](https://github.com/balena-io-examples/balena-node-hello-world)
2. `balena push` [python-hello-world](https://github.com/balena-io-examples/balena-python-hello-world)
3. `balena images -a`, notice that the node image from the first push is still there.

Upon checking the image table in the DB following step 3, the only image listed there is the python service image.

Verifying now & will edit post with results.

20k-ultra:
[...]

Issue #1579, opened 2021-01-26
Incorrect release commit


20k-ultra:
Trying to piece together the required conditions here but I am pretty confident as I've heard it twice now that a user moved their device from 1 application to another and it shows the first application's release id pointing to the new application's release id like so:

![image](https://user-images.githubusercontent.com/3946250/105799580-e4809d80-5f62-11eb-8447-5faffd12ebea.png)

This is not always the case though as seen in https://github.com/balena-io/balena-ui/pull/4513 which has an example of the same release ID pointing to itself. 

The endpoint for this data is coming from the cloud's `/v6/device` but I'm not sure what endpoint on the Supervisor that uses or if the Supervisor is pushing up local state.

20k-ultra:
On a device that had this issue I saw the service which was showing this issue to be running 1 container but it had 3 images for the service downloaded. 2 were in "Downloaded" state and 1 in "Running". In addition the UI showed that the device was not running on the target release. I saw that the services running were all the target release. This was due to the API response containing different commits for is_running__release and should_be_running__release. 

I believe by containing old images for a service we can recreate this release ID -> release ID and show different running release commits.

20k-ultra:
Issue reported in FD: https://www.flowdock.com/app/rulemotion/r-beginners/threads/oMk2P00r1XQ3I4BYl6LORt90VJA

20k-ultra:
This issue is also related or duplicate of https://github.com/balena-io/balena-supervisor/issues/1542

20k-ultra:
After querying the database on device I saw that there were 3 entries in the `images` table for `wifi-connect` service. Once I removed the 2 oldest records the device reported the correct commit running.

cywang117:
Moving the device between applications (staging to prod for instance) while the Supervisor is downloading app updates is a common contributing factor according to the ticket I linked above ^

pipex:
Solved by #1749

PR #1749, opened 2021-07-13
Use tags instead of dockerIds to track supervised images in docker

pipex:
The image manager module now uses tags instead of docker IDs as the main
way to identify docker images on the engine. That is, if the target
state image has a name `imageName:tag@digest`, the supervisor will always use
the given `imageName` and `tag` (which may be empty) to tag the image on
the engine after fetching. This PR also adds checkups to ensure
consistency is maintained between the database and the engine.

Using tags allows to simplify query and removal operations, since now
removing the image now means removing tags matching the image name.

Before this change the supervisor relied only on information in the
supervisor database, and used that to remove images by docker ID. However, the docker
id is not a reliable identifier, since images retain the same id between
releases or between services in the same release.

Relates-to: #1616 #1579
Change-type: patch

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -7,20 +7,20 @@ import StrictEventEmitter from 'strict-event-emitter-types';
 import * as config from '../config';
 import * as db from '../db';
 import * as constants from '../lib/constants';
-import {
-	DeltaFetchOptions,
-	FetchOptions,
-	docker,
-	dockerToolbelt,
-} from '../lib/docker-utils';
+import { DeltaFetchOptions, FetchOptions, docker } from '../lib/docker-utils';
 import * as dockerUtils from '../lib/docker-utils';
-import { DeltaStillProcessingError, NotFoundError } from '../lib/errors';
+import {
+	DeltaStillProcessingError,
+	NotFoundError,
+	StatusError,
+} from '../lib/errors';
 import * as LogTypes from '../lib/log-types';
 import * as validation from '../lib/validation';
 import * as logger from '../logger';
 import { ImageDownloadBackoffError } from './errors';
 
 import type { Service } from './service';
+import { strict as assert } from 'assert';
 
 import log from '../lib/supervisor-console';
 
@@ -43,11 +43,6 @@ export interface Image {
 	status?: 'Downloading' | 'Downloaded' | 'Deleting';
 	downloadProgress?: number | null;
 }
-
-// TODO: Remove the need for this type...
-type NormalisedDockerImage = Docker.ImageInfo & {
-	NormalisedRepoTags: string[];
-};
 
 // Setup an event emitter
 interface ImageEvents {
@@ -140,14 +135,18 @@ export async function triggerFetch(
 
 	let success: boolean;
 	try {
-		const imageName = await normalise(image.name);
+		const imageName = normalise(image.name);
 		image = _.clone(image);
 		image.name = imageName;
 
-		await markAsSupervised(image);
-
+		// Look for a matching image on the engine
 		const img = await inspectByName(image.name);
-		await db.models('image').update({ dockerImageId: img.Id }).where(image);
+
+		// If we are at this point, the image may not have the proper tag so add it
+		await tagImage(img.Id, image.name);
+
+		// Create image on the database if it already exists on the engine
+		await markAsSupervised({ ...image, dockerImageId: img.Id });
 
 		onFinish(true);
 		return;
@@ -171,7 +170,11 @@ export async function triggerFetch(
 				id = await fetchImage(image, opts, onProgress);
 			}
 
-			await db.models('image').update({ dockerImageId: id }).where(image);
+			// Tag the image with the proper reference
+			await tagImage(id, image.name);
+
+			// Create image on the database
+			await markAsSupervised({ ...image, dockerImageId: id });
 
 			logger.logSystemEvent(LogTypes.downloadImageSuccess, { image });
 			success = true;
@@ -219,24 +222,18 @@ export async function removeByDockerId(id: string): Promise<void> {
 	await remove(image);
 }
 
-export async function getNormalisedTags(
-	image: Docker.ImageInfo,
-): Promise<string[]> {
-	return await Bluebird.map(
-		image.RepoTags != null ? image.RepoTags : [],
-		normalise,
-	);
+export function getNormalisedTags(image: Docker.ImageInfo): string[] {
+	return (image.RepoTags || []).map(normalise);
 }
 
 async function withImagesFromDockerAndDB<T>(
-	cb: (dockerImages: NormalisedDockerImage[], composeImages: Image[]) => T,
+	cb: (dockerImages: Docker.ImageInfo[], composeImages: Image[]) => T,
 ) {
 	const [normalisedImages, dbImages] = await Promise.all([
-		Bluebird.map(docker.listImages({ digests: true }), async (image) => {
-			const newImage = _.clone(image) as NormalisedDockerImage;
-			newImage.NormalisedRepoTags = await getNormalisedTags(image);
-			return newImage;
-		}),
+		Bluebird.map(docker.listImages({ digests: true }), (image) => ({
+			...image,
+			RepoTag: getNormalisedTags(image),
+		})),
 		db.models('image').select(),
 	]);
 	return cb(normalisedImages, dbImages);
@@ -252,10 +249,10 @@ function addImageFailure(imageName: string, time = process.hrtime()) {
 
 function matchesTagOrDigest(
 	image: Image,
-	dockerImage: NormalisedDockerImage,
+	dockerImage: Docker.ImageInfo,
 ): boolean {
 	return (
-		_.includes(dockerImage.NormalisedRepoTags, image.name) ||
+		_.includes(dockerImage.RepoTags, dockerUtils.getImageWithTag(image.name)) ||
 		_.some(dockerImage.RepoDigests, (digest) =>
 			hasSameDigest(image.name, digest),
 		)
@@ -264,7 +261,7 @@ function matchesTagOrDigest(
 
 function isAvailableInDocker(
 	image: Image,
-	dockerImages: NormalisedDockerImage[],
+	dockerImages: Docker.ImageInfo[],
 ): boolean {
 	return _.some(
 		dockerImages,
@@ -288,7 +285,7 @@ export function getDownloadingImageIds(): number[] {
 	) as number[];
 }
 
-export async function cleanupDatabase(): Promise<void> {
+export async function cleanImageData(): Promise<void> {
 	const imagesToRemove = await withImagesFromDockerAndDB(
 		async (dockerImages, supervisedImages) => {
 			for (const supervisedImage of supervisedImages) {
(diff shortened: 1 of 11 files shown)


pipex:
@balena-ci retest

test/42-device-api-v2.spec.ts
```diff
@@ -289,12 +289,12 @@ describe('SupervisorAPI [V2 Endpoints]', () => {
 							.body,
 					);
 				});
-			// Deactivate localmode
```
pipex, line 292: Removed this because it was interfering with the splash tests, since the config listener on the firewall was being triggered 🙄

cywang117: In the short term, let's stub `.on` / `.once` listeners in another PR?

src/compose/images.ts
```diff
@@ -417,35 +434,75 @@ async function getImagesForCleanup(): Promise<string[]> {
...
-					.where('name', 'like', `%@${digest}`);
-			} else {
-				imagesFromDb = await db
-					.models('image')
-					.where({ name: imageName })
-					.select();
-			}
+export async function inspectByName(imageName: string) {
+	// Fail fast
+	assert(!!imageName, 'image name to inspect cannot be null');
```
cywang117, line 439: If `inspectByName` is passed a falsey value, this error message will be displayed in the journal logs right? Should we have a more user-friendly message? 

This error is also going to be of type AssertionError, what about changing it to `NotFoundError` which we already have, and which would enable more targeted handling of this class of errors? Or is there some complication associated with this error type change?

src/compose/images.ts
```diff
@@ -417,35 +434,75 @@ async function getImagesForCleanup(): Promise<string[]> {
...
+				filters: { reference: [reference] },
+			})
+			.then(([img]) =>
+				!!img
+					? docker.getImage(img.Id).inspect()
+					: Promise.reject(
+							new StatusError(
+								404,
+								`Failed to find an image matching ${imageName}`,
+							),
```
cywang117, line 471: `NotFoundError` seems to cover the 404 case already

src/compose/images.ts
```diff
@@ -417,35 +434,75 @@ async function getImagesForCleanup(): Promise<string[]> {
...
+			// the first one. If this assumption is false, there is a bug with cleanup
+			.then(([img]) =>
+				!!img
+					? docker.getImage(img.dockerImageId).inspect()
+					: Promise.reject(
+							new StatusError(
+								404,
+								`Failed to find an image matching ${imageName}`,
+							),
+					  ),
```
cywang117, line 498: Same as the other `NotFoundError` comment

src/compose/images.ts
```diff
@@ -514,73 +571,56 @@ async function removeImageIfNotNeeded(image: Image): Promise<void> {
 
 	const img = images[0];
 	try {
-		if (img.dockerImageId == null) {
```
cywang117, line 517: For `removeImageIfNotNeeded` changes, I'm really liking the reduction in if/else statements 👍🏼

test/42-device-api-v2.spec.ts
```diff
@@ -289,12 +289,12 @@ describe('SupervisorAPI [V2 Endpoints]', () => {
...
-			await config.set({ localMode: false });
 		});
 	});
 
-	describe('POST /v2/applications/:appId/start-service', function () {
+	// TODO: setup for this test is wrong, which leads to inconsistent data being passed to
+	// manager methods. A refactor is needed
```
cywang117, line 296: What do you mean? Is this an issue with some of the mocked manager methods returning data in an out-of-date format?

test/lib/mockerode.ts
```diff
@@ -778,6 +807,7 @@ export class MockEngine {
 	// an image is removed by tag.
 	removeImage(name: string, { force } = { force: false }) {
 		const image = this.findImage(name);
+		console.log('REMOVE', name);
```
cywang117, line 810: Remove this console.log?

test/42-device-api-v2.spec.ts
cywang117: We could migrate mock service/image setup to use mockerode here

src/compose/images.ts
pipex: `NotFoundError` is a checker function, not a subclass of error

src/compose/images.ts
pipex: If this function is passed a falsey value we should catch it before it getting to production because it's most likely a bug, this assertion is to catch those cases. For instance, the API tests I disabled were not failing before even though they were passing a null to this function. I'm ok with changing to a better message.

test/42-device-api-v2.spec.ts
pipex: Yeah, I think we should try that on another PR and fix whatever breaks.

test/lib/mockerode.ts
pipex: 😳

test/42-device-api-v2.spec.ts
pipex: There are multiple issues. For one, the mock managers are not setting all the required data, so the assertion in the `inspectByName` was getting triggered (it wasn't getting caught before that). I tried to add it but I kept running with missing pieces for the tests. I think when we work on API parity we should make sure to rewrite some of the API methods to make them more testable, right now we have to stub/mock everything and that is a bad sign.

test/src/compose/application-manager.spec.ts
```diff
@@ -900,19 +931,80 @@ describe('compose/application-manager', () => {
...
+			currentApps,
+			targetApps,
+			{
+				downloading,
+				availableImages,
+				containerIdsByAppId,
+			},
+		);
+
+		// A start step shoud happen for the depended service first
```
20k-ultra, line 991:
```suggestion
		// A start step should happen for the depended service first
```

20k-ultra (commented):
absolutely amazing work Felipe.

src/compose/images.ts
```diff
@@ -272,7 +286,7 @@ export function getDownloadingImageIds(): number[] {
 	) as number[];
 }
 
-export async function cleanupDatabase(): Promise<void> {
+export async function normaliseImageData(): Promise<void> {
```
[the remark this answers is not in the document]
pipex: It doesn't just change the database anymore, it also tags images. That's the reason for the change

src/compose/images.ts
20k-ultra: yeah, we want it to throw if `imageName` is null though right ?

src/compose/images.ts
```diff
@@ -398,35 +434,75 @@ async function getImagesForCleanup(): Promise<string[]> {
...
+					: Promise.reject(
+							new StatusError(
+								404,
+								`Failed to find an image matching ${imageName}`,
+							),
+					  ),
+			);
+
+	// Convert the promise array into a promise chain
+	return await [inspectByURI, inspectByReference, inspectByDigest].reduce(
```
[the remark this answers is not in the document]
pipex: Didn't see the point in declaring them outside really, since they are not being used elsewhere. Doing it this way means we only need to perform the `assert` once. 

If we ever need to use them individually we can always refactor, if you feel strongly about it I can make the change but I don't think is that much of an improvement.

src/compose/images.ts
```diff
@@ -398,35 +434,75 @@ async function getImagesForCleanup(): Promise<string[]> {
...
+		(res, p) => res.catch(() => p()),
+		Promise.reject('You should never see this'),
```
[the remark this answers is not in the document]
pipex: The message is a placeholder really, unless someone breaks the implementation it will never make it to the user. I think the message should reflect that idea, although I don't necessarily think my current message reflects that well

src/compose/images.ts
pipex: I mean we won't get an assertion error, we will get something like 

```
Uncaught TypeError: Cannot read property 'length' of null
```

src/compose/images.ts
```diff
@@ -330,6 +359,10 @@ export async function update(image: Image): Promise<void> {
 
 export const save = async (image: Image): Promise<void> => {
 	const img = await inspectByName(image.name);
+
+	const { repo, tag } = dockerUtils.getRepoAndTag(image.name);
```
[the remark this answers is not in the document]
pipex: 👍

src/compose/images.ts
20k-ultra: oh yeah true. Well then I have no input besides considering the error message thrown that I suggested.

src/compose/images.ts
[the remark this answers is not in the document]
20k-ultra: I see. I just can't see normalise being used as a way to describe "delete". I think `cleanImageData` is ok ? Clean could imply modifying (normalising) data and removing.

src/compose/images.ts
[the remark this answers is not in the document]
pipex: I think is the same as with the assertion on top of the method, if it does fail we want it to fail in a way that ensures we catch it before it can get to production. It really won't fail since we are using a static array as source for the `reduce`

Since it's causing confusing let me see if I can find a way to perform the `reduce` without passing an initial value. I couldn't think of a clean way to do it but I can give it another try

src/compose/images.ts
cywang117: I suppose we're removing this line after all the E2E tests pass?

src/compose/images.ts
cywang117: Oops! My mistake

test/42-device-api-v2.spec.ts
cywang117: Sounds good, thanks for clarifying.

src/compose/images.ts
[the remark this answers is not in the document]
20k-ultra: I think declaring them outside the function would improve the readability of that function with no cost

src/compose/images.ts
20k-ultra: I'm wasn't sure what Felipe meant by "catch it before it getting to production". Reading Christina's comment now is confirming I'm definitely out of the loop. @pipex why is the assert added ? I thought it was to prevent any issues at runtime that the TS compiler isn't catching.

pipex: @cywang117 no we won't remove them, this serves the same purpose as the `InternalConsistencyError` in other places of the code, which is to show bugs instead of hiding them. As @20k-ultra says, this is to catch issues at runtime that the compiler isn't catching.

See this also http://wiki.c2.com/?ShipWithAssertionsOn

src/lib/docker-utils.ts
```diff
@@ -48,14 +47,51 @@ export const dockerProgress = new DockerProgress({
...
+	}
+
+	const [, registry, imageName, tag, digest] = match;
+	const tagName = !digest && !tag ? 'latest' : tag;
+	const digestMatch = digest?.match(
+		/^[A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*:[0-9a-f-A-F]{32,}$/,
+	);
+	if (!imageName || (digest && !digestMatch)) {
+		throw new Error(
+			'Invalid image name, expected [domain.tld/]repo/image[:tag][@digest] format',
```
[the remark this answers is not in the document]
pipex: I think `domain.tld` is more technically accurate, agreed on adding the value though

src/compose/images.ts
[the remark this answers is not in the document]
pipex: Opted to update the reject message with something that makes it totally obvious that is a bug

src/compose/images.ts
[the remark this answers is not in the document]
pipex: That’s what normalizing means to me, cleaning and fixing missing data. But I might be wrong, cleanImageData works too

test/src/compose/application-manager.spec.ts
20k-ultra: @pipex you marked as resolved but the comment still has the type

pipex: No it doesn't. Not sure why it shows like this, but on the final file it has the correct spelling

cywang117:
Tested after bea82c2 and database entries are deleted as intended. I can't think of any other test cases so this LGTM. Self-certify or wait on @20k-ultra's OK 👍🏼

20k-ultra (approved):

PR #1753, opened 2021-07-21
Remove comparisons based on image, release, and service ids

pipex:
Preparing for the new v3 target state, where the supervisor will make environment
dependent ids optional and rely on using general UUIDs and user known identifiers
for comparison. This PR moves forward in that direction by removing some of those
comparisons for v2 target state.

- `imageId` to be replaced with `imageName`
- `serviceId` to be replace by `serviceName`
- `releaseId` to be replaced by `commit` (future `releaseUuid`)

This is a backwards compatible change, meaning it doesn't completely get rid of
these identifiers (which are still being used by supervisor API and for state
patch), but will not depend on those identifiers for calculating steps to target state.

One important change, although one that should not affect running applications is that this PR renames
service containers from `<serviceName>_<serviceId>_<releaseId>` to `<serviceName>_<serviceId>_<releaseId>_<commit>`

Once we can phase supervisor API v1 and v2 endpoints and replace them by a feature complete v3 API, then we'll be able to rename the containers to `<serviceName>_<commit>`

Change-type: minor

### DIFF
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -125,7 +125,6 @@ let targetVolatilePerImageId: {
 export const initialized = (async () => {
 	await config.initialized;
 
-	await imageManager.initialized;
 	await imageManager.cleanImageData();
 	const cleanup = async () => {
 		const containers = await docker.listContainers({ all: true });
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -179,7 +179,7 @@ export async function getRequiredSteps(
 ): Promise<CompositionStep[]> {
 	// get some required data
 	const [downloading, availableImages, currentApps] = await Promise.all([
-		imageManager.getDownloadingImageIds(),
+		imageManager.getDownloadingImageNames(),
 		imageManager.getAvailable(),
 		getCurrentApps(),
 	]);
@@ -199,7 +199,7 @@ export async function inferNextSteps(
 	targetApps: InstancedAppState,
 	{
 		ignoreImages = false,
-		downloading = [] as number[],
+		downloading = [] as string[],
 		availableImages = [] as Image[],
 		containerIdsByAppId = {} as { [appId: number]: Dictionary<string> },
 	} = {},
@@ -674,7 +674,7 @@ function saveAndRemoveImages(
 			(svc) =>
 				_.find(availableImages, {
 					dockerImageId: svc.config.image,
-					imageId: svc.imageId,
+					name: svc.imageName,
 				}) ?? _.find(availableImages, { dockerImageId: svc.config.image }),
 		),
 	) as imageManager.Image[];
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -121,30 +121,30 @@ function createTask(initialContext: Image) {
 	return running(initialContext);
 }
 
-const runningTasks: { [imageId: number]: ImageTask } = {};
+const runningTasks: { [imageName: string]: ImageTask } = {};
 function reportEvent(event: 'start' | 'update' | 'finish', state: Image) {
-	const { imageId } = state;
+	const { name: imageName } = state;
 
 	// Emit by default if a start event is reported
 	let emitChange = event === 'start';
 
 	// Get the current task and update it in memory
 	const currentTask =
-		event === 'start' ? createTask(state) : runningTasks[imageId];
-	runningTasks[imageId] = currentTask;
+		event === 'start' ? createTask(state) : runningTasks[imageName];
+	runningTasks[imageName] = currentTask;
 
 	// TODO: should we assert that the current task exists at this point?
 	// On update, update the corresponding task with the new state if it exists
 	if (event === 'update' && currentTask) {
 		const [updatedTask, changed] = currentTask.update(state);
-		runningTasks[imageId] = updatedTask;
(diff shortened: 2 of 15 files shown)


cywang117:
@pipex Should we start taking a look or are you going to make more changes?

pipex:
@cywang117 not yet, I'm making a couple of more changes to tests that were waiting for #1749 to avoid conflicts. I think this should be ready for review today though

src/compose/images.ts
```diff
@@ -68,8 +67,92 @@ const imageFetchLastFailureTime: Dictionary<ReturnType<
...
+		emitChange = changed;
+	}
+
+	// On update, update the corresponding task with the new state if it exists
+	if (event === 'finish' && currentTask) {
+		[, emitChange] = currentTask.finish();
+		delete runningTasks[imageId];
+	}
+
+	if (emitChange) {
```
cywang117, line 149: if (emitChange && currentTask)

Should we allow change events if there happens to be an undefined currentTask?

src/compose/images.ts
```diff
@@ -68,8 +67,92 @@ const imageFetchLastFailureTime: Dictionary<ReturnType<
...
+		events.emit('change');
+	}
+}
+
 // A store of volatile state for images (e.g. download progress), indexed by imageId
-const volatileState: { [imageId: number]: Image } = {};
+// const volatileState: { [imageId: number]: Image } = {};
```
cywang117: Let's just delete this?

src/compose/images.ts
pipex: Good point, probably not

src/compose/images.ts
pipex: I think I deleted it on another commit

src/compose/service-manager.ts
```diff
@@ -611,7 +611,7 @@ async function prepareForHandover(service: Service) {
 	const container = docker.getContainer(svc.containerId);
 	await container.update({ RestartPolicy: {} });
 	return await container.rename({
-		name: `old_${service.serviceName}_${service.imageId}_${service.imageId}_${service.releaseId}`,
+		name: `old_${service.serviceName}_${service.imageId}_${service.commit}`,
```
cywang117, line 614: I see that service.releaseId is omitted in the container name here, but not on line 154. Is this intended?

src/compose/service.ts
```diff
@@ -600,7 +600,7 @@ export class Service {
 				'Attempt to build Service class from container with malformed labels',
 			);
 		}
-		const nameMatch = container.Name.match(/.*_(\d+)_(\d+)$/);
+		const nameMatch = container.Name.match(/.*_(\d+)_(\d+)(?:_(.*?))?$/);
 		if (nameMatch == null) {
 			throw new InternalInconsistencyError(
 				'Attempt to build Service class from container with malformed name',
```
cywang117, line 606: We can start adapting the Go-style error reporting pattern by adding expected & actual container name to this error message

src/compose/service-manager.ts
pipex: Nice catch, that's an error

src/compose/service.ts
pipex: That's a good idea, although I'd only expect to see this error if you try running an older supervisor with the new container names

src/migrations/M00007.js
```diff
@@ -0,0 +1,41 @@
+export async function up(knex) {
```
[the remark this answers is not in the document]
pipex: I agree, but that will get easier once tests are using the memory db, I have tried it and right now is way too hard.

src/device-api/common.js
```diff
@@ -121,13 +121,6 @@ export async function doPurge(appId, force) {
 		});
 }
 
-export function serviceAction(action, serviceId, current, target, options) {
```
20k-ultra, line 124: death to this file

20k-ultra (changes requested):
if we switch to service name wouldn't there be issues with multi-app when someone runs applications with the same service name ?

src/compose/application-manager.ts
```diff
@@ -125,7 +125,6 @@ let targetVolatilePerImageId: {
 export const initialized = (async () => {
 	await config.initialized;
 
-	await imageManager.initialized;
```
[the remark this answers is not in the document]
pipex: There only thing imageManager.initialized did was setup the listener. I removed that and the function

pipex:
@20k-ultra Is the same service name within an app. Although that is a good thing to double check 
> if we switch to service name wouldn't there be issues with multi-app when someone runs applications with the same service name ?

src/compose/service-manager.ts
```diff
@@ -88,7 +88,7 @@ export async function get(service: Service) {
 	// Get the container ids for special network handling
 	const containerIds = await getContainerIdMap(service.appId!);
 	const services = (
-		await getAll(`service-id=${service.serviceId}`)
+		await getAll(`service-name=${service.serviceName}`)
```
pipex, line 91: This is the only one that could be suspect, but it also checks matching appId @20k-ultra so I thin we are I’m the clear

20k-ultra (approved):
my concerns about switching to service isn't an issue because I wasn't aware it's done in the context of application id as well.

PR #1760, opened 2021-08-02
Fix regression with local mode push

pipex:
PR #1749 introduced a bug when pushing local target state. An update to
the [image name normalization](https://github.com/balena-os/balena-supervisor/blob/f1bd4b8d9bcef29e326cbf97eaddd837c2704d19/src/lib/docker-utils.ts#L81)
failed to consider the local image name format. This results in mangling
of image names in the database, i.e. the image `ubuntu:latest` is stored
as `/ubuntu:latest`. This causes an exception to be returned by the
dockerode `getImage('/ubuntu:latest').inspect()` call.

This sends the supervisor into a crash loop and is shown on the supervisor
journal logs as

```
Error: getaddrinfo ENOTFOUND images
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:64:26)
```

Unfortunately if this happens on a user device, since the mangled image
name is already on the database, the easiest way to fix is to remove the
supervisor database and let the supervisor recreate it. Deleting the
database should be side effect free.

Change-type: patch

### DIFF
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -674,7 +674,9 @@ function saveAndRemoveImages(
 			(svc) =>
 				_.find(availableImages, {
 					dockerImageId: svc.config.image,
-					name: svc.imageName,
+					// There is no 1-1 mapping between services and images
+					// on disk, so the only way to compare is by imageId
+					imageId: svc.imageId,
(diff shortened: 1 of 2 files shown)


cywang117:
I assume you've tested that this works, so this looks good to me. We should deploy this to production after merge 👍🏼

pipex:
@balena-ci I self-certify!

Issue #1754, opened 2021-07-22
Failed container SIGKILL with `(HTTP code 409) unexpected - You cannot remove a running container`


cywang117:
Currently, when executing a service kill step, the Supervisor throws any daemon HTTP codes that aren't 304 Not Modified (container already stopped) or 404 Not Found (container removed). There is a condition where the engine will fail on both the SIGTERM (`docker stop`) and the SIGKILL 10 seconds later (`docker kill`). 

With this in mind, the Supervisor should force-terminate the old container so that the device isn't stuck in an invalid state of the Supervisor attempting to kill the container. Either that, or https://github.com/balena-os/balena-supervisor/issues/1743 would also avoid the invalid state by ensuring that the Supervisor wouldn't continue attempting to kill a container that doesn't want to be killed after a number of attempts and revert the state.

cywang117:
@20k-ultra @pipex Do you guys think forcing a container kill with:

1. Get the "ctr-pid" using docker inspect -f "{{.State.Pid}}" <container-name-or-id>
2. Then force-kill the "ctr-pid" process, using kill -9 <ctr-pid>; this should stop the container

Source: https://github.com/moby/moby/issues/32938#issuecomment-642205013

pipex:
I think the engine already does that @cywang117, when telling a container to stop, it sends a SIGTERM, waits 10 seconds and then sends a SIGKILL. If that is not working I'm not sure the supervisor would succeed. If we see a device in that state, it might be worth trying those steps manually to see if they have an effect.

20k-ultra:
I think the obvious issue is that the Supervisor is trying to rm a container which is still running so make the Supervisor handle containers which are refusing to exit better before trying to remove it.

cywang117:
^ I saw this again on a user's device from that ticket. Reporting that `kill -9 $(balena inspect CONTAINER_NAME -f "{{.State.Pid}}")` did *not* work, however, `balena rm --force CONTAINER_NAME` did. Maybe when we catch 2 classes of error codes [here](https://github.com/balena-io/balena-supervisor/blob/4a2ac557ef11d334154bafa0cab4e7272cbed749/src/compose/service-manager.ts#L547-L560), we should also catch a 409 and execute a force removal -- what do you guys think @pipex @20k-ultra?

pipex:
[...]

PR #1761, opened 2021-08-02
Force remove container on Docker daemon 409

cywang117:
If removeContainer is true, force the removal. However,
this doesn't cover a 409 case when removeContainer is false.

Change-type: patch
Relates-to: #1754

### DIFF
--- a/src/compose/service-manager.ts
+++ b/src/compose/service-manager.ts
@@ -555,6 +555,11 @@ function killContainer(
 					logger.logSystemEvent(LogTypes.stopRemoveServiceNoop, {
 						service,
 					});
+				} else if (statusCode === 409 && removeContainer) {
+					// 409 means the container is refusing to be stopped, however this will lead to a service
+					// stop error loop. If the container needs to be removed anyway, we force-remove the container.
+					logger.logSystemEvent(LogTypes.stopRemoveServiceNoop, { service });
+					return containerObj.remove({ v: true, force: true });
 				} else {
 					throw e;
 				}


pipex (changes requested):
I commented on the issue as well, but I think we should try to find the underlying cause for the problem instead of adding an extra catch clause here 

Looking at this, the catch clause should be invoked either if the stop() call fails or the remove call fails. Given the error message, it looks like the stop call is returning misleading information, which I think is the root of the problem and what we should figure out first.

* Test the result of 
```
 curl -v  --unix-socket /var/run/balena-engine.sock -X POST http://localhost/containers/:containerId/stop
```
and
```
 curl -v  --unix-socket /var/run/balena-engine.sock -X POST http://localhost/containers/:containerId/kill
```
If they return 2xx, then I'd expect this to be an engine issue
* Test the same operation with the engine, stop the container and see the result `balena stop <container> && echo "OK"`. If we see "OK" and the container didn't really stop, then is a engine CLI issue
* Within a docker image with node and dockerode, try to stop the engine manually `docker.getContainer(<containerId>).stop().then(() => console.log("OK"))` if this returns OK, then the issue is with dockerode.
* Otherwise the issue might be with this code on the supervisor

cywang117:
As I haven't seen this issue on support with the frequency that would indicate a change to Supervisor is needed, I'll close this PR for now.

cywang117:
Appreciate the thought you've put into testing steps though, @pipex. I'll be watching out for this kind of error in user devices to try to get a consistent reproduction, then I'll test the items you've listed here.

PR #1762, opened 2021-08-02
Removed fire emoji from firewall logs.

### LINKED ISSUE #1751: Remove fire emoji 🔥 from supervisor firewall logs
Our logs generally don't contain emojis, with one exception: The firewall logs from the supervisor (independently of the severity level).

I think we should remove that emoji, because it distracts readers of the logs, unless they are explicitly searching for FW issues

peakyDicers:
Closes #1751 

I removed the fire emoji prefix from the firewall logs.

### DIFF
--- a/src/lib/firewall.ts
+++ b/src/lib/firewall.ts
@@ -21,7 +21,6 @@ export const initialised = (async () => {
 })();
 
 const BALENA_FIREWALL_CHAIN = 'BALENA-FIREWALL';
-const LOG_PREFIX = '🔥';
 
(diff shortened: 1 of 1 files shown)


20k-ultra:
Hey Ian, thanks for the PR. Checkout the contribution docs which mentions that your commit message should contain a `Change-type: <type>` footer. The type following semver.

Alternatively, you can just change your commit message to b prefixed with "patch:"

20k-ultra (approved):
congrats on your first PR to balena

PR #1765, opened 2021-08-03
Update supervisor to use new v3 target state format

pipex:
The API and the supervisor are moving towards a more multi-app compatible format for the target state (and away from database dependent ids). This PR introduces support for the new target state for the supervisor (see below for the new target state format).

This is large change as the supervisor needs add support for the new state format on the GET request, as well as on `apps.json`, modifying everywhere in the code that used to rely on the old format (e.g. validations, contracts, migrations,etc. ) In addition to that it adds some conversions from the old target state format into the new target state format for backwards compatibility on the supervisor API. The change also adds support for the new target state format on target state PATCH. Finally it performs some refactoring on most of the places where `appIds` where used.

This is a MAJOR change, as installing this new supervisor version, will restart all the currently running services as metadata inside the services needs to change (the restart still waits for users to lift the lock though).

```typescript
export type StateV3 = {
	[uuid: string]: {
		name: string;
		parent_device?: string;
		config?: {
			[varName: string]: string;
		};
		apps: {
			[uuid: string]: {
				/**
				 * @deprecated to be removed in state v4
				 */
				id: number;
				name: string;
				class: 'fleet' | 'block' | 'app';
				parent_app?: string;
				is_host?: boolean;
				releases?: {
					[uuid: string]: {
						/**
						 * @deprecated to be removed in state v4
						 */
						id: number;
						services?: {
							[name: string]: {
								/**
								 * @deprecated to be removed in state v4
								 */
								id: number;
								/**
								 * @deprecated to be removed in state v4
								 */
								image_id: number;
								image: string;
								/**
								 * Defaults to true if undefined
								 */
								running?: boolean;
								environment: {
									[varName: string]: string;
								};
								labels: {
									[labelName: string]: string;
								};
								contract?: AnyObject;
								composition?: ServiceComposition;
							};
						};
						volumes?: AnyObject;
						networks?: AnyObject;
					};
				};
			};
		};
	};
};
```


src/migrations/M00008.js
```diff
@@ -0,0 +1,54 @@
...
+					uuid: app.appId.toString(),
+					services: JSON.stringify(services),
+				});
+		}),
+	);
+}
+
+export function down() {
+	return Promise.reject(new Error('Not Implemented'));
+}
```
Page-, line 54:
```suggestion
export async function down() {
	throw new Error('Not Implemented');
}
```

Page- (commented):
I haven't reviewed everything but this is what I've done so far

src/compose/app.ts
```diff
@@ -782,11 +771,29 @@ export class App {
...
+			!svc.labels ||
+			!svc.labels['io.balena.image.store'] ||
+			svc.labels['io.balena.image.store'] === 'data';
+
 		// In the db, the services are an array, but here we switch them to an
 		// object so that they are consistent
 		const services: Service[] = await Promise.all(
-			(JSON.parse(app.services) ?? []).map(
-				async (svc: ServiceComposeConfig) => {
+			(JSON.parse(app.services) ?? [])
```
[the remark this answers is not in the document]
pipex: I think that was there before, but that's a good point

src/device-state/db-format.ts
```diff
@@ -37,73 +38,109 @@ export async function getApps(): Promise<InstancedAppState> {
...
-					serviceId: checkInt(sId),
-					commit: app.commit,
-					image: await images.normalise(s.image),
-				})),
-			);
+	const dbApps = Object.keys(apps).map((uuid) => {
+		const { id: appId, ...app } = apps[uuid];
+
+		// Get the first uuid
+		const [commit] = Object.keys(app.releases);
```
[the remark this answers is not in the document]
pipex: You mean rename `commit` to `releaseUuid` wherever possible? I can do that

src/compose/app.ts
```diff
@@ -65,18 +69,21 @@ export class App {
...
 		this.volumes = opts.volumes;
 		this.networks = opts.networks;
+		this.isHost = !!opts.isHost;
 
 		if (this.networks.default == null && isTargetState) {
 			// We always want a default network
 			this.networks.default = Network.fromComposeObject(
 				'default',
 				opts.appId,
+				opts.appUuid!, // app uuid always exists on the target state
```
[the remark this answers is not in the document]
pipex: @Page- I could not get this to work. I thought maybe it was something that would work in TS 4.4, but no luck there either :(

I still get `Type undefined is not assignable to string`

src/device-api/v2.ts
```diff
@@ -557,19 +561,19 @@ export function createV2Api(router: Router) {
...
+				const [release] = Object.values(app.releases);
+				// Return a list of the volume names
+				return Object.keys(release?.volumes ?? {}).map((volumeName) =>
+					Volume.generateDockerName(app.id, volumeName),
 				);
-			});
-		});
+			})
+			// Flaten the list
+			.reduce((list, v) => list.concat(v), []);
```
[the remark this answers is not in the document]
pipex: `flat()` and `flatMap()` are only available with `lib` set to `es2019` in `tsconfig.json`. Should we make the change?

src/compose/app.ts
```diff
@@ -145,17 +152,13 @@ export class App {
 		// Generate volume steps
 		steps = steps.concat(
 			this.generateStepsForComponent(volumeChanges, servicePairs, (v, svc) =>
-				// TODO: Volumes are stored without the appId prepended, but networks are stored
-				// with it prepended. Sort out this inequality
-				svc.config.volumes.includes(v.name),
+				svc.hasVolume(v.name),
```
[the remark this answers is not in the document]
pipex: Done https://github.com/balena-os/balena-supervisor/pull/1802

src/device-state.ts
```diff
@@ -42,52 +44,17 @@ import * as apiKeys from './lib/api-keys';
...
-	// to be accessed
-	if (!_.isObject((state as any).local)) {
-		throw new Error('Local state must be an object');
-	}
-	validateLocalState((state as any).local);
-	if ((state as any).dependent != null) {
-		return validateDependentState((state as any).dependent);
-	}
+	const errors = [
+		'Invalid target state. Please ensure you are using the most up-to-date version of Balena CLI.',
```
cywang117, line 56: We may need to make another PR when we know the specific balenaCLI version that moves to v3 target state

pipex: I might remove this, the PR also converts between v2 and v3 target state compatibility in order to avoid breaking the supervisor API so it really it should not depend on the CLI version

src/types/basic.ts
```diff
@@ -1,7 +1,182 @@
...
+ */
+export const StringIdentifier = new t.Type<string, string>(
+	'StringIdentifier',
+	(i: unknown): i is string =>
+		t.string.is(i) && !isNaN(+i) && +i === parseInt(i, 10) && +i >= 0,
+	(i, c) =>
+		pipe(
+			t.string.validate(i, c),
+			chain((s) =>
+				!isNaN(+s) && +s === parseInt(s, 10) && +s >= 0
```
cywang117, line 48: Why is `!isNaN(+1) && +i === parseInt(i, 10) && +i >= 0` (or `s` instead of `i` validated twice?

pipex: That's just the way [io-ts custom types work](https://github.com/gcanti/io-ts/blob/master/index.md#custom-types) one is the typescript guard and the second one is the actual validator

src/compose/service.ts
```diff
@@ -122,27 +126,31 @@ export class Service {
...
-		delete appConfig.commit;
+		service.appUuid = appConfig.appUuid;
+
+		// dependsOn is used by other parts of the step
+		// calculation so we delete it from the composition
+		service.dependsOn = appConfig.composition?.dependsOn || null;
+		delete appConfig.composition?.dependsOn;
 
-		delete appConfig.contract;
+		// Get remaining fields from appConig
```
cywang117, line 143: `appConig` --> `appConfig`

src/device-state.ts
pipex: This message will appear if target state from the cloud is wrong for some reason. Should we just replace it with `Invalid target state`?

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

20k-ultra (approved):
huge milestone congrats

PR #1766, opened 2021-08-06
patch: Update fleet terminology for application update locks

vipulgupta2048:
Just changed this over on Docs before realizing the actual location for this file is here.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -39,7 +39,7 @@ Here's the full list of endpoints implemented so far. In all examples, replace e
 Responds with a simple "OK", signaling that the supervisor is alive and well.
 
 #### Examples:
-From a service:
+From an app container:
 ```bash
 $ curl -X GET --header "Content-Type:application/json" \
 	"$BALENA_SUPERVISOR_ADDRESS/ping"
@@ -67,7 +67,7 @@ Starts a blink pattern on a LED for 15 seconds, if your device has one.
 Responds with an empty 200 response. It implements the "identify device" feature from the dashboard.
 
 #### Examples:
-From a service:
+From an app container:
 ```bash
 $ curl -X POST --header "Content-Type:application/json" \
 	"$BALENA_SUPERVISOR_ADDRESS/v1/blink?apikey=$BALENA_SUPERVISOR_API_KEY"
@@ -102,7 +102,7 @@ Can be a JSON object with a `force` property. If this property is true, the upda
 ```
 
 #### Examples:
-From a service:
+From an app container:
 ```bash
 $ curl -X POST --header "Content-Type:application/json" \
 	--data '{"force": true}' \
@@ -140,7 +140,7 @@ When successful, responds with 202 accepted and a JSON object:
 Can contain a `force` property, which if set to `true` will cause the update lock to be overridden.
 
 #### Examples:
-From a service:
+From an app container:
 ```bash
 $ curl -X POST --header "Content-Type:application/json" \
 	"$BALENA_SUPERVISOR_ADDRESS/v1/reboot?apikey=$BALENA_SUPERVISOR_API_KEY"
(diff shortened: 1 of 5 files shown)


pipex:
@vipulgupta2048 are they fleet update blocks or app update blocks?

vipulgupta2048:
@pipex Looking at this PR again, I think we can do away with the mention of application in the heading. And, yes you are correct these are infact app update locks. My bad.

docs/update-locking.md
```diff
@@ -1,17 +1,17 @@
...
 
 For multicontainer applications, a release will only be updated if all of the services can be updated. While locks are per-service, having the update lock in a single service will prevent all services from updating to a new release.
 
-The presence of a lockfile will ensure that your application does not get killed, but updates will still be downloaded by the supervisor, ready to be applied once the lockfile no longer exists.
+The presence of a lockfile will ensure that your services does not get killed, but updates will still be downloaded by the supervisor, ready to be applied once the lockfile no longer exists.
```
pipex, line 14:
```suggestion
The presence of a lockfile will ensure that your services do not get killed, but updates will still be downloaded by the supervisor, ready to be applied once the lockfile no longer exists.
```

docs/API.md
```diff
@@ -39,7 +39,7 @@ Here's the full list of endpoints implemented so far. In all examples, replace e
 Responds with a simple "OK", signaling that the supervisor is alive and well.
 
 #### Examples:
-From an application container:
+From a service:
```
pipex, line 42: I'm divided if this should be just `service` or using something `service container` is better. In my understanding, the service is the abstraction, but the container is the code actually running there. I'm not sure if that is the common view though

docs/API.md
```diff
@@ -210,7 +210,7 @@ $ curl -X POST --header "Content-Type:application/json" \
 
 > **Note:** This route will remove and recreate all service containers, as volumes can only be removed when their associated containers are removed. On devices with supervisor version lower than 7.22.0, replace all `BALENA_` variables with `RESIN_`, e.g. `RESIN_SUPERVISOR_ADDRESS` instead of `BALENA_SUPERVISOR_ADDRESS`.
 
-Clears the user application's `/data` folder.
+Clears the device's `/data` folder and named volumes.
```
pipex, line 213: I'm not sure about this change, it might require some archeology, but I think for single container, the supervisor used a `/data/` folder for the app instead of a volume.

pipex:
Looks good in general @vipulgupta2048, only a couple of observations

pipex:
@vipulgupta2048 Are you still working on this? I took a look at my newer comments and I think they are pretty minor, I think we should probably just merge.

vipulgupta2048:
@balena-ci rebase

docs/API.md
chrisys: `From an app container` works for me

docs/update-locking.md
```diff
@@ -1,17 +1,17 @@
 ---
-title: Application update locks
-excerpt: Locking application updates on your balenaOS devices
+title: Update locks
+excerpt: Locking the updates to your application that is running on balenaOS devices.
```
chrisys, line 3:
```suggestion
excerpt: Locking updates to the release that is running on your balenaOS devices.
```

docs/update-locking.md
```diff
@@ -1,17 +1,17 @@
...
 ---
 
-# Application update locks
+# Update locks
 
-Locking updates means that the balena supervisor will not be able to kill your application. This is meant to be used at critical sections of your code where you don't want to be interrupted, or to ensure that updates are only installed at certain times.
+Locking updates means that the balena supervisor will not be able to kill your application running on the device. This is meant to be used at critical sections of your code where you don't want to be interrupted, or to ensure that updates are only installed at certain times.
```
chrisys, line 8:
```suggestion
Locking updates means that the balena supervisor will not be able to kill the services running on the device. This is meant to be used at critical sections of your code where you don't want to be interrupted, or to ensure that updates are only installed at certain times.
```

docs/update-locking.md
chrisys: This still seems to be incorrect

vipulgupta2048: Resolved

docs/API.md
vipulgupta2048: @pipex I changed this to something else. how about now?

vipulgupta2048:
@balena-ci I self-certify!

PR #1774, opened 2021-08-19
Created new dind tool, works with latest supervisor.

peakyDicers:
Change-type: minor

I've re-written the dind tool to fix a number of issues and in hopes to remove unmaintained/broken and "no-longer-useful" code.

**What's wrong with dindctl?**
1) The big issue with dindctl is that trying to run versions of the supervisor that uses `device-type.json` (so any of the newer versions) will not work because dindctl uses an outdated `balenaos-in-container` submodule which doesn't cp the `device-type.json` into the boot directory for the supervisor container. The latest version of `balenaos-in-container` does. 
2) Dindctl contains code for its own version of 'live-push' through the use of dindctl refresh. I believe it would be better if the project just keeps to a single live-push feature (the node one; which seems much more maintained). Both dindctl and dind2 works with node live-push, but again, there isn't much use of live-push for dindctl due to bullet number 1. 
3) Dindctl also just seems unmaintained, and filled with a bunch of small bugs. After some time, I just gave up trying to hack things together and decided to attempt this re-write. 

**Dind2** 
Dind2 is simple to use and works with the latest supervisors and balenaos-es.
```
dind2 start \
    --os-version ${balena_os_version} \
    --device-type ${balena_device_type}
```
```
dind2 stop 
```

Then, you just run `npm sync ${docker_container_ip}` and you're good to go. 

**Notes** 
Dindctl has a --preload flag which I haven't tried, and haven't re-implemented in dind2. 

I haven't removed dindctl from this PR yet, but it should be removed if dind2 is to be pulled in. This projects README would also need to be updated.

### DIFF
--- a/tools/dind/dind2
+++ b/tools/dind/dind2
@@ -1,0 +1,102 @@
+#!/bin/bash
+
+#
+# This tool can be used to facilitate supervisor development. This tools runs a supervisor
+# on a Docker instance of a Balena OS. It can be used the the live-push feature in the
+# supervisors code base.
+#
+# Usage: dind2 action [options]
+#
+# Actions:
+#   start [options]           Start docker in docker.
+#   stop                      Stop the docker in docker instance.
+#   logs [-f]                 Prints our supervisor logs. Use -f to follow or any other arguments you'd wish to send to journalctl.
+#
+# Options:
+#   --os-version              Specify the balena-os version to use. See note 1. Default: 2.58.6_rev1
+#   --device-type             Specify the device type. See note 1. Default: genericx86-64-ext
+#
+#
+# Notes:
+#   1: Will be used as so: resin/resinos:${OS_VERSION}.dev-${DEVICE_TYPE}
+#
+
+THIS_FILE=$0
+OS_VERSION="2.58.6_rev1"
+DEVICE_TYPE="genericx86-64-ext"
+DIND_PREFIX="balenaos-in-container"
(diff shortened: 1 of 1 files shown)


cywang117:
Hi @peakyDicers , thanks for all the investigations you've done surrounding dind. Some background that we currently use a specific development method for Supervisor development, which consists of livepushing Supervisor onto a local device / Intel NUC VM. The local device/VM are provisioned to either balenaCloud production, balenaCloud staging, or an internal tool which runs the entire balena platform stack on an Intel NUC running balenaOS and in local mode. You can read more about this tool [here](https://www.balena.io/blog/how-we-develop-balenacloud-using-balenacloud/), but it's still in development and as such not publicly available. Because of that, the use cases for dind are few and far between, but we certainly appreciate the time you've taken on this. 

Some questions to help us determine how to proceed with reviewing this PR -- have you used the original dind for Supervisor development? How did you find it? Functionally, are there any differences between dind2 and the original? What tests have you run with dind2 to determine that it's working as intended? Does it work with device types that aren't in the x86-64 arch family?

Finally, I remember our colleague @Page- mentioning that he uses/used dind for some functionality, so Page, do you have any input on this PR?

20k-ultra:
closing in favour of https://github.com/balena-os/balena-supervisor/pull/1774

pipex:
Closing in favor of #1985.

Issue #1775, opened 2021-08-19
Supervisor not restarting proxy in balenaOS >= v2.82.0


pipex:
In balenaOS [v2.82.0](https://github.com/balena-os/meta-balena/blob/master/CHANGELOG.md#v2820) the `resin-proxy-config` service was renamed to `balena-proxy-config` but the supervisor was not updated  accordingly here

https://github.com/balena-os/balena-supervisor/blob/master/src/host-config.ts#L155-L156

Same thing for `resin-hostname` here https://github.com/balena-os/balena-supervisor/blob/master/src/host-config.ts#L167

Fix should be easy but we'll need to try both versions until we have a specific OS cutoff

klutchell:
I'll fix this up since we have releases in production with this name change.

Going forward, the OS will actually restart the related services automatically when certain files change.

1. https://github.com/balena-os/meta-balena/pull/2294
2. `resin-proxy-config` and `redsocks` TODO

In that case we actually want to ensure the supervisor does not try to restart these services. Not sure of the best way to approach that offhand without checking the OS version. Maybe there's some other state we could check before attempting to restart these services via the supervisor.

pipex:
@klutchell I think you already found the solution in #1776?

klutchell:
> @klutchell I think you already found the solution in #1776?

Yeah, checking if they are part of certain target units will determine if they need to be restarted or not. I haven't tested that PR yet so it's still in Draft ATM.

PR #1776, opened 2021-08-20
Skip restarting services if they are part of conf targets

klutchell:
Some recent changes to the OS allowed some services to restart
automatically when the associated config files are changed.

In these cases we want to avoid restarting the same services
manually from the supervisor.

Change-type: patch
Signed-off-by: Kyle Harding <kyle@balena.io>
Resolves: https://github.com/balena-os/balena-supervisor/issues/1775

Tested with 3 manually edited versions of the OS:
1. `balena-hostname` and `balena-proxy-config` both restarted automatically via the hostOS => supervisor does not restart
2. `balena-hostname` and `balena-proxy-config` not restarted automatically via the hostOS => supervisor restarts services
3. `resin-hostname` and `resin-proxy-config` not restarted automatically via the hostOS => supervisor restarts services

### DIFF
--- a/src/host-config.ts
+++ b/src/host-config.ts
@@ -152,8 +152,28 @@ async function setProxy(maybeConf: ProxyConfig | null): Promise<void> {
 		await writeFileAtomic(redsocksConfPath, redsocksConf);
 	}
 
-	await dbus.restartService('resin-proxy-config');
-	await dbus.restartService('redsocks');
+	// restart balena-proxy-config if it is loaded and NOT PartOf redsocks-conf.target
+	if (
+		(
+			await Bluebird.any([
+				dbus.servicePartOf('balena-proxy-config'),
+				dbus.servicePartOf('resin-proxy-config'),
+			])
+		).includes('redsocks-conf.target') === false
+	) {
+		await Bluebird.any([
+			dbus.restartService('balena-proxy-config'),
+			dbus.restartService('resin-proxy-config'),
+		]);
+	}
+
+	// restart redsocks if it is loaded and NOT PartOf redsocks-conf.target
+	if (
+		(await dbus.servicePartOf('redsocks')).includes('redsocks-conf.target') ===
+		false
+	) {
+		await dbus.restartService('redsocks');
+	}
 }
 
 const hostnamePath = path.join(constants.rootMountPoint, '/etc/hostname');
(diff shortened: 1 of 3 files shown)


src/host-config.ts
```diff
@@ -164,7 +178,36 @@ async function readHostname() {
...
+	}
+}
+
+async function isServicePartOf(
+	serviceName: string,
+	unitName: string,
+): Promise<boolean> {
+	try {
+		const partOf = await dbus.servicePartOf(serviceName);
+		return _.includes(unitName, partOf);
```
klutchell, line 204: At this point `partOf` is an unknown type and I can't do much with it. The following line always fails, even when I print `partOf` and confirm the substring exists.

pipex: think the problem is that `getUnitProperty` doesn't have a return type. Can you add `Promise<string>` as return type to the function and try

test/41-device-api-v1.spec.ts
```diff
@@ -973,9 +973,10 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 						expect(await exists(noProxyPath)).to.be.false;
 					});
 
-				expect(restartServiceSpy.callCount).to.equal(2);
-				expect(restartServiceSpy.args).to.deep.equal([
-					['resin-proxy-config'],
+				// Should query service(s) to see if a restart is required
+				expect(servicePartOfSpy.callCount).to.equal(2);
```
pipex, line 977: I think the code is fine, but the tests you modified now don't test the `restartService` branch of the code. Could you add a test to check that that the function `restartService` is called when the `servicePartOf` throws an exception? 

I know the implementation is not great, because to test we need to know the internal details of the method, but for the time being is probably better to add that extra test case.

src/host-config.ts
```diff
@@ -164,7 +192,22 @@ async function readHostname() {
...
-	await dbus.restartService('resin-hostname');
+
+	try {
+		// restart balena-hostname if it is loaded and NOT PartOf config-json.target
+		if (
+			!(await dbus.servicePartOf('balena-hostname')).includes(
+				'config-json.target',
+			)
+		) {
+			await dbus.restartService('balena-hostname');
```
pipex, line 203: To avoid the nested try/catch you could do 

```
await Bluebird.any([dbus.restartService('balena-hostname'), dbus.restartService('resin-hostname')])
```

Which will try both alternatives and only throw if both fail.

Unfortunately that is not implemented yet on standard node https://www.npmjs.com/package/promise.any

pipex (approved):
Nice work @klutchell. LGTM!

PR #1777, opened 2021-08-23
backends/extra-uEnv: Extend custom DTB support for Jetson TX2 NX

acostach:
We just added support for the TX2 NX, which supports u-boot
thus allows for using custom device-trees. Let's allow
for both the Jetson TX2, Jetson TX2 NX and future derived
device types to have device-trees configured from the dashboard.

Change-type: Patch


acostach:
@20k-ultra the newly released TX2 NX module has the slug 'jetson-tx2-nx-devkit' and we'll be merging a derived community device soon, the CTI Photon carrier board that will support the TX2 NX modules. Please let me know if the slug needs to also be added in other places apart from the var schema: https://github.com/balena-io/open-balena-api/pull/744

cywang117:
@balena-ci rebase

cywang117:
@20k-ultra @pipex It looks good to me -- should we proceed with merging this?

acostach:
@balena-ci rebase

Issue #1231, opened 2020-03-30
Race condition between user containers and balena-engine api


robertgzr:
... leads to `/var/run/balena.sock` being created as a directory on the host.

This can happen after the user containers have been created and started when we reboot the device/ restart balena-engine.
The engine brings up previously started containers and it's own API (including the socket) in parallel, leading to a race condition when the `--volume` flag was used for creating the containers.

One solution on the supervisor side would be to switch to the `--mount` notation (or it's equivalent in docker-compose) which does not auto-create the source of the bind mount as a directory if it is missing on the host.

Related: 
* https://github.com/balena-os/meta-balena/issues/1748
* https://github.com/moby/moby/issues/30348

robertgzr:
another quick solution would be to force containers that mount the balena socket to have the restart policy set to off. This would move the responsibility of starting the container to the supervisor rather than the engine and thus avoid the race

robertgzr:
it looks like this can still happen with application containers that have access to the balena socket

see https://www.flowdock.com/app/rulemotion/balenalabs/threads/chG_M67A0Va0Bl2cgLiZ4pdCelh

robertgzr:
The supervisor should use the `mounts` field instead of [`binds`](https://github.com/balena-io/balena-supervisor/blob/f8cc2bc0cea362ea0d3dec3e564028f918f64dca/src/compose/service.ts#L692
) for mounting host paths into the container: 

see:
* https://github.com/moby/moby/issues/30348#issuecomment-532285233
* https://docs.docker.com/engine/api/v1.40/#operation/ContainerCreate

robertgzr:
this comes with the caveat that container start will fail if the source path doesn't exist, which needs to be handled

pipex:
This issue was fixed by in #1946

PR #1780, opened 2021-08-25
Use mount instead of binds to bind mount paths into the container

robertgzr:
When using the binds api the docker engine will auto-create the host
path as a directory if it doesn't exist. This can lead to unintended
side-effects when the source is a file (like when bind-mounting the
docker socket), as seen in the linked issue.

Closes https://github.com/balena-os/balena-supervisor/issues/1231

Change-type: major
Signed-off-by: Robert Günzler <robertg@balena.io>
Co-authored-by: Christina Wang <christina@balena.io>

### DIFF
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -528,10 +528,29 @@ export class Service {
 		}
 		expose = _.uniq(expose);
 
-		const tmpfs: string[] = [];
-		_.each((container.HostConfig as any).Tmpfs, (_v, key) => {
-			tmpfs.push(key);
-		});
+		const tmpfs: string[] = _.uniq(
+			([] as string[]).concat(
+				Object.keys(container.HostConfig?.Tmpfs || {}),
+				// also consider mount api
+				(container.HostConfig?.Mounts || [])
+					.filter(({ Type }) => Type === 'tmpfs')
+					.map(({ Target }) => Target),
+			),
+		);
+
+		// bind and volume mounts
+		const mounts: string[] = _.uniq(
+			([] as string[]).concat(
+				Object.keys(container.Config?.Volumes || {}),
+				// also consider mount api
+				(container.HostConfig?.Mounts || [])
+					.filter(({ Type }) => Type === 'bind' || Type === 'volume')
+					.map(
+						(mount) =>
+							`${mount.Source}:${mount.Target}${mount.ReadOnly ? ':ro' : ''}`,
+					),
+			),
+		);
 
 		// We cannot use || for this value, as the empty string is a
 		// valid restart policy but will equate to null in an OR
@@ -558,10 +577,7 @@ export class Service {
 			hostname,
 			command: container.Config.Cmd || '',
 			entrypoint: container.Config.Entrypoint || '',
-			volumes: _.concat(
-				container.HostConfig.Binds || [],
-				_.keys(container.Config.Volumes || {}),
-			),
+			volumes: mounts,
 			image: container.Config.Image,
 			environment: Service.omitDeviceNameVars(
 				conversions.envArrayToObject(container.Config.Env || []),
@@ -667,7 +683,7 @@ export class Service {
 		deviceName: string;
 		containerIds: Dictionary<string>;
 	}): Dockerode.ContainerCreateOptions {
-		const { binds, volumes } = this.getBindsAndVolumes();
+		const { mounts, volumes } = this.getMountsAndVolumes();
 		const { exposedPorts, portBindings } = this.generateExposeAndPorts();
 
 		const tmpFs: Dictionary<''> = {};
@@ -723,7 +739,6 @@ export class Service {
 			HostConfig: {
 				CapAdd: this.config.capAdd,
 				CapDrop: this.config.capDrop,
-				Binds: binds,
 				CgroupParent: this.config.cgroupParent,
 				Devices: this.config.devices,
 				DeviceRequests: this.config.deviceRequests,
@@ -751,6 +766,7 @@ export class Service {
 				CpusetCpus: this.config.cpuset,
 				Memory: this.config.memLimit,
 				MemoryReservation: this.config.memReservation,
+				Mounts: mounts,
 				OomKillDisable: this.config.oomKillDisable,
 				OomScoreAdj: this.config.oomScoreAdj,
 				Privileged: this.config.privileged,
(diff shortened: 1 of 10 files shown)


robertgzr:
if this all passes and you approve I'd still like to squash everything down into the original commit :)

pipex:
@balena-ci rebase

pipex:
@robertgzr did a quick test pushing an app with a volume with this branch and got

```
(HTTP code 400) unexpected - invalid mount config for type "bind": invalid mount path: '1_my-data' mount path must be absolute
```

I'm not sure exactly where the issue is, I'll try to take a deeper look in the next couple of days

robertgzr:
@pipex sounds like it tries to make a bind from something that's actually a volume, I'll add a volume to the test case so we can also see it on ci

robertgzr:
@pipex yeah it was named volumes that I just ignored in my implementation. they usually appear under the `binds` key as well but need to be moved to the `mounts` format as well. I also fixed that in the tests. Let me know if you would rather have that as a single commit

src/compose/service.ts
```diff
@@ -507,6 +507,16 @@ export class Service {
 			tmpfs.push(key);
 		});
 
+		const mounts: string[] = (container.HostConfig?.Mounts || [])
```
robertgzr, line 510: this should use `container.Mounts` instead

src/compose/service.ts
```diff
@@ -507,6 +507,16 @@ export class Service {
...
+			.filter(
+				({ Type, Source, Target }) =>
+					Type === 'bind' && path.isAbsolute(Source) && path.isAbsolute(Target),
```
robertgzr, line 513: we can probably just trust `type` here

20k-ultra:
just to be in the loop...what work is waiting for this ?

robertgzr:
@20k-ultra for me to find some time :p

robertgzr:
@pipex @20k-ultra @cywang117 can I get another review here?

pipex:
Have you tested this on a device @robertgzr ? I would be interested in checking that these scenarios work

- container that already exists with `Binds` configuration for a volume. What happens on supervisor update?
- newly created container, does the `Mount` get added properly?

pipex:
thank you @cywang117

robertgzr:
@pipex so in the second case, just using a fresh qemu device with the SV from this branch (`npm run sync -- 127.0.0.1`) I get:
```
Error: Failed to apply state transition steps. (HTTP code 400) unexpected - invalid mount config for type "bind": bind source path does not exist: /tmp/balena-supervisor/services/1/idle
```

Looks like we need to actually handle the creation ourselves now, as opposed to docker doing this for us. That is expected (and the intended effect of this PR)

cywang117:
@balena-ci rebase

cywang117:
compose/service test suite not passing.. taking a look

cywang117:
@pipex, from your 2 bullet points [a few comments ago](https://github.com/balena-os/balena-supervisor/pull/1780#issuecomment-1074259832),

> container that already exists with Binds configuration for a volume. What happens on supervisor update?

Bind mounts for volumes are not allowed through balena's docker-compose, so does this question still require testing?

pipex:
> Bind mounts for volumes are not allowed through balena's docker-compose, so does this question still require testing?

We use bind mounts when configuring labels, so the same label would create `Bind` before the update and a `Mount` after the update (I expect). If that's the case, will the container be recreated?

src/compose/service.ts
```diff
@@ -528,10 +528,30 @@ export class Service {
...
+					.filter(({ Type }) => Type === 'tmpfs')
+					.map(({ Target }) => Target),
+			),
+		);
+
+		// bind and volume mounts
+		const mounts: string[] = _.uniq(
+			([] as string[]).concat(
+				Object.keys(container.Config?.Volumes || {}),
+				container.HostConfig?.Binds || [],
```
cywang117, line 545: Removing this line may force recreate of container

test/data/docker-states/simple/inspect.json
```diff
@@ -34,9 +34,18 @@
...
+			{
+				"Type": "bind",
+				"Source": "/tmp/balena-supervisor/services/1011165/main",
+				"Target": "/tmp/resin"
+			},
+			{
+				"Type": "bind",
+				"Source": "/tmp/balena-supervisor/services/1011165/main",
+				"Target": "/tmp/balena"
+			}
```
cywang117, line 48: Changed from `Binds` to `Mounts` in these 3 test JSON files to ensure the `Creating service instances from docker configuration` ([ref](https://github.com/balena-os/balena-supervisor/blob/3a315db354c2c937b2a3a8d37d3fe00196b6a192/test/src/compose/service.spec.ts#L1015-L1046)) test suites passed.

src/compose/service.ts
cywang117: I removed this line in https://github.com/balena-os/balena-supervisor/commit/3a315db354c2c937b2a3a8d37d3fe00196b6a192 and updated some test data files as a result. Unit tests pass, on-device tests pending.

test/data/docker-states/simple/inspect.json
cywang117: On a device, when inspecting a container that's created with the corrrect HostConfig.Mounts config, this is what you would expect to see, so changing the inspect.json data in this way is in line with how the change should be reflected on devices.

cywang117: Reverted the JSON changes since we're using Binds instead of Mounts for update lock bind mounts.

src/compose/service.ts
cywang117: Added this line back with https://github.com/balena-os/balena-supervisor/commit/c8f2c21e84971bd893a273578f4f81683f23386c, since update lock bind mounts are configured as HostConfig.Binds.

cywang117:
@pipex @20k-ultra Ready for review, all on-device and unit tests pass.

The summary of the final changes are that containers are recreated to have all volumes be configured for HostConfig.Mounts instead of HostConfig.Binds, with the exception of update lock bind mounts in `/tmp/balena-supervisor/services` which still use HostConfig.Binds. The reason update lock bind mounts still use this is to make use of the Binds feature wherein directories are auto-created if they don't exist on host. This means the Supervisor won't have to manually create update lock directories and can leave that to the engine, as it was done before.

Should squash merge following approval.

cywang117:
Closing in favor of two PRs mentioned in https://github.com/balena-os/balena-supervisor/issues/1933

PR #1782, opened 2021-08-31
Include issues with downgrading versions in README

20k-ultra:
We removed the upgrade-policy document in https://github.com/balena-os/balena-supervisor/pull/1642 so I wanted to still mention that the Supervisor is not downgrade-able


20k-ultra:
@balena-ci I self-certify!

PR #1784, opened 2021-09-01
Remove "variable list" heading in configuration doc

20k-ultra:
cleans up the doc


20k-ultra:
@balena-ci I self-certify!

PR #1797, opened 2021-09-24
Improve network interface filter

pipex:
The supervisor filters out some network interfaces for mac address
reporting, to remove (balena*,lo,tun*,etc). The previous filter was
matching any interface containing in one of the defined filters, making
it stricter than necessary. This commit fixes the issue

Change-type: patch


pipex:
@balena-ci I self certify!

PR #1798, opened 2021-09-27
Add backoff to state reporting when 503 is received

pipex:
Current state reporting has a backoff when network or inconsistency
errors were found, but not on API errors. This change adds a backoff
using RetryAfter header if present to reduce load on API

Change-type: patch


pipex:
@balena-ci retest

src/device-state/current-state.ts
```diff
@@ -130,100 +140,124 @@ const getStateDiff = (): DeviceStatus => {
...
+const backoff = (
+	retry: (attempts: number) => void,
+	attempts = 0,
+	maxDelay: number,
+	minDelay = MINIMUM_BACKOFF_DELAY,
+) => {
+	const delay = Math.min(2 ** attempts * minDelay, maxDelay);
+	log.info(`Retrying request in ${delay / 1000} seconds`);
+	setTimeout(() => retry(attempts + 1), delay);
+};
```
cywang117, line 166: Is there a way to reset this backoff delay in current state reporting, such as with an API call that uses `force`?

cywang117: I'm imagining a scenario where current state reporting has failed and is at the maximum backoff delay time. The user then changed something in their network, since perhaps that was the issue that caused state reporting to fail, and would like to force a retry of the current state report. How would they achieve this?

cywang117: That's understandable, and since this is the case, it sounds fine to not allow a force option. This LGTM

src/lib/errors.ts
```diff
@@ -10,7 +10,11 @@ export interface StatusCodeError extends Error {
 }
 
 export class StatusError extends Error {
```
20k-ultra, line 12: Can we change this class to be called `HttpError` because StatusError doesn't make sense. I saw that its only referenced in 6 places so should be easy to refactor.

20k-ultra (commented):
Looks great, why did you delete the entire current-state.spec.ts file though ?

src/lib/errors.ts
pipex: I think it makes sense as is tbh. `StatusError` is a node error that can also have a status code, we use it some places to provide more info about the error. 

`HttpError` is one class of status error, although I agree that adding the `retryAfter` moves it towards HTTP camp. 

Maybe I could create an `HttpError` that extends `StatusError` with the `retryAfter` and is used by the current state report, but I don't see it as a urgent improvement.

pipex:
@20k-ultra this type of module doesn't lend itself well to testing, it has too many dependencies and the behavior is not really easy to describe. Christina did a good job testing the module before, but at the expense of making the test  dependent on the implementation details of the module, so it immediately broke. 

On #1765 I am making the module very simple so testing the reporting is really not as important as testing the functions the reporting depends on.

src/lib/errors.ts
20k-ultra: if StatusError is used in places that are not related to HTTP then that makes sense. Extending it seems nice but we can leave it for now I suppose since it's only 1 spot.

src/device-state/current-state.ts
```diff
@@ -86,13 +93,16 @@ const sendReportPatch = async (
...
 	};
 
-	const [{ statusCode, body: statusMessage }] = await request
-		.patchAsync(endpoint, params)
-		.timeout(apiTimeout);
+	const [
+		{ statusCode, body: statusMessage, headers },
+	] = await request.patchAsync(endpoint, params).timeout(apiTimeout);
 
 	if (statusCode < 200 || statusCode >= 300) {
```
20k-ultra, line 100: The current condition allows 300 for some reason.

```suggestion
// If statusCode is not 2xx range then throw an error
	if (statusCode < 200 || statusCode > 300) {
```

src/device-state/current-state.ts
```diff
@@ -130,100 +140,119 @@ const getStateDiff = (): DeviceStatus => {
...
+
+				reportPending = false;
+			},
+			(e) => {
+				if (e instanceof StatusError) {
+					// We don't want these errors to be classed as a report error, as this will cause
+					// the watchdog to kill the supervisor - and killing the supervisor will
+					// not help in this situation
+					log.error(
+						`Non-200 response from the API! Status code: ${e.statusCode} - message:`,
```
20k-ultra, line 232: When this error is caught it never says that the device state report failed. On Line 249 the eventTracker is passed a similar message but it doesn't go to the journal logs.

```suggestion
						`Device state report failure! Status code: ${e.statusCode} - message:`,
```

I don't think mentioning Non-200 response from the API explains what's happening.

src/device-state/current-state.ts
pipex: 300 is an error response, so the code is correct no? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/300

src/device-state/current-state.ts
20k-ultra: I've just realized the eventTracker is logged to journal but it wasn't triggered because I was testing statusCodes. The eventTracker is called when the error is not of StatusError type

src/device-state/current-state.ts
```diff
@@ -130,100 +140,119 @@ const getStateDiff = (): DeviceStatus => {
...
+					backoff(
+						reportCurrentState,
+						// Do not do exponential backoff if the API reported a retryAfter
+						e.retryAfter ? 0 : attempts,
+						maxDelay,
+						// Start the polling at the value given by the API if any
+						e.retryAfter ?? MINIMUM_BACKOFF_DELAY,
+					);
+				} else {
+					eventTracker.track('Device state report failure', { error: e });
```
20k-ultra, line 249: if we change the following line we can turn

```
[event]   Event: Device state report failure {"error":{"message":""}}
```

into

```
[event]   Event: Device state report failure {"error":"getaddrinfo ENOTFOUND api.xgho.st"}
```

```suggestion
					eventTracker.track('Device state report failure', { error: e.message });
```

src/device-state/current-state.ts
20k-ultra: ah, I had a brain fart. Yes we WANT to include 300... this would allow 200...300 but I meant 200..299.

Can you add the comment though, I think it helps

20k-ultra (approved):

PR #1799, opened 2021-09-28
Only deploy on commits from `balena-ci` and versionbot

pipex:
Change-type: patch


pipex:
@balena-ci I self certify!

PR #1801, opened 2021-09-29
Create supervisor docker-compose

pipex:
Creates supervisor docker-compose for creating releases on balena-cloud. This will allow the supervisor service to be correctly named when querying the API.

Change-type: patch


pipex:
@balena-ci I self certify!

PR #1802, opened 2021-09-29
Check config for networks and volumes inside Service

pipex:
This removes the need for the app module to know about the naming
conventions for networks and volumes since those exist now within the
service module itself. This also fixes a small bug where the volume would be
removed before the service itself had been successfully stopped.

Change-type: patch


src/compose/app.ts
```diff
@@ -657,33 +651,22 @@ export class App {
...
 		}
 
+		// Wait for networks to be created before starting the service
 		if (
 			_.some(
 				networkPairs,
-				(pair) =>
-					`${this.appId}_${pair.target?.name}` === target.config.networkMode,
+				(pair) => pair.target && target.hasNetworkMode(pair.target.name),
 			)
```
cywang117, line 659: Could just substitute for native method here as `networkPairs` is of type Array.
```
networkPairs.some((pair) => pair.target && target.hasNetworkMode(pair.target.name))
```

src/compose/app.ts
```diff
@@ -657,33 +651,22 @@ export class App {
...
-
-				return _.some(
-					volumePairs,
-					(pair) => `${target.appId}_${pair.target?.name}` === sourceName,
-				);
-			})
+			_.some(
+				volumePairs,
+				(pair) => pair.target && target.hasVolume(pair.target.name),
+			)
```
cywang117, line 669: Native Array.prototype.some substitution could be used as in previous comment since volumePairs is also an Array type.

src/compose/service.ts
```diff
@@ -1086,6 +1086,8 @@ export class Service {
 					log.warn(`Ignoring invalid bind mount ${volume}`);
 				}
 			} else {
+				// Push anonymous volume. Anonymous volumes can only be
+				// set by using the `VOLUME` instruction inside a dockerfile
```
cywang117, line 1090: This is different than unmanaged volumes?

src/compose/app.ts
pipex: yeah, that's a good point, I was trying to minimize the changes on the v3 PR but it makes sense to do it here

src/compose/service.ts
pipex: what are unmanaged volumes? Anonymous volumes are a docker thing https://docs.docker.com/storage/volumes/#remove-anonymous-volumes, I think they are the reason we see sometimes those weirdly named volumes

PR #1816, opened 2021-11-05
Do not build for staging in circleci job

20k-ultra:
Since we have balena-ci action building releases for staging we can remove the circleci code that does this as well.

I didn't remove all the staging code in the job because we are going to delete the entire file once we use the action for production.


20k-ultra:
@balena-ci I self-certify!

PR #1820, opened 2021-11-10
Log entire Exception if message is empty

20k-ultra:
The log message `Event: Device state report failure {"error":{"message":""}}` is only found here. It doesn't find the root cause but this will help to pin it down instead of seeing this incredibly annoying message. Ideally this change just leads to the discovery of the root cause.

related to https://github.com/balena-os/balena-supervisor/issues/1680

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -246,8 +246,9 @@ function reportCurrentState(attempts = 0) {
 						e.retryAfter ?? MINIMUM_BACKOFF_DELAY,
 					);
 				} else {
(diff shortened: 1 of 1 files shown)


20k-ultra:
@balena-ci rebase

20k-ultra:
@balena-ci rebase

20k-ultra:
@balena-ci rebase

20k-ultra:
@balena-ci rebase

20k-ultra:
I want this PR to try to catch the error better and log to stdio (journal) but continue to send the current error message to the dashboard

PR #1825, opened 2021-11-11
Change repo type to skip concourse build

pipex:
Testing and building is now done on balena builders. This sets the repo type to generic to avoid unnecessary checks

Change-type: patch


pipex:
@balena-ci I self-certify!

Issue #1233, opened 2020-03-30
User reports that delete-then-download strategy isn't deleting the images beforehand.


thgreasi:
Steps to reproduce:
* Set app's `RESIN_SUPERVISOR_UPDATE_STRATEGY` to `delete-then-download`
* Pinning the device to a different release

Expected results:
* images & containers getting deleted
* then the new download starts

Observed results:
* `balena ps -a` shows the containers being deleted
* dowload starts
* `balena images` still shows the old images
* download finishes
* the images of the old release get deleted

Supervisor logs:
```
Downloaded image 'registry2.balena-cloud.com/v2/faabd6c2ee0f68eef24f2a96b0eff77e@sha256:7d42ad0a910adaa67c2b4a47b6042927aa8758c0f1c680f2e640eab346f9d4c2'
Downloaded image 'registry2.balena-cloud.com/v2/f058387cf0cff9c098faab6e547932d0@sha256:e50314ff8c5b587753e47e365cd76c5efd33a462155f39b25775df03140a9bf9'
Deleting image 'registry2.balena-cloud.com/v2/f9c4be01fa5b8bb34b23270f19e97dbb@sha256:adad5135f662107ef25e8286c06b2bd6a5e7a689189befed38941538a1d74ad6'
Deleting image 'registry2.balena-cloud.com/v2/aac894f3f1b98f121fb53166bb4c9c7f@sha256:44fceb7229fca7c01b6a896b7875345797f5b626f0d55b702defee1ddcd708ec'
Deleted image 'registry2.balena-cloud.com/v2/f9c4be01fa5b8bb34b23270f19e97dbb@sha256:adad5135f662107ef25e8286c06b2bd6a5e7a689189befed38941538a1d74ad6'
Deleted image 'registry2.balena-cloud.com/v2/aac894f3f1b98f121fb53166bb4c9c7f@sha256:44fceb7229fca7c01b6a896b7875345797f5b626f0d55b702defee1ddcd708ec'
Installing service 'data sha256:c70ec64ff11be1e6d2180936396cd6935469811869c992c9f4997b1d1e6c8f84'
Installed service 'data sha256:c70ec64ff11be1e6d2180936396cd6935469811869c992c9f4997b1d1e6c8f84'
Starting service 'data sha256:c70ec64ff11be1e6d2180936396cd6935469811869c992c9f4997b1d1e6c8f84'
[...]

PR #1826, opened 2021-11-11
Fix `delete-then-download` strategy

pipex:
The strategy has been broken for a while but it was not clear how to
fix it before the changes to image management. This PR fixes application
manager to remove images before downloading the new image. This will
only have an effect on changing images.

Closes: #1233
Change-type: patch

### DIFF
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -680,9 +680,9 @@ function saveAndRemoveImages(
 				}) ?? _.find(availableImages, { dockerImageId: svc.config.image }),
 		),
 	) as imageManager.Image[];
-	const targetImages = _.flatMap(target, (app) =>
-		_.map(app.services, imageForService),
-	);
+
+	const targetServices = Object.values(target).flatMap((app) => app.services);
+	const targetImages = targetServices.map(imageForService);
 
 	const availableAndUnused = _.filter(
 		availableWithoutIds,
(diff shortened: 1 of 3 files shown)


src/compose/application-manager.ts
```diff
@@ -680,9 +680,9 @@ function saveAndRemoveImages(
 				}) ?? _.find(availableImages, { dockerImageId: svc.config.image }),
 		),
 	) as imageManager.Image[];
-	const targetImages = _.flatMap(target, (app) =>
-		_.map(app.services, imageForService),
-	);
+
+	const targetServices = _.flatMap(target, (app) => app.services);
```
20k-ultra, line 684:
```suggestion
	const targetServices = target.flatMap(app => app.services);
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

pipex: We need to upgrade to `es2019` in `tsconfig.json` to be able to use those operators

pipex: We need to set `es2019` on `tsconfig.json` for that.

pipex: We need to add `es2019` to `tsconfig.json` for that to work

pipex: I thought it would be painful but it was straightforward, I'll add it to the PR

pipex:
@balena-ci rebase

20k-ultra:
@balena-ci rebase

src/compose/application-manager.ts
```diff
@@ -680,9 +680,9 @@ function saveAndRemoveImages(
 				}) ?? _.find(availableImages, { dockerImageId: svc.config.image }),
 		),
 	) as imageManager.Image[];
-	const targetImages = _.flatMap(target, (app) =>
-		_.map(app.services, imageForService),
-	);
+
+	const targetServices = Object.values(target).flatMap((app) => app.services);
```
20k-ultra, line 684: this is unfortunate we have to do Object.values but i'm sure lodash was doing exactly this.

src/compose/application-manager.ts
```diff
@@ -741,32 +741,35 @@ function saveAndRemoveImages(
...
+
+	const imagesToRemove = _.filter(availableAndUnused, (image) => {
+		const notUsedForDelta = !_.includes(deltaSources, image.name);
+		const notUsedByProxyvisor = !_.some(proxyvisorImages, (proxyvisorImage) =>
+			imageManager.isSameImage(image, {
+				name: proxyvisorImage,
+			}),
+		);
+		return notUsedForDelta && notUsedByProxyvisor;
+	});
```
20k-ultra, line 772:
```suggestion
	const imagesToRemove = availableAndUnused.filter(image => {
		const notUsedForDelta = !deltaSources.includes(image.name);
		const notUsedByProxyvisor = !proxyvisorImages.some(proxyvisorImage => {
			imageManager.isSameImage(image, {
				name: proxyvisorImage
			})
		})
		return notUsedForDelta && notUsedByProxyvisor;
	})
```

test/src/compose/application-manager.spec.ts
```diff
@@ -389,6 +389,110 @@ describe('compose/application-manager', () => {
 			.that.deep.includes({ serviceName: 'main' });
 	});
 
+	it('infers a kill step when a service has to be updated but the strategy is delete-then-download', async () => {
```
20k-ultra, line 392: I don't think this level of unit testing could not have been sanely done if it weren't for all the testing improvements we've made to the Supervisor...this is super cool to see.

20k-ultra (commented):
See my 1 code suggestion to use standard lib functions then can self cert

src/compose/application-manager.ts
pipex: That's probably right. There is no other option as target is an object

src/compose/application-manager.ts
pipex: Sure, will do the change.

test/src/compose/application-manager.spec.ts
pipex: I think it could still be made simpler, there is a lot of boilerplate needed now, but is a better situation than what we had before for sure.

pipex:
@balena-ci I self-certify!

PR #1827, opened 2021-11-12
Re-add resinci config to control node checks

pipex:
Change-type: patch


pipex:
@balena-ci retest

cywang117:
@pipex LGTM, after this is merged and works as expected, we should be able to merge things again right?

pipex:
@cywang117 almost, the second issue currently going on is related to stale branches in versionbot

pipex:
@balena-ci I self certify!

PR #1829, opened 2021-11-17
Bootstrap supervisor configuration on container start

pipex:
This change goes in line with making the supervisor more like an app. This
reduces the number of parameters that need to be given to start the
supervisor container, moving it closer to something that could be
started by the supervisor itself. This should allow to greatly simplify
the supervisor [start script](https://github.com/balena-os/meta-balena/blob/master/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor/start-balena-supervisor)

This is now the minimal startup configuration for the container

```
balena run --privileged --name balena_supervisor \
	--restart=always \
	--net=host \
	--cidenv=SUPERVISOR_CONTAINER_ID \
	--mount type=bind,source=/var/run/balena-engine.sock,target=/var/run/balena-engine.sock \
	--mount type=bind,source=/,target=/mnt/root \
	-e DOCKER_SOCKET=/var/run/balena-engine.sock \
	"${SUPERVISOR_IMAGE}:${SUPERVISOR_TAG}"
```

Change-type: patch

### DIFF
--- a/setenv.sh
+++ b/setenv.sh
@@ -1,0 +1,63 @@
+#!/bin/sh
+
+set -a
+
+# read key from <key>="<value>" config file
+read_config() {
+  sed -n "s/^$1=\"\(.*\)\"$/\1/p" $2 2>/dev/null
+}
+
+# apply filter to json file and return raw output
+# return default if the filter does not exist
+read_json() {
+  [ "$(jq -r $1 $2 2>/dev/null)" != "null" ] && jq -r $1 $2 || echo -n $3
+}
+
+# Setup paths
+ROOT_MOUNTPOINT=${ROOT_MOUNTPOINT:-/mnt/root}
+BOOT_MOUNTPOINT=${BOOT_MOUNTPOINT:-/mnt/boot}
+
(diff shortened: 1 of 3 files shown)


setenv.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Dbus socket
+DBUS_SYSTEM_BUS_ADDRESS="unix:path=${ROOT_MOUNTPOINT}/run/dbus/system_bus_socket"
+
+# Use a trick to get the container id from inside the container itself in
+# in case the supervisor is started some other way than with the start-resin-supervisor
+# script. This will not work with cgrous v2
+# https://stackoverflow.com/a/25729598
+if [ -z "${SUPERVISOR_CONTAINER_ID}" ]; then
+	SUPERVISOR_CONTAINER_ID=$(cat /proc/self/cgroup | grep -o  -e "docker-.*.scope" | head -n 1 | sed "s/docker-\(.*\).scope/\\1/")
```
pipex, line 42: @klutchell do you know of a way to do this in cgroups2 by any chance?

PR #1835, opened 2021-11-24
add-supervisor-update-info

fisehara:
Adding clarity to the supervisor /v1/update endpoint that the endpoint will not trigger a supervisor update itself.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -89,7 +89,7 @@ $ curl -X POST --header "Content-Type:application/json" \
 
 > **Note:** on devices with supervisor version lower than 7.22.0, replace all `BALENA_` variables with `RESIN_`, e.g. `RESIN_SUPERVISOR_ADDRESS` instead of `BALENA_SUPERVISOR_ADDRESS`.
 
-Triggers an update check on the supervisor. Optionally, forces an update when updates are locked.
+Triggers a check for the target state of configurations and app services. Optionally, if a new target is available, the `force` option tells the supervisor to ignore locks and install the update. This will NOT update the supervisor service.
 
 Responds with an empty 204 (No Content) response.
 


cywang117:
LGTM @fisehara, did someone in forums / paid support get confused by the endpoint before?

docs/API.md
```diff
@@ -89,7 +89,7 @@ $ curl -X POST --header "Content-Type:application/json" \
 
 > **Note:** on devices with supervisor version lower than 7.22.0, replace all `BALENA_` variables with `RESIN_`, e.g. `RESIN_SUPERVISOR_ADDRESS` instead of `BALENA_SUPERVISOR_ADDRESS`.
 
-Triggers an update check on the supervisor. Optionally, forces an update when updates are locked.
+Triggers an update check on the supervisor. Optionally, forces an update when updates are locked. This update check will check for service updates and will not update the supervisor itself.
```
pipex, line 92: I think the whole phrase is a bit long and not too clear. How about something like  
```suggestion
Triggers a check for the target state of configurations and app services. Optionally, if a new target is available, the `force` options forces the supervisor to ignore locks and install the update. 
```

@cywang117 what do you think?

fisehara:
@cywang117 Yes, a customer thought he can trigger a supervisor update by triggering this SV api endpoint.

docs/API.md
fisehara: @pipex I'd like to have somewhere the explicit information that his endpoint will not trigger the supervisor to update itself.

pipex: Fair enough. Let's add something like `This will NOT update the supervisor service.`.

docs/API.md
```diff
@@ -89,7 +89,7 @@ $ curl -X POST --header "Content-Type:application/json" \
 
 > **Note:** on devices with supervisor version lower than 7.22.0, replace all `BALENA_` variables with `RESIN_`, e.g. `RESIN_SUPERVISOR_ADDRESS` instead of `BALENA_SUPERVISOR_ADDRESS`.
 
-Triggers an update check on the supervisor. Optionally, forces an update when updates are locked.
+Triggers a check for the target state of configurations and app services. Optionally, if a new target is available, the `force` options forces the supervisor to ignore locks and install the update. This will NOT update the supervisor service.
```
pipex, line 92: Last suggestion @fisehara :p 
```suggestion
Triggers a check for the target state of configurations and app services. Optionally, if a new target is available, the `force` option tells the supervisor to ignore locks and install the update. This will NOT update the supervisor service.
```

pipex (approved):

PR #1839, opened 2021-12-02
Improve target state validation

pipex:
This moves target state type definitions to io-ts which also improves validation and reporting. This will ensures target state validation errors are passed to the log backend.

This PR also extracts some commits from PR #1765 to simplify review. 

Change-type: patch


pipex:
@balena-ci I self certify!

PR #1841, opened 2021-12-03
Skip initial apply until a target has been set

pipex:
The supervisor always applies target state on start to ensure that the
device is at the correct in case of a crash or another reason. This had
the side effect that if the database is deleted, the supervisor would
apply target state (which is empty), stopping services and possibly
causing volume data loss.

This prevents that behavior and ensures that the supervisor only
applies target state if a target has been set either by the cloud, preload or local
mode.

Change-type: patch

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -406,24 +406,17 @@ export async function loadInitialState() {
 		update_downloaded: false,
 	});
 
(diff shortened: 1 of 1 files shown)


pipex:
@balena-ci I self certify!

PR #1842, opened 2021-12-06
Move all Supervisor API related routes & code to device-api directory

cywang117:
This is a code organization change that doesn't add new features. This will allow the API to be more easily separated into its own module / app in the long term, and easier API testing in the medium term.

Change-type: patch


cywang117:
Last possible task: the current API tests pass but they're not great, so there's room for improvement. Maybe not in this PR though

src/device-api/middleware/validation.ts
```diff
@@ -0,0 +1,501 @@
...
+		// already been sent, so it's safe to return from the middleware here.
+		if (res.writableEnded) {
+			return;
+		}
+	}
+
+	// Check auth scope against decoded appId
+	if (res.locals.appId && !req.auth.isScoped({ apps: [res.locals.appId] })) {
+		// If not scoped, return error message for auth validator
+	}
```
cywang117, line 498: TODO

src/device-api/middleware/validation.ts
```diff
@@ -0,0 +1,501 @@
...
+		}
+
+		// We don't want to `return handleLeft` in case it's not an instance
+		// of a 4xx or 5xx response being sent, since that may cut off the loop
+		// prematurely. res.writableEnded only returns true when a response has
+		// already been sent, so it's safe to return from the middleware here.
+		if (res.writableEnded) {
+			return;
+		}
+	}
```
cywang117, line 493: Can probably use fp-ts's `fold` here

src/device-api/middleware/validation.ts
cywang117: done

src/device-api/middleware/validation.ts
cywang117: Done

cywang117:
All pending TODOs are done, with the exception of maybe improving API tests

src/device-api/middleware.ts
```diff
@@ -0,0 +1,141 @@
...
+import log from '../lib/supervisor-console';
+import { UpdatesLockedError } from '../lib/errors';
+
+import type { Request, Response, NextFunction } from 'express';
+import type { AuthorizedRequestHandler } from './types';
+
+/**
+ * Request logger
+ */
+export const apiLogger = morgan(
```
pipex: Maybe we can rename this middleware to `logging`? As we have `middleware.auth`, then we would have `middleware.logging`

src/device-api/middleware.ts
```diff
@@ -0,0 +1,141 @@
+import * as morgan from 'morgan';
+import * as _ from 'lodash';
+
+import * as config from '../config';
+import * as apiKeys from '../lib/api-keys';
```
pipex: Should we move the `lib/api-keys` module to `./device-api`?

test/lib/mocked-device-api.ts
```diff
@@ -1,17 +1,14 @@
 import * as _ from 'lodash';
```
pipex, line 1: Do you know why we need this file? Have you maybe give it some though on how we can get rid from it?

src/compose/application-manager.ts
```diff
@@ -55,19 +53,9 @@ export const removeAllListeners: typeof events['removeAllListeners'] = events.re
 	events,
 );
 
-const proxyvisor = new Proxyvisor();
+export const proxyvisor = new Proxyvisor();
```
pipex, line 56: I don't like this, and the fact that the API has to import application manager.  What if we modify the `proxyvisor` module to have a default export ? Then both application manager and API can import the proxyvisor

```
// proxyvisor.ts

const proxyvisor = new Proxyvisor();
export default proxyvisor
```

src/types/basic.ts
```diff
@@ -182,5 +182,83 @@ export const DeviceName = new t.Type<string, string>(
...
+			chain((s) => {
+				if (['true', '1', 'on'].includes(s)) {
+					return t.success(true);
+				}
+
+				if (['false', '0', 'off'].includes(s)) {
+					return t.success(false);
+				}
+
+				return t.failure(i, c, 'must be a valid string boolean');
```
pipex, line 246: Is this too strict? Would it suffice with making anything not in `['true', '1', 'on']` as false?

src/types/basic.ts
```diff
@@ -47,7 +47,7 @@ export const StringIdentifier = new t.Type<string, string>(
 			chain((s) =>
 				!isNaN(+s) && +s === parseInt(s, 10) && +s >= 0
 					? t.success(s)
-					: t.failure(s, c, 'must be be an positive integer'),
```
pipex, line 50: oops, nice catch

src/types/basic.ts
```diff
@@ -182,5 +182,83 @@ export const DeviceName = new t.Type<string, string>(
...
+		),
+	Number,
+);
+export type BooleanFromInt = t.TypeOf<typeof BooleanFromInt>;
+
+/**
+ * A stringified or regular boolean, one of the set of (VALID_STRING_BOOLS | [true, false] | [0, 1]).
+ * Decodes to a boolean.
+ *
+ * TODO: replace PermissiveBoolean in src/config/types.ts
```
pipex, line 277: I was going to point this out too

src/types/basic.ts
```diff
@@ -182,5 +182,83 @@ export const DeviceName = new t.Type<string, string>(
...
+ */
+export const BooleanIdentifier = new t.Type<
```
pipex, line 279: I don't think `Identifier` is the right suffix here. Since `PermissiveBoolean` is taken, maybe `Bool` or `GenericBoolean` could valid options?

src/device-api/index.ts
```diff
@@ -10,7 +10,7 @@ import * as apiKeys from '../lib/api-keys';
 import * as middleware from './middleware';
 
 import type { Server } from 'http';
-import type { AuthorizedRequest } from './types';
+import type { AuthorizedRequest } from '../types';
```
pipex, line 13: I'm not sure about this commit. I think, whenever possible, we need to think as separate modules/subfolders as independent entities. Each module can have its own type definitions, its own internal helper functions, and their main API methods. 

Imho, a couple of good questions to ask are
- can I easily extract this submodule into its own repo? 
- does this piece of functionality of this submodule need to be exported? Is it used somewhere else in the code?

In this case, having the `AuthorizedRequest` as part of the main types does not help the rest of the code, so I don't think it belongs in `src/types`.

src/device-api/middleware/apiLogger.ts
```diff
@@ -0,0 +1,21 @@
+import * as morgan from 'morgan';
```
pipex, line 1: This is a bit nit-picky, but how about naming the folder `middlewares` instead of `middleware` as it is a collection of middlewares instead of a single middleware functionality? Similarly, I would expect if we were ever to split the API into multiple files, we would put them in a `routes` folder as that would communicate better purpose of that subfolder.

src/device-api/middleware/apiLogger.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import log from '../../lib/supervisor-console';
+
+/**
+ * Request logger
+ */
+export const apiLogger = morgan(
```
pipex, line 8: Also I would call this middleware `logging` (I think I mentioned before)

src/device-api/middleware.ts
cywang117: I wanted to, but saw that they are used in some other places, notably in [reporting initial state in device-state.ts](https://github.com/balena-os/balena-supervisor/blob/234a46c3cb163af959dce7969dfa09db78156560/src/device-state.ts#L249-L263) and [when generating a SV API key when the compose label is present](https://github.com/balena-os/balena-supervisor/blob/104a8006fbab4bad0f2da923eade0e64002dd59f/src/compose/utils.ts#L374-L377). Is the purpose of this move to indicate that "api-keys.ts belongs to device-api but can be used elsewhere"? If so then I can move it, no problem.

test/lib/mocked-device-api.ts
cywang117: I think if we follow your suggestion of having each endpoint call `doXX` as with `doRestart` and `doPurge` for example, we can move towards removing mocked-device-api.ts. The current existence of this file seems be to because the Supervisor API interacts too closely with its underlying interfaces and interacts at multiple points in some routes, too, so shrinking these interfaces down to one function call will go a long way. The subject of another PR, perhaps?

src/types/basic.ts
cywang117: I like GenericBoolean!

src/device-api/middleware/apiLogger.ts
cywang117: I shall do both

cywang117: After moving v1/v2 into a routes directory, a test in v1 API is failing. I'll just rename middlewares for now. Something is wrong in the tests that needs to be debugged, in the API testing PR..

src/device-api/index.ts
cywang117: That makes sense and was an oversight on my part, I'll move it to device-api.

src/types/basic.ts
cywang117: TODO: Christina will address this by itself in another PR

src/compose/application-manager.ts
cywang117: Done with https://github.com/balena-os/balena-supervisor/commit/5714830fceef45714312d08aeb5229b8324eb2c7

test/lib/mocked-device-api.ts
cywang117: Moved with https://github.com/balena-os/balena-supervisor/commit/5714830fceef45714312d08aeb5229b8324eb2c7

src/device-api/middleware.ts
cywang117: Moved with https://github.com/balena-os/balena-supervisor/commit/5714830fceef45714312d08aeb5229b8324eb2c7

test/lib/mocked-device-api.ts
cywang117: This will be done in the next API PR which will also move tests to unit / integration.

src/device-api/middleware/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './logging';
+export * from './auth';
+export * from './errorHandler';
```
pipex, line 3: How about renaming `errorHandler` to `errors`?

src/device-state.ts
```diff
@@ -45,7 +45,7 @@ import {
 	AppState,
 } from './types';
 import * as dbFormat from './device-state/db-format';
-import * as apiKeys from './lib/api-keys';
+import * as apiKeys from './device-api/api-keys';
```
pipex: Can we try to export the specific module that `device-state` (and maybe others) on `device-api/index.ts`, I would love if we can get to a point where modules can work more like black boxes, so only what is exported can be used. 

In this case the import would change to

```
import {initialised as initialiseApiKeys, cloudApiKey} from './device-api'
``` 

Also, it is probably not necessary in this PR, but maybe the API doesn't need to be initialised here, but the initialised method can be called when the cloudApiKey is obtained. Let's say we make `cloudApiKey` a function `getGlobalApiKey`

That function can be implemented like this

```
export async function getGlobalApiKey() { 
    await initialised();
    return generateCloudKey();
}
```

And only `getGlobalApiKey` needs to be exported by `device-api` from this module.

src/device-api/v1.ts
```diff
@@ -1,190 +1,213 @@
...
+	}
+
+	return doPurge(appId, force)
+		.then(() => res.status(200).json({ Data: 'OK', Error: '' }))
+		.catch(next);
+});
+
+router.post('/v1/update', (req, res, next) => {
+	eventTracker.track('Update notification');
+	if (isReadyForUpdates()) {
```
pipex, line 225: This is probably out of scope for this PR, but I am not sure this is necessary. Regardless of target state polling being started, the device API should be able to receive this request and try to trigger a target state request. If the device has not yet authenticated with the API, then the request will fail and the endpoint could return a 500

test/legacy/41-device-api-v1.spec.ts
```diff
@@ -667,10 +667,12 @@ describe('SupervisorAPI [V1 Endpoints]', () => {
 	describe('POST /v1/update', () => {
 		let configStub: SinonStub;
 		let targetUpdateSpy: SinonSpy;
+		let readyForUpdatesStub: SinonStub;
 
 		before(() => {
 			configStub = stub(config, 'get');
 			targetUpdateSpy = spy(TargetState, 'update');
+			readyForUpdatesStub = stub(apiBinder, 'isReadyForUpdates').returns(true);
```
pipex, line 675: I'm looking forward to us removing all this stubbing once we can improve the API tests

src/compose/app.ts
```diff
@@ -51,15 +51,26 @@ interface ChangingPair<T> {
 	target?: T;
 }
 
-export class App {
+export interface IApp {
```
pipex, line 54: Why do we need this? Why not just use `App` in `common.ts`?

src/device-api/v1.ts
cywang117: Makes sense. I'll add this logic in a separate PR

src/compose/app.ts
cywang117: That was the original intent, however type checks complain of missing public methods from the App class, such as `isTargetState`, `nextStepsForAppUpdate`, etc. I think this is due to the slight differences between current & target state, which are relevant here because we apply current state as target state in `doPurge` in common.ts.

Instead of declaring the `IApp` interface, I could specify the i/o types as:
```
export function safeAppClone(app: Partial<App>): Partial<App> {
``` 
or instantiate a new `App` instance for each `app` input into `safeAppClone`. Thoughts?

src/device-state.ts
cywang117: I added 2 new commits in https://github.com/balena-os/balena-supervisor/commit/a105417e89d4a33282e313ebb87bd0eaef6e4761 to address this. The dbus commit was because using deviceApi directly instead of apiKeys in compose/utils broke the unit tests for it.

src/compose/app.ts
pipex: Hmm, I think we need to improve the implementation of `doPurge`, I'm not sure why it needs to do the `safeStateClone` and eventually we need to clean up the way "intermediate" state is applied. In the meantime, I think instantiating an app in `safeAppClone` would solve the issue

`﻿``
export function safeAppClone(app: App): App {
   // ...
   return new App({appId: app.appId, /* add missing data here ... */}, true);
}
```

cywang117: Done with https://github.com/balena-os/balena-supervisor/pull/1842/commits/966d9574656bb393031ba64dd596f933eb972947!

src/device-state.ts
cywang117: Additionally the dbus commit is just better overall to decomplicate the test environment, and makes importing lib/dbus side-effect-free.

pipex: Yeah, I think this works in line with the latest changes. Some day we'll move to a dependency injection pattern at which point I hope these sort of singleton type initializations won't be necessary anymore 🙏

cywang117:
Tested on-device & works normally.

cywang117:
@balena-ci I self-certify!

pipex:
Awesome work @cywang117 !

PR #1843, opened 2021-12-07
Device state patch controls

20k-ultra:
Previously there was no way to control how frequently the device reported its current state. This PR adds config vars that let the cloud specify intervals for:

 - [x] current state patching via `BALENA_SUPERVISOR_PATCH_INTERVAL`
 - [ ] device metric patching via `BALENA_SUPERVISOR_METRICS_INTERVAL`
 - [ ] download progress patching via `BALENA_SUPERVISOR_DOWNLOAD_UPDATE_INTERVAL` ?

### DIFF
--- a/docs/configurations.md
+++ b/docs/configurations.md
@@ -25,6 +25,7 @@ This list contains configuration variables that can be used with all balena devi
 | BALENA_HOST_DISCOVERABILITY | boolean | false | true | Enable / Disable Avahi to run which will allow the device to respond to requests such as network scans. | v11.9.2 |
 | BALENA_HOST_SPLASH_IMAGE | integer | true | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
 | BALENA_SUPERVISOR_HARDWARE_METRICS | boolean | false | true | Toggle hardware metrics reporting to the cloud, which occurs every 10 seconds when there are changes. Metrics include CPU utilization, CPU temperature, memory usage, and disk space. Useful for devices with bandwidth sensitivity. | v12.8.0 |
+| BALENA_SUPERVISOR_PATCH_INTERVAL | integer | false | 60000 | Define the frequency of updating the cloud with device's statel in milliseconds. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v12.12.0 |
 
 ---
 
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -2,20 +2,17 @@ import * as _ from 'lodash';
 import * as url from 'url';
 import { CoreOptions } from 'request';
 
-import * as constants from '../lib/constants';
+import { backoffPromise } from '../lib/backoff';
 import { log } from '../lib/supervisor-console';
 import { InternalInconsistencyError, StatusError } from '../lib/errors';
 import { getRequestInstance } from '../lib/request';
 import * as sysInfo from '../lib/system-info';
+import sleep from '../lib/sleep';
 
 import { DeviceStatus } from '../types/state';
 import * as config from '../config';
-import { SchemaTypeKey, SchemaReturn } from '../config/schema-type';
 import * as eventTracker from '../event-tracker';
 import * as deviceState from '../device-state';
-
-// The exponential backoff starts at 15s
-const MINIMUM_BACKOFF_DELAY = 15000;
 
 const INTERNAL_STATE_KEYS = [
 	'update_pending',
@@ -24,33 +21,28 @@ const INTERNAL_STATE_KEYS = [
 ];
 
 export let stateReportErrors = 0;
+
 const lastReportedState: DeviceStatus = {
 	local: {},
 	dependent: {},
 };
-const stateForReport: DeviceStatus = {
-	local: {},
-	dependent: {},
-};
-let reportPending = false;
 
-type CurrentStateReportConf = {
-	[key in keyof Pick<
-		config.ConfigMap<SchemaTypeKey>,
-		| 'uuid'
-		| 'apiEndpoint'
-		| 'apiTimeout'
-		| 'deviceApiKey'
-		| 'deviceId'
-		| 'localMode'
-		| 'hardwareMetrics'
-	>]: SchemaReturn<key>;
-};
+/**
+ * patchInterval is set when startReporting successfuly queries the config
+ */
+let patchInterval: number;
 
-type StateReport = {
-	stateDiff: DeviceStatus;
-	conf: Omit<CurrentStateReportConf, 'deviceId' | 'hardwareMetrics'>;
-};
+/**
+ * Listen for config changes to statePatchInterval
+ */
+(async () => {
+	await config.initialized;
+	config.on('change', (changedConfig) => {
+		if (changedConfig.statePatchInterval) {
+			patchInterval = changedConfig.statePatchInterval;
+		}
+	});
+})();
 
 /**
  * Returns an object that contains only status fields relevant for the local mode.
@@ -64,7 +56,20 @@ const stripDeviceStateInLocalMode = (state: DeviceStatus): DeviceStatus => {
 	};
 };
 
-const report = async ({ stateDiff, conf }: StateReport) => {
+/**
+ * Try to patch difference in device state
+ *
+ * report attempts to patch the difference in device state from the last
+ * report successfully sent.
+ */
+async function report(stateDiff: Partial<DeviceStatus>): Promise<void> {
+	const conf = await config.getMany([
+		'deviceApiKey',
+		'apiTimeout',
+		'apiEndpoint',
+		'uuid',
+		'localMode',
+	]);
 	let body = stateDiff;
 	const { apiEndpoint, apiTimeout, deviceApiKey, localMode, uuid } = conf;
 	if (localMode) {
@@ -104,14 +109,20 @@ const report = async ({ stateDiff, conf }: StateReport) => {
 			headers['retry-after'] ? parseInt(headers['retry-after'], 10) : undefined,
 		);
 	}
-};
+}
 
-const getStateDiff = (): DeviceStatus => {
-	const lastReportedLocal = lastReportedState.local;
-	const lastReportedDependent = lastReportedState.dependent;
-	if (lastReportedLocal == null || lastReportedDependent == null) {
+/**
+ * Calculates difference in 2 DeviceStatus objects
+ *
+ * getStateDiff omits values in newState where value is present in previousState
+ */
+function getStateDiff(
+	previousState: DeviceStatus,
+	newState: DeviceStatus,
+): Partial<DeviceStatus> {
+	if (previousState.local == null || previousState.dependent == null) {
 		throw new InternalInconsistencyError(
-			`No local or dependent component of lastReportedLocal in CurrentState.getStateDiff: ${JSON.stringify(
+			`No local or dependent component of previousState in CurrentState.getStateDiff: ${JSON.stringify(
 				lastReportedState,
 			)}`,
 		);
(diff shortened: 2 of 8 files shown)


docs/configurations.md
```diff
@@ -25,6 +25,7 @@ This list contains configuration variables that can be used with all balena devi
 | BALENA_HOST_DISCOVERABILITY | boolean | false | true | Enable / Disable Avahi to run which will allow the device to respond to requests such as network scans. | v11.9.2 |
 | BALENA_HOST_SPLASH_IMAGE | integer | true | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
 | BALENA_SUPERVISOR_HARDWARE_METRICS | boolean | false | true | Toggle hardware metrics reporting to the cloud, which occurs every 10 seconds when there are changes. Metrics include CPU utilization, CPU temperature, memory usage, and disk space. Useful for devices with bandwidth sensitivity. | v12.8.0 |
+| BALENA_SUPERVISOR_PATCH_INTERVAL | integer | false | 60000 | Define the frequency of updating the cloud with device's statel in milliseconds. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v12.12.0 |
```
20k-ultra, line 28: Will have to confirm what the default should be.

src/device-state/current-state.ts
```diff
@@ -2,55 +2,47 @@ import * as _ from 'lodash';
 import * as url from 'url';
 import { CoreOptions } from 'request';
 
-import * as constants from '../lib/constants';
+import { backoffPromise } from '../lib/backoff';
```
20k-ultra, line 5: Moved backoff to it's own module since it could be useful for other areas of the codebase (like adding it to target-state fetching since it uses its own mechanism)

src/device-state/current-state.ts
```diff
@@ -2,55 +2,47 @@ import * as _ from 'lodash';
...
 import { log } from '../lib/supervisor-console';
 import { InternalInconsistencyError, StatusError } from '../lib/errors';
 import { getRequestInstance } from '../lib/request';
 import * as sysInfo from '../lib/system-info';
+import sleep from '../lib/sleep';
```
20k-ultra, line 10: Creating this instead of using bluebird's delay. I will make a commit to replace all delays with our sleep.

src/device-state/current-state.ts
```diff
@@ -2,55 +2,47 @@ import * as _ from 'lodash';
...
-		| 'localMode'
-		| 'hardwareMetrics'
-	>]: SchemaReturn<key>;
-};
+/**
+ * patchInterval is set when startReporting successfuly queries the config
+ */
+let patchInterval: number;
 
-type StateReport = {
```
20k-ultra, line 50: Keeping a local copy of state for report didn't make sense because it would become outdated theoretically right away if something changes. The other types didn't seem necessary either.

src/device-state/current-state.ts
```diff
@@ -104,170 +109,149 @@ const report = async ({ stateDiff, conf }: StateReport) => {
...
-	// But check once every max report frequency to ensure that changes in system
-	// info are picked up (CPU temp etc)
-	setInterval(doReport, constants.maxReportFrequency);
-	// Try to perform a report right away
-	return doReport();
+export const startReporting = async (): Promise<void> => {
+	// Set config values we need to avoid multiple db hits
+	patchInterval = await config.get('statePatchInterval');
+	// All config values fetch, start reporting!
+	return reportHandler();
```
20k-ultra, line 256: Might be worth discussing the consequences of removing the device state change listener which provides more immediate updates to the dashboard.

20k-ultra (commented):
I mostly just moved existing logic/code into functions and called those. I think the only significant logical change is removing the device state change event listener. I will be pushing more commits with the ability to control device metric and download progress patching intervals.

src/device-state/current-state.ts
```diff
@@ -2,55 +2,47 @@ import * as _ from 'lodash';
...
 	'update_failed',
 ];
 
 export let stateReportErrors = 0;
+
 const lastReportedState: DeviceStatus = {
 	local: {},
 	dependent: {},
 };
-const stateForReport: DeviceStatus = {
```
20k-ultra, line 31: Keeping a local copy of state for report didn't make sense because it would become outdated theoretically right away if something changes. The other types didn't seem necessary either.

docs/configurations.md
pipex, line 28: Can we call this `BALENA_SUPERVISOR_REPORT_INTERVAL`? Patch seems too specific. Also I'm wondering where did you get 60s? It used to be 10s?

src/device-state/current-state.ts
```diff
@@ -2,55 +2,47 @@ import * as _ from 'lodash';
...
-	stateDiff: DeviceStatus;
-	conf: Omit<CurrentStateReportConf, 'deviceId' | 'hardwareMetrics'>;
-};
+/**
+ * Listen for config changes to statePatchInterval
+ */
+(async () => {
+	await config.initialized;
+	config.on('change', (changedConfig) => {
+		if (changedConfig.statePatchInterval) {
```
pipex, line 41: I don't see the advantage of doing this instead of doing `config.get('statePatchInterval')` at the beginning of the report handler

src/device-state/current-state.ts
```diff
@@ -64,7 +56,20 @@ const stripDeviceStateInLocalMode = (state: DeviceStatus): DeviceStatus => {
...
 
-const report = async ({ stateDiff, conf }: StateReport) => {
+/**
+ * Try to patch difference in device state
+ *
+ * report attempts to patch the difference in device state from the last
+ * report successfully sent.
+ */
+async function report(stateDiff: Partial<DeviceStatus>): Promise<void> {
+	const conf = await config.getMany([
```
pipex, line 66: I sort of liked that this function was unaware of the `config` module and it just got the configs as parameter. I think it makes it cleaner but that's just an opinion really

docs/configurations.md
pipex: Sorry, just saw your comment. Ignore the part about the interval.

pipex, line 28:
```suggestion
| BALENA_SUPERVISOR_PATCH_INTERVAL | integer | false | 60000 | Define the frequency of updating the cloud with device's state in milliseconds. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v12.12.0 |
```

src/device-state/current-state.ts
```diff
@@ -104,170 +109,149 @@ const report = async ({ stateDiff, conf }: StateReport) => {
...
+		// Check if it was an HTTP status error
+		if (e instanceof StatusError) {
+			// Supervisor got an HTTP error from API
+			log.error(
+				`Device state report failure! Status code: ${e.statusCode} - message:`,
+				e.message,
+			);
+			// Use retryAfter as maxDelay if the header is present
+			await backoffPromise(
+				reportHandler,
```
pipex, line 219: I think you might be missing a `return` statement here. If you don't then the report will be called at least twice on the iteration. This also creates a branching point where a parallel recursion is started leading to a memory leak.

src/device-state/current-state.ts
```diff
@@ -104,170 +109,149 @@ const report = async ({ stateDiff, conf }: StateReport) => {
...
-					eventTracker.track('Device state report failure', {
-						error: e.message,
-					});
-					backoff(reportCurrentState, stateReportErrors++, maxDelay);
-				}
-			},
-		);
-	})();
+	// Begin report loop again with 0 attempts
+	return reportHandler(0);
```
pipex, line 249: Do you know if node performs some cleanup of the stack on this type of call? This is technically an infinite recursion over the lifetime of the supervisor, unless node performs some cleanup, eventually you will run out of call stack space right?

src/lib/backoff.ts
```diff
@@ -0,0 +1,37 @@
...
+	setTimeout(() => retry(attempts + 1), delay);
+};
+
+/**
+ * Perform exponential backoff on an async function, increasing the attempts
+ * counter on each call
+ *
+ * If attempts is 0 then, it will delay for min(minDelay, maxDelay)
+ */
+export async function backoffPromise(
```
pipex, line 27: I don't see the point of having both functions, they both do exactly the same thing. We should go with the promise version which is the right way we want to do this type of operation, what do you think?

src/device-state/current-state.ts
Page-: cc @toochevere

src/lib/sleep.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Delays for input ms before resolving
+ */
+export default async function sleep(ms: number): Promise<void> {
+	return new Promise((resolve) => setTimeout(resolve, ms));
+}
```
Page-, line 6: Just use the built in promisified `setTimeout`
```suggestion
import { promisify } from 'util';

const sleep = promisify(setTimeout);
export default sleep;
```
And if you update to nodejs 16 you can do
```suggestion
export { setTimeout as sleep } from 'timers/promises';

Comment (Page-) [reply src/lib/sleep.ts:6]: Also I'd probably name it just `setTimeout` since it's less confusing when using the native apis

File: src/lib/backoff.ts (hunk 11 above)
Comment (Page-) [reply src/lib/backoff.ts:27]: I agree, if we did want to keep both versions we could just do:
```ts
export const backoff = (
	...args: Parameters<typeof backoffPromise>
) => {
	backoffPromise(...args);
};
```
and have the same result with less code

src/device-state/current-state.ts
Page-: I'd definitely lean towards keeping the immediate report check, if the device hasn't reported anything recently then showing it's started updating asap makes sense imo. In general the behavior should be one of reporting at most once every X seconds rather than only reporting when hitting the X seconds interval. It's a somewhat minor difference but keeping that first update happening immediately if it's otherwise quiet can help the UX a lot imo without adding additional load

src/device-state/current-state.ts
Page-: It doesn't seem necessary but it does predate this separated reporting module even so I'd be a little wary of changing it without ensuring it's safe and testing it. I think it makes sense to remove if it's unnecessary but I'd lean towards delaying it for the moment since if it does break something it's likely to do it in a way that puts unnecessary load on the API, and removing it is not urgent/necessary other than to clean up the code

src/device-state/current-state.ts
```diff
@@ -104,170 +109,149 @@ const report = async ({ stateDiff, conf }: StateReport) => {
...
+ * Start reporting current state
+ *
+ * reportHandler will initialize a loop that trys to patch the current
+ * state using statePatchInterval as the delay between attempts.
+ * If there is an issue with the network request the loop will exponentially
+ * back off until successful then reset interval to original time.
+ */
+export async function reportHandler(attempts = 0): Promise<void> {
+	// Generate latest report repesenting the device's current state
+	const stateForReport = await generateReport();
```
Page-, line 201: This could in theory fail and we'd never report again

src/device-state/current-state.ts
Page-: I agree, it also opened up the possibility of only querying for these values on startup and then reacting to them changing rather than querying them on every single report attempt (a bunch/all of them hit the disk right? and could fail/be slow due to that if things are busy)

src/device-state/current-state.ts
```diff
@@ -104,170 +109,149 @@ const report = async ({ stateDiff, conf }: StateReport) => {
...
+	// Calculate differences between current state and last reported state
+	const stateDiff = getStateDiff(lastReportedState, stateForReport);
+	// Try to send report now
+	try {
+		await report(stateDiff);
+	} catch (e) {
+		// Use appUpdatePollInterval as maxDelay since we encountered an error and will backoff now
+		const maxDelay = await config.get('appUpdatePollInterval');
```
Page-, line 209: If this fails for any reason we'd never report again

src/device-state/current-state.ts
Page-: I think because it's a promise (via async/await) that it'd reset the stack regularly and so it would be ok, but also there's no benefit to returning it so just removing the `return` avoids the issue altogether

Page- (commented):
I'd say there's a lot of additional changes/refactoring beyond just the report frequency config and at least some of them make the state reporting far more brittle and likely to fail and have the device never report again - I'd highly appreciate if the refactorings could be split out to separate PRs to get the reviews they deserve and so the headline change of this PR can also be more easily/thoroughly reviewed in isolation

20k-ultra:
Superseded by https://github.com/balena-os/balena-supervisor/pull/1848 because this PR introduced too many changes

PR #1844, opened 2021-12-09
Separate backoff from current state reporting

Page-:
Change-type: patch

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -13,6 +13,7 @@ import * as config from '../config';
 import { SchemaTypeKey, SchemaReturn } from '../config/schema-type';
 import * as eventTracker from '../event-tracker';
 import * as deviceState from '../device-state';
+import { backoff } from '../lib/backoff';
 
 // The exponential backoff starts at 15s
 const MINIMUM_BACKOFF_DELAY = 15000;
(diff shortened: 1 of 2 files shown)


src/device-state/current-state.ts
```diff
@@ -244,12 +228,19 @@ function reportCurrentState(attempts = 0) {
...
 					eventTracker.track('Device state report failure', {
 						error: e.message,
 					});
-					backoff(reportCurrentState, stateReportErrors++, maxDelay);
+					backoff(
+						reportCurrentState,
+						stateReportErrors++,
+						maxDelay,
+						MINIMUM_BACKOFF_DELAY,
+						'Device state report',
```
Page-, line 242: I added this bit so the log message is more useful when used in more places

src/device-state/current-state.ts
```diff
@@ -244,12 +228,19 @@ function reportCurrentState(attempts = 0) {
 						maxDelay,
 						// Start the polling at the value given by the API if any
 						e.retryAfter ?? MINIMUM_BACKOFF_DELAY,
```
Page-, line 230: I don't like this as it's a bit of a hack really and is working around the backoff api

src/lib/backoff.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export const backoff = (
+	retry: (attempts: number) => void,
+	attempts = 0,
+	maxDelay: number,
+	minDelay: number,
+	name: string,
+) => {
+	const delay = Math.min(2 ** attempts * minDelay, maxDelay);
+	log.info(`${name}: Retrying in ${delay / 1000} seconds`);
```
pipex, line 17: I'm not sure is that important right now, but I think it would be better if we passed something like an `onError` callback to this function, so this module would be independent of the `supervisor-console` module.

I we did that we could call this function as

```
backoff(
  reportCurrentState,
  stateReportErrors++,
  maxDelay,
  MINIMUM_BACKOFF_DELAY,
 (delay). => `Retrying in ${delay/1000} seconds`)
```

which would make it a bit more generic

Page-: I agree

src/device-state/current-state.ts
pipex: see my comment below on how it maybe this could be a callback instead of a string

src/device-state/current-state.ts
pipex: I changed the backoff function which leaves this bit like this. Better, worst?

20k-ultra:
Closing since https://github.com/balena-os/balena-supervisor/pull/1844 implements backoff module

PR #1845, opened 2021-12-09
Add exponential backoff on errors for logs reporting

Page-:
Change-type: patch

### DIFF
--- a/src/logging/balena-backend.ts
+++ b/src/logging/balena-backend.ts
@@ -10,7 +10,8 @@ import { LogBackend, LogMessage } from './log-backend';
 import log from '../lib/supervisor-console';
 
 const ZLIB_TIMEOUT = 100;
-const COOLDOWN_PERIOD = 5 * 1000;
+const MIN_COOLDOWN_PERIOD = 5 * 1000; // 5 seconds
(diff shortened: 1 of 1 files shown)


src/logging/balena-backend.ts
```diff
@@ -118,53 +119,83 @@ export class BalenaLogBackend extends LogBackend {
...
+					'LogBackend: server responded with status code:',
+					res.statusCode,
+				);
+				setupFailed();
+			});
+
+			this.req.on('timeout', setupFailed);
+			this.req.on('close', setupFailed);
+			this.req.on('error', (err) => {
+				log.error('LogBackend: unexpected error:', err);
```
pipex, line 167: I prefer to avoid dumping error stacks in logs as it gets noisy on support
```suggestion
				log.error(`LogBackend: unexpected error: "${err.message}"`);
```

Page-: I'd prefer to leave this as-is (it's unchanged except for an extra indentation level). I think it should be pretty rare and in the cases it does happen I expect we'll be happy to have more info rather than less, although if you have seen this being too noisy I think it'd be fine having it ignore expected errors

Page-:
@balena-ci I self-certify!

PR #1847, opened 2021-12-10
Remove apps.json after initial preload

pipex:
This avoids the supervisor trying to get back to the preloaded target
state if the database is deleted by mistake. It does this by moving the
used apps.json to a backup location.

Change-type: patch
Depends-on: #1841


src/device-state/preload.ts
```diff
@@ -8,21 +9,21 @@ import * as deviceConfig from '../device-config';
...
-	appsPath: Nullable<string>,
-): Promise<void> {
-	log.info('Attempting to load any preloaded applications');
-	if (!appsPath) {
-		appsPath = constants.appsJsonPath;
-	}
+export function appsJsonBackup(appsPath: string) {
+	const [dirname, basename] = [path.dirname(appsPath), path.basename(appsPath)];
+
+	return path.join(dirname, `preloaded-${basename}`);
```
20k-ultra, line 22: Consider adding extension to file instead of changing the name. The logic is because when we make backups we do ${file}.backup. This doesn't really matter though.

```suggestion
	return path.join(dirname, `${basename}.preloaded`);
```

src/device-state/preload.ts
```diff
@@ -114,5 +115,23 @@ export async function loadTargetFromFile(
...
+		const targetPath = appsJsonBackup(appsPath);
+		if (!(await fsUtils.exists(targetPath))) {
+			// Try to rename the path so the preload target state won't
+			// be used again if the database gets deleted for any reason.
+			// If the target file already exists or something fails, just debug
+			// the failure.
+			await fsUtils
+				.safeRename(appsPath, targetPath)
+				.then(() => fsUtils.writeFileAtomic(appsPath, '{}'))
+				.then(() => log.debug(`Removed old apps.json`))
```
20k-ultra, line 128:
```suggestion
				.then(() => log.debug(`Migrated previous apps.json`))
```

src/device-state/preload.ts
```diff
@@ -114,5 +115,23 @@ export async function loadTargetFromFile(
 				error: e,
 			});
 		}
+	} finally {
+		const targetPath = appsJsonBackup(appsPath);
+		if (!(await fsUtils.exists(targetPath))) {
```
20k-ultra, line 120: I acknowledge that this current code could still exhibit the issue it tries to mitigate if another apps.json file appears since it won't create a backup if a backup already exists. 

I can't find anything in our docs about apps.json but I believe this file is only created when you preload a device so it would be really weird if someone preloaded a device again after it has ran making this issue not that big of a deal. Is this correct?

20k-ultra (commented):
If a preloaded apps.json already exists then

src/device-state/preload.ts
pipex: Is not really a problem, as you cannot preload a device, you preload an image which then you flash into a device, so modification of apps.json should be a one time operation.

src/device-state/preload.ts
pipex: The only reason I did it this way was to keep the correct extension in case you want to read the file, but that may not be necessary really.

20k-ultra: Do we expect people to interact with this file ? Otherwise, a "file extension" isn't meaningful here like on other systems such as Windows where it the OS will behave differently based on the extension.

src/device-state/preload.ts
20k-ultra: great! no further thought needed then. Thanks for clarifying.

src/device-state/preload.ts
pipex: I would expect to be useful for us to read it while on support. I was under the impression that `vi` on device would use the extension for syntax highlighting but I tested and it doesn't do syntax highlighting at all. 

I'll make the change

test/05-device-state.spec.ts
```diff
@@ -283,14 +289,22 @@ describe('deviceState', () => {
 			.to.have.property('labels')
 			.that.has.property('io.balena.something')
 			.that.equals('bar');
+
+		// Restore renamed apps.json
+		await fsUtils.safeRename(appsJsonBackup(appsJson), appsJson);
```
cywang117, line 294: Is it possible to use mock-fs for these fs tests since we are trying to incrementally move all tests over to mock-fs anyway?

src/device-state/preload.ts
20k-ultra: that's a good point my bad!

20k-ultra: for me the .preloaded does prevent my vim (not vi) from identifying the filetype as JS

test/05-device-state.spec.ts
pipex: I think I tried at some point for the v3 PR for this whole suite, but there were too many dependencies. It might be possible to do it just for preloaded files.

cywang117: That would be great if it's not too big a change!

pipex: I remembered why I ended up not using mockfs. The `device-state` module has too many dependencies on config files, meaning if I cannot just mock `apps.json`, I have to mock `config.json`, `/etc/hostname`, etc.

Otherwise I start getting stuff like
```
 1) deviceState
       loads a target state from an apps.json file and saves it as target state, then returns it:
     Error: ENOENT, no such file or directory 'test/data/etc/hostname'
      at Binding.<anonymous> (node_modules/mock-fs/lib/binding.js:571:15)
```

pipex:
@balena-ci rebase

cywang117:
LGTM @pipex, self-certify if you're ready

pipex:
@balena-ci I self certify!

PR #1848, opened 2021-12-10
Added `SUPERVISOR_REPORT_INTERVAL` config var

20k-ultra:
Tried to make the logical changes as small as possible but by loading a interval value that can be updated at runtime we had to modify the throttle and and interval timing which ran the report.


src/lib/throttle.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Creates a curried function for throttling a function
+ *
+ * default returns a instance of lodash's throttle function with a function and wait time set.
+ * This throttle handles promise rejections better then a regular lodash throttle by allowing
+ * the caller to handle the exception instead of lodash.
+ */
+export default function throttle(
+	fn: (opts: any) => Promise<void>,
```
20k-ultra, line 11: I'd like to think about how to make the fn arguments more generic

pipex: You could use generics and all type checks should work
```
export default function throttle<T>(
	fn: (opts: T) => Promise<void>,
```

cywang117: Having this function in lib/throttle seems to imply that it should be used over directly using lodash's throttle. What happens if you pass a function to throttle that's not a Promise?

docs/configurations.md
```diff
@@ -24,7 +24,8 @@ This list contains configuration variables that can be used with all balena devi
...
 | BALENA_HOST_SPLASH_IMAGE | integer | true | /boot/splash/balena-logo-default.png | Allows changing splash screen on boot to user defined file from userspace. | v12.3.0 |
-| BALENA_SUPERVISOR_HARDWARE_METRICS | boolean | false | true | Toggle hardware metrics reporting to the cloud, which occurs every 10 seconds when there are changes. Metrics include CPU utilization, CPU temperature, memory usage, and disk space. Useful for devices with bandwidth sensitivity. | v12.8.0 |
+| BALENA_SUPERVISOR_HARDWARE_METRICS | boolean | false | true | Toggle hardware metrics reporting to the cloud, which occurs every 10 seconds when there are changes. Metrics include CPU utilization, CPU temperature, memory usage, and disk space. Useful for devices with bandwidth sensitivity. | v12.8.0 | 
+| BALENA_SUPERVISOR_REPORT_INTERVAL | integer | false | 10000 | Define the frequency of updating the cloud with device's state in milliseconds. The minimum value for this variable is defined by the balenaCloud backend, and may vary. | v12.12.0 |
```
cywang117, line 28: open-balena-api PR for this next?

src/lib/throttle.ts
Page-: If you want to nicely inherit the args of the function regardless of what they are you'd be better off with something like
```ts
export default function throttle<T extends (...args: any[]) => Promise<void>>(
	fn: T,
	wait: number,
): (
	args: Parameters<T>,
	resolve: () => void,
	reject: (e: Error) => void,
) => void {
	return _throttle(
		(
			args: Parameters<T>,
			resolve: () => void,
			reject: (e: Error) => void,
		) => {
			fn(...args)
				.then(resolve)
				.catch(reject);
		},
		wait,
	);
}
```

src/lib/throttle.ts
```diff
@@ -0,0 +1,20 @@
+import { throttle as _throttle } from 'lodash';
```
Page-, line 1: The `_` prefix in typescript means to not warn if it's unused so it's better not to use it except for that specific case:
```suggestion
import { throttle as $throttle } from 'lodash';
```

docs/configurations.md
20k-ultra: of course

src/device-state/current-state.ts
```diff
@@ -256,7 +255,38 @@ function reportCurrentState(attempts = 0) {
...
+			// Stop current report interval
+			clearInterval(reportLoop);
+			// Start interval with new time
+			reportLoop = setInterval(doReport, reportInterval);
+		}
+	});
+
+	try {
+		reportInterval = await config.get('stateReportInterval');
+	} catch {
```
pipex, line 280: I would not catch the `config.get`. I would expect if something fails during `config.get` it should return the default instead of throwing an exception.

20k-ultra: I don't like the try/catch also, I just copied how Target state fetch is done to work towards both mechanisms behaving the same.

20k-ultra:
Moved PR to draft because I am going to do some more work on how throttle is handled.

20k-ultra:
will implement from scratch again since https://github.com/balena-os/balena-supervisor/pull/1888 has refactored a lot

PR #1849, opened 2021-12-13
Write SBVR model for Supervisor

cywang117:
Mainly for Supervisor v3 API. Expect this model to change.


src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+Term: update lock
+	Concept Type: Short Text (Type)
+	-- Path of update lockfile (Short Text (Type))?
+	-- Or integer written into lockfile (Integer (Type))?
+
+	Fact type: application has update lock
+		Necessity: each application has at most one update lock.
+
+	Fact type: service has update lock
+		Necessity: each service has at most one update lock.
```
cywang117, line 368: There must be a better way to translate the concept of "update locks can be on an application level or a service level" into SBVR

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+
+		Fact type: reboot is force
+
+	Term: regenerate api key
+	-- POST /v1/regenerate-api-key
+
+		Synonym: regenerate api secret
+		-- The current name is regenerate api key but "api key" is an ambiguous term since balenaAPI also has api keys.
+		-- We should use api secret which is the database table name as well as the name for the SV's api key in the open-balena-api model
+		Concept Type: action
```
cywang117, line 413: I haven't expanded more on the concept of scoped API secrets in this model, because I haven't had a chance to think about what a more ideal API key model could be for the Supervisor.

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+
+Term: application
+
+	Fact type: application has slug
+	-- Not currently provided to Supervisor, optional, but would be nice for LocalUI
+		Necessity: each application has exactly one slug
+		Necessity: each slug is of exactly one application
+
+	Fact type: application has api port
+		Necessity: each application that has a slug that is equal to "balena-os/balena-supervisor", has exactly one api port that is equal to 48484.
```
cywang117, line 285: Hardcoded port 48484 for now

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+Term: download progress
+	Concept Type: Integer (Type)
+
+Term: ip address
+	Concept Type: Short Text (Type)
+
+Term: name
+	Concept Type: Short Text (Type)
+
+Term: port
```
pipex, line 16: What is this port?

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+	Concept Type: Integer (Type)
+
+Term: slug
+	Concept Type: Short Text (Type)
+
+Term: status
+	Concept Type: Short Text (Type)
+
+Term: version
```
pipex, line 25: Same, is this meant to be a generic version term? Or is it a specific version (e.g. for the supervisor)

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+	Fact type: device has status
+		Necessity: each device has exactly one status.
+		Definition: "Provisioning" or "Post-Provisioning" or "Stopping" or "Starting" or "Downloading" or "Installing" or "Idle" 
+		-- TODO: these are partially taken from v2 API reference which may not be ideal. Decide on status strings for v3.
+
+	Fact type: device has name
+		Necessity: each device has exactly one name.
+
+	Fact type: device has version
+		-- OS version
```
pipex, line 104: I would maybe remove this, as we probably want to think about the future model where the OS is another app and the version belongs to the release of that app. Same for the os variant below

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+Term: version
+	Concept Type: Short Text (Type)
+
+Term: uuid
+	Concept Type: Text (Type)
+
+
+
+-- HOST CONFIG PROXY
+-- Discuss: we may not want to incorporate this into the Supervisor model
```
pipex, line 34: I think we should definitely incorporate this in the future, but it might not be needed for a proof of concept.

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+
+	Fact type: image is stored at image location
+		Synonymous Form: image location hosts image
+		Necessity: each image is stored at exactly one image location.
+		Necessity: each image location hosts exactly one image.
+
+	Fact type: image has content hash
+		Necessity: each image has at most one content hash.
+
+	Fact type: image is downloaded by device
```
pipex, line 215: This is confusing for me, I think for the device we need to distinguish between target images (i.e. the description of an image hosted on the cloud), and current images (what's installed on the device right now). I'm not sure how to solve that with the model, but I think is probably something we need to investigate. 

The same applies for all entities, app, release, etc.

pipex: The relationship should probably be between app and device, app has a target release and release has images (or something like that)

src/supervisor.sbvr
pipex: Why do we need this

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+	Fact type: application has api port
+		Necessity: each application that has a slug that is equal to "balena-os/balena-supervisor", has exactly one api port that is equal to 48484.
+
+	Fact type: application has name
+	-- Not currently provided to Supervisor, optional, but would be nice for LocalUI
+		Necessity: each application has exactly one name.
+		Necessity: each application has an app name that has a Length (Type) that is greater than or equal to 4 and is less than or equal to 100.
+
+	Fact type: application has uuid
+	-- Optional: wait for v3 target state?
```
pipex, line 293: Not optional :p

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+
+Term: config var
+	Concept Type: Short Text (Type)
+
+	Fact type: config var has value
+		Necessity: each config var has exactly one value.
+
+	Fact type: application owns config var
+		Synonymous Form: config var belongs to application
+		Necessity: each application that has a config var must have a slug equal to "balena-os/balena-supervisor".
```
pipex, line 343: I would like to see if we can get rid of config vars at least for the supervisor for now

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+	Fact type: application has download progress
+	-- Maybe should be computed term of all image download progresses?
+		Necessity: each application has at most one download progress.
+
+	Fact type: application should be running release
+		Synonymous Form: release should be running on application
+		Necessity: each application should be running at most one release.
+
+	Fact type: application has version
+		Necessity: each application that has a slug that is equal to "balena-os/balena-supervisor" has exactly one version.
```
pipex, line 314: Isn't version a property of the release? Why only the supervisor needs to have an app version ?

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+	-- Optional: wait for v3 target state?
+		Necessity: each application has exactly one uuid.
+		Necessity: each uuid is of exactly one application.
+		Necessity: each application has a uuid that has a Length (Type) that is equal to 32.
+
+	Fact type: application is host
+
+	Fact type: application is healthy
+	-- GET /v1/healthy
+		Necessity: each application that has a slug that is equal to "balena-os/balena-supervisor" is healthy or not healthy.
```
pipex, line 302: All applications should have a health status no? In any case maybe health is a property of a service install and the application health is a computed term. We don't need to think about that for a proof of concept though

src/supervisor.sbvr
pipex: Do we need an API for update locks?

src/supervisor.sbvr
```diff
@@ -0,0 +1,461 @@
...
+	Fact type: application has update lock
+		Necessity: each application has at most one update lock.
+
+	Fact type: service has update lock
+		Necessity: each service has at most one update lock.
+
+
+
+-- ACTION: TODO
+-- Since this is theoretical, this probably shouldn't initially be included in the model.
```
pipex, line 373: Yes, I think we should take it out, let's see how far we can get with data first

src/supervisor.sbvr
cywang117: I added this for the actions section of the SBVR file, but I've since commented out actions so this shouldn't be necessary anymore.

src/supervisor.sbvr
cywang117: Done

src/supervisor.sbvr
cywang117: Removed it

src/supervisor.sbvr
cywang117: Removed this section and anything else related to Supervisor API keys (API secret, not deviceApiKey) for the time being

src/supervisor.sbvr
cywang117: Moved this to a separate file for now that's not used in the proof of concept.

src/supervisor.sbvr
cywang117: Updated to reflect this

src/supervisor.sbvr
cywang117: Removed the config var section

src/supervisor.sbvr
cywang117: Fixed, removed this Fact and included the following:
```
Fact type: release has version
		Necessity: each release has at most one version.
```

model/sbvr/supervisor.sbvr
```diff
@@ -0,0 +1,298 @@
...
+	Fact type: device has hostname
+		Necessity: each device has exactly one hostname.
+
+	Fact type: device is connected to vpn
+
+
+
+-- DEVICE METRICS
+
+Term: cpu id
```
pipex, line 67: This might make more sense on the device model rather than the device metrics model as it is a constant.

model/sbvr/supervisor.sbvr
```diff
@@ -0,0 +1,298 @@
...
+	Fact type: service instance has created at timestamp
+		Necessity: each service instance has exactly one created at timestamp.
+
+
+-- COMMIT
+
+Term: commit
+	Concept Type: Short Text (Type)
+
+	Fact type: commit has uuid
```
pipex, line 159: Will this create a commit table? Should commit just be a synonym for uuid (or something like that?)

model/sbvr/supervisor.sbvr
```diff
@@ -0,0 +1,298 @@
...
+
+	Fact type: release has version
+		Necessity: each release has at most one version.
+
+	Fact type: release has commit
+		Synonymous Form: commit belongs to release
+		Necessity: each release has exactly one commit.
+		Necessity: each commit belongs to exactly one release.
+
+	Fact type: device should be operated by release
```
pipex, line 181: We don't have to change this right now, but I think we need a `device is being operated by release` and a `device is being managed by release` to identify the current state of the host app and the supervisor. Since we have the same pattern 3 times, we might as well look up a way to make that a generic multi-app relation. Just something to think about

Issue #1758, opened 2021-07-26
Update locks created by the Supervisor not being removed


20k-ultra:
I have seen at least twice now where a device is not updating its release because of update locks but after asking the user if they are using update locks they say no.

We know the Supervisor can create locks itself when apply updating to prevent other processes from changing state during the funnel so something is happening which causes those to exist. 

When does the Supervisor create lock files ?
When does the Supervisor remove the lock files ?
Is there any mechanism to mitigate being stuck from its own lock files ?

cywang117:
^ From that ticket, the lockfiles were for two application services, but the user stated that they don't create lockfiles. I looked in the logs but it looks like logs have rotated since update lock creation. The Supervisor version is 12.7.0 so I wonder if it had anything to do with https://github.com/balena-os/balena-supervisor/pull/1663 which I PR'd (v12.6.1)..

shawaj:
We have seen this many many times on our fleets. 

https://github.com/NebraLtd/helium-miner-software/issues/103

Is there any progress on a fix?

cywang117:
Per internal discussions, this is current proposed solution:

- Writing to the file might not work unless it can be done in an atomic way.
- Setting the UID ownership on the lockfile is a preferred solution. This could be achieved by running the supervisor with specific UID (but still privileged), or having the lock be created by a separate user within the supervisor container.
- The supervisor can then perform cleanup on startup and remove lockfiles that have the specified UID.

cywang117:
A fix for this issue is available with Supervisor v13.1.0.

PR #1851, opened 2021-12-15
Clean leftover locks after supervisor crash

pipex:
This will prevent the supervisor being unable to update target state
after a crash due to leftover locks

Change-type: patch
Closes: #1758


pipex:
Needs testing

pipex:
The fix will be performed by #1893

PR #1858, opened 2021-12-22
Added circleci config to publish v12.11.16

20k-ultra:
using this to trigger circleci to publish images to docker hub


20k-ultra:
@balena-ci retest

PR #1863, opened 2021-12-30
Add troubleshooting notice for macOS

fisehara:
When developing on macOS (Darwin) DBus may have to be installed.

Change-type: patch
Signed-off-by: fisehara <harald@balena.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,8 +274,12 @@ node -v       # >= 12.16.2
 npm -v        # >= 6.14.4
 git --version # >= 2.13.0
 ```
-
 Also, ensure you're installing dependencies with `npm ci` as this will perform a clean install and guarantee the module versions specified are downloaded rather then installed which might attempt to upgrade!
+
+### DBus
+When developing on macOS you may need to install DBus on the development host.
+1. `brew install dbus`
+2. `npm ci`
 
 #### Downgrading versions
 


20k-ultra:
You are running the Supervisor on a machine with OS X ? The normal development workflow is to livepush the Supervisor to a device running balenaOS so developing on the machine that also has the git repository is new to me. Even in that situation we run the Supervisor using docker in docker (run balenaOS inside docker)

fisehara:
The npm install is failing when I have not installed the dbus package on macos.

cywang117:
@fisehara We could probably make the docs section more generic by adding what libraries are needed for different OS's. I know I definitely encountered the `dbus-1 not found` error when first trying to set up the repo:
- Debian: `sudo apt-get install libdbus-1-dev`
- CentOS: `sudo yum install dbus-devel`
- Arch: Maybe @20k-ultra knows?
- MacOS: `brew install dbus`
- WSL: Same as Linux comands above, depends on distro installed

20k-ultra:
installing all these dependencies just to develop the Supervisor sucks...I dislike that just to get the types for software you'll never run you have to install it....I wonder if there's another way.

pipex:
@balena-ci rebase

pipex (approved):

PR #1864, opened 2022-01-06
Use dmidecode to read cpuid in non ARM devices

pipex:
Cpu id is set to null so far for non ARM devices (e.g. Intel NUC). This
parses the output of dmidecode to get the cpu id and system model.

Change-type: patch

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -33,6 +33,7 @@ RUN apk add --no-cache \
 	libuv \
 	sqlite-libs \
 	sqlite-dev \
+	dmidecode \
 	dbus-dev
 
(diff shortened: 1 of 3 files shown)


pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
@balena-ci I self certify!

pipex:
I merged this without realizing I had pushed two commits with the same name (instead of doing `--force-with-lease`). For that reason the change log for this PR shows a repeated line. The version on the latest commit passes tests and works on devices, so it should not cause any issues.

PR #1866, opened 2022-01-17
Fix memoization of registry token request

Page-:
Change-type: patch

### DIFF
--- a/src/lib/docker-utils.ts
+++ b/src/lib/docker-utils.ts
@@ -367,13 +367,27 @@ export async function isV2DeltaImage(imageName: string): Promise<boolean> {
 	return inspect.Size < 40 && inspect.VirtualSize < 40;
 }
 
-const getAuthToken = memoizee(
-	async (
+const getAuthToken = (() => {
+	const memoizedGetToken = memoizee(
+		async (tokenUrl: string, tokenOpts) => {
+			const tokenResponseBody = (
+				await (await request.getRequestInstance()).getAsync(tokenUrl, tokenOpts)
+			)[1];
+			const token = tokenResponseBody?.token;
+
+			if (token == null) {
+				throw new ImageAuthenticationError('Authentication error');
+			}
+
+			return token;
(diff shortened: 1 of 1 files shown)


src/lib/docker-utils.ts
```diff
@@ -367,13 +367,27 @@ export async function isV2DeltaImage(imageName: string): Promise<boolean> {
...
+			)[1];
+			const token = tokenResponseBody?.token;
+
+			if (token == null) {
+				throw new ImageAuthenticationError('Authentication error');
+			}
+
+			return token;
+		},
+		{ maxAge: DELTA_TOKEN_TIMEOUT, promise: true, primitive: true },
```
Page-, line 384: The really key thing here is that `primitive` (or a custom resolver) must be used when the parameters contain objects that you want to cache by content rather than by object identity. The problem with object identity being that it requires them to be the same object instantiation and as a bonus it'll also ignore if the contents have changed, eg
```
let x = { foo: 'bar' }
let y = { foo: 'bar' }
memoizedFn(x)
memoizedFn(y) // Will not match memoization for the previous `{ foo: 'bar' }`
x.foo = 'baz'
memoizedFn(x) // Will match the memoization for `{ foo: 'bar' }` even though it's now `{ foo: 'baz' }`

File: src/lib/docker-utils.ts [hunk 2]
```diff
@@ -367,13 +367,27 @@ export async function isV2DeltaImage(imageName: string): Promise<boolean> {
 	return inspect.Size < 40 && inspect.VirtualSize < 40;
 }
 
-const getAuthToken = memoizee(
-	async (
+const getAuthToken = (() => {
+	const memoizedGetToken = memoizee(
```
Page-, line 371: I separated this out so as to easily include only the values actually relevant to the request within the memoization (and ignore any extra info that may be passed in but is unnecessary)

Page-:
@balena-ci I self-certify!

PR #1867, opened 2022-01-18
Increase delta request timeout to 59s to better align with our backends

Page-:
Change-type: patch


Page-:
This addresses a timeout that was missed in https://github.com/balena-os/balena-supervisor/pull/1836

Page-:
@balena-ci I self-certify!

PR #1868, opened 2022-01-18
Ensure the `finish` event is always reported when fetching images

Page-:
Change-type: patch

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -172,7 +172,7 @@ export function imageFromService(service: ServiceInfo): Image {
 export async function triggerFetch(
 	image: Image,
 	opts: FetchOptions,
-	onFinish = _.noop,
+	onFinish: (success: boolean) => void,
 	serviceName: string,
 ): Promise<void> {
 	const appUpdatePollInterval = await config.get('appUpdatePollInterval');
@@ -202,8 +202,7 @@ export async function triggerFetch(
 	let success: boolean;
 	try {
 		const imageName = normalise(image.name);
-		image = _.clone(image);
-		image.name = imageName;
+		image = { ...image, name: imageName };
 
 		// Look for a matching image on the engine
 		const img = await inspectByName(image.name);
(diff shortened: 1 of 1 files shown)


src/compose/images.ts
```diff
@@ -172,7 +172,7 @@ export function imageFromService(service: ServiceInfo): Image {
 export async function triggerFetch(
 	image: Image,
 	opts: FetchOptions,
-	onFinish = _.noop,
```
Page-, line 175: A value is always passed in so I removed the default for clarity and improved the typing for specifically what we expect (as opposed to inheriting `_.noop`s typing which is pretty generic with `any`s)

src/compose/images.ts
```diff
@@ -214,8 +213,7 @@ export async function triggerFetch(
 		// Create image on the database if it already exists on the engine
 		await markAsSupervised({ ...image, dockerImageId: img.Id });
 
-		onFinish(true);
-		return;
+		success = true;
```
Page-, line 216: This means it'll now fall through to the normal finish procedure which handles both the `finish` event and the `onFinish` callback

Page-:
@balena-ci rebase

src/compose/images.ts
pipex: This sounds like a good solution, but I think we should probably modify the report function a bit. If the execution gets to the end of the first try, the report will get a 'finish' event without ever triggering a start, which right now will fail. I'm trying to think of a scenario that will get to that run path so I can test.

pipex: I guess it depends on whether or not we want a report to be triggered when the fetch is skipped due to the image being on disk. I can see value in that too and if that's the case, the change I made to the reportEvent function prevents other issues.

pipex: I remembered the issue we saw now. The way the `reportEvent` function was implemented, if an event were to be called out of order (e.g. finish without a start), it would create a `runningTask[<imageName>] = undefined` which would cause the `getDownloadingImageNames` function (line 352) to fail.

pipex:
@balena-ci rebase

pipex:
@balena-ci rebase

pipex:
@Page- this looks good like a good change IMHO, should we LGTM?

pipex:
@balena-ci rebase

pipex (approved):
LGTM

PR #1874, opened 2022-01-25
Throw if target states gets a 304 without an ETAG

pipex:
The API uses 304 as a mechanism for load management on target state
requests. This may cause that the supervisor receives a 304 response
without having received a copy of the target state first, leading to
issues. This change checks for an etag when receiving a 304, throwing an
exception otherwise.

Change-type: patch

### DIFF
--- a/src/device-state/target-state.ts
+++ b/src/device-state/target-state.ts
@@ -150,7 +150,7 @@ export const update = async (
 			.getAsync(endpoint, params)
 			.timeout(apiTimeout);
 
-		if (statusCode === 304) {
+		if (statusCode === 304 && cache?.etag != null) {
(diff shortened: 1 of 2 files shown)


src/device-state/target-state.ts
```diff
@@ -150,7 +150,7 @@ export const update = async (
 			.getAsync(endpoint, params)
 			.timeout(apiTimeout);
 
-		if (statusCode === 304) {
+		if (statusCode === 304 && !!cache?.etag) {
```
dfunckt, line 153:
```suggestion
		if (statusCode === 304 && cache?.etag != null) {
```

Page- (approved):

PR #1877, opened 2022-01-26
Use write+sync when writing to `/mnt/boot`

pipex:
`/mnt/boot` is a vfat partition which does not support atomic file  rename. The best course of action is to write and sync as fast as possible to prevent corruption (although it still may happen). This PR modifies all writes to that directory to use the fs-utils `writeAndSyncFile`.

Change-type: patch


thgreasi:
Resolves: #1879

cywang117:
@pipex I think this PR would help reduce instances of https://github.com/balena-os/balena-supervisor/issues/1872, are there any blockers?

pipex:
@cywang117 this was a big PR and I ran into some blockers when trying to implement tests, I'll try to take another look. In any case I think we need to change the implementation for the splash image feature, is way too brittle.

test/12-device-config.spec.ts
```diff
@@ -216,58 +332,51 @@ describe('Device Backend Config', () => {
...
+				[extlinuxConf]: stripIndent`
+				DEFAULT primary
+				TIMEOUT 30
+				MENU TITLE Boot Options
+				LABEL primary
+      	MENU LABEL primary Image
+      	LINUX /Image
+      	APPEND \${cbootargs} \${resin_kernel_root} ro rootwait				`,
+			});
+		};
```
cywang117, line 364: Something is fishy about the indents here :eyes:

test/12-device-config.spec.ts
```diff
@@ -402,46 +561,26 @@ describe('Device Backend Config', () => {
...
-						}),
-					);
-				}
-				return Promise.resolve('');
-			});
-
 			await configFsBackend.initialise();
 			expect(fsUtils.exec).to.be.calledWith('modprobe acpi_configfs');
+
+			// If the module performs this call is because all the prior checks succeeded
```
cywang117, line 577: "If the module performs this call is because all the prior checks succeeded" to "If the module performs this call, it's because all the prior checks succeeded"

test/12-device-config.spec.ts
```diff
@@ -642,137 +804,124 @@ describe('Device Backend Config', () => {
...
+		// doing the file initialisation on 00-init.ts
+		const mockFs = () => {
+			mock({
+				// This is only needed so config.get doesn't fail
+				[configJson]: JSON.stringify({}),
+				[configTxt]: stripIndent`
 				enable_uart=true
 			`,
-			[osRelease]: stripIndent`
+				[osRelease]: stripIndent`
```
cywang117, line 820: Anything off about indents here?

cywang117 (commented):
Nice work! Good to see more mock-fs and less stubbing as a result. Only comments are related to host-utils deduplication and code formatting, otherwise LGTM.

test/12-device-config.spec.ts
pipex: I don't know why the indents got messed up here, they looked ok in my editor.

cywang117:
One more thing I noticed, when I ran test locally, following all tests passing, `test/data/mnt/boot/config.txt` gets changed to the following:
```
dtoverlay=balena-fin
```

Whereas before it was a 1195-line file.

src/lib/host-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+// the host partition mount inside the container
+export function pathOnHost(relPath: string) {
+	return path.join(constants.rootMountPoint, relPath);
+}
+
+// Takes a path relative to the boot partition and returns the
+// correct path under the root partition inside the container
+export function pathOnBoot(relPath: string) {
+	return pathOnHost(path.join(constants.bootMountPoint, relPath));
+}
```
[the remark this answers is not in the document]
pipex: Sounds good, but let me see how much of a change it is. I don't want to make this PR bigger than it already is.

src/lib/host-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Check if a path exists on the root partition
+export async function existsOnHost(relPath: string) {
+	return fsUtils.exists(pathOnHost(relPath));
+}
```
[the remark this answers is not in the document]
pipex: Same here, let me see how much of a change it is. I don't want to make this PR bigger than it already is.

test/12-device-config.spec.ts
```diff
@@ -27,190 +28,302 @@ const splashImageBackend = new SplashImage();
...
+				SUPERVISOR_DELTA_VERSION: '2',
+				SUPERVISOR_INSTANT_UPDATE_TRIGGER: 'true',
+				SUPERVISOR_OVERRIDE_LOCK: 'false',
+				SUPERVISOR_PERSISTENT_LOGGING: 'false',
+				SUPERVISOR_HARDWARE_METRICS: 'true',
+			});
+		});
+	});
+
+	describe('config.txt', () => {
```
cywang117, line 117: One more thing I noticed, when I ran test locally, following all tests passing, test/data/mnt/boot/config.txt gets changed to the following:
```
dtoverlay=balena-fin
```
Whereas before it was a 1195-line file. I would guess something goes wrong with the setup/teardown here, maybe? Alternatively, something going on with the balena fin test suite starting on line 411 of this spec.

pipex: hmmm, I might have missed something in the setup/teardown, let me check

src/lib/host-utils.ts
```diff
@@ -0,0 +1,44 @@
+import * as path from 'path';
+import * as constants from './constants';
+import * as fsUtils from './fs-utils';
+
+// Takes a path relative to root `/` and returns a path in
+// the host partition mount inside the container
```
20k-ultra, line 6:
```suggestion
// Returns an absolute path starting from the hostOS root partition
// This path is accessible from within the Supervisor container
```

src/lib/host-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+export function pathOnHost(relPath: string) {
+	return path.join(constants.rootMountPoint, relPath);
+}
+
+// Takes a path relative to the boot partition and returns the
+// correct path under the root partition inside the container
```
20k-ultra, line 12:
```suggestion
// Returns an absolute path starting from the hostOS boot partition
// This path is accessible from within the Supervisor container
```

src/lib/host-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Check if a path exists on the root partition
+export async function existsOnHost(relPath: string) {
+	return fsUtils.exists(pathOnHost(relPath));
+}
+
+// Check if a path exists under the boot partition
+export async function existsOnBoot(relPath: string) {
+	return fsUtils.exists(pathOnBoot(relPath));
+}
```
20k-ultra, line 25: Apply same comments on existsOnHost function

src/lib/host-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Receives an absolute path for a file under the boot partition (e.g. `/mnt/root/mnt/boot/config.txt`)
+// and writes the given data. This function uses the best effort to write a file trying to minimize corruption
+// due to a power cut. Given that the boot partition is a vfat filesystem, this means
+// using write + sync
+export async function writeToBootAbsolute(file: string, data: string | Buffer) {
```
20k-ultra, line 31: I would just delete this and make other sections us `fsUtils.writeAndSyncFile`

20k-ultra (commented):
test/12-device-config.spec.ts changes were so large it was pretty difficult to tell what even changed but looks just like things were moved around

src/lib/host-utils.ts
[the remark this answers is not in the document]
pipex: I disagree with this last comment @20k-ultra, having an `existsOnHost` adds readability to the code without any real drawbacks.

src/config/backends/config-fs.ts
```diff
@@ -116,7 +112,10 @@ export class ConfigFs extends ConfigBackend {
 	}
 
 	private async writeConfigJSON(config: ConfigfsConfig): Promise<void> {
-		await remountAndWriteAtomic(this.ConfigFilePath, JSON.stringify(config));
+		await hostUtils.writeToBootAbsolute(
+			this.ConfigFilePath,
+			JSON.stringify(config),
+		);
```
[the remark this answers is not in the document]
pipex: I do agree that the method is unnecessary, but I think I'll keep it, since it the next step in this improvement is to move to https://github.com/balena-os/fatrw. When that happens, all we will need is to modify this single method to have all writes on boot adopt this new secure writing mechanism.

I will rename it to `writeToBoot` as the existing `writeToBoot` method is not used.

src/lib/host-utils.ts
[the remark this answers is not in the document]
pipex: @cywang117 I think I'll just remove the host-utils function, which is not used anywhere. I don't remember why I added it. On some other refactor we can move the fs-utils function.

test/12-device-config.spec.ts
pipex: I still get the issue if I skip this test suite, still trying to figure out where the issue is

pipex: I could not figure out where the config.txt was getting removed, it is some weird interaction between the target state tests and the API tests. It doesn't seem to affect subsequent tests so I'm leaving it for now.

src/config/backends/config-txt.ts
```diff
@@ -70,7 +66,7 @@ export class ConfigTxt extends ConfigBackend {
 		if (await exists(ConfigTxt.bootConfigPath)) {
 			configContents = await fs.readFile(ConfigTxt.bootConfigPath, 'utf-8');
 		} else {
-			await fs.writeFile(ConfigTxt.bootConfigPath, '');
+			await hostUtils.writeToBootAbsolute(ConfigTxt.bootConfigPath, '');
```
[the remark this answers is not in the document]
pipex: See my comment [here](https://github.com/balena-os/balena-supervisor/pull/1877#discussion_r854276680). 

I don't think using `this` is necessarily better for a static variable.

src/host-config.ts
```diff
@@ -133,7 +134,7 @@ async function setProxy(maybeConf: ProxyConfig | null): Promise<void> {
 		const conf = maybeConf as ProxyConfig;
 		await mkdirp(proxyBasePath);
 		if (_.isArray(conf.noProxy)) {
-			await writeFileAtomic(noProxyPath, conf.noProxy.join('\n'));
+			await writeToBootAbsolute(noProxyPath, conf.noProxy.join('\n'));
```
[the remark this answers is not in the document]
pipex: See my comment [here](https://github.com/balena-os/balena-supervisor/pull/1877#discussion_r854276680) regarding the use of `writeAndSyncFile`.

pipex:
> test/12-device-config.spec.ts changes were so large it was pretty difficult to tell what even changed but looks just like things were moved around.

That test suite changed a lot because I modified the suite to use mockfs instead of writing to disk. No other changes to the tests were made

pipex:
@balena-ci I self certify!

PR #1878, opened 2022-02-01
Write sync config json

pipex:
Use `writeAndSync` when writing to config.json

`/mnt/boot` is a vfat partition which does not support atomic file
rename. The best course of action is to write and sync as fast as
possible to prevent corruption (although it still may happen)

Change-type: patch

### DIFF
--- a/src/config/configJson.ts
+++ b/src/config/configJson.ts
@@ -4,7 +4,7 @@ import { promises as fs } from 'fs';
 import * as path from 'path';
 
 import * as constants from '../lib/constants';
-import { writeAndSyncFile, writeFileAtomic } from '../lib/fs-utils';
+import { writeAndSyncFile } from '../lib/fs-utils';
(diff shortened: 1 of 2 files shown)


src/config/configJson.ts
```diff
@@ -86,21 +86,14 @@ export default class ConfigJsonConfigBackend {
...
-				} else {
-					return writeAndSyncFile(configPath, JSON.stringify(this.cache));
-				}
-			});
+	private async write(): Promise<void> {
+		// We use writeAndSyncFile since /mnt/boot partition is a vfat
+		// filesystem which dows not provide atomic file renames. The best
+		// course of action on that case is to write and sync as soon as possible
+		return writeAndSyncFile(
+			await this.pathOnHost(),
```
pipex, line 94: `pathOnHost()` does not throw so no need for a catch here

alexgg (approved):
lgtm

Issue #1876, opened 2022-01-26
Endpoints that modify config.json should check for update locks


cywang117:
These endpoints will need to check for application locks before modifying, otherwise reject the modification and send an error in the response (or at least warn to the console) -- this will be a breaking change. The API documentation also needs to be updated to specify that the endpoints will restart the user application (see https://github.com/balena-os/balena-supervisor/issues/1875 for why).

The following is a (working) list of endpoints that modify config.json:
- PATCH /v1/device/host-config, if it changes the `hostname` of the device

pipex:
@cywang117 is this relevant anymore? With the host-config change and the os-config improvements I think we are covered.

cywang117:
@pipex I believe it's still relevant -- I haven't investigated every field, but `persistentLogging` immediately comes to mind.

pipex:
@cywang117 we can keep it open while we check. I don't think persistentLogging can be modified via an API endpoint, and config vars that require reboots do check for logs before rebooting. 

https://github.com/balena-os/balena-supervisor/blob/master/src/device-config.ts#L589-L595

PR #1881, opened 2022-02-09
Add update lock checks to PATCH /v1/device/host-config endpoint

cywang117:
Tests done:
- Set update locks, change hostname on OS >=2.82.6 and <2.82.6, with and without force
- Set update locks, change redsocks config on OS >=2.82.6 and <2.82.6, with and without force

This PR also includes a small commit for merging the functionality of `updateLock.lock` and `lockingIfNecessary` in `applicationManager`, as the 2 functions are very similar. As such I also ran some generic update lock tests:
- Set update locks, attempt release pin with and without force
- Set update locks, attempt config change with and without force

Change-type: patch
Relates-to: #1876

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -548,16 +548,16 @@ $ curl -X POST --header "Content-Type:application/json" \
 
 > **Note:** on devices with supervisor version lower than 7.22.0, replace all `BALENA_` variables with `RESIN_`, e.g. `RESIN_SUPERVISOR_ADDRESS` instead of `BALENA_SUPERVISOR_ADDRESS`.
 
-This endpoint allows setting some configuration values for the host OS. Currently it supports
-proxy and hostname configuration.
+This endpoint allows setting some configuration values for the host OS. Currently it supports proxy and hostname configuration.
 
 For proxy configuration, balenaOS 2.0.7 and higher provides a transparent proxy redirector (redsocks) that makes all connections be routed to a SOCKS or HTTP proxy. This endpoint allows services to modify these proxy settings at runtime.
 
 
 #### Request body
 
-Is a JSON object with several optional fields. Proxy and hostname configuration go under a "network" key. If "proxy" or "hostname" are not present (undefined), those values will not be modified, so that a request can modify hostname
-without changing proxy settings and viceversa.
+A JSON object with several optional fields. Proxy and hostname configuration go under a `network` key. If `proxy` or `hostname` are not present (undefined), those values will not be modified, so that a request can modify hostname without changing proxy settings and vice versa.
+
+By default, with [balenaOS 2.82.6](https://github.com/balena-os/meta-balena/blob/master/CHANGELOG.md#v2826) or newer, host config PATCH requests will cause a balenaEngine restart. Therefore, with Supervisor v12.11.34 and newer, the PATCH request will respect the presence of update locks. Specify the `force` boolean (`false` by default) property in the request body to ignore this.
 
 ```json
 {
@@ -570,7 +570,8 @@ without changing proxy settings and viceversa.
 			"password": "password",
 			"noProxy": [ "152.10.30.4", "253.1.1.0/16" ]
 		},
-		"hostname": "mynewhostname"
+		"hostname": "mynewhostname",
+		"force": true
 	}
 }
 ```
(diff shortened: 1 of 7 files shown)


src/lib/update-lock.ts
```diff
@@ -115,47 +121,62 @@ const lockExistsErrHandler = (err: Error, release: () => void) => {
...
-export function lock(
+export function lock<T extends unknown>(
 	appId: number | null,
-	{ force = false }: { force: boolean },
-	fn: () => PromiseLike<void>,
-): Bluebird<void> {
+	{ force = false, skipLock = false }: { force: boolean; skipLock?: boolean },
+	fn: () => Resolvable<T>,
+): Bluebird<T> {
+	if (skipLock) {
```
pipex, line 132: Not a comment on your code, I really hate this `skipLock` stuff. I hope we can remove it one day. I think it's only necessary because of the implementation of those functions with the `pausingApply`.

pipex: I think this might be my only argument about merging `lockingIfNecessary` into this module. If `update-lock` is though as an independent locking module, I'm not sure it would make sense to have the `skipLock` option.

I don't think this is necessarily a good enough reason to remove this duplication, but at least we should make a note to remove this option in the future. Modules should either use the lock or not use the lock, but having `skipLock` to bypass the lock on some occasions is just a bad API

docs/API.md
```diff
@@ -548,16 +548,16 @@ $ curl -X POST --header "Content-Type:application/json" \
...
 For proxy configuration, balenaOS 2.0.7 and higher provides a transparent proxy redirector (redsocks) that makes all connections be routed to a SOCKS or HTTP proxy. This endpoint allows services to modify these proxy settings at runtime.
 
 
 #### Request body
 
-Is a JSON object with several optional fields. Proxy and hostname configuration go under a "network" key. If "proxy" or "hostname" are not present (undefined), those values will not be modified, so that a request can modify hostname
-without changing proxy settings and viceversa.
+A JSON object with several optional fields. Proxy and hostname configuration go under a `network` key. If `proxy` or `hostname` are not present (undefined), those values will not be modified, so that a request can modify hostname without changing proxy settings and vice versa.
+
+By default, with [balenaOS 2.82.6](https://github.com/balena-os/meta-balena/blob/master/CHANGELOG.md#v2826) or newer, host config PATCH requests will cause a balenaEngine restart, so the request will respect the presence of update locks. Specify the `force` boolean property in the request body to ignore this.
```
pipex, line 560:
```
By default, with [balenaOS 2.82.6](https://github.com/balena-os/meta-balena/blob/master/CHANGELOG.md#v2826) or newer, host config host config PATCH requests may cause a balena engine restart, in order for the container to load the new network configuration, so the request will respect the presence of update locks```
```

Probably this line should include the supervisor version as well

pipex: Also 

```
Specify the `force` boolean property (`false` by default) ...
```

src/host-config.ts
```diff
@@ -214,15 +215,30 @@ export function get(): Bluebird<HostConfig> {
...
 	}
-	return Bluebird.all(promises).return();
+
+	const skipLock =
+		(
+			await Bluebird.any([
+				dbus.servicePartOf('balena-hostname'),
+				dbus.servicePartOf('resin-hostname'),
+			])
+		).includes('config-json.target') === false;
```
pipex, line 230: I would remove this, I don't see much of an advantage in adding this extra complexity. The API from this PR always uses `updateLock`, no matter the OS version.

src/lib/update-lock.ts
cywang117: > I don't think this is necessarily a good enough reason to remove this duplication

Could you clarify? I'm taking this to mean, we shouldn't remove lockingIfNecessary in favor of the base lock function because of skipLock. If this is not the case, I can add a TODO comment here to remove skipLock?

pipex: Oh, yes, I said the opposite of what I meant, which is that we keep these changes but add a `FIXME` comment to remove `skipLock`

docs/API.md
cywang117: > Probably this line should include the supervisor version as well

You mean the Supervisor version that this PR will be? The Supervisor version that comes with 2.82.6 will not include the update lock check.

cywang117: I've force pushed a docs change which specifies that Supervisor v12.11.34 and newer will check for update locks.

cywang117:
@balena-ci rebase

cywang117:
@balena-ci retest

pipex (approved):

PR #1883, opened 2022-02-09
Make the supervisor more resistant to restarts during config changes

pipex:
In newer OS versions changes to config.json may restart the engine (and hence the supervisor). This PR makes some changes to reduce config.json changes when possible and reduce the impact of a crash while writing configuration changes.

Change-type: patch


src/device-config.ts
```diff
@@ -118,7 +118,7 @@ const configKeys: Dictionary<ConfigOption> = {
 	appUpdatePollInterval: {
 		envVarName: 'SUPERVISOR_POLL_INTERVAL',
 		varType: 'int',
-		defaultValue: '60000',
+		defaultValue: '900000',
```
pipex, line 133: @Page- do you know of a reason this value was set to 60 seconds instead of the 15 minutes we set on the API?

src/lib/fs-utils.ts
```diff
@@ -82,3 +83,24 @@ export async function unlinkAll(...paths: string[]): Promise<void> {
 export function getPathOnHost(...paths: string[]): string[] {
 	return paths.map((p: string) => path.join(constants.rootMountPoint, p));
 }
+
+/**
+ * Change modification and access time of the given file.
+ * It creates an empty file if it does not exist
+ */
+export const touch = (file: string, time = new Date()) =>
+	fs.utimes(file, time, time).catch((e) =>
```
cywang117, line 98: Are there supposed to be 2 `time` params here, or should they be `atime` and `mtime` instead for clarity?

src/lib/fs-utils.ts
```diff
@@ -80,8 +80,14 @@ export async function unlinkAll(...paths: string[]): Promise<void> {
...
-export function getPathOnHost(...paths: string[]): string[] {
-	return paths.map((p: string) => path.join(constants.rootMountPoint, p));
+export function getPathOnHost(path: string): string;
+export function getPathOnHost(...paths: string[]): string[];
+export function getPathOnHost(...paths: string[]): string[] | string {
+	if (paths.length === 1) {
+		return path.join(constants.rootMountPoint, paths[0]);
+	} else {
+		return paths.map((p: string) => path.join(constants.rootMountPoint, p));
+	}
```
cywang117, line 90: This looks like an improvement to how I did it 👍🏼

cywang117 (commented):
LGTM apart from the minor question about the params for `fs.utimes`.

src/lib/fs-utils.ts
pipex: You mean as input for the touch function? Not really, this is mimicking the touch command on Unix that modifies both mtime and atime to the current time. I added the time as an arg since I think it might be useful in the future

pipex: Oh wait, I think I misunderstood, but part of my answer still applies. We want to pass the same value for the `atime` and `mtime` args to the `fs.utimes` function. We could create two variables with the same value for clarity but that might be a little redundant.

PR #1885, opened 2022-02-16
Implement `SUPERVISOR_REPORT_INTERVAL` config var

20k-ultra:
This PR simply lifts all the intervals/listeners/queues to a function that monitors for changes to device configs. If `SUPERVISOR_REPORT_INTERVAL` changes, the reporting passed back from `startReporting` is cancelled and started again with the new value.

### DIFF
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -2,7 +2,6 @@ import * as _ from 'lodash';
 import * as url from 'url';
 import { CoreOptions } from 'request';
 
-import * as constants from '../lib/constants';
 import { log } from '../lib/supervisor-console';
 import { InternalInconsistencyError, StatusError } from '../lib/errors';
 import { getRequestInstance } from '../lib/request';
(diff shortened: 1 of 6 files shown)


src/device-state/current-state.ts
```diff
@@ -165,7 +156,7 @@ const backoff = (
 	setTimeout(() => retry(attempts + 1), delay);
 };
 
-function reportCurrentState(attempts = 0) {
+function reportCurrentState(reporter: any, attempts = 0) {
```
20k-ultra, line 159: Unsure how to make reporter have the correct typings so that it is a throttled report function shown in https://github.com/balena-os/balena-supervisor/pull/1885/files#diff-4201a9052aedc0dfdf3d8eb0566ecc3ac822602e1a872e858cddecb2ac772b8cR257-R261

20k-ultra:
The backoff function only passes retry counter back to the `reportCurrentState` function but now it needs to pass the reporter function too....will have to figure this out.

src/device-state/current-state.ts
Page-: It'd be:
```suggestion
function reportCurrentState(reporter: (opts: StateReport, resolve: () => void, reject: (e: Error) => void) => void, attempts = 0) {
```
but if you wanted to infer it then it'd be
```ts
function reportCurrentState(reporter:(...args: Parameters<typeof $fn>) => void, attempts = 0) {
```
Either way it wouldn't enforce that it's throttled but merely that it looks like throttled function (but that's pretty much function types in general, it takes/returns the right types and hopefully does the right thing to transition between them)

20k-ultra:
will implement from scratch again since https://github.com/balena-os/balena-supervisor/pull/1888 has refactored a lot

PR #1888, opened 2022-02-17
Move report throttle out of reporting logic

20k-ultra:
This PR is related to https://github.com/balena-os/balena-supervisor/pull/1885 where I am trying to add more controls to current state reporting. This PR was required to make controlling (setting and resetting) the various intervals/throttles from other modules.

Before the report logic handled collecting, throttling, and backing off. Now, the report just reports with backoff and that function is throttled from the report caller.

I ran this on a device and it works. Will add tests soon for backoff module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "dbus": "^1.0.7",
     "mdns-resolver": "^1.0.0",
     "semver": "^7.3.2",
+    "sinon": "^11.1.2",
     "sqlite3": "^4.1.1",
     "systeminformation": "^5.6.10"
   },
@@ -117,7 +118,6 @@
     "rimraf": "^2.7.1",
     "rwlock": "^5.0.0",
     "shell-quote": "^1.7.2",
-    "sinon": "^10.0.0",
     "sinon-chai": "^3.6.0",
     "strict-event-emitter-types": "^2.0.0",
     "supertest": "^6.1.3",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@types/rwlock": "^5.0.2",
     "@types/semver": "^7.3.3",
     "@types/shell-quote": "^1.7.0",
-    "@types/sinon": "^10.0.0",
+    "@types/sinon": "^10.0.11",
     "@types/sinon-chai": "^3.2.5",
     "@types/supertest": "^2.0.11",
     "@types/terser-webpack-plugin": "^3.0.0",
--- a/src/device-state/current-state.ts
+++ b/src/device-state/current-state.ts
@@ -3,6 +3,7 @@ import * as url from 'url';
 import { CoreOptions } from 'request';
 
 import * as constants from '../lib/constants';
+import { withBackoff, OnFailureInfo } from '../lib/backoff';
 import { log } from '../lib/supervisor-console';
 import { InternalInconsistencyError, StatusError } from '../lib/errors';
 import { getRequestInstance } from '../lib/request';
@@ -14,9 +15,6 @@ import { SchemaTypeKey, SchemaReturn } from '../config/schema-type';
 import * as eventTracker from '../event-tracker';
 import * as deviceState from '../device-state';
 
-// The exponential backoff starts at 15s
-const MINIMUM_BACKOFF_DELAY = 15000;
-
 const INTERNAL_STATE_KEYS = [
 	'update_pending',
 	'update_downloaded',
@@ -25,10 +23,6 @@ const INTERNAL_STATE_KEYS = [
 
 export let stateReportErrors = 0;
 const lastReportedState: DeviceStatus = {
-	local: {},
-	dependent: {},
-};
-const stateForReport: DeviceStatus = {
 	local: {},
 	dependent: {},
 };
@@ -43,6 +37,7 @@ type CurrentStateReportConf = {
 		| 'deviceApiKey'
 		| 'deviceId'
 		| 'localMode'
+		| 'appUpdatePollInterval'
 		| 'hardwareMetrics'
 	>]: SchemaReturn<key>;
 };
@@ -64,7 +59,7 @@ const stripDeviceStateInLocalMode = (state: DeviceStatus): DeviceStatus => {
 	};
 };
 
-const report = async ({ stateDiff, conf }: StateReport) => {
+async function report({ stateDiff, conf }: StateReport): Promise<boolean> {
 	let body = stateDiff;
 	const { apiEndpoint, apiTimeout, deviceApiKey, localMode, uuid } = conf;
 	if (localMode) {
@@ -73,7 +68,7 @@ const report = async ({ stateDiff, conf }: StateReport) => {
 
 	if (_.isEmpty(body.local)) {
 		// Nothing to send.
-		return;
+		return false;
 	}
 
 	if (conf.uuid == null || conf.apiEndpoint == null) {
@@ -104,9 +99,11 @@ const report = async ({ stateDiff, conf }: StateReport) => {
 			headers['retry-after'] ? parseInt(headers['retry-after'], 10) : undefined,
 		);
 	}
-};
+	// State was reported
+	return true;
+}
 
-const getStateDiff = (): DeviceStatus => {
+function newStateDiff(stateForReport: DeviceStatus): DeviceStatus {
 	const lastReportedLocal = lastReportedState.local;
 	const lastReportedDependent = lastReportedState.dependent;
 	if (lastReportedLocal == null || lastReportedDependent == null) {
@@ -138,128 +135,110 @@ const getStateDiff = (): DeviceStatus => {
 	};
 
 	return _.omitBy(diff, _.isEmpty);
-};
-
-const throttledReport = _.throttle(
-	// We define the throttled function this way to avoid UncaughtPromise exceptions
-	// for exceptions thrown from the report function
-	(opts: StateReport, resolve: () => void, reject: (e: Error) => void) =>
-		report(opts).then(resolve).catch(reject),
-	constants.maxReportFrequency,
-);
-
-/**
- * Perform exponential backoff on the function, increasing the attempts
- * counter on each call
- *
- * If attempts is 0 then, it will delay for min(minDelay, maxDelay)
- */
-const backoff = (
-	retry: (attempts: number) => void,
-	attempts = 0,
-	maxDelay: number,
-	minDelay = MINIMUM_BACKOFF_DELAY,
-) => {
-	const delay = Math.min(2 ** attempts * minDelay, maxDelay);
-	log.info(`Retrying request in ${delay / 1000} seconds`);
-	setTimeout(() => retry(attempts + 1), delay);
-};
-
-function reportCurrentState(attempts = 0) {
-	(async () => {
-		const {
-			hardwareMetrics,
-			appUpdatePollInterval: maxDelay,
-			...conf
-		} = await config.getMany([
-			'deviceApiKey',
-			'apiTimeout',
-			'apiEndpoint',
-			'uuid',
-			'localMode',
-			'appUpdatePollInterval',
-			'hardwareMetrics',
-		]);
-
-		reportPending = true;
-		const currentDeviceState = await deviceState.getStatus();
-		// If hardwareMetrics is false, send null patch for system metrics to cloud API
-		const info = {
-			...(hardwareMetrics
-				? await sysInfo.getSystemMetrics()
-				: {
-						cpu_usage: null,
-						memory_usage: null,
-						memory_total: null,
-						storage_usage: null,
-						storage_total: null,
-						storage_block_device: null,
-						cpu_temp: null,
-						cpu_id: null,
-				  }),
-			...(await sysInfo.getSystemChecks()),
-		};
-
-		stateForReport.local = {
-			...stateForReport.local,
-			...currentDeviceState.local,
-			...info,
-		};
-		stateForReport.dependent = {
-			...stateForReport.dependent,
-			...currentDeviceState.dependent,
-		};
-
-		const stateDiff = getStateDiff();
-
-		// report state diff
-		throttledReport(
-			{ stateDiff, conf },
-			() => {
-				// If the patch succeeds update lastReport and reset the error counter
-				_.assign(lastReportedState.local, stateDiff.local);
-				_.assign(lastReportedState.dependent, stateDiff.dependent);
-				stateReportErrors = 0;
-
-				reportPending = false;
-			},
-			(e) => {
-				if (e instanceof StatusError) {
-					// We don't want these errors to be classed as a report error, as this will cause
-					// the watchdog to kill the supervisor - and killing the supervisor will
-					// not help in this situation
-					log.error(
-						`Device state report failure! Status code: ${e.statusCode} - message:`,
-						e.message,
-					);
-
-					// We want to backoff on all errors, but without having the healthchecks
-					// get triggered.
-					// This will use retryAfter as maxDelay if the header is present in the
-					// response by the API
-					backoff(
-						reportCurrentState,
-						// Do not do exponential backoff if the API reported a retryAfter
-						e.retryAfter ? 0 : attempts,
-						maxDelay,
-						// Start the polling at the value given by the API if any
-						e.retryAfter ?? MINIMUM_BACKOFF_DELAY,
-					);
-				} else {
-					eventTracker.track('Device state report failure', {
-						error: e.message,
-					});
-					backoff(reportCurrentState, stateReportErrors++, maxDelay);
-				}
-			},
-		);
-	})();
 }
 
-export const startReporting = () => {
-	const doReport = () => {
+async function reportCurrentState(conf: CurrentStateReportConf) {
+	// Ensure no other report starts
+	reportPending = true;
+	// Wrap the report with fetching of state so report always has the latest state diff
+	const getStateAndReport = async () => {
+		// Get state to report
+		const stateToReport = await generateStateForReport();
+		// Get diff from last reported state
+		const stateDiff = newStateDiff(stateToReport);
+		// Report diff
+		if (await report({ stateDiff, conf })) {
+			// Update lastReportedState
+			_.assign(lastReportedState.local, stateDiff.local);
+			_.assign(lastReportedState.dependent, stateDiff.dependent);
+			// Log that we successfully reported the current state
+			log.info('Reported current state to the cloud');
+		}
+	};
+	// Create a report that will backoff on errors
+	const reportWithBackoff = withBackoff(getStateAndReport, {
+		maxDelay: conf.appUpdatePollInterval,
+		minDelay: 15000,
+		onFailure: handleRetry,
+	});
+	// Run in try block to avoid throwing any exceptions
+	try {
+		await reportWithBackoff();
+		stateReportErrors = 0;
+	} catch (e) {
+		log.error(e);
+	}
+	reportPending = false;
+}
+
+function handleRetry(retryInfo: OnFailureInfo) {
+	if (retryInfo.error instanceof StatusError) {
+		log.error(
+			`Device state report failure! Status code: ${retryInfo.error.statusCode} - message:`,
+			retryInfo.error?.message ?? retryInfo.error,
+		);
+	} else {
+		eventTracker.track('Device state report failure', {
+			error: retryInfo.error?.message ?? retryInfo.error,
+		});
+	}
+	stateReportErrors++;
+	log.info(
+		`Retrying current state report in ${retryInfo.delay / 1000} seconds`,
+	);
+}
+
+async function generateStateForReport() {
+	const { hardwareMetrics } = await config.getMany(['hardwareMetrics']);
+
+	const currentDeviceState = await deviceState.getStatus();
+
+	// If hardwareMetrics is false, send null patch for system metrics to cloud API
+	const info = {
+		...(hardwareMetrics
+			? await sysInfo.getSystemMetrics()
+			: {
+					cpu_usage: null,
+					memory_usage: null,
+					memory_total: null,
+					storage_usage: null,
+					storage_total: null,
+					storage_block_device: null,
+					cpu_temp: null,
+					cpu_id: null,
+			  }),
+		...(await sysInfo.getSystemChecks()),
+	};
+
+	return {
+		local: {
+			...currentDeviceState.local,
+			...info,
+		},
+		dependent: currentDeviceState.dependent,
+	};
+}
+
+export async function startReporting() {
+	// Get configs needed to make a report
+	const reportConfigs = (await config.getMany([
+		'uuid',
+		'apiEndpoint',
+		'apiTimeout',
+		'deviceApiKey',
+		'deviceId',
+		'localMode',
+		'appUpdatePollInterval',
+		'hardwareMetrics',
+	])) as CurrentStateReportConf;
+	// Throttle reportCurrentState so we don't query device or hit API excessively
+	const throttledReport = _.throttle(
+		reportCurrentState,
+		constants.maxReportFrequency,
+	);
+	const doReport = async () => {
 		if (!reportPending) {
-			reportCurrentState();
+			throttledReport(reportConfigs);
 		}
 	};
 
@@ -270,4 +249,4 @@ export const startReporting = () => {
 	setInterval(doReport, constants.maxReportFrequency);
 	// Try to perform a report right away
 	return doReport();
-};
+}
--- a/src/lib/backoff.ts
+++ b/src/lib/backoff.ts
@@ -1,0 +1,105 @@
+import { strict as assert } from 'assert';
+import { promisify } from 'util';
+
+export type OnFailureInfo = {
+	failures: number;
+	delay: number;
+	error: any;
+};
+
+export type Options = {
+	retryCount: number;
+	maxDelay: number;
+	minDelay: number;
+	maxRetries: number;
+	onFailure?: (info: OnFailureInfo) => void;
+};
+
+const DEFAULT_OPTIONS: Partial<Options> = {
+	retryCount: 0,
+	maxDelay: 15000,
+	minDelay: 15000,
+	maxRetries: Infinity,
+};
+
+/**
+ * Retries a function with exponential delay between calls.
+ * If attempts is 0 then, it will delay for min(minDelay, maxDelay)
+ *
+ * Supports exceptions which return `retryAfter` value to specify the backoff duration.
+ * Will call `retryCallback` if passed in options as a way for callers to react to retries
+ */
+export function withBackoff<T extends (...args: any[]) => any>(
+	func: T,
+	options: Partial<Options> = {},
+) {
+	assert(typeof func === 'function', 'expected a function as parameter');
+	// If the function returns a promise, unwrap the promise return type
+	// otherwise use the actual return
+	type TReturn = ReturnType<T> extends Promise<infer R> ? R : ReturnType<T>;
+
+	// TODO use standard lib async setTimout (requires node 16)
+	const sleep = promisify(setTimeout);
+
+	const normalizedOptions: Options = {
+		...DEFAULT_OPTIONS,
+		...options,
+	} as Options;
+
+	return async function wrapped(...args: Parameters<T>): Promise<TReturn> {
+		try {
+			return await func(...args);
+		} catch (error) {
+			if (normalizedOptions.retryCount >= normalizedOptions.maxRetries) {
+				throw new Error(
+					`Reached max number of retries: ${normalizedOptions.maxRetries}`,
+				);
+			}
+
+			const wait = isRetryError(error)
+				? error.retryAfter
+				: exponentialRange(
+						normalizedOptions.retryCount,
+						normalizedOptions.minDelay,
+						normalizedOptions.maxDelay,
+				  );
+
+			normalizedOptions.retryCount++;
+
+			if (normalizedOptions.onFailure) {
+				normalizedOptions.onFailure({
+					failures: normalizedOptions.retryCount,
+					delay: wait,
+					error,
+				});
+			}
+
+			await sleep(wait);
+
+			return wrapped(...args);
+		}
+	};
+}
+
+// This is the expected interface for the error type
+export interface RetryError extends Error {
+	retryAfter: number;
+}
+
+// Type guard only checks that retryAfter is present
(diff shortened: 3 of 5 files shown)


src/lib/backoff.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	if (config.retryCallback) {
+		config.retryCallback(`Retrying request in ${wait / 1000} seconds`);
+	}
+
+	await sleep(wait);
+
+	if (config.maxRetries && config.attempt + 1 >= config.maxRetries) {
+		throw new Error(`Reached max number of retries: ${config.maxRetries}`);
+	}
```
20k-ultra, line 59: why throw error of retries reached after waiting ? move this to before retryCallback since the function is not going to retry.

src/lib/backoff.ts
```diff
@@ -0,0 +1,66 @@
+import { promisify } from 'util';
+
+const sleep = promisify(setTimeout);
+
+export type Options = {
+	attempt: number;
+	maxDelay: number;
+	minDelay: number;
+	maxRetries: number;
+	retryCallback?: (msg: string) => void;
```
pipex, line 10: I would make the `retryCallback` signature as 
```
(timeToNextAttempt: number, e: Error, attempts: number) => void
``` 

So the caller can act on the error information (e.g. whether is a network or HTTP error) and update its internal state

src/lib/backoff.ts
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const DEFAULT_OPTIONS: Partial<Options> = {
+	attempt: 0,
+	maxDelay: 15000,
+	minDelay: 15000,
+	maxRetries: 10,
```
pipex, line 17: Maybe `maxRetries` could default to `+Infinity`, which would make your logic below still work but will have the default behavior that it will retry forever.

src/lib/backoff.ts
20k-ultra: review new signature

src/device-state/current-state.ts
```diff
@@ -64,7 +59,7 @@ const stripDeviceStateInLocalMode = (state: DeviceStatus): DeviceStatus => {
 	};
 };
 
-const report = async ({ stateDiff, conf }: StateReport) => {
+async function report({ stateDiff, conf }: StateReport): Promise<boolean> {
```
20k-ultra, line 62: This makes updating state easier since we now know if the report actually sent something or not.

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+			throw new Error(
+				`Reached max number of retries: ${normalizedOptions.maxRetries}`,
+			);
+		}
+
+		const wait = delay(
+			normalizedOptions.attempt,
+			normalizedOptions.minDelay,
+			normalizedOptions.maxDelay,
+		);
```
20k-ultra, line 68: we can easily modify the backoff options to allow passing a delay algorithm so the caller and use exponential, fibonacci, etc.

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
-					eventTracker.track('Device state report failure', {
-						error: e.message,
-					});
-					backoff(reportCurrentState, stateReportErrors++, maxDelay);
-				}
-			},
-		);
-	})();
+		},
+		dependant: currentDeviceState.dependent,
```
cywang117, line 217:
```suggestion
		dependent: currentDeviceState.dependent,
```

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	return async function wrapped(...args: Parameters<T>): Promise<TReturn> {
+		let error: any;
+		try {
+			return await func(...args);
+		} catch (e) {
+			error = e;
+			if (e.retryAfter) {
+				normalizedOptions.attempt = 0;
+				normalizedOptions.minDelay = e.retryAfter;
```
pipex, line 54: if `maxDelay < e.retryAfter` this won't have the intended effect.

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+		...options,
+	} as Options;
+
+	return async function wrapped(...args: Parameters<T>): Promise<TReturn> {
+		let error: any;
+		try {
+			return await func(...args);
+		} catch (e) {
+			error = e;
+			if (e.retryAfter) {
```
pipex, line 52: Might want to check that `retryAfter` is of the correct type.

You could do something like

```
export class RetryError extends Error {
    constructor(public statusMessage?: string, public retryAfter: number) {
        super(statusMessage);
    }
}

// Type guard
const isRetryError = (x: unknown): x is RetryError => x != null && x instanceof Error && Number.isInteger(x.retryAfter);
```

And then inside the if, you can use the type guard
```
if (isRetryError(e)) {
    // e.retryAfter will be available here
}
```

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+	// otherwise use the actual return
+	type TReturn = ReturnType<T> extends Promise<infer R> ? R : ReturnType<T>;
+
+	const normalizedOptions: Options = {
+		...DEFAULT_OPTIONS,
+		...options,
+	} as Options;
+
+	return async function wrapped(...args: Parameters<T>): Promise<TReturn> {
+		let error: any;
```
pipex, line 47: This is a nitpick but if you move the backoff logic to the catch clause you don't need this `let`

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Retries a function with exponential delay between calls.
+ * If attempts is 0 then, it will delay for min(minDelay, maxDelay)
+ *
+ * Supports exceptions which return `retryAfter` value to specify the backoff duration.
+ * Will call `retryCallback` if passed in options as a way for callers to react to retries
+ */
+export function withBackoff<T extends (...args: any[]) => any>(
+	func: T,
+	options: Partial<Options> = {},
+) {
```
pipex, line 36: Although is not necessary with typescript, it would not hurt to still assert that the passed argument is a function


```
import { strict as assert } from 'assert';

export function withBackoff.... {
   assert(typeof func === 'function', 'expected a function as parameter');
}
```

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
-			'hardwareMetrics',
-		]);
+function handleRetry(retryInfo: BackoffCallback) {
+	if (retryInfo.error instanceof StatusError) {
+		log.error(
+			`Device state report failure! Status code: ${retryInfo.error.statusCode} - message:`,
+			retryInfo.error.message,
+		);
+	} else {
+		eventTracker.track('Device state report failure', {
```
pipex, line 181: I think you still need to increase `stateReportErrors` here as they are checked as part of the healthcheck.

src/device-state/current-state.ts
```diff
@@ -104,9 +99,11 @@ const report = async ({ stateDiff, conf }: StateReport) => {
...
-};
+	// State was reported
+	return true;
+}
 
-const getStateDiff = (): DeviceStatus => {
+function newStateDiff(stateForReport: DeviceStatus): DeviceStatus {
 	const lastReportedLocal = lastReportedState.local;
 	const lastReportedDependent = lastReportedState.dependent;
 	if (lastReportedLocal == null || lastReportedDependent == null) {
```
20k-ultra, line 109: could be removed since the lastREportedState is initialized with an object

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
-) => {
-	const delay = Math.min(2 ** attempts * minDelay, maxDelay);
-	log.info(`Retrying request in ${delay / 1000} seconds`);
-	setTimeout(() => retry(attempts + 1), delay);
-};
+async function reportCurrentState(conf: CurrentStateReportConf) {
+	// Ensure no other report starts
+	reportPending = true;
+	// Wrap the report with fetching of state so report always has the latest state diff
+	const wrappedReport = async () => {
```
cywang117, line 144: Can this function name be a bit more semantic? `getStateAndReport` maybe?

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
+			_.assign(lastReportedState.local, stateDiff.local);
+			_.assign(lastReportedState.dependent, stateDiff.dependent);
+			// Log that we successfully reported the current state
+			log.info('Reported current state to the cloud');
+		}
+	};
+	// Create a report that will backoff on errors
+	const reportWithBackoff = withBackoff(wrappedReport, {
+		maxDelay: conf.appUpdatePollInterval,
+		minDelay: 15000, // If no retryAfter passed, start at 15 seconds
```
20k-ultra, line 161: remove comment

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
+import { promisify } from 'util';
+
+const sleep = promisify(setTimeout);
+
+export type BackoffCallback = {
```
pipex, line 5: `OnRetryCallback` ? Since `backoff` is in the name of the module?

pipex: `OnRetry` might be enough too

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
+		retryCallback: handleRetry,
+	});
+	// Run in try block to avoid throwing any exceptions
+	try {
+		await reportWithBackoff();
+	} catch (e) {
+		log.debug(e);
+	}
+	stateReportErrors = 0;
+	reportPending = false;
```
20k-ultra, line 171: if the try catch throws this code will execute when it should not...return in the catch block.

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
-			'uuid',
-			'localMode',
-			'appUpdatePollInterval',
-			'hardwareMetrics',
-		]);
+function handleRetry(retryInfo: BackoffCallback) {
+	if (retryInfo.error instanceof StatusError) {
+		log.error(
+			`Device state report failure! Status code: ${retryInfo.error.statusCode} - message:`,
+			retryInfo.error.message,
```
cywang117, line 178:
```suggestion
			retryInfo.error?.message ?? retryInfo.error,
```
In case an API response doesn't contain a message

src/lib/backoff.ts
pipex: Actually the type does not point to a callback so `OnRetryOpts` might make more sense

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+
+		await sleep(wait);
+
+		normalizedOptions.attempt++;
+
+		return wrapped(...args);
+	};
+}
+
+function delay(attempts: number, minDelay: number, maxDelay: number): number {
```
pipex, line 86: Another nitpick, `delay` is a verb as well as a noun, so I would expect this function to perform a delay, maybe replace with `calculateDelay` or something like that?

src/device-state/current-state.ts
```diff
@@ -138,128 +135,109 @@ const getStateDiff = (): DeviceStatus => {
...
+	// Get configs needed to make a report
+	const reportConfigs = (await config.getMany([
+		'uuid',
+		'apiEndpoint',
+		'apiTimeout',
+		'deviceApiKey',
+		'deviceId',
+		'localMode',
+		'appUpdatePollInterval',
+		'hardwareMetrics',
```
20k-ultra, line 231: don't need this in report config since it's only used for getting the target state...but maybe we want to pass this config in so getting target state doesn't have to query the config...?

src/lib/backoff.ts
20k-ultra: good point...the delay does not really consider that the value is from the exception which should be respected regardless of what the min and max are. I will modify the delay to use min and max but retryAfter over rules them all.

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+			return await func(...args);
+		} catch (e) {
+			error = e;
+			if (e.retryAfter) {
+				normalizedOptions.attempt = 0;
+				normalizedOptions.minDelay = e.retryAfter;
+			}
+		}
+
+		if (normalizedOptions.attempt + 1 >= normalizedOptions.maxRetries) {
```
pipex, line 58:
```suggestion
		if (normalizedOptions.attempt++ >= normalizedOptions.maxRetries) {
```

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+			normalizedOptions.attempt,
+			normalizedOptions.minDelay,
+			normalizedOptions.maxDelay,
+		);
+
+		if (normalizedOptions.retryCallback) {
+			normalizedOptions.retryCallback({
+				retryCount: normalizedOptions.attempt,
+				delay: wait,
+				error,
```
20k-ultra, line 74: what if error is undefined ?

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+			}
+		}
+
+		if (normalizedOptions.attempt + 1 >= normalizedOptions.maxRetries) {
+			throw new Error(
+				`Reached max number of retries: ${normalizedOptions.maxRetries}`,
+			);
+		}
+
+		const wait = delay(
```
pipex, line 64:
```suggestion
		const wait = isRetryError(e) ? e.retryAfter : delay(
```

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+			normalizedOptions.retryCallback({
+				retryCount: normalizedOptions.attempt,
+				delay: wait,
+				error,
+			});
+		}
+
+		await sleep(wait);
+
+		normalizedOptions.attempt++;
```
20k-ultra, line 80: change this name to retry not attempt because it will be 1 at this point however the function is about to be called twice.

src/device-state/current-state.ts
20k-ultra: I'm going to leave this since @pipex mention in another PR he removes this.

src/device-state/current-state.ts
20k-ultra: good catch, I did not remove this intentionally and I actually had no code which increments the error count so this was important

src/device-state/current-state.ts
20k-ultra: leaving this for now since I don't want to add more files being modified to this PR

src/lib/backoff.ts
20k-ultra: I usually try to make try/catch blocks be as specific/small as possible but there is no other other to do after the try/catch so I'll move the backoff into the catch to remove initializing error.

src/lib/backoff.ts
20k-ultra: I actually found that there is an existing StatusError type in errors module which has a retryAfter parameter!

src/lib/backoff.ts
20k-ultra: I added this and removed the increment down below, thanks

src/lib/backoff.ts
20k-ultra: I removed setting minDelay to retryAfter because I just set the wait time equal to the retryAfter value. I am still setting attempt = 0 so if the API stops sending retryAfter the exponential curve does not resume a previously high value.

src/lib/backoff.ts
20k-ultra: decided to leave it as singular

src/lib/backoff.ts
20k-ultra: changed to exponentialRange

src/lib/backoff.ts
```diff
@@ -0,0 +1,88 @@
...
+
+		return wrapped(...args);
```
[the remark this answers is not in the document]
20k-ultra: It does not since async functions go onto another stack I think? Either way, I performed a test and recursive calls on an async function do not exceed the call stack.

src/lib/backoff.ts
```diff
@@ -0,0 +1,101 @@
+import { strict as assert } from 'assert';
+import { promisify } from 'util';
+
+import { StatusError } from './errors';
```
pipex, line 4: I don't think this module should depend on StatusError, since it only needs a specific error interface. 

```
// This is just because is nice to have for users of the module
export interface RetryError extends Error {
	retryAfter: number;
}

// Type guard only checks that retryAfter is present
const isRetryError = (x: unknown): x is RetryError =>
	x != null && x instanceof Error && Number.isInteger((x as RetryError).retryAfter);
```

The type guard will work with any error with the right interface, including `StatusError`

src/lib/backoff.ts
```diff
@@ -0,0 +1,101 @@
...
+	delay: number;
+	error: any;
+};
+
+export type Options = {
+	attempt: number;
+	maxDelay: number;
+	minDelay: number;
+	maxRetries: number;
+	retryCallback?: (info: OnRetryOpts) => void;
```
pipex, line 20: how about calling this `onRetry`?

src/lib/backoff.ts
```diff
@@ -0,0 +1,101 @@
...
+		...DEFAULT_OPTIONS,
+		...options,
+	} as Options;
+
+	return async function wrapped(...args: Parameters<T>): Promise<TReturn> {
+		try {
+			return await func(...args);
+		} catch (error) {
+			if (isRetryError(error)) {
+				normalizedOptions.attempt = 0;
```
pipex, line 56: Why reset the number of attempts on this case? Even though you are using a different backoff time, having the right number of attempts will still be useful.

src/lib/backoff.ts
```diff
@@ -0,0 +1,101 @@
...
+			}
+
+			if (++normalizedOptions.attempt >= normalizedOptions.maxRetries) {
```
pipex, line 59: Does `maxRetries = 1`, means you backoff at least once? (i.e. perform 2 tries), or does it mean you just try once? 

If your intention is the first one, I think this should be a `>` instead of a `>=`

src/lib/backoff.ts
pipex: I see you answered this above. Even though I see your point, I’m still not sure about the reset, since that might result in a weird interface when using the `retryCallback`, since the value passed for the attempts opt is non monotonic.

src/lib/backoff.ts
20k-ultra: that's a great point..going with your suggestion

src/lib/backoff.ts
20k-ultra: I removed the attempt = 0 since the maxDelay will prevent the big attempt number from causing the device to wait a long time. Additionally, as we spoke in our check-in, it's weird if an observer of the onRetry callback logs attempt = 0 over and over again even though it is incrementing...

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+		minDelay: 1,
+	};
+
+	it('resolves after 3 retries', async () => {
+		// Create a function that will fail 3 times so it succeeds on the 4th
+		const failer = new Failer(3);
+		// Wrap function withBackoff
+		const fnWithBackoff = withBackoff(async () => {
+			return failer.willResolve('data that fails a lot...');
+		}, DEFAULT_OPTIONS);
```
20k-ultra, line 21: I have to call the failer inside a closure because for whatever reason if I don't the references to this inside the class are undefined. I spent some time debugging it and couldn't figure it out.

cywang117: @20k-ultra Could you clarify: the references to what are undefined?

20k-ultra: The failer class has functions that increment an internal counter for number of fails. When this `fnWithBackoff` runs and tries to do `this.count++` the reference to this is undefined. I hope that makes sense.

20k-ultra:
@balena-ci rebase

test/src/lib/backoff.spec.ts
pipex: As a theory, it is possible that when the Promise is loaded into the event loop, the object context is no longer available. Maybe doing something like the following might work? Using a closure sounds fine imho too

```
	public async willResolve(resolvesWith: string): Promise<string> {
                const that = this; 
		return new Promise((resolve, reject) => {
			if (++that.counter <= that.maxFails) {
				return reject('Not ready!');
			}
			return resolve(resolvesWith);
		});
	}
```

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,163 @@
+import { expect } from 'chai';
+
+import Failer from '../../lib/failer';
```
pipex, line 3: Maybe is better to define the `Failer` object within this same file. If we need it somewhere else we can always refactor. Storing it in lib provides no immediate benefit imho

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+					} else if (data.delay < minRecordedDelay) {
+						minRecordedDelay = data.delay;
+					}
+				},
+			},
+		);
+		// Call the function
+		await expect(fnWithBackoff()).to.eventually.equal('ok');
+		// Check that the smallest recorded delay is greater than the minDelay passed in the options
+		expect(minRecordedDelay!).to.be.greaterThan(DEFAULT_OPTIONS.minDelay);
```
pipex, line 53: I think using Sinon fake timers is a better option here and would probably end up in simpler tests. Right now this test is checking that the delay passed to `onRetry` is the not smaller than `minDelay` but it does not test that the function actually backs off for `minDelay` ms. Same thing with the other tests

https://sinonjs.org/releases/latest/fake-timers/

pipex: Testing that the data reported by `onRetry` is also a good test to make but is not the most important behavior.

test/src/lib/backoff.spec.ts
20k-ultra: moved failer into backoff tests

test/src/lib/backoff.spec.ts
20k-ultra: I have seen that = this pattern used before for situations like this but that still doesn't work :/

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+		lib.__set__('sleep', sleepStub);
+	});
+
+	beforeEach(() => {
+		sleepStub.reset();
+	});
+
+	it('resolves after 3 retries', async () => {
+		// Create a function that will fail 3 times so it succeeds on the 4th
+		const failer = new Failer(3);
```
cywang117, line 30: I see that the test failer module was deleted, are you waiting on discussions surrounding implementation before changing this code?

test/src/lib/backoff.spec.ts
cywang117: For future reference, when referring to the JS/TS keyword `this` could we wrap that in backticks for clarity? It's too ambiguous of a word otherwise

cywang117: @20k-ultra What is the non-wrapped test code that you tried before, that doesn't work?

src/lib/backoff.ts
```diff
@@ -0,0 +1,103 @@
...
+	const normalizedOptions: Options = {
+		...DEFAULT_OPTIONS,
+		...options,
+	} as Options;
+
+	return async function wrapped(...args: Parameters<T>): Promise<TReturn> {
+		try {
+			return await func(...args);
+		} catch (error) {
+			if (++normalizedOptions.attempt > normalizedOptions.maxRetries) {
```
pipex, line 53: I suggested making the comparison like this, but this is an error, because this means that the function never backoffs with `minDelay` as the calls to `exponentialRange` start at `1`

```suggestion
			if (normalizedOptions.attempt >= normalizedOptions.maxRetries) {
```

The attempts should be increased before the check for `onFailure`

src/lib/backoff.ts
```diff
@@ -0,0 +1,103 @@
...
+			}
+
+			const wait = isRetryError(error)
+				? error.retryAfter
+				: exponentialRange(
+						normalizedOptions.attempt,
+						normalizedOptions.minDelay,
+						normalizedOptions.maxDelay,
+				  );
+
```
pipex, line 66: Increasing here means that what gets passed to onFailure is equal to `attempts+1` which is the right value

```suggestion
normalizedOptions.attempt++;
```

src/device-state/current-state.ts
```diff
@@ -138,128 +135,110 @@ const getStateDiff = (): DeviceStatus => {
...
-		]);
+function handleRetry(retryInfo: OnFailureInfo) {
+	if (retryInfo.error instanceof StatusError) {
+		log.error(
+			`Device state report failure! Status code: ${retryInfo.error.statusCode} - message:`,
+			retryInfo.error?.message ?? retryInfo.error,
+		);
+	} else {
+		eventTracker.track('Device state report failure', {
+			error: retryInfo.error.message,
```
cywang117, line 182: `retryInfo.error?.message ?? retryInfo.error` may be a better catch-all

src/lib/backoff.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Type guard only checks that retryAfter is present
+const isRetryError = (x: unknown): x is RetryError =>
+	x != null &&
+	x instanceof Error &&
+	Number.isInteger((x as RetryError).retryAfter);
+
+/**
+ * Calculates a number with exponential growith given {retryCount} within
```
cywang117, line 96:
```suggestion
 * Calculates a number with exponential growth given {retryCount} within
```

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+			},
+			{
+				minDelay,
+			},
+		);
+		// Function should have been called 0 times to start
+		expect(failer.callCount).to.equal(0);
+		// Call function
+		myBackoffFunc();
+		// Check that function is ran atleast once
```
cywang117, line 55:
```suggestion
		// Check that function was run at least once
```
nitpicky, feel free to ignore

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+		expect(failer.callCount).to.equal(1);
+		// Tick clock by minDelay
+		// This will not be enough to call the function because retryAfter was in the
+		// exception thrown and it is greater then minDelay
+		await clock.tickAsync(minDelay);
+		// Check that function wasn't called yet
+		expect(failer.callCount).to.equal(1);
+		// Tick clock again just before retryAfter
+		await clock.tickAsync(retryAfter - minDelay - 1);
+		// Check that calls is still only 1 since retryAfter time has not elapsed
```
cywang117, line 169:
```suggestion
		// Check that call count is still only 1 since retryAfter time has not elapsed
```
nit, ignore if desired

package.json
```diff
@@ -32,6 +32,7 @@
     "dbus": "^1.0.7",
     "mdns-resolver": "^1.0.0",
     "semver": "^7.3.2",
+    "sinon": "^11.1.2",
```
cywang117, line 35: Do we need to update @types/sinon too? It's currently still at `^10.0.0`

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+						failures: counter,
+						delay:
+							counter === 1 ? minDelay : exponentialize(minDelay, counter - 1),
+						error: 'Not ready!',
+					});
+				},
+			},
+		);
+		// Call function and allow clock to trigger all events
+		clock.runAllAsync();
```
pipex, line 136: does this work? I have only seen this being done the oppossite way

```
const p = fnWithBackoff();
await clock.runAllAsync();
await p;
```

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+		);
+		const fnWithBackoff = withBackoff(
+			async () => {
+				return failer.willResolve('ok');
+			},
+			{
+				minDelay,
+			},
+		);
+		assert(retryAfter > minDelay, 'retryAfter must be greater then minDelay');
```
cywang117, line 155:
```suggestion
		assert(retryAfter > minDelay, 'retryAfter must be greater than minDelay');
```

test/src/lib/backoff.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+	afterEach(() => {
+		clock.restore();
+	});
+
+	it('resolves after 3 retries', async () => {
+		// Create a function that will fail 3 times so it succeeds on the 4th
+		const failer = new Failer(3);
+		// Wrap function withBackoff
+		const fnWithBackoff = withBackoff(async () => {
+			return failer.willResolve('fails 3 times then resolves on 4th');
```
pipex, line 29: Is not a deal breaker and I'm find to leave it as is, but I feel like the `Failer` class doesn't add much more than a stub would and it probably makes it a bit more difficult to read the code. 

As I said though I don't want to nitpick more on this as I think this works and we can always refactor in the future if needed.

package.json
20k-ultra: latest @types/sinon is v10.0.11 so I can't upgrade to 11.1.2 unfortunately.

I had to check what the convention was for definitely typed package versions which https://github.com/DefinitelyTyped/DefinitelyTyped/issues/13225 confirmed that they should match the version of the package it is adding types for.

test/src/lib/backoff.spec.ts
20k-ultra: I don't like that I have to curry the function to make it work though so I'm interested in removing it if we can. I don't know how a stub could be made to fail N number of times

test/src/lib/backoff.spec.ts
20k-ultra: "clock.runAll(); / await clock.runAllAsync()

This runs all pending timers until there are none remaining. If new timers are added while it is executing they will be run as well."

20k-ultra:
@balena-ci I self-certify!

test/src/lib/backoff.spec.ts
20k-ultra: right I see now. You'd just tick the clock and then make the stub resolve or reject at the callCount you want.

Issue #1890, opened 2022-02-23
Handle new default security-opts (balena-engine 20.10.x)


robertgzr:
## v19.03.24
```
root@balena:~# balena inspect balena_supervisor | grep -i security
            "SecurityOpt": null,
```
## v20.10.12

```
balena inspect balena_supervisor | jq '.[].HostConfig.SecurityOpt'
[
  "label=disable"
]
```

Connects-to: https://github.com/balena-os/meta-balena/pull/2498
Internal thread: https://www.flowdock.com/app/rulemotion/r-resinos/threads/qkzdgm9f6k8O0JLINsJvdif4BQB

PR #1891, opened 2022-02-23
Ignore selinux security opts when comparing services

pipex:
The moby engine v20.x.y adds some selinux [security configurations](https://docs.docker.com/engine/reference/run/#security-configuration)
depending on the [container configuration](https://github.com/moby/moby/blob/master/daemon/create.go#L214).
This would cause the supervisor to enter a service restart loop as the
current and target service configurations will never match. The
supervisor now ignores selinux specific security options since those are
not supported by balenaOS and have no effect.

Any existing selinux configurations will not be removed so updating to this supervisor won't cause any service restarts.

Closes: #1890
Change-type: patch


robertgzr (approved):
nice thanks :tada:

Issue #1806, opened 2021-10-08
Error: (HTTP code 404) no such network - network supervisor0 not found


pdcastro:
Supervisor version 12.10.15:

```
[error]   Scheduling another update attempt in 900000ms due to failure:  Error: (HTTP code 404) no such network - network supervisor0 not found 
[error]         at /usr/src/app/dist/app.js:10:2333800
[error]       at /usr/src/app/dist/app.js:10:2333732
[error]       at Modem.buildPayload (/usr/src/app/dist/app.js:10:2333752)
[error]       at IncomingMessage.<anonymous> (/usr/src/app/dist/app.js:10:2333005)
[error]       at IncomingMessage.emit (events.js:322:22)
[error]       at endReadableNT (_stream_readable.js:1187:12)
[error]       at processTicksAndRejections (internal/process/task_queues.js:84:21)
[error]   Device state apply error Error: (HTTP code 404) no such network - network supervisor0 not found 
[error]         at /usr/src/app/dist/app.js:10:2333800
[error]       at /usr/src/app/dist/app.js:10:2333732
[error]       at Modem.buildPayload (/usr/src/app/dist/app.js:10:2333752)
[error]       at IncomingMessage.<anonymous> (/usr/src/app/dist/app.js:10:2333005)
[error]       at IncomingMessage.emit (events.js:322:22)
[error]       at endReadableNT (_stream_readable.js:1187:12)
[error]       at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

The state was:

* `ifconfig` showed that a `supervisor0` interface existed
* `balena-engine network ls` showed that a `supervisor0` network did **not** exist
* Restarting the supervisor would not fix it.

The following manual intervention "fixed it":

```sh
$ systemctl stop resin-supervisor
$ ip link del supervisor0
$ systemctl start resin-supervisor
```

The `ip link del` command deletes the interface as listed by `ifconfig`.

Tentative conclusion: The `supervisor0` entity exists both as a system interface, as listed by `ifconfig`, and as a balenaEngine network, as listed by `balena-engine network ls`. The existence of the former prevents the supervisor from creating the latter. The solution would thus be for the supervisor to independently inspect and create both `supervisor0` entities.

alexszilagyi:
@pdcastro , @pipex : Any updates on this?

toochevere:
[...]

PR #1892, opened 2022-02-25
Fix check for supervisor0 network

pipex:
The check for the docker network supervisor0 assumed that if the
interface supervisor0 existed, then the network would exist too. However this is not
true on the case of docker directory corruption, which would lead to a
loop with `Error: (HTTP code 404) no such network - network supervisor0 not found`.

Change-type: patch
Closes: #1806


cywang117:
@pipex LGTM, looks like a solid upgrade.

20k-ultra (approved):

PR #1893, opened 2022-03-08
Differentiate Supervisor-created lockfiles from userapp-created lockfiles

cywang117:
This ensures that the Supervisor is able to clean up any lockfiles it created on process startup, preventing the device from getting stuck in an invalid state caused by the Supervisor getting killed before it can remove its lockfiles.

Tested:
- Restarting/purging app:
This calls pausingApply which calls updateLock.lock, meaning these are actions where the Supervisor "takes the lock". After the Supervisor action which requires updateLock.lock is completed, updateLock.dispose is called. In `dispose` method, inspected the files that are to be unlocked here, to ensure that they've been created with the correct UID (they were)

- Cleaning up locks on start: stopped Supervisor process, created a lock with UID 65534, started Supervisor, observed that the lockfile was removed as intended

TODO, separate PR to docs repo:
    - Update balena docs for update locks to indicate that nobody UID is expected for Supervisor-created lockfiles as of version 12.x.y

See: https://www.flowdock.com/app/rulemotion/r-resinos/threads/gWMgK5hmR26TzWGHux62NpgJtVl
Change-type: minor
Closes: https://github.com/balena-os/balena-supervisor/issues/1758


cywang117:
@balena-ci retest

src/lib/lockfile.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export async function lock(path: string) {
+	/**
+	 * Set parent directory permissions to `drwxrwxrwt` (octal 1777), which are needed
+	 * for lockfile binary to run successfully as the `nobody` (UID 65534) user.
+	 * Otherwise the `nobody` user will not have write permissions to the necessary
+	 * bind-mounted lockfile directories. `chmod` does not fail or throw if the
+	 * directory already has the proper permissions.
+	 */
+	await fs.promises.chmod(dirname(path), STICKY_WRITE_PERMISSIONS);
```
cywang117, line 18: As /tmp contents are stored in memory, calling chmod on each lockfile.lock call will not affect storage media and is of negligible overhead.

src/lib/update-lock.ts
```diff
@@ -8,30 +7,20 @@ import * as Lock from 'rwlock';
...
-	unlockAsync(path: string): Bluebird<void>;
-	lockAsync(path: string): Bluebird<void>;
-};
-const lockFile = Bluebird.promisifyAll(lockFileLib) as asyncLockFile;
-export type LockCallback = (
-	appId: number,
-	opts: { force: boolean },
-	fn: () => PromiseLike<void>,
-) => Bluebird<void>;
+const BASE_LOCK_DIR = '/tmp/balena-supervisor/services';
```
cywang117, line 20: Need to account for legacy lock paths, see https://www.balena.io/docs/learn/deploy/release-strategy/update-locking/#location-of-the-lockfile

src/lib/update-lock.ts
```diff
@@ -75,13 +64,52 @@ const locksTaken: { [lockName: string]: boolean } = {};
...
+export async function cleanupLocks() {
+	// Should be /mnt/root/tmp/balena-supervisor/services
+	const lockfileParentDir = getPathOnHost(BASE_LOCK_DIR);
+
+	try {
+		// Unix `find` command for listing all Supervisor-created lockfiles
+		const findCommand = `find ${lockfileParentDir} -type f -user ${lockfile.NOBODY_UID} -name "*updates.lock"`;
+
+		// Find paths for leftover Supervisor lockfiles
+		const { stdout, stderr } = await exec(findCommand);
```
cywang117, line 90: Using Unix `find` is more efficient than implementing a recursive directory search in TypeScript

src/lib/update-lock.ts
```diff
@@ -94,28 +122,56 @@ export const readLock: LockFn = Bluebird.promisify(locker.async.readLock, {
...
+	if (LockfileExistsError(err as ChildProcessError)) {
+		let errMsg = (err as ChildProcessError).stderr;
+
 		// Extract appId|appUuid and serviceName from lockfile path for log message
-		// appId: [0-9]{7}, appUuid: [0-9a-w]{32}, short appUuid: [0-9a-w]{7}
-		const pathMatch = err.message.match(
-			/\/([0-9]{7}|[0-9a-w]{32}|[0-9a-w]{7})\/(.*)\/(?:resin-)?updates.lock/,
+		// appId: [0-9]{7}, appUuid: [0-9a-f]{32}, short appUuid: [0-9a-f]{7}
+		const pathMatch = errMsg.match(
+			/\/([0-9]{7}|[0-9a-f]{32}|[0-9a-f]{7})\/(.*)\/(?:resin-)?updates.lock/,
```
cywang117, line 141: I made this silly mistake before and added 1204c40bbd21595cc18c124dab4daa9068338c1f to rectify.. UUID regex should be `[0-9a-f]` and I don't know why I wrote it as `[0-9a-w]`

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	beforeEach(() => {
+		// @ts-ignore
+		execStub = stub(fsUtils, 'exec').callsFake(async (command, opts) => {
+			// Extract target path
+			const targetPath = command.replace('lockfile -r 0 ', '');
+
+			// Emulate the lockfile binary exec call
+			await fsUtils.touch(targetPath);
+			await fs.chown(targetPath, opts!.uid!, 0);
```
cywang117, line 68: Just as a note for future reference, we *could* be using the `lockfile` npm library and then running `fs.chown` on the resulting lockfile, but that operation is not atomic. The decision to move to Debian/Alpine's `lockfile` from the `procmail` package is because we can lock a file as a specific user atomically, thus eliminating the chance of something interrupting the lock-taking operation.

src/lib/update-lock.ts
```diff
@@ -75,13 +64,52 @@ const locksTaken: { [lockName: string]: boolean } = {};
...
+ * Supervisor-created lockfiles are distinguished by their UID of 65534,
+ * corresponding to the `nobody` user.
+ */
+export async function cleanupLocks() {
+	// Should be /mnt/root/tmp/balena-supervisor/services
+	const lockfileParentDir = getPathOnHost(BASE_LOCK_DIR);
+
+	try {
+		// Unix `find` command for listing all Supervisor-created lockfiles
+		const findCommand = `find ${lockfileParentDir} -type f -user ${lockfile.NOBODY_UID} -name "*updates.lock"`;
```
pipex, line 87: Here is how I was doing it on my other PR

https://github.com/balena-os/balena-supervisor/blob/7f539547d02515b32cd28d65e542fb06c1e794c6/src/lib/update-lock.ts#L39-L69

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+		}
+	};
+
+	let execStub: SinonStub;
+
+	beforeEach(() => {
+		// @ts-ignore
+		execStub = stub(fsUtils, 'exec').callsFake(async (command, opts) => {
+			// Extract target path
+			const targetPath = command.replace('lockfile -r 0 ', '');
```
pipex, line 64: Although this works, I don't like that it is depending on an internal implementation detail of the lockfile module. If we ever want to add another argument to the `lockfile` call then this will fail, which goes against the idea of tests being resistant to refactoring. 

It's a bit extra, but maybe something like this?

```suggestion
                        // Sanity check for the command call
                        expect(command.trim().startsWith('lockfile')).to.be.true
                        
                        // Remove any command options to leave just the command and the target file
			const [,targetPath] = command.replace(/-v|-nnn|-r\s+\d+|-l\s+\d+|-s\s+\d+|-!|-ml|-mu/g,'').split(/\s+/);
```

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+		await lockfile.lock(lockPath);
+
+		// Verify lockfile exists
+		await checkLockDirFiles(lockPath, { shouldExist: true });
+
+		// Verify lockfile UID
+		expect((await fs.stat(lockPath)).uid).to.equal(NOBODY_UID);
+	});
+
+	it('should change lockfile parent directory permissions to 1777 on lock', async () => {
```
pipex, line 90: This might be an unnecessary test, if the file can be written, then the directory has to have the right permissions. This also falls into the test going into implementation details.

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+		// Verify correct permissions
+		await checkDirPermissions(lockPath, DEFAULT_PERMISSIONS.node);
+
+		await lockfile.lock(lockPath);
+
+		// Verify directory permissions have changed
+		await checkDirPermissions(lockPath, STICKY_WRITE_PERMISSIONS.node);
+	});
+
+	it('should not error if lockfile parent directory permissions are already 1777 on lock', async () => {
```
pipex, line 103: Same thing as above, this might be unnecessary, since if trying to write the lockfile fails, the lock test will fail too.

src/lib/lockfile.ts
```diff
@@ -0,0 +1,34 @@
+import * as fs from 'fs';
+import { dirname } from 'path';
+
+import { exec, unlinkAll } from './fs-utils';
+
+// Equivalent to `drwxrwxrwt`
+const STICKY_WRITE_PERMISSIONS = parseInt('1777', 8);
```
pipex, line 7: You can do octal literals in javascript by prefixing with a `0`. In this case 

```
const STICKY_WRITE_PERMISSIONS = 01777
```

src/lib/lockfile.ts
```diff
@@ -0,0 +1,34 @@
...
+export const NOBODY_UID = 65534;
+
+export async function lock(path: string) {
```
pipex, line 10: Something like this would make this function a bit more generic.

```suggestion
export async function lock(path: string, uid = NOBODY_UID) {
```

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+import * as lockfile from '../../../src/lib/lockfile';
+import * as fsUtils from '../../../src/lib/fs-utils';
+
+describe('lib/lockfile', () => {
+	const lockPath = '/tmp/balena-supervisor/services/1234567/updates.lock';
+	// mock-fs expects an octal file mode, however, Node's fs.stat.mode returns a bit field:
+	//   - 16877 (Node) == octal 0755 (drwxr-xr-x)
+	//   - 17407 (Node) == octal 1777 (drwxrwxrwt)
+	const DEFAULT_PERMISSIONS = {
+		unix: parseInt('0755', 8),
```
pipex, line 16: Here is also better to use octal literals

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+		node: 16877,
+	};
+	const STICKY_WRITE_PERMISSIONS = {
+		unix: parseInt('1777', 8),
```
pipex, line 20: Prefer octal literal

test/src/lib/lockfile.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+		// Verify correct permissions
+		await checkDirPermissions(lockPath, STICKY_WRITE_PERMISSIONS.node);
+
+		await lockfile.lock(lockPath);
+
+		// Verify directory permissions have not changed
+		await checkDirPermissions(lockPath, STICKY_WRITE_PERMISSIONS.node);
+	});
+
```
pipex, line 115: A test that is missing here is something that checks that the file is not created if the exec call throws.

src/lib/errors.ts
```diff
@@ -23,6 +23,12 @@ interface CodedSysError extends Error {
 	code?: string;
 }
 
+export interface ChildProcessError extends Error {
```
pipex, line 26: We should try to move towards making modules more self contained. This error could probably be declared inside the `update-lock` module, where it is going to be used.

Even better it could be declared in the `lockfile` module and the lock function could catch an error on the exec and return a `LockfileExistsError` if `code == 73` or a generic error otherwise.

src/lib/errors.ts
```diff
@@ -45,6 +51,13 @@ export function UnitNotLoadedError(err: string[]): boolean {
 	return endsWith(err[0], 'not loaded.');
 }
 
+export function LockfileExistsError(err: ChildProcessError): boolean {
```
pipex, line 54: Same thing here, this is better moved to where it makes more sense (probably the lockfile module)

src/lib/update-lock.ts
```diff
@@ -94,18 +122,22 @@ export const readLock: LockFn = Bluebird.promisify(locker.async.readLock, {
...
 }
 
-const lockExistsErrHandler = (err: Error, release: () => void) => {
-	let errMsg = err.message;
-	if (EEXIST(err)) {
+const lockfileErrHandler = (err: Error, release: () => void) => {
+	let error: any;
+
+	// If error code is 73, updates.lock file already exists.
+	if (LockfileExistsError(err as ChildProcessError)) {
```
pipex, line 135: If the conversion to `LockFileExistsError` happens inside the `lockfile` module then this just needs to do

```
if (err instanceof LockFileExistsError)
```

pipex: Actually you don't need to check the error type, any error should cause the update-lock function to throw.

src/lib/update-lock.ts
```diff
@@ -114,8 +146,32 @@ const lockExistsErrHandler = (err: Error, release: () => void) => {
...
+		 * In theory, we should get a child process error with code 73,
+		 * indicating the lockfile already exists. Therefore any other process
+		 * error code should be thrown and stop the parent process as something
+		 * unexpected has gone wrong. Other errors that are not the typical "file exists"
+		 * errors include but aren't limited to:
+		 *   - running out of file descriptors
+		 *   - binary corruption
+		 *   - other systems-based errors
+		 */
+		error = new Error(
```
pipex, line 162: Same here, this probably should go inside the `lockfile` module

src/lib/lockfile.ts
cywang117: Changed to 0o1777 per `Octal literals are not available when targeting ECMAScript 5 and higher. Use the syntax '0o1777'.ts(1085)`

test/src/lib/lockfile.spec.ts
cywang117: Good point.

test/src/lib/lockfile.spec.ts
cywang117: No arguments against this.

test/src/lib/lockfile.spec.ts
cywang117: Added test, will force push soon

test/src/lib/update-lock.spec.ts
```diff
@@ -108,40 +112,101 @@ describe('lib/update-lock', () => {
...
+						three: {
+							'updates.lock': mockFs.file({ uid: lockfile.NOBODY_UID }),
+						},
+						four: {
+							'updates.lock': mockFs.file({ uid: 0 }),
+						},
+					},
+				},
+			});
+			stub(fsUtils, 'exec').resolves({
```
pipex, line 140: This test doesn't seem terribly useful, since is basically checking that everything returned by this call (which breaks the black box of the implementation) is removed of the filesystem. The call could be anything which means we are not really testing the cleanup locks behavior. 

This might be an argument for using node operations to find instead of calling `find` as that way we will be able to test the state of the filesystem before and after as a black box.

Alternatively we can do 

```
find ${lockfileParentDir} -type f -user ${lockfile.NOBODY_UID} -name "*updates.lock" | xargs rm
```

on `entry.sh` and just forget about the testing maybe?

test/src/lib/update-lock.spec.ts
```diff
@@ -108,40 +112,101 @@ describe('lib/update-lock', () => {
...
+			(fsUtils.exec as SinonStub).restore();
+		});
+	});
+
 	describe('Lock/dispose functionality', () => {
-		const lockFile = updateLock.__get__('lockFile');
 		const locksTaken = updateLock.__get__('locksTaken');
 		const dispose = updateLock.__get__('dispose');
-		const lockExistsErrHandler = updateLock.__get__('lockExistsErrHandler');
+		const lockfileErrHandler = updateLock.__get__('lockfileErrHandler');
```
pipex, line 171: Another argument for moving the error handing to the `lockfile` module is that we can get away from using `rewire` here.

test/src/lib/update-lock.spec.ts
```diff
@@ -108,40 +112,101 @@ describe('lib/update-lock', () => {
...
+				await fs.readdir(path.join(BASE_PATH, '7654321', 'four')),
+			).to.have.length(1);
+
+			mockFs.restore();
+			(fsUtils.exec as SinonStub).restore();
+		});
+	});
+
 	describe('Lock/dispose functionality', () => {
-		const lockFile = updateLock.__get__('lockFile');
```
pipex, line 112: Can we re-do these tests without using rewire?

test/src/lib/update-lock.spec.ts
pipex: `dispose` is not used so we can remove it, and the rest should be removable if we move the lock tracking and error management functionality to `lockfile` where I think it makes more sense.

src/lib/update-lock.ts
pipex: Fwiw I don't think is such a good idea to be parsing the error. The locking function can tell which one of the locking operations failed, and can return an `UpdatesLockedError` with the relevant info.

src/lib/update-lock.ts
```diff
@@ -75,13 +64,52 @@ const locksTaken: { [lockName: string]: boolean } = {};
 process.on('exit', () => {
```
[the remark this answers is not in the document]
cywang117: Moving this, will be viewable with next force-push

src/lib/update-lock.ts
cywang117: For convenience when building a Supervisor-specific lockfile exists error message, checking may still help to avoid some regex `.match` calls:
```
const lockfileErrHandler = (err: Error, release: () => void) => {
	let error = err;

	if (err instanceof lockfile.LockfileExistsError) {
		let errMsg = (err as lockfile.LockfileExistsError).stderr;

		// Extract appId|appUuid and serviceName from lockfile path for log message
		// appId: [0-9]{7}, appUuid: [0-9a-f]{32}, short appUuid: [0-9a-f]{7}
		const pathMatch = errMsg.match(
			/\/([0-9]{7}|[0-9a-f]{32}|[0-9a-f]{7})\/(.*)\/(?:resin-)?updates.lock/,
		);
		if (pathMatch && pathMatch.length === 3) {
			errMsg = `Lockfile exists for ${JSON.stringify({
				serviceName: pathMatch[2],
				[/^[0-9]{7}$/.test(pathMatch[1]) ? 'appId' : 'appUuid']: pathMatch[1],
			})}`;
		}

		error = new UpdatesLockedError(errMsg);
	}

	return dispose(release).throw(error);
};
```

src/lib/errors.ts
cywang117: Done, moved to inside lockfile. The Supervisor-specific "Lockfile exists for { service: ..., appId: ... }" message creation is still in update-lock.ts though.

test/src/lib/update-lock.spec.ts
cywang117: Yeah, that's basically what I'm testing though - that anything returned by `find` is removed from the filesystem. Here we are working on the assumption that `find` always runs as expected, since this is a unit test.

I like the idea of running the `find` command in entry.sh instead though.

test/src/lib/update-lock.spec.ts
cywang117: We still do need the `lockfileErrHandler` (that I pasted above) though, since the error message if a lockfile exists is specific to the Supervisor in that it references serviceName and appId, and requires calling `dispose` which is also an `update-lock` module method.

src/lib/update-lock.ts
pipex: I think we need to move away from this, we can discuss tomorrow in a call if you like. Parsing the error message is a not a good practice imho.

pipex: From this I mean, from parsing the error message :p

test/src/lib/update-lock.spec.ts
cywang117: Depending on the outcome of our discussion on `lockfileErrHandler` above, we could remove that, but I don't think `dispose` could be modified to not use rewire unless the method is moved to `lib/lockfile.ts` as well. I opted to keep `dispose` in update-lock.ts as it has more to do with the Supervisor-specific update-lock functionality than the generic lockfile lib.

src/lib/update-lock.ts
cywang117: The main upside to adding error message parsing to extract the serviceName and appId/uuid is for the clarity of the message which I rather liked.. we could move away from parsing but I'd still want the exact source of the error communicated clearly, so it would mean surfacing a file path directly in the error message. I like surfacing the lockfile path directly because it tells the user exactly where to remove the lockfile, but dislike it because it's overly verbose (the path name is too long a block of letters together, leading to less than ideal UX).

cywang117: I do not know how we can do the following things simultaneously:
- Remove the lockfile error handler from update-lock while still letting users know the exact source of the lockfile error
- Throw a generic error in the lockfile module that doesn't reference serviceName or appId in any way, but still communicate this with good UX to the user, without sending a long file path

entry.sh
```diff
@@ -56,6 +56,10 @@ fi
 # not a problem.
 modprobe ip6_tables || true
 
+# Cleanup leftover Supervisor-created lockfiles from any previous processes.
+# Supervisor-created lockfiles have a UID of 65534.
+find /mnt/root/tmp/balena-supervisor/services -type f -user 65534 -name "*updates.lock" | xargs rm
```
pipex, line 61: I think is better if we add something like

```suggestion
export BASE_LOCK_DIR="/tmp/balena-supervisor/services"
export LOCKFILE_UID=65534
find "/mnt/root/${BASE_LOCK_DIR}" -type f -user "${LOCKFILE_UID}" -name "*updates.lock" | xargs rm
```

And then use these exported variables in the update-lock module

src/lib/lockfile.ts
```diff
@@ -0,0 +1,111 @@
...
+		await exec(`lockfile -r 0 ${path}`, { uid });
+		// Store a lock in memory as taken
+		locksTaken[path] = true;
+	} catch (error) {
+		// Code 73 refers to EX_CANTCREAT (73) in sysexits.h, or:
+		// A (user specified) output file cannot be created.
+		// See: https://nxmnpg.lemoda.net/3/sysexits
+		if (error instanceof LockfileExistsError) {
+			// If error code is 73, updates.lock file already exists.
+			// Throw this error directly where it's parsed into a Supervisor-specific
```
pipex, line 74: Is not really parsed anymore right.

src/lib/update-lock.ts
```diff
@@ -8,30 +7,17 @@ import * as Lock from 'rwlock';
...
-	unlockAsync(path: string): Bluebird<void>;
-	lockAsync(path: string): Bluebird<void>;
-};
-const lockFile = Bluebird.promisifyAll(lockFileLib) as asyncLockFile;
-export type LockCallback = (
-	appId: number,
-	opts: { force: boolean },
-	fn: () => PromiseLike<void>,
-) => Bluebird<void>;
+const BASE_LOCK_DIR = '/tmp/balena-supervisor/services';
```
pipex, line 17: If you make the change to entry 
```suggestion
const BASE_LOCK_DIR = process.env.BASE_LOCK_DIR || '/tmp/balena-supervisor/services';
```

src/lib/update-lock.ts
```diff
@@ -92,54 +65,32 @@ export const readLock: LockFn = Bluebird.promisify(locker.async.readLock, {
...
 	{ force = false, skipLock = false }: { force: boolean; skipLock?: boolean },
 	fn: () => Resolvable<T>,
 ): Bluebird<T> {
-	if (skipLock) {
+	if (skipLock || appId == null) {
 		return Bluebird.resolve(fn());
 	}
 
 	const takeTheLock = () => {
-		if (appId == null) {
```
pipex, line 140: Since you removed this, the `if` in line 252 is no longer necessary right?

src/lib/update-lock.ts
```diff
@@ -152,20 +103,41 @@ export function lock<T extends unknown>(
...
+													return lockfile.unlock(tmpLockName);
+												}
 											})
-											.catchReturn(ENOENT, undefined);
+												.then(() => {
+													return lockfile.lock(tmpLockName);
+												})
+												// If lockfile exists, dispose of writeLock and throw a user-friendly error
+												.catch(lockfile.LockfileExistsError, () => {
+													return dispose(release).throw(
```
pipex, line 117: Why is the `dispose` needed here if there is a disposer set at the end of the promise chain?

src/lib/update-lock.ts
```diff
@@ -152,20 +103,41 @@ export function lock<T extends unknown>(
...
+														new UpdatesLockedError(
+															`Lockfile exists for ${JSON.stringify({
+																serviceName,
+																appId,
+															})}`,
+														),
+													);
+												})
+												// Else, dispose of writeLock and throw error as-is
```
pipex, line 126: Do we need this extra `catch`? What happens if you don't add this? Where is the error handled?

src/lib/update-lock.ts
pipex: Depends, can those devices do a supervisor update?

test/src/lib/update-lock.spec.ts
```diff
@@ -109,152 +105,168 @@ describe('lib/update-lock', () => {
...
+			// so mock it here as we don't have access to the binary in the test env
 			// @ts-ignore
-			unlockSyncStub = stub(lockFile, 'unlockSync').callsFake((lockPath) => {
-				// Throw error on process.exit for one of the two lockpaths
-				if (lockPath === testLockPaths[1]) {
-					throw new Error(
-						'handled unlockSync error which should not crash test process',
-					);
-				}
+			execStub = stub(fsUtils, 'exec').callsFake(async (command, opts) => {
```
pipex, line 129: I think it would be enough to stub the `lockfile` instead of stubbing `exec` here. The internals of the `lockfile` module have already been tested in the other file and here we are testing that this module responds well to that interface.

If we could test with the actual binary (e.g. within a container) we would not need any stubs and we could be sure that both modules behave as expected with a real filesystem, but that's a future improvement :)

pipex: On second though, this might be too much of a nitpick, just think about whether it simplifies your tests.

test/src/lib/update-lock.spec.ts
cywang117: Updated tests have removed the dependency on `rewire`.

test/src/lib/update-lock.spec.ts
cywang117: This sounds like a good improvement for when we move or add integration testing in an environment with the OS (balenaOS-in-container, eventually testbot), as discussed.

entry.sh
cywang117: Done, will be in the next force-push

src/lib/update-lock.ts
cywang117: After discussing, the paths in that link are the lock paths from the user app POV, so they are not relevant here. The paths on host will be either `/tmp/balena-supervisor/services/$APP_ID/$SERVICE_NAME/*updates.lock`. `entry.sh` already symlinks `/tmp/resin-supervisor` to `/tmp/balena-supervisor`, if this older path exists.

src/lib/lockfile.ts
cywang117: Yeah, the service name and app ID that are included in the lockfile error message are in  the scope of the parent function, so no longer need to be parsed from the error message thrown here. Updated this in the next force push.

src/lib/update-lock.ts
cywang117: The disposer is a meta method that only calls the function provided to `.disposer` when invoked in the context of `Bluebird.using`. However, when an error is thrown in `takeTheLock`, `Bluebird.using` is never reached in updateLock.lock.. so an additional call to `dispose(release)` is needed outside of `disposer(dispose)`. I changed it to be a bit more concise though, by attaching a `.finally(() => dispose(release)` to the end of the inner `Bluebird.mapSeries` like so:
```
return Bluebird.mapSeries(
	lockFilesOnHost(appId, serviceName),
	(tmpLockName) => {
        // abbreviated for brevity 
	}
).finally(() => dispose(release));
```

it'd be nice to move away from this nested structure though, because it's not easy to follow. The only thing preventing this is that there are some Bluebird-specific features used here that aren't present with native Promises..

src/lib/update-lock.ts
cywang117: Before I added it because of the `dispose(release)` call, but after moving the `dispose(release)` call to `.finally` this 2nd catch is not needed anymore.

src/lib/update-lock.ts
cywang117: I'm not sure which file you're referring to?

test/src/lib/update-lock.spec.ts
cywang117: Removed lockfileErrHandler and explicit tests for the `dispose` method, since it's an implementation detail.

test/src/lib/update-lock.spec.ts
cywang117: Changed to `find "/mnt/root${BASE_LOCK_DIR}" -type f -user "${LOCKFILE_UID}" -name "*updates.lock" -delete` which avoids any errors with piping `find` to `rm`

src/lib/update-lock.ts
cywang117: No longer parsing the error message, we're just checking if the thrown error is an instance of LockfileExistsError or not.

src/lib/update-lock.ts
cywang117: Changed to performing this `find` in entry.sh

src/lib/update-lock.ts
cywang117: No longer parsing the error 💯

src/lib/update-lock.ts
```diff
@@ -152,22 +110,38 @@ export function lock<T extends unknown>(
...
+												.then(() => {
+													return lockfile.lock(tmpLockName);
+												})
+												// If lockfile exists, throw a user-friendly error.
+												// Otherwise throw the error as-is.
+												// This will interrupt the call to Bluebird.using, so
+												// dispose needs to be called even though it's referenced
+												// by .disposer later.
+												.catch((error) => {
+													return dispose(release, appId).throw(
```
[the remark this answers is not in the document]
cywang117: Additionally, passing `appId` was the way I found to make lock disposing multi-app aware. The last 2 unit tests in `lockfile.spec.ts` test that the underlying functionality that only lists locks of the `appId` passed here when calling `lockfile.getLocksTaken`.

This resolves https://github.com/balena-os/balena-supervisor/pull/1893#discussion_r833746558 which for some reason I can't reply to anymore.

cywang117 (commented):
@pipex @20k-ultra, Ready for another look. Addressed all the comments:

- https://github.com/balena-os/balena-supervisor/pull/1893#discussion_r844495750
- `lockfile.getLocksTaken` / dispose functionality is multi-app compatible in 6a697bcab139b731339a67f2ea46958fab7679f8
- Cleaned up some inefficiencies with 2 `.catch`'s in `updateLock.lock`

Pending questions - https://github.com/balena-os/balena-supervisor/pull/1893#discussion_r833750065

Thanks!

src/lib/update-lock.ts
pipex: I meant the `if` in

```
if (disposer) {
  return Bluebird.using(disposer, fn as () => PromiseLike<T>);
} else {
  return Bluebird.resolve(fn());
}
```

Since `takeTheLock` never returns undefined, then you don't need this branching.

src/lib/lockfile.ts
```diff
@@ -0,0 +1,117 @@
+import * as fs from 'fs';
+import { dirname } from 'path';
+import { isRight } from 'fp-ts/lib/Either';
+
+import { exec, unlinkAll } from './fs-utils';
+import { NumericIdentifier } from '../types';
+
+// Equivalent to `drwxrwxrwt`
+const STICKY_WRITE_PERMISSIONS = 0o1777;
+export const BASE_LOCK_DIR =
```
pipex, line 10: I think this constant needs to go on `update-lock.ts` as this module provides generic lock functionality independent from the supervisor.

src/lib/lockfile.ts
```diff
@@ -0,0 +1,117 @@
...
+			 * indicating the lockfile already exists. Any other error's child process
+			 * code should be included in the error message to more clearly signal
+			 * what went wrong. Other errors that are not the typical "file exists"
+			 * errors include but aren't limited to:
+			 *   - running out of file descriptors
+			 *   - binary corruption
+			 *   - other systems-based errors
+			 */
+			throw new Error(
+				`Got code ${(error as ChildProcessError).code} while locking updates: ${
```
pipex, line 100: This module is a generic lockfile implementation so it should not reference update locks or `update.lock`

```suggestion
				`Got code ${(error as ChildProcessError).code} while trying to take lock: ${
```

src/lib/lockfile.ts
```diff
@@ -0,0 +1,117 @@
...
+					(error as ChildProcessError).stderr
+				}`,
+			);
+		}
+	}
+}
+
+export async function unlock(path: string): Promise<void> {
+	// Removing the updates.lock file releases the lock
```
pipex, line 109: Same here, avoid referencing `updates.lock` inside this module

cywang117: Addressed all comments with 0e13ee16e6758b1c4a549667b030e804e8bfee66! I'll probably squash when merging.

src/lib/lockfile.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export async function lock(path: string, uid = LOCKFILE_UID) {
+	/**
+	 * Set parent directory permissions to `drwxrwxrwt` (octal 1777), which are needed
+	 * for lockfile binary to run successfully as the `nobody` (UID 65534) user.
+	 * Otherwise the `nobody` user will not have write permissions to the necessary
+	 * bind-mounted lockfile directories. `chmod` does not fail or throw if the
+	 * directory already has the proper permissions.
+	 */
+	await fs.promises.chmod(dirname(path), STICKY_WRITE_PERMISSIONS);
```
pipex, line 70: Maybe this needs to go into `update-lock.ts` as well, this does not make much sense in a generic module. 

Same thing with the `uid` argument, it should probably default to `os.userInfo().uid` and it is the responsibility of the caller to make sure that the given uid has write permissions on the directory.

src/lib/lockfile.ts
```diff
@@ -0,0 +1,115 @@
...
+import { isRight } from 'fp-ts/lib/Either';
+
+import { exec, unlinkAll } from './fs-utils';
+import { NumericIdentifier } from '../types';
+
+// Equivalent to `drwxrwxrwt`
+const STICKY_WRITE_PERMISSIONS = 0o1777;
+
+const decodedUid = NumericIdentifier.decode(process.env.LOCKFILE_UID);
+export const LOCKFILE_UID = isRight(decodedUid) ? decodedUid.right : 65534;
```
pipex, line 12: As I mention in a comment below, maybe this should also go into `update-lock.ts` as the generic module should not presume a

pipex:
Nice work @cywang117! I think my last nitpick is the `LOCKFILE_UID` that should go into `update-lock.ts`, other than that no more comments :)

src/lib/lockfile.ts
cywang117: I made the default uid `os.userInfo().uid`, but I'm not sure about moving `STICKY_WRITE_PERMISSIONS` outside of lockfile.ts when it's only referenced in lockfile.ts. I settled on
```
if (uid !== 0) {
    await fs.promises.chmod(dirname(path), STICKY_WRITE_PERMISSIONS);
}
```
But otherwise moved BASE_LOCK_DIR and LOCKFILE_UID consts outside of lockfile. Thoughts?

pipex: I think this is probably fine for now, since we are only using this in the supervisor and I don't want to keep blocking you

As a generic module, the following could potentially cause some issues
- `exec('<cmd>', uid)` will only succeed if `uid == os.userInfo().uid` or `os.userInfo().uid` is a privileged user. 
- On the first case, the call to `chmod` will fail unless the user already has the right permissions to the directory, in which case the call is useless (but benign probably)
- On the second case the call to `chmod` will succeed, but now the module has unilaterally changed the permissions of a directory, what if that directory is important? 

As I said, these considerations are thinking about this as an independent module, and if we ever move this somewhere else we will have to take that into account, but for now I think this works

cywang117: When thinking about lockfile as a independent module, the directory permissions change being unexpected or even undesired makes sense. I've added this as a consideration in the comments for lockfile, for future reference. Thanks!

cywang117:
@balena-ci I self-certify!

vipulgupta2048:
Hey just checking @cywang117 if you did create that docs PR for this TODO message you have here:

> TODO, separate PR to docs repo:  Update balena docs for update locks to indicate that nobody UID is expected for Supervisor-created lockfiles as of version 12.x.y

cywang117:
Thanks for the heads up @vipulgupta2048 , I created https://github.com/balena-os/balena-supervisor/pull/2086 to document this.

PR #1900, opened 2022-03-16
update packages for vulnerabilities

20k-ultra:
Seeing dependabot PRs made me want to try and update out deps. Running npm update broke tests but npm audit fix did not.


20k-ultra:
@balena-ci I self-certify!

PR #1901, opened 2022-03-16
Only count report connectivity errors for healthcheck

pipex:
Change-type: patch


20k-ultra:
To clarify, this issue made the healthchecks kill the Supervisor for any error caused by reporting current state. We however only want to indicate the Supervisor is unheathly if the error is not a StatusError which are HTTP errors. This is because if the Supervisor is getting an HTTP response from the API, having the healthchecks kill the SV will not fix that. The fact the Supervisor is reaching the API and getting a response means that it is working correctly.

20k-ultra:
gh actions are having an incident atm btw https://www.githubstatus.com/

20k-ultra (approved):

PR #1903, opened 2022-03-16
Moved test setup into file included for all tests

20k-ultra:
just doing some test cleanup


20k-ultra:
@balena-ci I self-certify!

Issue #1905, opened 2022-03-21
Supervisor patches device state without status in service body


fisehara:
Recently the API faces a device state patch that sends a body with a missing service `status` property which is a mandatory field.

```
{
    "local": {
        "apps": {
            "blacked": {
                "services": {
                    "blacked": {
                        "download_progress": None,
                        "releaseId": blacked,
                        "status": "Running"
                    },
                    "blacked": {
                        "download_progress": None,
                        "releaseId": blacked,
                        "status": "Running"
                    },
                    "blacked": {
                        "download_progress": None,
                        "releaseId": blacked,
                        "status": "Running"
                    },
                    "blacked": {
                        "download_progress": None,
                        "releaseId": blacked,
                        "status": "Running"
                    },
                    "blacked": {
                        "releaseId": blacked,
                        "status": "Downloaded"
                    },
                    "blacked": {
                        "releaseId": blacked,
                        "status": "Downloaded"
                    },
>                    "blacked": {
>                        "download_progress": 48,
>                        "releaseId": blacked
>                    },
                    "blacked": {
                        "releaseId": blacked,
                        "status": "Downloaded"
                    },
                    "blacked": {
                        "releaseId": blacked,
                        "status": "Downloaded"
                    },
                    "blacked": {
                        "releaseId": blacked,
                        "status": "Downloaded"
                    },
                    "blacked": {
                        "releaseId": blacked,
                        "status": "Downloaded"
                    }
                }
            }
        },
        "cpu_temp": 50,
        "cpu_usage": 3,
        "memory_usage": 484
    }
}
```
Internal data have been blacked.

cywang117:
Closed with https://github.com/balena-os/balena-supervisor/pull/1908

PR #1908, opened 2022-03-21
Always add status to image download report

pipex:
It seems that in some cases the supervisor can report
an image without a `status` field leading to a cloud side 401 response. This should fix the issue even 
though we don't have a clear replication.

Relates to: #1905

Change-type: patch

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -193,7 +193,11 @@ export async function triggerFetch(
 	}
 
 	const onProgress = (progress: FetchProgressEvent) => {
-		reportEvent('update', { ...image, downloadProgress: progress.percentage });
+		reportEvent('update', {
+			...image,
+			downloadProgress: progress.percentage,
+			status: 'Downloading',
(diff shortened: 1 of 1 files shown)


20k-ultra (approved):
I have spent most of my day reviewing the code which sets the status for a service. The application manager loops over all the services and images installed and sets the status during these 2 loops. This makes it a bit harder to understand where the status comes from. I focused more on the image loop though since in the issue's log it said the service was downloading.

The code for an image status being set is very difficult to follow but it uses Tasks which combine data into the state [here](https://github.com/balena-os/balena-supervisor/blob/master/src/compose/images.ts#L99-L103). This combination makes it impossible for the status to be removed since you can't unset a key via object de-structuring. Once the status gets set it will only be updated.

I am really confused how this change will fix the issue but I cannot understand how the issue even exists even after testing on device several situations of images being downloaded and watching what the state is. I cannot for see this being an issue so will approve.

PR #1910, opened 2022-03-29
Added `BALENA_SUPERVISOR_REPORT_INTERVAL` config

20k-ultra:


src/api-binder/index.ts
```diff
@@ -265,13 +265,27 @@ export async function provisionDependentDevice(
...
+	// Store reporting cancelation if reportInterval changes
+	let reporting = await startReporting(reportInterval);
+
+	// If stateReportInterval changed then reset reporting interval
+	config.on('change', async (changedConfig: any) => {
+		if (changedConfig.stateReportInterval != null) {
+			// Clear previous reporting state
+			reporting.cancel();
+			// Set reporting again with new value
+			reporting = await startReporting(changedConfig.stateReportInterval);
```
20k-ultra, line 286: From my manual testing I saw that if we cancel the reporting and start again than a report is sent right away. We may want to change this so that when cancelled we can somehow ensure the next report started respected the old or new interval (probably new interval)

20k-ultra:
This will be pending a product call to discuss https://github.com/balena-os/balena-supervisor/pull/1910#discussion_r837985419

src/api-binder/index.ts
20k-ultra: The product call decided to keep this behavior of sending a report right away https://jel.ly.fish/brainstorm-topic-6597323c-d100-4205-a257-7713735c0a58

20k-ultra:
I plan to get https://github.com/balena-os/balena-supervisor/pull/1936 merged so than the Supervisor just uses an in memory value to reference for the delay between reports. Clearing timeouts will no longer be necessary.

PR #1912, opened 2022-03-31
Remove restrictions on docker env vars

pipex:
Docker doesn't impose any restrictions on the contents of an env
var, and neither does the CLI and API so it doesn't make much sense
for the env var to be as restrictive.

Change-type: patch

### DIFF
--- a/src/types/basic.ts
+++ b/src/types/basic.ts
@@ -82,13 +82,20 @@ export const NumericIdentifier = new t.Type<number, StringOrNumber>(
 export type NumericIdentifier = t.TypeOf<typeof NumericIdentifier>;
 
 /**
- * Valid variable names are between 0 and 255 characters
- * and match /^[a-zA-Z_][a-zA-Z0-9_]*$/
+ * Valid variable names are between 0 and 255 characters and do not contain `=`
+ *
+ * Docker has no restriction on what an environment variable name can be
+ * the shell may not be able to read non alphanumeric variable names,
+ *  see https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html,
+ * but other processes (e.g. Node) may be able to read them.
+ *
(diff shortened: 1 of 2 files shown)


pipex:
Setting this as draft pending outcome of discussion in https://www.flowdock.com/app/rulemotion/r-beginners/threads/6Tq8dB8RVoOPCdl07X3ufYR7wTO

The main drawback with this is that, although docker does not restrict the variable name, the entrypoint shell might so users might find that their variable still does not show in their process. For instance, for this env var config in a `balenalib/intel-nuc-node` image

```
root@bb48ddc:~# balena inspect 
 | jq .[].Config.Env
[
  "3333=lalaa",
  "RESIN_DEVICE_NAME_AT_INIT=long-flower",
  "BALENA_DEVICE_NAME_AT_INIT=long-flower",
  "a:bbb=lalaa",
  "🙄=lalaa",
  "aaa\nbbb=lalaa",
  "HELLO=goodbye",
  "BALENA_APP_ID=1920094",
...
```

Looking at the shell environment variables we don't see the extra variables, which means any process using this `sh` as entrypoint (e.g.) won't get the variables either

```
root@bb48ddc:~# balena exec -ti  80470d93df30  sh
# env
BALENA_DEVICE_TYPE=intel-nuc
USER=root
HELLO=goodbye
BALENA_APP_ID=1920094
NODE_VERSION=10.24.1
HOSTNAME=80470d93df30
```

However if we run the `node` binary directly

```
root@bb48ddc:~# balena exec -ti  80470d93df30  node
> process.env
{ PATH:
   '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
  HOSTNAME: '80470d93df30',
  TERM: 'xterm',
  '3333': undefined,
  RESIN_DEVICE_NAME_AT_INIT: 'long-flower',
  BALENA_DEVICE_NAME_AT_INIT: 'long-flower',
  'a:bbb': 'lalaa',
  '🙄': 'lalaa',
  'aaa\nbbb': 'lalaa',
  HELLO: 'goodbye',
```

PR #1914, opened 2022-04-01
Fix database migration for legacyApps

pipex:
Migration `M00008` had a bug with the check for legacy apps, which
resulted in devices that had at some point been updated from a single
container supervisor to get the error

```
Undefined binding(s) detected when compiling UPDATE. Undefined column(s): [appUuid] query
```

This adds a new migration with the fix to ensure broken fix the
inconsistent database state.

Change-type: patch
Closes: #1913


pipex:
@balena-ci retest

src/migrations/M00008.js
```diff
@@ -10,45 +10,6 @@ export async function up(knex) {
...
 		table.string('commit');
 	});
-
-	const legacyAppsPresent = await knex('config')
-		.where({ key: 'legacyAppsPresent' })
-		.select('value');
-
-	// If there are legacy apps we let the database normalization function
-	// populate the correct values
-	if (legacyAppsPresent && legacyAppsPresent.length > 0) {
```
pipex, line 20: Here is the bug, if the legacy migration ran, then`legacyAppsPresent` would be set to `'false'`

src/migrations/M00009.js
```diff
@@ -0,0 +1,45 @@
+export async function up(knex) {
+	const legacyAppsPresent = await knex('config')
+		.where({ key: 'legacyAppsPresent' })
+		.select('value')
+		.first();
+
+	// If there are legacy apps we let the database normalization function
+	// populate the correct values
+	if (legacyAppsPresent && legacyAppsPresent.value === 'true') {
```
pipex, line 9: Now we explicitly check for the value being `true`

pipex:
@balena-ci I self certify!

Issue #1916, opened 2022-04-06
Core dumps filling the data partition on Compulab IOT-gate-imx8 boards


majorz:
In d06b8e0 we started using `dmidecode` to obtain system ID, but it is crashing and creating core dumps on `Compulab IOT-gate-imx8` boards. Since the crashes are happening repeatedly, eventually the whole data partition fills up with core dumps and the device becomes dysfunctional.

Running a newer `dmidecode` on the device does not crash. I tried version 3.3 from Alpine. It does not return a system ID though:
```
# dmidecode 3.3
# No SMBIOS nor DMI entry point found, sorry.
```

That device does not have `/proc/device-tree/serial-number` exposed, which leads to running `dmidecode`. On that device a `/proc/device-tree/product-sn` exists though together with some related fields:
```
/proc/device-tree/product-name 
UCM-iMX8M

/proc/device-tree/product-options 
C1800Q-D2-N16-E

/proc/device-tree/product-revision 
1.11

/proc/device-tree/product-sn 
5a1066895
```

It looks like we need to upgrade `dmidecode` and possibly use `/proc/device-tree/product-sn` as alternative to `/proc/device-tree/serial-number`.

PR #1917, opened 2022-04-06
Add system id/model support for Compulab IOT-gate

pipex:
dmidecode for alpine 3.11 doesn't work in this device type. This change
moves to using `/proc/device-tree/product-sn` and
`/proc/device-tree/product-name` for these devices.

Resolves: #1916
Change-type: patch

### DIFF
--- a/src/lib/system-info.ts
+++ b/src/lib/system-info.ts
@@ -3,6 +3,10 @@ import * as _ from 'lodash';
 import * as memoizee from 'memoizee';
 import { promises as fs } from 'fs';
 
+// TODO: remove this once we can update Node to v16 and typescript
+// to an es2021 compatible version for native Promise.any
+import { Promise } from 'bluebird';
+
 import { exec } from './fs-utils';
 
 export async function getCpuUsage(): Promise<number> {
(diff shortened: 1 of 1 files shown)


src/lib/system-info.ts
```diff
@@ -3,6 +3,10 @@ import * as _ from 'lodash';
 import * as memoizee from 'memoizee';
 import { promises as fs } from 'fs';
 
+// TODO: remove this once we can update Node to v16 and typescript
+// to an es2021 compatible version for native Promise.any
+import { Promise } from 'bluebird';
```
cywang117, line 8: `Promise.any` should be available natively, no?

cywang117: Otherwise LGTM!

20k-ultra: available in node 15 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any#browser_compatibility

20k-ultra (approved):

src/lib/system-info.ts
cywang117: Ahh, oh no

Issue #1918, opened 2022-04-07
Supervisor reports a service downloading without any download progress


20k-ultra:
I suspected https://github.com/balena-os/balena-supervisor/commit/21c1c006f7b8dfcf566f3e4a8e853e5e9978edf4 to be the culprit but I found the behavior to persist even before that change.

![image](https://user-images.githubusercontent.com/3946250/162286470-9ccc2e70-6a9a-431f-8aea-6c3005473465.png)

![image](https://user-images.githubusercontent.com/3946250/162286802-f60c0ea1-afc5-44d5-a296-a4981185ef17.png)

20k-ultra:
This issue was introduced in v13.0.0

Issue #1898, opened 2022-03-14
e2e tests for regressions


20k-ultra:
List for regressions to try and get e2e testing to avoid from happening again:

https://github.com/balena-os/balena-supervisor/issues/1895

cywang117:
https://github.com/balena-os/balena-supervisor/pull/1928 (v13.1.1) which fixes an incorrect exit in entry.sh introduced by https://github.com/balena-os/balena-supervisor/pull/1893 (v13.1.0).

cywang117:
https://github.com/balena-os/balena-supervisor/pull/1944 (v13.1.? //TODO, add version) which allows preloading v2 state apps without internet to proceed. Issue detailed here: https://github.com/balena-os/balena-supervisor/issues/1943

Note that when preloaded devices reconnect to the internet, user containers will restart to reflect the appUuid data from target state.

pipex:
#1959 we added name as part of v13 current state report which caused a race with the cloud side device name

cywang117:
https://github.com/balena-os/balena-supervisor/pull/1980 which fixes host config patch failing with no running apps, introduced in [v14.0.2](https://github.com/balena-os/balena-supervisor/blob/master/CHANGELOG.md#v1402) and resolved with this PR in v14.0.9.

PR #1920, opened 2022-04-08
Correctly evaluate downloadProgress when computing current state

20k-ultra:
Change-type: patch
Closes: #1918
Signed-off-by: 20k-ultra <3946250+20k-ultra@users.noreply.github.com>

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


20k-ultra:
@balena-ci I self-certify!

Issue #1668, opened 2021-04-21
Evaluate container's restart policy and exit status before starting


20k-ultra:
If a container is stopped either from the dashboard (which uses the Supervisor API) or on device via the cli, the Supervisor should not start a stopped container just because the target state wants it running. The target state actually never changes the running state to stopped (there is no way to make it stopped) which makes respecting unless-stopped impossible for the current Supervisor behaviour.

The Supervisor behaves this way because it remembers which containers it has started so on subsequent target state polls, even though the running state wants the stopped container running, the Supervisor won't try to start it because it has already started it. This mechanism fails when the Supervisor is restarted and thus losing memory of the start.

This mechanism could be improved to evaluate the container it wants to start by noting the restart policy and exit code. For example, if the restart policy was unless-stopped and the exit code is 0 then we know the container was stopped by a user and we shouldn't start it again. Additionally, if the restart policy is no the exit code doesn't matter, etc for each policy. 

tl;dr how Supervisor will evaluate stopped containers:

- unless-stopped: if exit code is 0 then start
- always: exit code is irrelevant, always start
- on-failure if exit code is 1 then start
- no: exit code is irrelevant, never start

https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy

pipex:
I think this is a bit more tricky than that @20k-ultra, what you are describing is already implemented by the engine, so we effectively could be racing against the engine.

What happens if the container was deleted? The exit code also persists after a reboot, so the supervisor should never start a stopped container? I agree there is a product discussion we want to have, because there is a pattern here, but we might have to think about the implications of this for other users.

pwt:
> What happens if the container was deleted?

Under what conditions is a container deleted?

> The exit code also persists after a reboot, so the supervisor should never start a stopped container?

Correct. It should comply with the restart semantics specified in the compose file, and should match how Docker would behave.

20k-ultra:
@pipex I think you're misunderstanding. The Supervisor will ignore restart policies and start containers that the engine isn't trying to start so there's no race condition. In the case where the engine does want to restart a container that race condition exists today because the Supervisor wants to always start containers and with a restart: always policy the engine will too. The key difference is that we shouldn't start if the policy is configured to not want that.

If the container was deleted then the Supervisor wouldn't be trying to decide if it should start it since it would be trying to create it. I'd hope exit codes persist after a reboot so that the Supervisor and engine continue to respect the restart-policy specified.

[...]

PR #1926, opened 2022-04-12
Prevent restart policy violation

20k-ultra:
See https://github.com/balena-os/balena-supervisor/issues/1668 for discussion.

This PR solves that issue by evaluating the status. This works because the only time we want to start a container is when it is `Installed`. This is because any other state will allow the engine to start the container based on the restart policy.

After that change I saw I could cleanup some of the in memory objects used for tracking what containers have been started or stopped. This is no longer necessary since the Supervisor uses the container data.

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -8,7 +8,6 @@ import Service from './service';
 
 import * as imageManager from './images';
 import type { Image } from './images';
-import * as applicationManager from './application-manager';
 import {
 	CompositionStep,
 	generateStep,
@@ -358,9 +357,7 @@ export class App {
 			// check that we want to stop it, and that it isn't stopped
 			return (
 				serviceTarget.config.running === false &&
-				// When we issue a stop step, we remove the containerId from this structure.
-				// We check if the container has been removed first, so that we can ensure we're not providing multiple stop steps.
-				applicationManager.containerStarted[serviceCurrent.containerId!] == null
+				serviceCurrent.status !== 'Stopped'
 			);
 		};
 
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -336,15 +336,13 @@ export class App {
 				return false;
 			}
 
-			// Check if we previously remember starting it
-			if (
-				applicationManager.containerStarted[serviceCurrent.containerId!] != null
-			) {
-				return false;
-			}
-
-			// If the config otherwise matches, then we should be running
-			return isEqualExceptForRunningState(serviceCurrent, serviceTarget);
+			// Only start a Service if we have never started it before and the service matches target!
+			// This is so the engine can handle the restart policy configured for the container.
+			return (
+				(serviceCurrent.status === 'Installing' ||
+					serviceCurrent.status === 'Installed') &&
+				isEqualExceptForRunningState(serviceCurrent, serviceTarget)
+			);
 		};
 
 		/**
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -74,9 +74,6 @@ export const router = (() => {
 	return $router;
 })();
 
-// We keep track of the containers we've started, to avoid triggering successive start
-// requests for a container
-export let containerStarted: { [containerId: string]: boolean } = {};
 export let fetchesInProgress = 0;
 export let timeSpentFetching = 0;
 
@@ -94,12 +91,6 @@ export function resetTimeSpentFetching(value: number = 0) {
 const actionExecutors = getExecutors({
 	lockFn: lock,
 	callbacks: {
-		containerStarted: (id: string) => {
-			containerStarted[id] = true;
-		},
-		containerKilled: (id: string) => {
-			delete containerStarted[id];
-		},
 		fetchStart: () => {
 			fetchesInProgress += 1;
 		},
@@ -347,9 +338,6 @@ export async function stopAll({ force = false, skipLock = false } = {}) {
 		services.map(async (s) => {
 			return lock(s.appId, { force, skipLock }, async () => {
 				await serviceManager.kill(s, { removeContainer: false, wait: true });
-				if (s.containerId) {
-					delete containerStarted[s.containerId];
-				}
 			});
 		}),
 	);
--- a/src/compose/composition-steps.ts
+++ b/src/compose/composition-steps.ts
@@ -128,8 +128,6 @@ type LockingFn = (
 interface CompositionCallbacks {
 	// TODO: Once the entire codebase is typescript, change
 	// this to number
-	containerStarted: (containerId: string | null) => void;
-	containerKilled: (containerId: string | null) => void;
 	fetchStart: () => void;
 	fetchEnd: () => void;
 	fetchTime: (time: number) => void;
@@ -155,7 +153,6 @@ export function getExecutors(app: {
 						removeContainer: false,
 						wait,
 					});
-					app.callbacks.containerKilled(step.current.containerId);
 				},
 			);
 		},
@@ -168,7 +165,6 @@ export function getExecutors(app: {
 				},
 				async () => {
 					await serviceManager.kill(step.current);
-					app.callbacks.containerKilled(step.current.containerId);
 				},
 			);
 		},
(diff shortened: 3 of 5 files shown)


src/compose/app.ts
```diff
@@ -355,9 +355,7 @@ export class App {
 			// check that we want to stop it, and that it isn't stopped
 			return (
 				serviceTarget.config.running === false &&
-				// When we issue a stop step, we remove the containerId from this structure.
-				// We check if the container has been removed first, so that we can ensure we're not providing multiple stop steps.
-				applicationManager.containerStarted[serviceCurrent.containerId!] == null
+				serviceCurrent.status !== 'Stopped'
```
20k-ultra, line 358: I quickly made this change so might need more thought.

src/compose/app.ts
```diff
@@ -336,15 +336,10 @@ export class App {
...
-			) {
-				return false;
-			}
-
-			// If the config otherwise matches, then we should be running
-			return isEqualExceptForRunningState(serviceCurrent, serviceTarget);
+			// Only start a Service if we have never started it before!
+			// Otherwise, all other states regarding starting will be handled by the engine.
+			// This is so the engine can enforce the restart policy configured for the container.
+			return serviceCurrent.status === 'Installed';
```
klutchell, line 342: Service changes count as an install, correct? Like changing env vars or updating images?

What about service restarts via the dashboard, does that count as an install?

I'm sure the answer is obvious if I dug through the code but it might be helpful to mention here as a comment for context.

pipex: Yeah, all of those operations recreate the container so they have an intermediate `Install` step @klutchell 

I wonder if this should include `Dead` containers, not sure if the engine restarts those.

src/compose/app.ts
pipex: That sounds correct, although this will only work for now while we don't have a "stopped" target. If we did, this would cause the supervisor to keep stopping the container, despite the user starting it through the command line or API, for instance.

The correct way to have stopped targets will be to use `finishedAt` 

Just leaving the comment for posterity (although we could add a comment in the code).

pipex:
This change looks great, and I love all the red :). Once you are confident about the testing in different scenarios (remember to test the dashboard service actions), I think we can merge.

src/compose/app.ts
20k-ultra: `Dead` containers cannot be started again https://github.com/docker/cli/issues/502#issuecomment-330361748

> What about service restarts via the dashboard, does that count as an install?

It's possible this change might break that so I'm doing some testing specifically for that situation

20k-ultra:
Dashboard start/stop actions still work with this change. Restarting the Supervisor with containers stopped from the dashboard also do not cause them to be started again.

All I need to do now is figure out why the tests are failing.

20k-ultra:
The tests are failing because the [case](https://github.com/balena-os/balena-supervisor/blob/master/test/src/compose/app.spec.ts#L867) which ensures the SV starts containers depended on first is not happy. 

So far, I have observed that new services not in the current state follow a different code path to determine if they should start. This is because the `shouldBeStarted` function modified in this PR is using existing docker resources, meaning it uses the current state and new non-existent containers simply check if they should download or start.

In that logic of download or start, the Supervisor checks if the dependencies are met or the Services. This is where you'd expect that function to return false because in the test case the container is [installing](https://github.com/balena-os/balena-supervisor/blob/master/test/src/compose/app.spec.ts#L884) but it returns true! This makes the next steps produce a step to start a container that has dependencies not running which the tests catch (nice).

For some experimenting, I changed the depended on service from `Installing` to `Installed` state and the tests pass. This is because the shouldBeStarted returns true. Also interestingly, with the old logic, shouldBeStarted would return true for a service that is in "Installing" status.

src/compose/app.ts
```diff
@@ -336,15 +335,10 @@ export class App {
...
-			if (
-				applicationManager.containerStarted[serviceCurrent.containerId!] != null
-			) {
-				return false;
-			}
-
-			// If the config otherwise matches, then we should be running
-			return isEqualExceptForRunningState(serviceCurrent, serviceTarget);
+			// Only start a Service if we have never started it before!
+			// Otherwise, all other states regarding starting will be handled by the engine.
```
cywang117, line 339: What is meant by "all other states regarding starting"?

20k-ultra: I mean that there are N possible status states and the only time we want to start a container is when it is Installed. Otherwise, all other states will be determined by the engine if it should be started.

I agree it's not clear >.<

20k-ultra:
So, the Supervisor builds a list of [removePairs, updatePairs, and installPairs](https://github.com/balena-os/balena-supervisor/blob/master/src/compose/app.ts#L122). These are services to remove, update, and install grouped into pairs of current and target.

Remove pairs are just removed right away. The update and install pairs though get concatenated into a single list of pairs. 

 These pairs go through [generateStepsForService](https://github.com/balena-os/balena-supervisor/blob/81a8705ff83e7f9d48573f911b980e92f88ef445/src/compose/app.ts#L450) which basically checks various states on the current and target and adds a step to reach the target...for example, if the current service is stopping a noop step is added. If target needs to be downloaded and it's downloading then return noop step.

Eventually, it gets to the main logic where it branches into 3 paths:
 1. If target and current configs are the same it checks next steps.
 2. else if current is null fetch or start current
 3. else there is current but it doesn't match target so do some special stuff (this can be ignored)

Right now the tests fail because the path for current is null (2) is ran. This path sees that the target image is already downloaded so it will create a start step but just before that it makes one more check with [dependenciesMetForServiceStart](https://github.com/balena-os/balena-supervisor/blob/81a8705ff83e7f9d48573f911b980e92f88ef445/src/compose/app.ts#L637). This function is the logic for handling `depends-on` directive in docker compose. 

I suspect the problem lies in that function because it will say the dependencies are met even though the current state has the depended on container as Installing in the [test case](https://github.com/balena-os/balena-supervisor/blob/master/test/src/compose/app.spec.ts#L884). I think [this](https://github.com/balena-os/balena-supervisor/blob/81a8705ff83e7f9d48573f911b980e92f88ef445/src/compose/app.ts#L650-L656) specific section is flawed since I observed it returning dependencies installing but not running as FALSE when it should be TRUE since the current state say the containers are installing.

### Fuzzy observations (not 100% sure on yet)

This bug? in dependenciesMetForServiceStart is masked either intentionally or not simply because the list of steps generated [concats](https://github.com/balena-os/balena-supervisor/blob/81a8705ff83e7f9d48573f911b980e92f88ef445/src/compose/app.ts#L138) installs pairs to the end of update pairs so the steps for update pairs are calculated first. This matters because this allows the Supervisor to add a start step to the updatePairs.

Finally worth noting, if I modify the `shouldBeStarted` function to return true for `Installing` (originally just `Installed`) than the tests pass. This is because the [code that generates the updatePairs](https://github.com/balena-os/balena-supervisor/blob/81a8705ff83e7f9d48573f911b980e92f88ef445/src/compose/app.ts#L372-L388) uses the `shouldBeStarted` function to determine if the service belongs in that update list. If should be started == true then it goes into updatePairs. This is what allows the quirk from up top to add a start step since it calculates steps for updatePairs first.

**it seems like the Supervisor does not wait for Installing to complete so making shouldBeStarted pass true for Installed or Installing works**

*to be continued*

src/compose/app.ts
20k-ultra: updated

20k-ultra:
I've resolved the tests by making the `shouldBeStarted` to return true for containers with `Installing` status. This seems to be the same as the previous behaviour and I did not want to modify the `dependenciesMetForServiceStart` function since it was harder to understand why it has the logic is has. Therefore, my change felt the least risky.

I had to modify 2 test cases since they explicitly **wanted** a start step for stopped containers but with this PR we made it so the Supervisor only starts containers once and the engine takes over.

Some manual on device testing will be done for:
 - start, stop, restart action on dashboard still works
 - Supervisor does not start a container if you stop via the dashboard
 - Supervisor does not start a container which has exitted (try with a container that exits instantly too)
 - Supervisor starts containers when moving to another release
 - Supervisor starts multi-containers when moving to another release
 - Supervisor starts containers which depend-on other containers
 - Supervisor starts containers when the current state has none of the target state containers
 - Supervisor starts containers when the current state has all the target state containers but meta-data changes like imageId

20k-ultra:
Check out the following service states. I modified the Supervisor to sleep for 1 minute if the serviceName is frontend. This was to confirm the depends-on logic still worked since that was a bit confusing with the code I had reviewed...it looked like that feature didn't actually work but it works correctly :)

image

image

20k-ultra:
@balena-ci I self-certify!

PR #1929, opened 2022-04-15
Avoid splash image failures if image is corrupt

pipex:
Splash image backend would throw if the image is not a valid png during
the write step. This could prevent the device from provisioning if some
corruption happens at some point.

Change-type: patch


src/config/backends/splash-image.ts
```diff
@@ -88,7 +88,9 @@ export class SplashImage extends ConfigBackend {
 			// Write the buffer to the given location
 			await remountAndWriteAtomic(where, buffer);
 		} else {
-			throw new Error('Splash image should be a base64 encoded PNG image');
+			log.error(
+				'Splash image should be a base64 encoded PNG image. This could be due to corruption. Skipping write.',
```
pipex, line 92: I'm not sure if this should skip the write or just write whatever is given. I need to test that writing garbage to the splash image does not prevent the OS from booting.

pipex: Writing whatever is given will avoid the potential supervisor loop when trying to apply target state

pipex: So, here is my reasoning on this

The supervisor writes in two places the splash image, when first initializing the backend, 
- to copy the splash-image into the splash image default
- when changing the splash image to match the target state

This is how the two choices, skipping writing or writing whatever garbage is given affect 

If the supervisor skips the write
- If the preloaded splash image is corrupted or not a png, then the supervisor will not copy it to the `balena-logo-default`, which means it will decide it needs to set the variable to the cloud side (by the logic in the line 115 of the module). The API does not restrict what can be written on the `BALENA_HOST_SPLASH_IMAGE` config var, so the target state will settle. Each supervisor restart will keep giving the warning about the splash image being the wrong format, which is good. 
-  If the image coming from the cloud side is not a png, then the supervisor will refuse to write and fail to reach the target state, falling into a retry loop. However since we are no longer throwing an exception, then the message will never make it to the dashboard logs, which will just show `Applying boot config: {"image": "...."}` in a loop

If the supervisor writes whatever is given
- If the preloaded splash image is corrupted or not a png, the supervisor will copy it anyway to `balena-logo-default.png` and warn exactly once (in the supervisor logs) about the problems with the splash image. The supervisor wont create the config var, so the user will only know the splash is wrong by looking into the device.
- If the image coming from the cloud side is not a png, the supervisor will warn once (in the journal logs, not the dashboard) and write whatever is given, settling the target state.

I think skipping the write sounds like the right choice, as it should not prevent provisioning, but it will prevent setting a bad splash value in future config variable changes.

pipex: > The API does not restrict what can be written on the BALENA_HOST_SPLASH_IMAGE config var, so the target state will settle. Each supervisor restart will keep giving the warning about the splash image being the wrong format, which is good.

Actually, a correction on this. Since failing to find the default splash image will make the `getBootConfig` method to go into the `catch` clause, then the supervisor won't create a cloud side config var, and will just keep showing 

```
Apr 18 18:40:55 b6a85ae balena-supervisor[1935]: [success] Initialised splash image backend
Apr 18 18:40:55 b6a85ae balena-supervisor[1935]: [warn]    Failed to read splash image: Error: ENOENT: no such file or directory, open '/mnt/root/mnt/boot/sp>
Apr 18 18:40:55 b6a85ae balena-supervisor[1935]: [error]   Splash image should be a base64 encoded PNG image. Skipping write.
Apr 18 18:40:55 b6a85ae balena-supervisor[1935]: [success] Initialised splash image backend
Apr 18 18:40:55 b6a85ae balena-supervisor[1935]: [event]   Event: Service install {"service":{"appId":1920270,"serviceId":1508850,"serviceName":"hello","comm>
Apr 18 18:40:56 b6a85ae balena-supervisor[1935]: [event]   Event: Service installed {"service":{"appId":1920270,"serviceId":1508850,"serviceName":"hello","co>
Apr 18 18:40:56 b6a85ae balena-supervisor[1935]: [event]   Event: Service start {"service":{"appId":1920270,"serviceId":1508850,"serviceName":"hello","commit>
Apr 18 18:40:56 b6a85ae balena-supervisor[1935]: [event]   Event: Service started {"service":{"appId":1920270,"serviceId":1508850,"serviceName":"hello","comm>
Apr 18 18:40:57 b6a85ae balena-supervisor[1935]: [debug]   Spawning journald with: chroot  /mnt/root journalctl -a -S 2022-04-18 18:40:56 -o json CONTAINER_I>
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [error]   Splash image should be a base64 encoded PNG image. Skipping write.
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [success] Initialised splash image backend
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [warn]    Failed to read splash image: Error: ENOENT: no such file or directory, open '/mnt/root/mnt/boot/sp>
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [error]   Splash image should be a base64 encoded PNG image. Skipping write.
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [success] Initialised splash image backend
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [debug]   Finished applying target state
Apr 18 18:40:58 b6a85ae balena-supervisor[1935]: [success] Device state apply success
```

On the supervisor logs, each time the target state is applied.

pipex:
@balena-ci rebase

src/config/backends/splash-image.ts
cywang117: Skipping the write sounds right to me as well, and I like how those logs look. Between this change and https://github.com/balena-os/balena-supervisor/pull/1877 it looks like there are enough failsafes in place to ensure that the core functionality of devices won't be affected by splash image errors. When a Supervisor with both of these changes releases, we should make a support PSA.

20k-ultra: I agree with the sentiment here but just need clarification on:

> If the image coming from the cloud side is not a png, then the supervisor will refuse to write and fail to reach the target state, falling into a retry loop. However since we are no longer throwing an exception, then the message will never make it to the dashboard logs, which will just show Applying boot config: {"image": "...."} in a loop

Are you saying `Splash image should be a base64 encoded PNG image. Skipping write.` will not make it to the dashboard ? 

To be honest, if it doesn't this seems like such a rare case that we're continuously improving that seems ok.. but not ideal. However, if this message does make it to the dashboard does that guarantee less confused users and less support tickets ? I'm not sure (read doubtful)

pipex: yes, that message won't make it to the dashboard, because is the `device-config` module that generates system messages. This is a log message that only makes it to the journal logs. 

When we start working on the logging block as a replacement for the supervisor logs functionality, that distinction won't exist, so depending on the loglevel, errors will make it to the dashboard.

pipex: I agree having the logs loop on `Applying target state` is not ideal, but that is kind of the standard behavior of all configuration backends. I don't want to add to the hairball by making the configuration backend call `logSystemMessage`

pipex:
@balena-ci rebase

pipex:
@balena-ci I self certify!

PR #1931, opened 2022-04-18
Explain /v2/state/status's status field in its response

cywang117:
The status field does not indicate the current state of the Engine on the device.
Rather, it only indicates whether the HTTP response returned successfully or not.

To get the current state of the device, i.e. whether it has applied target state or not,
use the `appState` field in the response.

Change-type: patch

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -747,6 +747,8 @@ Response:
 
 This will return a list of images, containers, the overall download progress and the status of the state engine.
 
+> **Note:** The `status` field in the JSON response does not indicate the status of the device itself, but rather the status of the HTTP response. To get an accurate representation of whether the device state has settled, use the `appState` field. `appState` will be one of `applying` or `applied`.
+
 From an app container:
 ```bash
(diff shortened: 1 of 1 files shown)


cywang117:
A user has been using /v2/state/status's `status` field to determine when a device is safe to shutdown following provision. However, `status` will always be success if the request to /v2/state/status is successful, and not an accurate indicator of whether device state has settled and the device is safe to turn off.

docs/API.md
```diff
@@ -784,6 +784,8 @@ Response:
 }
 ```
 
+> **Note:** The `status` field in the JSON response does not indicate the status of the device itself, but rather the status of the HTTP response. To get an accurate representation of whether the device state has settled, use the `appState` field. `appState` will be one of `applying` or `applied`.
```
pipex, line 787: Maybe this should go before the example, it will make it more likely than people read it.

cywang117: Sure thing.

cywang117:
@balena-ci I self-certify!

Issue #1933, opened 2022-04-21
Supervisor should use volume long syntax for labels


cywang117:
Short vs long syntax: https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes

Currently, balena-compose-parse supports version 2.4, but everywhere else still does not. Per https://github.com/balena-os/balena-supervisor/pull/1780, we want to move to using Mounts instead of Binds, to avoid instances of https://github.com/balena-os/balena-supervisor/issues/1231. In [addFeaturesFromLabels](https://github.com/balena-os/balena-supervisor/blob/104a8006fbab4bad0f2da923eade0e64002dd59f/src/compose/utils.ts#L321), we add volumes derived from balena's docker-compose labels with the short syntax only. We should use the long syntax here instead, since some labels involve bind mounts to a socket. This will require adding support for long syntax.

This will allow us to distinguish between labels that should use the long syntax (Mounts API), and internal bind mounts that should stay with using the short syntax. The latter is namely for update locks:
```
/tmp/balena-supervisor/services/$APP_ID/$SERVICE_NAME:/tmp/balena
/tmp/balena-supervisor/services/$APP_ID/$SERVICE_NAME:/tmp/resin
```
which should remain as the short syntax, since bind mounts using HostConfig.Binds config auto-creates the directory on host if it does not exist, and /tmp directories will not exist following device reboots.

Flowdock: https://www.flowdock.com/app/rulemotion/r-supervisor/threads/6b66VzeJZynWcuevc4tuvCEaOrC

cywang117:
This issue can be broken up into two PRs, per @pipex:
- Add support for long volume syntax, where volumes being described using an object would go to Mount and those using a string would go to Bind (minor change)
- Use long volume syntax for supervisor label mounts (breaking change)

cywang117:
Resolved with Supervisor v14.0.0.

PR #1935, opened 2022-04-27
Add compose support for volumes defined with long syntax

cywang117:
balena-compose already supports long syntax, and with this PR, Supervisor can
have the option of using HostConfig.Mounts for internal bind mounts such as
ones added by feature labels. This will be handled in a future PR.

The only blocker to having users use long syntax is adding this feature
to the state machine. This PR does not add that feature.

Relates-to: https://github.com/balena-os/balena-supervisor/pull/1780
Relates-to: https://github.com/balena-os/balena-engine/issues/220
Relates-to: #1933
Change-type: patch

# How Has This Been Tested?

1. Place device in local mode
2. POST /v2/local/target-state with a target state containing the following compose for services.volumes:
```
{ 
  "volumes": [
      "/mnt/data:/home/bind", // should be ignored
      "named:/home/named",
      "/usr/src/app/anonymous", 
      {
          "type": "volume",
          "target": "/usr/src/app/anonymous2",
          "volume": {
              "nocopy": true
          }
      },
      {
          "type": "volume",
          "source": "named2",
          "target": "/home/named2"
      },
      {
          "type": "bind", // should be ignored
          "source": "/home/root",
          "target": "/home/bind2",
          "bind": {
              "propagation": "slave"
          }
      },
      {
          "type": "tmpfs",
          "target": "/usr/src/app/tmp",
          "tmpfs": {
              "size": 2000
          }
      },
      {
          "type": "bind", // should be invalid
          "target": "/no_source"
      },
      {
          "type": "bind", // should be invalid
          "source": "not_an_absolute_path",
          "target": "/bind"
      },
      {
          "type": "tmpfs", // should be invalid
          "source": "should_not_have_source",
          "target": "/home/tmp"
      },
      {
          "type": "invalid", // should be invalid
          "source": "test",
          "target": "test2"
      }
  ]
}
```
The expected behavior is that binds will be filtered out and log a warning, same with invalid compose definitions. Additionally, the following volume types will be exposed in the following sections of container inspect info:
- Short syntax anonymous volumes -- `Config.Volumes`
- Long syntax anonymous volumes -- `HostConfig.Mounts`
- `tmpfs` volumes defined under compose's `service.volumes` -- `HostConfig.Mounts`
- `tmpfs` volumes defined under compose's `service.tmpfs` -- `HostConfig.Tmpfs`
- Short syntax named volumes -- `HostConfig.Binds`
- Long syntax named volumes -- `HostConfig.Mounts`
 
3. When inspecting the resulting container, all is as expected:
```
root@nuc:~# balena inspect $(balena ps -aq -f name=lock) | jq -r '.[].HostConfig | {Mounts,Binds,Tmpfs}'

{                                                                                                                            
  "Mounts": [                                                                                                                
    {                                                                                                                        
      "Type": "volume",                                                                                                      
      "Target": "/usr/src/app/anonymous2",
      "VolumeOptions": {
        "NoCopy": true
      }
    },
    {
      "Type": "volume",
      "Source": "1907314_named2",
      "Target": "/home/named2"
    },
    {
      "Type": "tmpfs",
      "Target": "/usr/src/app/tmp",
      "TmpfsOptions": {
        "SizeBytes": 2000
      }
    }
  ],
  "Binds": [
    "1907314_named:/home/named",
    "/tmp/balena-supervisor/services/1907314/lock:/tmp/resin",
    "/tmp/balena-supervisor/services/1907314/lock:/tmp/balena",
  ],
  "Tmpfs": null
}

// And for Config.Volumes:
"Volumes": {                                                                                                     
    "/usr/src/app/anonymous": {}                                                                                 
}
```
Note that volumes defined with long syntax are also namespaced by appId.

Logs contain warnings for bind mounts and invalid compose definitions:
```
[warn]    Ignoring invalid bind mount /mnt/data:/home/bind
[warn]    Ignoring invalid bind mount {"type":"bind","source":"/home/root","target":"/home/bind2","bind":{"propagation":"slave"}}
[warn]    Ignoring invalid compose volume definition {"type":"bind","target":"/no_source"}
[warn]    Ignoring invalid compose volume definition {"type":"bind","source":"not_an_absolute_path","target":"/bind"}
[warn]    Ignoring invalid compose volume definition {"type":"tmpfs","source":"should_not_have_source","target":"/home/tmp"}
[warn]    Ignoring invalid compose volume definition {"type":"invalid","source":"test","target":"test2"}
```
The reasons each of the above compose definitions are invalid are commented in my changes in `service.spec.ts`.

`balena volume ls`:
```
DRIVER              VOLUME NAME
local               4372b9211cf90087fcd3bf08973be45b645547859f096fe76fd310ac86fe8974
local               1907314_named
local               1907314_named2
local               d3e096e24676688a3c8a74eaf1822e29bfb160a2df231c7debba6a15dc17f22d
```
The 2 hashes correspond to the 2 anonymous volumes declared in the compose. However, there is one quirk, in that the Supervisor will log `Found unmanaged Volume` for managed anonymous volumes:
```
[debug]   Found unmanaged Volume: d3e096e24676688a3c8a74eaf1822e29bfb160a2df231c7debba6a15dc17f22d
[debug]   Found unmanaged Volume: 4372b9211cf90087fcd3bf08973be45b645547859f096fe76fd310ac86fe8974
```
Perhaps we should change the wording to `Found unmanaged or anonymous Volume`?

Also checked that a release with balena feature labels specified will use binds with the short syntax, for now. In the next PR we can move to using long syntax.

Note that if pushing a docker-compose with invalid long syntax, such as type: volume without target, or type: volume where a separate volume stanza is not included in the docker-compose.yml, compose will error (therefore balena-compose should pick this up too, and the Supervisor shouldn't need to handle it). Even so, I've added the log of `Ignoring invalid compose volume definition` so the Supervisor is super clear about how it's handling various long syntax volume definitions.

Other manual tests include the usual ones:
- Dashboard actions - reboot, purge, shutdown, restart services
- Set & remove a config var
- Move device to another app, move it back
- Updating between multicontainer releases
- Updating between single container releases
- Enable local mode, push a release, trigger a change, disable local mode

### DIFF
--- a/src/compose/sanitise.ts
+++ b/src/compose/sanitise.ts
@@ -65,7 +65,7 @@ export function sanitiseComposeConfig(
 ): ServiceComposeConfig {
 	const filtered: string[] = [];
 	const toReturn = _.pickBy(composeConfig, (_v, k) => {
-		const included = _.includes(supportedComposeFields, k);
+		const included = supportedComposeFields.includes(k);
 		if (!included) {
 			filtered.push(k);
 		}
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -4,26 +4,34 @@ import Duration = require('duration-js');
 import * as _ from 'lodash';
 import * as path from 'path';
 
-import * as conversions from '../lib/conversions';
-import { checkInt } from '../lib/validation';
-import { InternalInconsistencyError } from '../lib/errors';
 import { DockerPortOptions, PortMap } from './ports';
-import {
-	ConfigMap,
-	DeviceMetadata,
-	DockerDevice,
-	ServiceComposeConfig,
-	ServiceConfig,
-	ServiceConfigArrayField,
-} from './types/service';
 import * as ComposeUtils from './utils';
-
 import * as updateLock from '../lib/update-lock';
 import { sanitiseComposeConfig } from './sanitise';
 import { getPathOnHost } from '../lib/fs-utils';
-
 import log from '../lib/supervisor-console';
+import * as conversions from '../lib/conversions';
+import { checkInt } from '../lib/validation';
+import { InternalInconsistencyError } from '../lib/errors';
+
 import { EnvVarObject } from '../types';
+import {
+	ServiceConfig,
+	ServiceConfigArrayField,
+	ServiceComposeConfig,
+	ConfigMap,
+	DeviceMetadata,
+	DockerDevice,
+	ShortMount,
+	ShortBind,
+	ShortAnonymousVolume,
+	ShortNamedVolume,
+	LongDefinition,
+	LongTmpfs,
+	LongBind,
+	LongAnonymousVolume,
+	LongNamedVolume,
+} from './types/service';
 
 const SERVICE_NETWORK_MODE_REGEX = /service:\s*(.+)/;
 const CONTAINER_NETWORK_MODE_REGEX = /container:\s*(.+)/;
@@ -528,10 +536,20 @@ export class Service {
 		}
 		expose = _.uniq(expose);
 
-		const tmpfs: string[] = [];
-		_.each((container.HostConfig as any).Tmpfs, (_v, key) => {
-			tmpfs.push(key);
-		});
+		const tmpfs: string[] = Object.keys(container.HostConfig?.Tmpfs || {});
+
+		const binds: string[] = _.uniq(
+			([] as string[]).concat(
+				container.HostConfig.Binds || [],
+				Object.keys(container.Config?.Volumes || {}),
+			),
+		);
+
+		const mounts: LongDefinition[] = (container.HostConfig?.Mounts || []).map(
+			ComposeUtils.dockerMountToServiceMount,
+		);
+
+		const volumes: ServiceConfig['volumes'] = [...binds, ...mounts];
 
 		// We cannot use || for this value, as the empty string is a
 		// valid restart policy but will equate to null in an OR
@@ -558,10 +576,7 @@ export class Service {
 			hostname,
 			command: container.Config.Cmd || '',
 			entrypoint: container.Config.Entrypoint || '',
-			volumes: _.concat(
-				container.HostConfig.Binds || [],
-				_.keys(container.Config.Volumes || {}),
-			),
+			volumes,
 			image: container.Config.Image,
 			environment: Service.omitDeviceNameVars(
 				conversions.envArrayToObject(container.Config.Env || []),
@@ -667,13 +682,13 @@ export class Service {
 		deviceName: string;
 		containerIds: Dictionary<string>;
 	}): Dockerode.ContainerCreateOptions {
-		const { binds, volumes } = this.getBindsAndVolumes();
+		const { binds, mounts, volumes } = this.getBindsMountsAndVolumes();
 		const { exposedPorts, portBindings } = this.generateExposeAndPorts();
 
-		const tmpFs: Dictionary<''> = {};
-		_.each(this.config.tmpfs, (tmp) => {
-			tmpFs[tmp] = '';
-		});
+		const tmpFs: Dictionary<''> = this.config.tmpfs.reduce(
+			(dict, tmp) => ({ ...dict, [tmp]: '' }),
+			{},
+		);
 
 		const mainNetwork = _.pickBy(
 			this.config.networks,
@@ -724,6 +739,7 @@ export class Service {
 				CapAdd: this.config.capAdd,
 				CapDrop: this.config.capDrop,
 				Binds: binds,
+				Mounts: mounts,
 				CgroupParent: this.config.cgroupParent,
 				Devices: this.config.devices,
 				DeviceRequests: this.config.deviceRequests,
@@ -923,21 +939,26 @@ export class Service {
 		);
 	}
 
-	private getBindsAndVolumes(): {
+	private getBindsMountsAndVolumes(): {
 		binds: string[];
+		mounts: Dockerode.MountSettings[];
 		volumes: { [volName: string]: {} };
 	} {
 		const binds: string[] = [];
+		const mounts: Dockerode.MountSettings[] = [];
 		const volumes: { [volName: string]: {} } = {};
-		_.each(this.config.volumes, (volume) => {
-			if (_.includes(volume, ':')) {
-				binds.push(volume);
-			} else {
-				volumes[volume] = {};
-			}
-		});
 
-		return { binds, volumes };
+		for (const volume of this.config.volumes) {
+			if (LongDefinition.is(volume)) {
+				// Volumes with the long syntax are translated into Docker-accepted configs
+				mounts.push(ComposeUtils.serviceMountToDockerMount(volume));
+			} else {
+				// Volumes with the string short syntax are acceptable as Docker configs as-is
+				ShortMount.is(volume) ? binds.push(volume) : (volumes[volume] = {});
+			}
+		}
+
+		return { binds, mounts, volumes };
 	}
 
 	private generateExposeAndPorts(): DockerPortOptions {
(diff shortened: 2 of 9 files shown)


src/compose/sanitise.ts
```diff
@@ -65,7 +65,7 @@ export function sanitiseComposeConfig(
 ): ServiceComposeConfig {
 	const filtered: string[] = [];
 	const toReturn = _.pickBy(composeConfig, (_v, k) => {
-		const included = _.includes(supportedComposeFields, k);
+		const included = supportedComposeFields.includes(k);
```
pipex, line 68: 🎉

src/compose/service.ts
```diff
@@ -667,13 +675,13 @@ export class Service {
 		deviceName: string;
 		containerIds: Dictionary<string>;
 	}): Dockerode.ContainerCreateOptions {
-		const { binds, volumes } = this.getBindsAndVolumes();
+		const { binds, mounts, volumes } = this.getBindsMountsAndVolumes();
 		const { exposedPorts, portBindings } = this.generateExposeAndPorts();
 
 		const tmpFs: Dictionary<''> = {};
-		_.each(this.config.tmpfs, (tmp) => {
+		for (const tmp of this.config.tmpfs) {
```
pipex, line 682: This is a NIT, but maybe using reduce makes the definition shorter?

```
const tmpFs: Dictionary<''> = this.config.tmpfs.reduce((res, tmp) => ({...res, [tmp]: ''}), {});
```

src/compose/service.ts
```diff
@@ -1124,26 +1147,50 @@ export class Service {
...
+			let isBindMount = false;
+
+			if (isAnonymousOrTmpfs) {
+				// Push anonymous volume. Anonymous volumes can only be
+				// set by using the `VOLUME` instruction inside a Dockerfile.
+				// Tmpfs volumes do not require namespacing by appId.
+				volumes.push(volume);
+			} else if (LongSyntaxVolume.is(volume)) {
+				// Handle long syntax volumes of type 'tmpfs' and 'volume'
+				isBindMount = path.isAbsolute(volume.source!);
```
pipex, line 1168: If `volume.type` is `tmpfs`, then `volume.source` will be undefined, meaning the call to `path.isAbsolute` will fail, right??

```
path.isAbsolute(undefined)
Uncaught:
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:120:11)
    at Object.isAbsolute (path.js:1029:5)
    at REPL10:1:6
    at Script.runInThisContext (vm.js:120:18)
    at REPLServer.defaultEval (repl.js:442:29)
    at bound (domain.js:427:14)
    at REPLServer.runBound [as eval] (domain.js:440:12)
    at REPLServer.onLine (repl.js:777:10)
    at REPLServer.emit (events.js:326:22)
    at REPLServer.EventEmitter.emit (domain.js:483:12) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

src/compose/service.ts
```diff
@@ -1124,26 +1147,50 @@ export class Service {
...
+				if (!isBindMount) {
+					volume.source = namespaceVolume(volume.source!);
```
pipex, line 1170: What happens if you assign a source for a `tmpfs` volume? Same as above, if volume is `tmpfs` this will assign as source `namespaceVolume(undefined)` resulting in `<appId>_undefined`, if I'm understanding correctly (which I'm not sure I am)

src/compose/types/service.ts
```diff
@@ -28,6 +29,44 @@ export interface ServiceNetworkDictionary {
...
+		source: t.string,
+		readOnly: t.boolean,
+		bind: t.type({
+			propagation: t.keyof({
+				private: null,
+				rprivate: null,
+				shared: null,
+				rshared: null,
+				slave: null,
+				rslave: null,
```
pipex, line 55: This might not be the right place to define this requirement, as we don't validate the syntax when receiving the target state. If someone where to set a wrong value for the `propagation` option, and that makes it all the way to the point where the code is configuring the volumes, then the call `LongSyntaxVolume.is` will return `false`, causing unintended side effects in the comparison.

I think setting `propagation: t.string` should be enough, as any issues will be handled by the engine in that case.

test/src/compose/service.spec.ts
```diff
@@ -1123,4 +1123,185 @@ describe('compose/service', () => {
...
+			 */
+			// Inject bind mounts (as feature labels would)
+			// Bind mounts added under feature labels use the short syntax (for now)
+			service.config.volumes.push('/var/log/journal:/var/log/journal:ro');
+			service.config.volumes.push(
+				`${constants.dockerSocket}:${constants.containerDockerSocket}`,
+			);
+
+			const ctn = service.toDockerContainer({
+				deviceName: 'thicc_nucc',
```
pipex, line 1191: 😂

pipex:
Great work @cywang117! Only minor comments and maybe you can clarify the behavior that I referenced in [this comment](https://github.com/balena-os/balena-supervisor/pull/1935#discussion_r863861196). I'm pretty sure I'm misunderstanding but I'd rather clarify.

src/compose/service.ts
cywang117: Good catch!

src/compose/service.ts
cywang117: I clarified the comments in the latest push, but the path.isAbsolute calls should only be for long syntax volumes of type `bind` and `volume`, since `tmpfs` long syntax volumes are handled in the first `if` at the same code branch as anonymous volumes. 

I did notice that compose's behavior allows for long syntax volumes without a `source` defined (I suppose they behave as anonymous volumes), so I've updated the `if`s to reflect this.

src/compose/service.ts
cywang117: If you assign `source` to a long syntax tmpfs volume, docker-compose will error (and therefore balena-compose should error), so we shouldn't need to handle this. Would you prefer the check to be here though?

src/compose/types/service.ts
cywang117: docker-compose won't allow an incorrect propagation mode for bind as well, so balena-compose should also handle this. However since we don't allow user-defined bind mounts anyway, there shouldn't be any issue to changing to t.string as suggested, so I've changed this in the latest force push.

src/compose/service.ts
```diff
@@ -1124,26 +1147,60 @@ export class Service {
...
+				volumes.push(volume);
+			} else if (
+				LongSyntaxVolume.is(volume) &&
+				(volume.type === 'bind' || volume.type === 'volume')
+			) {
+				// Handle long syntax volumes of type 'volume' or 'bind'.
+				// Other invalid mount types will raise a compose error,
+				// so we don't need to worry about other long syntax types.
+				// Volumes are added to config while binds are ignored and flagged for console warning.
+				isBindMount = path.isAbsolute(volume.source!);
```
cywang117, line 1176: Clarified here @pipex -- this look okay?

cywang117: On second thought, this all looks to be a bit too verbose.. how do we feel about making the `io-ts` long syntax types more specific, i.e. `LongSyntaxTmpfs`, `LongSyntaxVolume`, `LongSyntaxBind`? I'm asking this because, as it turns out, the options and required configs for these different long syntax types turn out to be more different than the same. Their only similarities are that they require `target` and `type` fields. We could even add `ShortSyntaxVolume`. It would make the code here more terse which is a good thing.

pipex: I like that idea, this definitely is looking too verbose. Using the io-ts type guards will also help you avoid the non-null assertions `volume.source!`

src/compose/service.ts
pipex: On the new code, there is already a check in this branch for the mount being a volume or bind, so it should not be necessary to add the check here.

src/compose/service.ts
cywang117: Latest push switches completely to using io-ts `brand` type refinements.

cywang117:
@pipex @20k-ultra This PR is ready for review. I've updated the parent comment to reflect the results of my updated manual tests. All out-facing functionality remains the same, the main changes to check here are the io-ts type refinements starting [here](https://github.com/balena-os/balena-supervisor/pull/1935/files#diff-5e8d7d0564b2b0fa71bc58bb4ec3a9a538fb76a31b0178817e4b837126100e33R33). I've written unit tests for these types and integration tested them on a device though, so they should be fine.

One thing that may require addressing is that, when declaring an anonymous volume, today's Supervisor (13.1.x)'s behavior includes logging this anonymous volume as an unmanaged volume with `Found unmanaged Volume`. I guess this is technically accurate as the Supervisor does not really "manage" these anonymous volumes in the same way as Supervisor manages named volumes (by namespacing them). However, I thought I'd note it in case we want to change the log.

cywang117:
@balena-ci rebase

cywang117:
@balena-ci I self-certify!

pipex:
We knew that @klutchell, we added support supervisor side because we needed it for another PR,  adding support on the builder and CLI needs to wait until we can figure out how to do it without breaking devices before this PR

PR #1936, opened 2022-04-27
Use delay instead of interval to recursively report state

20k-ultra:
Follows the same mechanism as fetching target state seen in https://github.com/balena-os/balena-supervisor/blob/master/src/device-state/target-state.ts#L185-L210.

Using a delay which waits for a report is much more appropriate since you could have reports happen in quicker succession than expected since the time elapsed during a report and shared with the interval to start another report.

### DIFF
--- a/src/api-binder/report.ts
+++ b/src/api-binder/report.ts
@@ -1,5 +1,6 @@
 import * as url from 'url';
 import * as _ from 'lodash';
+import { delay } from 'bluebird';
 import { CoreOptions } from 'request';
 
 import * as constants from '../lib/constants';
@@ -141,15 +142,24 @@ export async function startReporting() {
 	);
 	const doReport = async () => {
 		if (!reportPending) {
-			throttledReport(reportConfigs);
+			await throttledReport(reportConfigs);
 		}
 	};
 
 	// If the state changes, report it
 	deviceState.on('change', doReport);
-	// But check once every max report frequency to ensure that changes in system
-	// info are picked up (CPU temp etc)
-	setInterval(doReport, constants.maxReportFrequency);
-	// Try to perform a report right away
-	return doReport();
+
+	async function recursivelyReport(delayBy: number) {
+		try {
+			// Try to send current state
+			await doReport();
+		} finally {
+			// Wait until we want to report again
+			await delay(delayBy);
+			// Try to report again
+			await recursivelyReport(delayBy);
+		}
+	}
+
+	return recursivelyReport(constants.maxReportFrequency);
 }


src/api-binder/report.ts
```diff
@@ -141,15 +142,30 @@ export async function startReporting() {
 	);
 	const doReport = async () => {
 		if (!reportPending) {
-			throttledReport(reportConfigs);
+			return throttledReport(reportConfigs);
```
Page-, line 145: We don't care about the return value, only about awaiting it:
```suggestion
			await throttledReport(reportConfigs);
```

src/api-binder/report.ts
```diff
@@ -141,15 +142,30 @@ export async function startReporting() {
...
+			await recursivelyReport(delayBy);
+		};
+
+		try {
+			// Try to send current state
+			await doReport();
+		} finally {
+			// Wait to report again
+			await delayedLoop();
+		}
```
Page-, line 167: I think the extra function just makes things more confusing here because it's only used once in precisely one place and is so short that it's not really abstracting much useful (the amount of extra comments just to try to explain what this separate function is just adds more noise than in-lining it)
```suggestion
		try {
			// Try to send current state
			await doReport();
		} finally {
			// Wait until we want to report again
			await delay(delayBy);
			// Try to report again
			await recursivelyReport(delayBy);
		}
```

src/api-binder/report.ts
```diff
@@ -1,5 +1,6 @@
 import * as url from 'url';
 import * as _ from 'lodash';
+import { delay } from 'bluebird';
```
cywang117, line 3: Can we use `sleep` from backoff.ts? (There's also a sleep function in test/lib/helpers.ts but we previously agreed to prefer `sleep` from backoff over helpers)

src/api-binder/report.ts
```diff
@@ -141,15 +142,30 @@ export async function startReporting() {
 	);
 	const doReport = async () => {
 		if (!reportPending) {
-			throttledReport(reportConfigs);
+			return throttledReport(reportConfigs);
 		}
 	};
 
 	// If the state changes, report it
 	deviceState.on('change', doReport);
```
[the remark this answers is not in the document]
20k-ultra: there's no more setInterval so unsure if this comment is still required. The throttle is to prevent device state changes and the interval calling report too quickly and the interval is to call report when no device state events are emitted.

Am I missing something ?

src/api-binder/report.ts
20k-ultra: I wanted to and checked because I recalled this too but the sleep function in backoff isn't exported so I just did what target state fetching does by using bluebird delay instead of introducing an export sleep function when maybe we should try to just upgrade to node 16 for native setTimeout promise

src/api-binder/report.ts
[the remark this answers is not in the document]
Page-: Discussed further in fd/meet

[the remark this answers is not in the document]
20k-ultra: the delay and throttle are the same value so effectively the deviceState event listener is useless

Page- (approved):

PR #1937, opened 2022-04-28
Only throttle HTTP request when reporting state

20k-ultra:
Previous logic throttled asking for changes and reporting. This could lead to the device not sending a report for longer than intended.

1. 0 seconds - device asks for changes and nothing is found (throttle is now triggered)
2. 1 second later - device emits a state change because a container stopped...since asking for changes and HTTP request is throttled nothing happens even though no request was ever actually made

This PR resolves that by only throttling the HTTP request.


20k-ultra:
We will just cancel the throttle so that subsequent calls do not get "delayed"

PR #1942, opened 2022-05-11
Do not stop containers on reboot/shutdown

20k-ultra:
This solves the issue by removing the stop container step. This means the engine will handle kill signals to the containers. From what I've read, the stop command defaults to 10 seconds and when you shutdown the engine it also uses 10 second timeout when telling a container to stop before using a more aggressive signal so the stopping behaviour should be the same.

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -684,32 +684,44 @@ export function reportCurrentState(newState: DeviceReport = {}) {
 	emitAsync('change', undefined);
 }
 
-export async function reboot(force?: boolean, skipLock?: boolean) {
-	await updateLock.abortIfHUPInProgress({ force });
-	await applicationManager.stopAll({ force, skipLock });
-	logger.logSystemMessage('Rebooting', {}, 'Reboot');
-	const $reboot = await dbus.reboot();
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return await $reboot;
+export interface ShutdownOpts {
+	force?: boolean;
+	reboot?: boolean;
 }
 
-export async function shutdown(force?: boolean, skipLock?: boolean) {
+export async function shutdown({
+	force = false,
+	reboot = false,
+}: ShutdownOpts = {}) {
 	await updateLock.abortIfHUPInProgress({ force });
-	await applicationManager.stopAll({ force, skipLock });
-	logger.logSystemMessage('Shutting down', {}, 'Shutdown');
-	const $shutdown = await dbus.shutdown();
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return $shutdown;
+	// Get current apps to create locks for
+	const apps = await applicationManager.getCurrentApps();
+	const appIds = Object.keys(apps).map((strId) => parseInt(strId, 10));
+	// Try to create a lock for all the services before shutting down
+	return updateLock.lockAll(appIds, force, async () => {
+		let dbusAction;
+		switch (reboot) {
+			case true:
+				logger.logSystemMessage('Rebooting', {}, 'Reboot');
+				dbusAction = await dbus.reboot();
+				break;
+			case false:
+				logger.logSystemMessage('Shutting down', {}, 'Shutdown');
+				dbusAction = await dbus.shutdown();
+				break;
+		}
+		shuttingDown = true;
+		emitAsync('shutdown', undefined);
+		return dbusAction;
+	});
 }
 
 export async function executeStepAction<T extends PossibleStepTargets>(
 	step: DeviceStateStep<T>,
 	{
-		force,
-		initial,
-		skipLock,
+		force = false,
+		initial = false,
+		skipLock = false,
 	}: { force?: boolean; initial?: boolean; skipLock?: boolean },
 ) {
 	if (deviceConfig.isValidAction(step.action)) {
@@ -728,13 +740,13 @@ export async function executeStepAction<T extends PossibleStepTargets>(
 				// and if they do, we wouldn't know about it until after
 				// the response has been sent back to the API. Just return
 				// "OK" for this and the below action
-				await reboot(force, skipLock);
+				await shutdown({ force, reboot: true });
 				return {
 					Data: 'OK',
 					Error: null,
 				};
 			case 'shutdown':
-				await shutdown(force, skipLock);
+				await shutdown({ force });
 				return {
 					Data: 'OK',
 					Error: null,
--- a/src/lib/update-lock.ts
+++ b/src/lib/update-lock.ts
@@ -88,6 +88,34 @@ function dispose(
 }
 
 /**
+ * Attempts to take locks for list of application IDs provided
+ * Waits until all applications are locked before running the passed function
+ */
+export async function lockAll<T extends unknown>(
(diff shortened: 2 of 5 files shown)


src/device-state.ts
```diff
@@ -684,24 +684,29 @@ export function reportCurrentState(newState: DeviceReport = {}) {
...
-export async function reboot(force?: boolean, skipLock?: boolean) {
-	await updateLock.abortIfHUPInProgress({ force });
-	await applicationManager.stopAll({ force, skipLock });
-	logger.logSystemMessage('Rebooting', {}, 'Reboot');
-	const $reboot = await dbus.reboot();
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return await $reboot;
+export async function reboot(force: boolean) {
+	// Multi-app warning!
```
cywang117, line 688: To clarify, we'll need to change this when multiple user apps are possible, because updateLock.lock in this case needs to check that none of the multiple apps have locks taken, but right now the lock method only accepts 1 appId.

src/device-state.ts
```diff
@@ -684,32 +684,37 @@ export function reportCurrentState(newState: DeviceReport = {}) {
...
+export async function shutdown(force: boolean) {
 	await updateLock.abortIfHUPInProgress({ force });
-	await applicationManager.stopAll({ force, skipLock });
-	logger.logSystemMessage('Shutting down', {}, 'Shutdown');
-	const $shutdown = await dbus.shutdown();
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return $shutdown;
+	// Multi-app warning!
+	const appId = (await config.get('applicationId')) as number;
```
20k-ultra, line 702: get application IDs from the database and ensure you can create a lock in each of these apps.

src/lib/update-lock.ts
```diff
@@ -87,6 +87,29 @@ function dispose(
...
+	}
+	let lockCounter = 0;
+	const nextLock = async (): Promise<Resolvable<T>> => {
+		lockCounter++;
+		if (lockCounter === appIds.length) {
+			return lock(appIds[lockCounter + 1], { force }, fn);
+		}
+		return lock(appIds[lockCounter + 1], { force }, nextLock);
+	};
+	return lock(appIds[0], { force }, nextLock);
```
20k-ultra, line 110: If there is 1 application the lock is called twice which will result in the 2nd call throwing since locks still exist from the first one.

src/lib/update-lock.ts
```diff
@@ -87,6 +87,29 @@ function dispose(
 		.return();
 }
 
+/**
+ * Attempts to take locks for list of application IDs provided
+ * The locks will exist until all applications are locked and the passed function resolves or throws.
+ */
+export async function lockAll<T extends unknown>(
```
20k-ultra, line 94: write unit tests for this

src/lib/update-lock.ts
20k-ultra: resolved by calling fn if there is only 1 appId in the array

src/lib/update-lock.ts
```diff
@@ -61,30 +62,18 @@ export function abortIfHUPInProgress({
...
-			return lockfile.unlock(lockName);
-		},
-	)
-		.finally(release)
-		.return();
+async function dispose(appIdentifier: string | number): Promise<void> {
+	const locks = lockfile.getLocksTaken((p: string) =>
+		p.includes(`${BASE_LOCK_DIR}/${appIdentifier}`),
+	);
+	for await (const l of locks) {
```
cywang117, line 74: Is `await` necessary here?

src/lib/update-lock.ts
```diff
@@ -119,74 +108,65 @@ export async function lockAll<T extends unknown>(
...
+			} catch (e) {
+				if (lockfile.LockfileExistsError.is(e)) {
+					// Throw more descriptive error
+					throw new UpdatesLockedError(
+						`Lockfile exists for { appId: ${appId}, service: ${service} }`,
+					);
+				}
+				// Otherwise just throw the error
+				throw e;
+			}
```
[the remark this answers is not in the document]
cywang117: The throw here will short-cut the `finally` from what I can tell, at least it did with the Bluebird implementation (it short-cut the `.disposer` chain before when I tested it). You may need to test again to see if the behavior differs for native Promises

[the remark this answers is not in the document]
20k-ultra: I'll add a finally here to dispose locks if adding other locks fails :+1:

src/lib/update-lock.ts
```diff
@@ -119,74 +108,65 @@ export async function lockAll<T extends unknown>(
...
+		lockOverride = await config.get('lockOverride');
+	} catch (err) {
+		throw new InternalInconsistencyError(
+			`Error getting lockOverride config value: ${err?.message ?? err}`,
+		);
+	}
 
-	const disposer = takeTheLock();
+	// Acquire write lock for appId
+	await writeLock(appId);
```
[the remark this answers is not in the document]
20k-ultra: I was confused by this tbh. Even had a comment that said "not sure why this is called" lol.

I'll call the release in the finally blocks. Since i'm disposing in 2 finally blocks I'll make it cleaner so it's only written in 1 place.

src/lib/update-lock.ts
```diff
@@ -119,74 +108,65 @@ export async function lockAll<T extends unknown>(
...
-	return Bluebird.using(disposer, fn as () => PromiseLike<T>);
+	// Get list of service folders in lock directory
+	let serviceFolders: string[] = [];
+	try {
+		serviceFolders = await fs.readdir(lockDir);
+	} catch (e) {
+		log.error(`Error getting lock directory contents - ${e}`);
+	}
+
+	for await (const service of serviceFolders) {
```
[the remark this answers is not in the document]
20k-ultra: don't we have to wait for the lock to complete since it's async ? The reason I am doing this refactor is because I cannot reliable write tests when promises are not awaited.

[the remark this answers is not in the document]
cywang117: I meant if `await` is used in the next line of both `for await`s, are the awaits here necessary or are they redundant?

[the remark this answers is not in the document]
20k-ultra: I see. I never used this format, just added it because the linter yelled at me but you're right that only 1 await is needed!

src/lib/update-lock.ts
```diff
@@ -70,95 +71,126 @@ export const readLock: LockFn = Bluebird.promisify(locker.async.readLock, {
...
+ * Attempts to take locks for list of application IDs provided
+ * Waits until all applications are locked before running the passed function
+ */
+export async function lockAll<T extends unknown>(
+	appIds: number[],
+	force = false,
+	fn: () => Resolvable<T>,
+): Promise<T | undefined> {
+	if (appIds.length === 0) {
+		throw new UpdatesLockedError(
```
pipex, line 91: This should be an assertion or an `InternalInconsistencyError` as `UpdatesLockError` is caught upper in the call stack to report a service being locked.

20k-ultra: good one

Issue #1943, opened 2022-05-12
Running a preloaded app without having internet connectivity fails (RPI3 32 bits OS version 2.98.20)


floion:
Trying to run an OS with preloaded wifi connect app, without eth plugged in. The WiFi Connect AP won't show up in the scan list until eth is plugged in and at that same time the board appears online in the dashboard. Then, using the webterminal to check what's going on:

```
Aug 06 14:41:27 b1f4f0b balena-supervisor[2276]: [info]    Attempting to load any preloaded applications
Aug 06 14:41:42 b1f4f0b balena-supervisor[2276]: [error]   LogBackend: unexpected error: Error: getaddrinfo EAI_AGAIN api.balena-cloud.com
Aug 06 14:41:42 b1f4f0b balena-supervisor[2276]: [error]         at GetAddrInfoReqWrap.onlookupall [as oncomplete] (dns.js:76:26)
Aug 06 14:41:57 b1f4f0b balena-supervisor[2276]: [debug]   getaddrinfo EAI_AGAIN api.balena-cloud.com
Aug 06 14:41:57 b1f4f0b balena-supervisor[2276]: [event]   Event: Loading preloaded apps failed {"error":{"message":"getaddrinfo EAI_AGAIN api.balena-cloud.com","stack":"Error: getaddrinfo EAI_AGAIN api.balena-cloud.com\n    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (dns.js:76:26)"}}
```

20k-ultra:
Additional logs while reproducing this:

```
root@2d97200:~# journalctl -fn 100 -u balena-supervisor
-- Journal begins at Thu 2022-05-12 16:50:07 UTC. --
May 12 16:50:27 2d97200 balena-supervisor[1539]: Error response from daemon: No such container: resin_supervisor
May 12 16:50:27 2d97200 balena-supervisor[1549]: Error response from daemon: No such container: balena_supervisor
May 12 16:50:27 2d97200 balena-supervisor[1559]: active
May 12 16:50:27 2d97200 balena-supervisor[1578]: Error: No such object: balena_supervisor
May 12 16:50:28 2d97200 balena-supervisor[1619]: find: /mnt/root/tmp/balena-supervisor/services: No such file or directory
May 12 16:50:29 2d97200 balena-supervisor[1619]: [info]    Supervisor v13.1.6 starting up...
[...]

PR #1944, opened 2022-05-12
Do not migrate apps.json if preload v2 to v3 fails

cywang117:
Preloading for OS versions with supervisor v13 need to use a newer CLI. The supervisor has a mechanism to migrate apps.json from a v2 apps.json into a v3 apps.json, however it would fail if the device was preloading without connectivity and apps.json would get migrated anyway, meaning the supervisor would not retry the preloading step later.

This PR adds a more clarifying message if this condition happens pointing users to preload with a newer CLI, but it also makes it so the apps.json is not migrated, allowing the preloading step to be retried if the device is rebooted.

This PR also fixes the check for v2 apps.json, which was being triggered before by appUuids starting with a number `12deadbeef`

Closes: #1943
Relates-to: https://github.com/balena-io-modules/balena-preload/pull/273
Change-type: patch

### DIFF
--- a/src/device-state/preload.ts
+++ b/src/device-state/preload.ts
@@ -2,6 +2,7 @@ import * as _ from 'lodash';
 import { promises as fs } from 'fs';
 
 import { Image, imageFromService } from '../compose/images';
+import { NumericIdentifier } from '../types';
 import * as deviceState from '../device-state';
 import * as config from '../config';
 import * as deviceConfig from '../device-config';
@@ -66,11 +67,7 @@ export async function loadTargetFromFile(appsPath: string): Promise<boolean> {
 		}
 
 		// if apps.json apps are keyed by numeric ids, then convert to v3 target state
-		if (
-			Object.keys(stateFromFile.apps || {}).some(
-				(appId) => !isNaN(parseInt(appId, 10)),
-			)
-		) {
+		if (Object.keys(stateFromFile.apps || {}).some(NumericIdentifier.is)) {
 			stateFromFile = await fromV2AppsJson(stateFromFile as any);
 		}
 
--- a/src/device-state/preload.ts
+++ b/src/device-state/preload.ts
@@ -26,6 +26,26 @@ export function appsJsonBackup(appsPath: string) {
 	return `${appsPath}.preloaded`;
 }
 
+async function migrateAppsJson(appsPath: string) {
+	const targetPath = appsJsonBackup(appsPath);
+	if (!(await fsUtils.exists(targetPath))) {
+		// Try to rename the path so the preload target state won't
+		// be used again if the database gets deleted for any reason.
+		// If the target file already exists or something fails, just debug
+		// the failure.
+		await fsUtils
+			.safeRename(appsPath, targetPath)
+			.then(() => fsUtils.writeFileAtomic(appsPath, '{}'))
+			.then(() => log.debug(`Migrated existing apps.json`))
+			.catch((e) =>
+				log.debug(
+					`Continuing without migrating apps.json because of`,
+					e.message,
+				),
+			);
+	}
+}
+
 export async function loadTargetFromFile(appsPath: string): Promise<boolean> {
 	log.info('Attempting to load any preloaded applications');
 	try {
@@ -119,6 +139,7 @@ export async function loadTargetFromFile(appsPath: string): Promise<boolean> {
 		};
 
 		await deviceState.setTarget(localState);
+		await migrateAppsJson(appsPath);
 		log.success('Preloading complete');
 		if (preloadState.pinDevice) {
 			// Multi-app warning!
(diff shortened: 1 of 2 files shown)


cywang117:
The tests for this PR also deleted `./test/data/mnt/boot/config.txt` and I don't know why.. I'll revert that test file deletion before this merges, but we should definitely look at fixing the tests that are causing this leakage soon.

src/lib/legacy.ts
```diff
@@ -309,8 +309,18 @@ export async function fromV2TargetApps(
...
+						try {
+							// If local mode just use id as uuid
+							uuid = local ? id : await getUUIDFromAPI(appId);
+						} catch {
+							// If getUUIDFromAPI fails, set uuid to id.
+							// getUUIDFromAPI will fail with network connection errors,
+							// so when the device reconnects and sees a v3 target state,
+							// user containers will restart. Update locks should be leveraged
+							// during critical processes in container to avoid this.
+							uuid = id;
```
cywang117, line 322: Has unexpected behavior after the device connects to the internet post-preload. Will investigate

src/device-state/preload.ts
```diff
@@ -154,7 +156,7 @@ export async function loadTargetFromFile(appsPath: string): Promise<boolean> {
 		}
 	} finally {
 		const targetPath = appsJsonBackup(appsPath);
-		if (!(await fsUtils.exists(targetPath))) {
+		if (isTargetLoaded && !(await fsUtils.exists(targetPath))) {
```
pipex, line 159: I don't think it makes so much sense to do this check within the `finally`. I think we can probably move the whole `await fsUtils.safeRename...` call to after the `setTarget` call and we can avoid this variable. 

Since the call is a bit verbose, we can put it in an function and call that after the `setTarget`.

src/lib/legacy.ts
```diff
@@ -309,8 +309,18 @@ export async function fromV2TargetApps(
 						const appId = parseInt(id, 10);
 						const app = apps[appId];
 
-						// If local mode just use id as uuid
-						const uuid = local ? id : await getUUIDFromAPI(appId);
+						let uuid: string;
+						try {
+							// If local mode just use id as uuid
+							uuid = local ? id : await getUUIDFromAPI(appId);
```
pipex, line 315: How about using`uuid = local ? id : await getUUIDFromAPI(appId).catch(() => id);` to avoid the `let` and wrapping into a `try ... catch`?

src/device-state/preload.ts
cywang117: Good idea, will change

cywang117: Changed as of https://github.com/balena-os/balena-supervisor/pull/1944/commits/d92d696e0e599fe367729df679031a37d9503401

pipex:
How to test,

- Clone and build a CLI using the draft version of balena-preload from https://github.com/balena-io-modules/balena-preload/pull/273 (use Node 12)
```
git clone https://github.com/balena-io/balena-cli.git && cd balena-cli
npm i
npm i --save balena-preload@12.1.0-preload-v3-b212c5bcb5238989c94331992c9f1b46881c73db
npm run build
```
- Download a balenaOS image with supervisor v12 and preload it (we will livepush the supervisor later)
```
./bin/balena os download intel-nuc --version v2.89.15 -o ~/intel-nuc.img
./bin/balena preload ~/intel-nuc.img --fleet <my-fleet>
```
- Provision the image into an new device or VM
- Livepush this supervisor version (requires internet access), wait for it to start
```
npm run sync -- <address> -a amd64
```
- SSH into the image, reset the state of the device to factory
```
# On the balena device
systemctl stop balena-supervisor
systemctl stop update-balena-supervisor.timer
# Stop and remove any containers created by preload
balena ps -qa | xargs balena stop | xargs balena rm
docker network prune -f
docker volume prune  -f
rm /mnt/data/resin-data/balena-supervisor/database.sqlite
mv /mnt/data/apps.json.preloaded /mnt/data/apps.json
```
- Disconnect the device from the internet
- Livepush the supervisor again (this time it doesn't require internet access as it has the parent images downloaded)
```
npm run sync -- <address> -a amd64
```

You should see `Cannot migrate from v2 apps.json without Internet connectivity` and apps.json to not be migrated

src/device-state/preload.ts
```diff
@@ -46,11 +67,7 @@ export async function loadTargetFromFile(appsPath: string): Promise<boolean> {
 		}
 
 		// if apps.json apps are keyed by numeric ids, then convert to v3 target state
-		if (
-			Object.keys(stateFromFile.apps || {}).some(
-				(appId) => !isNaN(parseInt(appId, 10)),
-			)
-		) {
+		if (Object.keys(stateFromFile.apps || {}).some(NumericIdentifier.is)) {
```
cywang117, line 70: Nice improvement!

src/lib/legacy.ts
```diff
@@ -309,8 +309,15 @@ export async function fromV2TargetApps(
...
 
-						// If local mode just use id as uuid
-						const uuid = local ? id : await getUUIDFromAPI(appId);
+						// If local mode or connectivity is not available just use id as uuid
+						const uuid = local
+							? id
+							: await getUUIDFromAPI(appId).catch(() => {
+									throw new Error(
+										// TODO: use the proper CLI version when it is released
+										'Cannot migrate from v2 apps.json without Internet connectivity. Please use a newer CLI with v3 apps.json support.',
```
cywang117, line 318: I see the balena-cli PR bumping balena-preload to include https://github.com/balena-io-modules/balena-preload/pull/273 isn't out yet -- would it be better to include a specific version of the CLI in this error message, once the CLI PR goes through?

cywang117: Otherwise LGTM!

pipex: I think so, hence the TODO :). `balena-preload` was just merged, I'll bump the CLI

cywang117: Ahh sorry, I didn't see that 😅

pipex: @cywang117 updated the message

src/lib/legacy.ts
cywang117: Since we're resolving this by bumping balena-preload in balena CLI to preload v3 target state, this behavior isn't an issue anymore.

src/lib/legacy.ts
cywang117: @pipex Looks good! Feel free to self-certify.

pipex:
@balena-ci I self certify!

pipex (approved):
LGTM!

PR #1953, opened 2022-05-26
Shutdown with locks

20k-ultra:
The previous logic would shutdown each application before shutting down the device. I believe this was done because if you can't shutdown an application due to locks then you can't shutdown a device so it was a good idea. The problem was that this caused restart policy "unless-stopped" to be triggered incorrectly since the Supervisor was stopping containers..this caused the containers to not be started again on boot.

This PR solves that by just checking for locks. I wrote this PR to be multi-app safe which added more work.

Unit test coverage is pretty good for this part of the codebase. I performed some manual integration tests by adding locks to services then trying to reboot from the dashboard as well as ensuring that containers with `unless-stopped` are not stuck in stopped state after a reboot.

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -684,24 +684,36 @@ export function reportCurrentState(newState: DeviceReport = {}) {
 	emitAsync('change', undefined);
 }
 
-export async function reboot(force?: boolean, skipLock?: boolean) {
-	await updateLock.abortIfHUPInProgress({ force });
-	await applicationManager.stopAll({ force, skipLock });
-	logger.logSystemMessage('Rebooting', {}, 'Reboot');
-	const $reboot = await dbus.reboot();
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return await $reboot;
+export interface ShutdownOpts {
+	force?: boolean;
+	reboot?: boolean;
 }
 
-export async function shutdown(force?: boolean, skipLock?: boolean) {
+export async function shutdown({
+	force = false,
+	reboot = false,
+}: ShutdownOpts = {}) {
 	await updateLock.abortIfHUPInProgress({ force });
-	await applicationManager.stopAll({ force, skipLock });
-	logger.logSystemMessage('Shutting down', {}, 'Shutdown');
-	const $shutdown = await dbus.shutdown();
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return $shutdown;
+	// Get current apps to create locks for
+	const apps = await applicationManager.getCurrentApps();
+	const appIds = Object.keys(apps).map((strId) => parseInt(strId, 10));
+	// Try to create a lock for all the services before shutting down
+	return updateLock.lock(appIds, { force }, async () => {
+		let dbusAction;
+		switch (reboot) {
+			case true:
+				logger.logSystemMessage('Rebooting', {}, 'Reboot');
+				dbusAction = await dbus.reboot();
+				break;
+			case false:
+				logger.logSystemMessage('Shutting down', {}, 'Shutdown');
+				dbusAction = await dbus.shutdown();
+				break;
+		}
+		shuttingDown = true;
+		emitAsync('shutdown', undefined);
+		return dbusAction;
+	});
 }
 
 export async function executeStepAction<T extends PossibleStepTargets>(
@@ -728,13 +740,13 @@ export async function executeStepAction<T extends PossibleStepTargets>(
 				// and if they do, we wouldn't know about it until after
 				// the response has been sent back to the API. Just return
 				// "OK" for this and the below action
-				await reboot(force, skipLock);
+				await shutdown({ force, reboot: true });
 				return {
 					Data: 'OK',
 					Error: null,
 				};
 			case 'shutdown':
-				await shutdown(force, skipLock);
+				await shutdown({ force, reboot: false });
 				return {
 					Data: 'OK',
 					Error: null,
--- a/src/lib/update-lock.ts
+++ b/src/lib/update-lock.ts
@@ -71,20 +71,11 @@ export const readLock: LockFn = Bluebird.promisify(locker.async.readLock, {
 });
 
 // Unlock all lockfiles, optionally of an appId | appUuid, then release resources.
-function dispose(
-	release: () => void,
-	appIdentifier: string | number,
-): Bluebird<void> {
-	return Bluebird.map(
-		lockfile.getLocksTaken((p: string) =>
-			p.includes(`${BASE_LOCK_DIR}/${appIdentifier}`),
-		),
-		(lockName) => {
-			return lockfile.unlock(lockName);
-		},
-	)
-		.finally(release)
-		.return();
+async function dispose(appIdentifier: string | number): Promise<void> {
+	const locks = lockfile.getLocksTaken((p: string) =>
+		p.includes(`${BASE_LOCK_DIR}/${appIdentifier}`),
+	);
+	await Promise.all(locks.map((l) => lockfile.unlock(l)));
 }
 
 /**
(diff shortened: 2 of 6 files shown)


src/lib/update-lock.ts
```diff
@@ -4,6 +4,7 @@ import { promises as fs } from 'fs';
 import * as path from 'path';
 import * as Lock from 'rwlock';
 import { isRight } from 'fp-ts/lib/Either';
+import { isArray } from 'lodash';
```
cywang117, line 7: Array.isArray? :eyes:

src/lib/update-lock.ts
```diff
@@ -96,69 +88,94 @@ function dispose(
...
+		return Promise.resolve(fn());
+	} finally {
+		// Cleanup locks
+		for (const release of releases) {
+			release();
+		}
+		for await (const lockedApp of sortedIds) {
+			dispose(lockedApp);
+		}
+	}
```
cywang117, line 147: Since `release` & `dispose` will run the same number of times, i.e. once for each `appId`, why not deduplicate here by calling `release` in `dispose`? As far as semantics goes, `release`-ing the writeLock is part of `disposing` of the lock effects, so incorporating `release` into `dispose` makes sense. (I believe that's how it was done before too)

20k-ultra: yeah I wanted to think of how to clean this up for sure. It's weird though because you can't assume you'll have the same number of releases and applications to unlock. What if you never lock any services but you locked an application ? I know the cost is nothing (I think) but it feels wrong.

tl;dr I did plan to come back to this.

test/src/lib/update-lock.spec.ts
```diff
@@ -242,6 +242,45 @@ describe('lib/update-lock', () => {
...
+					serviceName,
+				)]: {},
+			});
+
+			await expect(
+				updateLock.lock(appIds, { force: false }, async () => {
+					// At this point the locks should be taken and not removed
+					// until this function has been resolved
+					// Both `updates.lock` and `resin-updates.lock` should have been taken
+					expect(lockSpy.args).to.have.length(6);
```
20k-ultra, line 271: other tests use the `expectLocks` function which I wanted to but for some reason when it reads the folders from mockfs the locks do not exist. This test logic is checking the same thing without requiring any mocks/stubs to be working (maybe something is wrong between lock, lockfile, and mockfs.

@cywang117

cywang117: @20k-ultra I think expectLocks only checks for the single `appId` const declared in this test suite, the `appId` should be 1234567, so it makes sense why it wouldn't work for other appIds. The way this test is formatted works as-is, or you could also modify `expectLocks` -- up to you.

20k-ultra: yeah I modified it to check multiple applications Ids. It was 100% checking fs.readdir for each service but for some reason was saying the other applications were empty (only the first service checked would have locks). It's as is mockFS could only mock 1 folder at a time...

src/lib/update-lock.ts
cywang117: To clarify, when you're talking about `locking applications`, you're referring to `rwlock`'s `writeLock`? But when you're talking about `locking services`, you're referring to `lockfile`?

So we know for sure that:

- Whenever we pass an `appId` to `writeLock`, that appId does not currently have a write lock on it, per `rwlock`'s stipulation that "there may be only one writer at a time". So every `appId` will definitely have a `release` function associated with it which needs to be called.
- Whenever we try to `unlock` then `lock` each app's services using `lockfile`, 0-1 locks for each service will be created. So 0-n locks will be created for any given appId, where n = number of services.

So, you're right that you could end up never locking any services, but locking an application. However, this is the only condition that's possible - you'll never be able to lock services but not lock an application.

Given the above, we know that if we want to combine the `lockfile.unlock` and the `release` part of `dispose`, we need to 
- always release the `writeLock` for an appId
- handle 0-n `lockfile.unlock`s for an appId's services, where n = number of services

The current implementation of the [lockfile.unlock](https://github.com/balena-os/balena-supervisor/pull/1953/files#diff-376068891dd599a79a4b6bf84d850688e14f1d5ebaf089f88518f08e40fc6f4bR75-R78) part of `dispose` is already capable of handling 0-n `lockfile.unlock`s for a service. Therefore we should be able to add `release` to this without anything erroring..

When you say that combining these two "feels wrong", I guess you mean that the `dispose` method originally combined the dispose effects from two different locking mechanisms: `lockfile` and `rwlock`, and that perhaps combining these 2 mechanisms in this way isn't a clear separate of concerns. If that's the case, how about separating them more distinctly?
```
async function dispose(appIdentifier: string | number, release: () => void): Promise<void> {
    const disposeServiceLockfiles = async () => {
        const locks = lockfile.getLocksTaken((p: string) =>
	    p.includes(`${BASE_LOCK_DIR}/${appIdentifier}`),
        );
        await Promise.all(locks.map((l) => lockfile.unlock(l)));
    }

    const disposeAppWriteLock = () => release();

    return await Promise.all([ disposeServiceLockfiles(), disposeAppWriteLock() ]);
}
```

test/src/lib/update-lock.spec.ts
cywang117: MockFS can definitely mock multiple, just like you declared it [here](https://github.com/balena-os/balena-supervisor/pull/1953/files#diff-3e38a1b1da5c2f1d2fbe196b5b98996a7f6b8fa9721fb8de1c0f6f9da8b14aceR248-R264).. probably the way I implemented `mockLockDir` can only mock 1 initial directory. I think if you call mockFs repeatedly without mockFs.restore(), which may have been what you tried, the subsequent mock calls won't do anything. Instead, to add other directories to the mock after mockFs has already been called, you may have to use fs.mkdir or other fs methods..

cywang117: Anyways, since this is the only test suite that mocks multiple lock dirs, it's probably not necessary to modify `mockLockDir`, and the implementation you have here works fine 👍🏼

src/lib/update-lock.ts
20k-ultra: I think the way I've done it is pretty nice

src/lib/update-lock.ts
```diff
@@ -96,69 +87,93 @@ function dispose(
...
+			} catch (err) {
+				if (ENOENT(err)) {
+					// Default to empty list of folders
+					serviceFolders = [];
+				} else {
+					throw err;
+				}
+			}
+			// Attempt to create a lock for each service
+			for await (const service of serviceFolders) {
```
cywang117, line 131: You don't need the `for await` since it only works for `async iterables` as Page said :eyes:

src/lib/update-lock.ts
```diff
@@ -96,69 +87,93 @@ function dispose(
...
+async function lockService(
+	appId: number,
+	service: string,
+	opts: LockOptions = {
+		force: false,
+		lockOverride: false,
+	},
+): Promise<void> {
+	const serviceLockFiles = lockFilesOnHost(appId, service);
+	for await (const file of serviceLockFiles) {
```
cywang117, line 162: Same as previous comment, the `for await` has no effect on promises in the loop here, and you're already using `await` for promises within the loop, so the `await isn't needed here.

src/lib/update-lock.ts
cywang117: Looks good in the newest version (https://github.com/balena-os/balena-supervisor/commit/195792a8c1f652b7c860b836b91fbb8301d653c1)!

src/device-state.ts
```diff
@@ -684,24 +684,36 @@ export function reportCurrentState(newState: DeviceReport = {}) {
...
-	shuttingDown = true;
-	emitAsync('shutdown', undefined);
-	return await $reboot;
+export interface ShutdownOpts {
+	force?: boolean;
+	reboot?: boolean;
 }
 
-export async function shutdown(force?: boolean, skipLock?: boolean) {
+export async function shutdown({
```
pipex, line 692: I don't like this `shutdown({reboot: true})` signature. A reboot and a shutdown are different things. I understand that the idea is to make the code more DRY, but I think in that case would make more sense to create a function

```
withLocking(fn: () => any,  opts: {force: boolean}) that calls the function passed as argument only if there are no locks on any app, then both reboot and shutdown can call `withLocking` changing just the callback.
```

I understand if you want to leave these changes for another PR though

src/lib/update-lock.ts
```diff
@@ -96,69 +87,93 @@ function dispose(
...
+	const releases = new Map<number, () => void>();
+	try {
+		for (const id of sortedIds) {
+			const lockDir = getPathOnHost(lockPath(id));
+			// Acquire write lock for appId
+			releases.set(id, await writeLock(id));
+			// Get list of service folders in lock directory
+			let serviceFolders: string[] = [];
+			try {
+				serviceFolders = await fs.readdir(lockDir);
```
pipex, line 121: Hoe about the following to avoid the let and nested `try...catch`?

```				
serviceFolders = await fs.readdir(lockDir).catch((e) => {
    if (ENOENT(err)) { return []; }
    throw e;
});
```

src/lib/update-lock.ts
```diff
@@ -96,69 +87,93 @@ function dispose(
...
+					throw err;
+				}
+			}
+			// Attempt to create a lock for each service
+			for (const service of serviceFolders) {
+				await lockService(id, service, { force, lockOverride });
+			}
+		}
+		// Resolve the function passed
+		return Promise.resolve(fn());
```
pipex, line 136: Again, I think the `Promise.resolve` is unnecessary here.

src/lib/update-lock.ts
```diff
@@ -96,69 +87,93 @@ function dispose(
...
+type LockOptions = {
+	force?: boolean;
+	lockOverride?: boolean;
+};
+
+async function lockService(
+	appId: number,
+	service: string,
+	opts: LockOptions = {
+		force: false,
```
pipex, line 157: Do you need the two options, `force` and `lockOverride` here? Since this function is only called internally within the module, you could just define a `force` option and call it as `lockService(appId, service, force || lockOverride)`

pipex: Maybe adding a catch all within `dispose()` will be enough

src/device-state.ts
pipex: I disagree, they are semantically different, and the operations are different. The interface is not intuitive either, if you are new to the codebase and you want to check how `reboot` is triggered, you won't find it easily.

I'm not going to argue if you want to leave it like this, I'm just giving you my opinion.

src/lib/update-lock.ts
20k-ultra: awesome. My brain has forgotten all the other ways to handle promises other then await..

src/lib/update-lock.ts
```diff
@@ -96,69 +87,93 @@ function dispose(
...
+		}
+		// Resolve the function passed
+		return Promise.resolve(fn());
+	} finally {
+		// Cleanup locks
+		for (const [id, release] of releases.entries()) {
+			// Release lock on application
+			release();
+			// Dispose of locks in each service within the application
+			await dispose(id);
```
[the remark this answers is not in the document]
20k-ultra: > release was done within a finally before.

Release removes application locks
Dispose removes service locks

The previous code passed release to dispose so that it (release) would be called after disposing (release services). It wasn't mentioned but this PR changes release to be called before disposing so I'll change that to be more like the old behaviour.

Regarding error handling though, this code is the same I believe. 

Previously, when locking a service, if it throws an error it was [caught and logged](https://github.com/balena-os/balena-supervisor/blob/33ad877c67a639024a35cd4aa8452d280f757a45/src/lib/update-lock.ts#L136-L145). The current behaviour catches the error when locking a service too but with the exception of not releasing the application yet. This is because the finally block always runs and that calls release. The old code would release twice in fact. That never threw an error because of [this](https://github.com/71104/rwlock/blob/master/src/lock.js#L180) check in the rwlock module.

So to clarify, the old code would have locks left in memory/disk since it just logs when there was a problem releasing service locks.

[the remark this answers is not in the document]
20k-ultra: > That is the reason the call to release was done within a finally before.

Unless there's another behaviour of this but it's hard to read since the old code didn't use finally it used dispose.

src/lib/update-lock.ts
20k-ultra: We want to remove lockOverride or force since they are basically duplicates but I didn't want to do that in this PR. I just copied exactly how the previous code worked which wanted to evaluated both conditions https://github.com/balena-os/balena-supervisor/blob/33ad877c67a639024a35cd4aa8452d280f757a45/src/lib/update-lock.ts#L123

20k-ultra: I realized after doing `lockService(appId, service, force || lockOverride)` provides the same logic while improving the lockService interface so I did it!

src/lib/update-lock.ts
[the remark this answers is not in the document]
20k-ultra: right, basically: 

old code: might not clear all the service locks but should always clear application lock
new code: might not clear all service locks AND might not clear application lock if service lock clearing has a problem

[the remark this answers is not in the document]
20k-ultra: resolving because we discussed in another chat. I passed release to dispose which uses finally to always call the release if any unlocking of a service fails. This is the exact same behavior (and code) as before.

src/device-state.ts
20k-ultra: I'm open to changing it but perhaps can do in another PR.

20k-ultra:
@balena-ci I self-certify!

PR #1957, opened 2022-06-06
Don't use config.get for appId when checking locks in host config PATCH

cywang117:
In a previous PR I added update lock checks to the PATCH /v1/device/host-config route, but this used `config.get('applicationId')` which we should avoid as it reads from `config.json`, and the appId does not update when moving the device between fleets. It's also not multi-app compatible. With changes from https://github.com/balena-os/balena-supervisor/pull/1953, the update lock check in host-config is now multi-app compatible and avoids going through the config module.

Change-type: patch

### DIFF
--- a/src/host-config.ts
+++ b/src/host-config.ts
@@ -5,6 +5,7 @@ import { promises as fs } from 'fs';
 import * as path from 'path';
 
 import * as config from './config';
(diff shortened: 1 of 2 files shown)


20k-ultra (commented):
lgtm

cywang117:
@balena-ci I self-certify!

PR #1958, opened 2022-06-06
Check for deferred reports after reporting current state

20k-ultra:
This ensures any reports attempted to be sent when a report is already being sent will be checked once the first (pending/active) report is completed.

This is so that any changes that occur during a report make it to the cloud in a more timely manner.

### DIFF
--- a/src/api-binder/report.ts
+++ b/src/api-binder/report.ts
@@ -20,6 +20,7 @@ import { shallowDiff, prune, empty } from '../lib/json';
 
 let lastReport: DeviceState = {};
 let lastReportTime: number = -Infinity;
+let checkForNewerState = false;
 export let stateReportErrors = 0;
 
 type StateReportOpts = {
@@ -70,6 +71,9 @@ async function reportCurrentState(opts: StateReportOpts) {
 		const now = performance.now();
 		// Only try to report if enough time has elapsed since last report
 		if (now - lastReportTime >= constants.maxReportFrequency) {
+			if (checkForNewerState) {
+				checkForNewerState = false;
+			}
 			const currentState = await deviceState.getCurrentForReport(lastReport);
 			const stateDiff = prune(shallowDiff(lastReport, currentState, 2));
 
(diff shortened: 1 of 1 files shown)


src/api-binder/report.ts
```diff
@@ -139,11 +139,26 @@ export async function startReporting() {
...
+				// Check if a report was requested while this one was processing
+				if (reportDeferred) {
+					reportDeferred = false;
+					// Try to report again in case something changed while the last report was running
+					doReport();
+				}
+			}
+		} else {
+			// Defer this report so we can try again once the current one is completed
+			reportDeferred = true;
```
thgreasi, line 161: You could also clear this inside `getStateAndReport()` , right below the `getCurrentForReport()`
See: https://github.com/balena-os/balena-supervisor/blob/e58683537afcc716d785c7632cb3ed3e1a22fdfa/src/api-binder/report.ts#L74

This way the deferred report is discarded, if the `getCurrentForReport()` is called after the change event, which means that the new data ended up being send already (eg b/c of a backoff).

src/api-binder/report.ts
```diff
@@ -139,11 +139,26 @@ export async function startReporting() {
...
-			await reportCurrentState(reportConfigs);
-			reportPending = false;
+			try {
+				// Report changes in the current state
+				await reportCurrentState(reportConfigs);
+			} finally {
+				reportPending = false;
+				// Check if a report was requested while this one was processing
+				if (reportDeferred) {
+					reportDeferred = false;
```
thgreasi, line 154: I like to think about this variable as `shouldReportNewerState` and we move this assignment right after the `getCurrentForReport()`, we should be fine.

20k-ultra: went with `checkForNewerState` but not 100% on it yet.

20k-ultra:
might use https://github.com/balena-os/balena-supervisor/pull/1962 instead

thgreasi:
@20k-ultra I have a feeling that this is more accurate thatn #1962, since a throttle can never know the internal state/progress of the exponential backoff mechanism, which we really need to know. I think this satisfies the requirements of the synchronization scheme that we try to achieve more accurately, regardless of not using a re-usable & generic `throttle` function.
It also works exactly how we describe the solution, which imo is a good sign.

Issue #1959, opened 2022-06-07
Supervisor sometimes sets device name as `local` for balenaOS v2.98.33


pipex:
balenaOS v2.98.33 is the first OS release using supervisor v13. On provisioning a race occurs where if the device gets the target state before the current state report is completed, then the device name on the supervisor database will follow the cloud.

If the device sends the current state report first, then the cloud device name will be set to `local` as that is the [default name from the config module](https://github.com/balena-os/balena-supervisor/blob/master/src/config/schema-type.ts#L90-L94). 

This race did not happen with supervisor v12, which used the v2 device state endpoint on the API. While the v2 endpoint accepts the `name` property as part of the report, the supervisor did not send it 

Here is v12 device current state calculation 
https://github.com/balena-os/balena-supervisor/blob/v12.11.43/src/device-state.ts#L531

vs v13 device current state calculation
https://github.com/balena-os/balena-supervisor/blob/master/src/device-state.ts#L652

zoobot:
Hi @pipex today I burned [balena-go-hello-world](https://github.com/balena-io-examples/balena-go-hello-world) to a device and it named it local. It has supervisor [13.1.11](https://github.com/balena-io/balena-supervisor/blob/master/CHANGELOG.md#v13111) Should I reopen this issue or do something else?

pipex:
@zoobot the fix for this is in 14.0.3, could you try with a newer OS and see if it works?

pipex:
Because of https://github.com/balena-os/balena-supervisor/blob/v13.1.11/src/device-state/preload.ts#L132, if the device is preloaded, this bug may present as the device getting an empty name instead of `local`. The fix in 14.0.3 solves both issues.

PR #1960, opened 2022-06-07
Do not send name as part of the current state

pipex:
This fixes a race condition that could occur with the first current
state report, where if the device managed to send the current state
report first, then the device name on the cloud would be set to `local`
(see #1959).

Closes: #1959
Change-type: patch

# How Has This Been Tested?

- Provision a new device with v2.98.33, there is a good chance the name will be set to `local`
- Sync this latest version of the supervisor using `npm run sync`
- SSH into the device, stop the supervisor and remove the database under `/mnt/data/resin-data/balena-supervisor`
- Change the device name on the cloud
- Start sync again, it should not change the device name

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -481,11 +481,7 @@ export async function setTarget(target: TargetState, localSource?: boolean) {
 
 	globalEventBus.getInstance().emit('targetStateChanged', target);
 
-	const { uuid, apiEndpoint } = await config.getMany([
-		'uuid',
-		'apiEndpoint',
-		'name',
-	]);
+	const { uuid, apiEndpoint } = await config.getMany(['uuid', 'apiEndpoint']);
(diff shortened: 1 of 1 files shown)


src/device-state.ts
```diff
@@ -481,11 +481,7 @@ export async function setTarget(target: TargetState, localSource?: boolean) {
 
 	globalEventBus.getInstance().emit('targetStateChanged', target);
 
-	const { uuid, apiEndpoint } = await config.getMany([
-		'uuid',
-		'apiEndpoint',
-		'name',
-	]);
+	const { uuid, apiEndpoint } = await config.getMany(['uuid', 'apiEndpoint']);
```
pipex, line 484: This change wasn't really necessary, but the name query was being done without any reason

PR #1966, opened 2022-06-14
Use compose long syntax for /lib/firmware and /lib/modules

cywang117:
As /lib/firmware and /lib/modules are always expected to be in the host
filesystem, using long syntax ensures that the Supervisor container
cannot start on an invalid host. (Long syntax volumes error when the host
path does not exist, whereas short syntax volumes create the host path.)

Change-type: major
Closes: #1965

This change is major, because when user applications move to this Supervisor, any containers with the kernel-modules or firmware docker-compose labels will be restarted by the Supervisor due to the short -> long syntax bind migration.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,3 +14,4 @@ services:
       io.balena.features.balena-api: '1'
       io.balena.features.dbus: '1'
       io.balena.features.balena-socket: '1'
+      io.balena.features.kernel-modules: '1'
--- a/entry.sh
+++ b/entry.sh
@@ -42,6 +42,9 @@ if [ -n "${BALENA_ROOT_CA}" ]; then
 	fi
 fi
 
+# TODO: Remove this symlink and modprobe call when the Supervisor
+# can update itself. At that point, we should run modprobe through
+# the bind mount as we have the kernel-modules docker-compose label.
 # Mount the host kernel module path onto the expected location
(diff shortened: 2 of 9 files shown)


docker-compose.yml
```diff
@@ -14,3 +14,4 @@ services:
       io.balena.features.balena-api: '1'
       io.balena.features.dbus: '1'
       io.balena.features.balena-socket: '1'
+      io.balena.features.kernel-modules: '1'
```
cywang117, line 17: This does not have any effect currently, but when the Supervisor is able to upgrade itself, it can bind mount /lib/modules and we can run modprobe through the bind mount instead of relying on the `/mnt/root/lib/modules` symlink in `entry.sh`. See TODO note [here](https://github.com/balena-os/balena-supervisor/pull/1966/files#diff-54a2cc8e7ef0905d7ef106e7af17d918725cb359ff1b082db64d2f1e7e82f30cR45-R47)

cywang117:
I'm marking this as draft temporarily. It's possible that we'll implement some more major changes shortly with the push for Supervisor-as-an-app. In theory, semver major bumps are not sacred, and I agree with this, but the user experience of seeing so many major versions so quickly may be less than ideal. 

When we have more changes that would be a major change, we can cherry-pick the commits into this PR and merge all at once so we only increment a major version once. This change is also not immediately needed so we can sleep on it.

PR #1967, opened 2022-06-14
Use throttle module for reporting current state

20k-ultra:
The `reportCurrentState` function (the function we want to throttle) calculated the diff internally, if we add a throttle and the diff was empty nothing will be reported but now the function is throttled. This causes the device to not send anything for no reason than (until the throttle is cleared).

Lifting the diff out of `reportCurrentState` is complex because `reportCurrentState` does not convey to the outside caller if something was reported on not. This means you don't know when to update `lastReport` which tracks the last reported current start. You can make the function return a boolean to indicate if something was returned but now the throttle needs to know to not throttle on this return type ? It's weird.

Therefore, in this PR I am bringing to the teams attention a solution which uses a callback in `reportCurrentState` to allow the caller to react if a report was sent or not while also being able to throttle `reportCurrentState`.

### DIFF
--- a/src/api-binder/report.ts
+++ b/src/api-binder/report.ts
@@ -2,10 +2,10 @@ import * as url from 'url';
 import * as _ from 'lodash';
 import { delay } from 'bluebird';
 import { CoreOptions } from 'request';
-import { performance } from 'perf_hooks';
 
 import * as constants from '../lib/constants';
 import { withBackoff, OnFailureInfo } from '../lib/backoff';
+import throttle from '../lib/throttle';
 import { log } from '../lib/supervisor-console';
 import { InternalInconsistencyError, StatusError } from '../lib/errors';
 import { getRequestInstance } from '../lib/request';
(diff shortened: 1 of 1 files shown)


src/api-binder/report.ts
```diff
@@ -138,13 +152,25 @@ export async function startReporting() {
...
+					log.info('Reported current state to the cloud');
+					stateReportErrors = 0;
+					lastReport = result.state.full;
+				} else {
+					// No report was sent to reset the throttle
+					throttledReport.cancel();
+				}
+				resolve();
+			});
+		});
```
20k-ultra, line 173: This returns a promise so that the callback and resolve the function in order to retain the recursivelyReport function below. That function will report and wait until it completes before delaying until the next report.

src/api-binder/report.ts
```diff
@@ -64,30 +63,46 @@ async function report({ body, opts }: StateReport) {
 	}
 }
 
-async function reportCurrentState(opts: StateReportOpts) {
+interface ReportResult {
+	reported: boolean;
+	state: StateDiff;
+}
```
20k-ultra, line 69: This is created to convey the result of a report attempt. Provides an interface to know what was reported (full and diff) and if it was reported. This easily allows the caller to update last reported states.

src/api-binder/report.ts
```diff
@@ -64,30 +63,46 @@ async function report({ body, opts }: StateReport) {
...
+
+interface StateDiff {
```
20k-ultra, line 71: Not a fan of this name

Issue #1968, opened 2022-06-14
Read hostname from config.json with container /etc/hostname as backup


cywang117:
Currently the Supervisor reads the host's hostname from `/mnt/root/etc/hostname`. We want to reduce reliance on the `/mnt/root` bind mount, so instead we should read hostname from `/mnt/boot/config.json`, with the container's `/etc/hostname` as a backup. The container `/etc/hostname` is a backup option because, even with `--network=host`, the container hostname file does not update on host hostname changes until the container is *recreated*.

This change is a workaround, until we can implement #1969.

PR #1970, opened 2022-06-14
Read hostname from config.json with container /etc/hostname as backup

cywang117:
We don't need to read the host's hostname through /mnt/root/etc/hostname,
because the hostname is written to config.json on a change. When the hostname
has never changed, it won't be found in config.json, so we can default to
the Supervisor container's /etc/hostname as it will match the host's
/etc/hostname, the network mode being `host`.

This is a step away from reliance on the `/mnt/root` bind mount, and can be used until https://github.com/balena-os/balena-supervisor/issues/1969.

Closes: #1968
Change-type: patch

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -1,6 +1,5 @@
 import * as _ from 'lodash';
 import { promises as fs } from 'fs';
-import * as path from 'path';
 
 import Network from './network';
 import Volume from './volume';
@@ -742,7 +741,7 @@ export class App {
 			opts,
 			supervisorApiHost,
 			hostPathExists,
-			hostnameOnHost,
+			hostname,
 		] = await Promise.all([
 			config.get('extendedEnvOptions'),
 			dockerUtils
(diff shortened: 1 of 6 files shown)


src/compose/app.ts
```diff
@@ -753,19 +752,15 @@ export class App {
...
 			}))(),
 			(async () =>
-				_.trim(
-					await fs.readFile(
-						path.join(constants.rootMountPoint, '/etc/hostname'),
-						'utf8',
-					),
-				))(),
+				(await config.get('hostname')) ??
+				(await (await fs.readFile('/etc/hostname', 'utf-8')).trim()))(),
```
20k-ultra, line 756: I can't make code suggestions on code you didn't modify so I have to past the whole thing... does the following work ? my compiler isn't complaining:

```
		const [
			opts,
			supervisorApiHost,
			hostPathExists,
			hostnameOnHost,
		] = await Promise.all([
			config.get('extendedEnvOptions'),
			dockerUtils
				.getNetworkGateway(constants.supervisorNetworkInterface)
				.catch(() => '127.0.0.1'),
			(async () => ({
				firmware: await pathExistsOnHost('/lib/firmware'),
				modules: await pathExistsOnHost('/lib/modules'),
			}))(),
      (await config.get('hostname') ?? await fs.readFile('/etc/hostname', 'utf-8')).trim(),
		]);
```

cywang117: Looks better, I'll change it

20k-ultra:
codewise lgtm. Not sure the risk in changing this (ie possible issues of the file being wrong or something)

cywang117:
@20k-ultra If hostname has never changed, the Supervisor container's hostname will be the same as the host's hostname. If the hostname has been changed with PATCH /v1/device/host-config (the official, supported method), the host's hostname will be in config.json (added by the `balena-hostname` systemd service). Any other methods to change hostname are not recommended so the Supervisor shouldn't need to take them into consideration. 

The upside of this change is that it should remove a barrier to Supervisor-as-an-app, and the only downside is that this change relies on config.json. However, it may be possible to mount partitions as a privileged container, so we may eventually be able to read from config.json without relying on the `/mnt/root` mount as well (this needs checking though). The best solution as far as getting the host's hostname goes is https://github.com/balena-os/balena-supervisor/issues/1969 but we need to wait until the Supervisor can update itself, so this change is an improvement in the short term.

pipex:
LGTM!

cywang117:
@balena-ci I self-certify!

PR #1979, opened 2022-06-27
Specify that Supervisor will not restore previous lock status from user containers

cywang117:
There is no functionality which 1) stores the filesystem lock status when toggling override
to on, and 2) restore the system to that status when toggling override to off.

Change-type: patch


cywang117:
@balena-ci I self-certify!

PR #1981, opened 2022-07-06
Refactor supervisor Dockerfile to remove custom dependencies

pipex:
Restructure the supervisor image to remove the dependency on the custom `alpine-supervisor-base`
image and the custom node build. The dockerfile is now a multi-stage
build that splits the process into two build stages and two runtime
stages. 

Here is the full list of changes

- The node binary is now copied from  `balenalib/${ARCH}-alpine-node:12-run`, the node binary
now supports running with a debugger.
- The runtime image now inherits from the official `alpine:3.16` image
- Tests are ran within the runtime image configuration instead of the
  build image
- Livepush is ran within the runtime image context
- Unnecessary packages have been removed
- Removed avahi-daemon.conf as that service is not being used
- Fix livepush to work with a multi-stage image. This also deprecates the `image-tag` argument to npm run sync as
`SUPERVISOR_TAG` is no longer used by new OSs
- Fix livepush build on old rpi devices. Allows passing a 'PREFIX'
  argument to let the builder pull images directly from docker hub arch
  repositories. Relates to https://github.com/balena-os/balena-engine/issues/269

Change-type: patch


pipex:
A couple of side effects of the image change. 

Since the image now has multiple stages, the supervisor will see the intermediate build images as dangling and it will try to [clean them up on startup](https://github.com/balena-os/balena-supervisor/blob/e542ca04b3cf622db0b658f8a51e8c176b4ac0f6/src/compose/images.ts#L512). However [livepush will create containers for each stage](https://github.com/balena-io-modules/livepush/blob/c182d1191d66e064ce0c1107320573b5edd5f257/lib/livepush.ts#L83-L89) to prevent that from happening, so the supervisor will report 

```
[debug]   Cleaning up sha256:fa94c6784e33fcb82b4d8786a1ab0a60c561acca5fad0f6c7acadea723b41e12
[event]   Event: Image cleanup error {"error":{"message":"(HTTP code 409) conflict - conflict: unable to delete 7aa9d2b348d5 (cannot be forced) - image is being used by running container 59d0bccaf0f0 ","stack":"Error: (HTTP code 409) conflict - conflict: unable to delete 7aa9d2b348d5 (cannot be forced) - image is being used by running container 59d0bccaf0f0 \n    at /usr/src/app/node_modules/docker-modem/lib/modem.js:257:17\n    at getCause (/usr/src/app/node_modules/docker-modem/lib/modem.js:287:7)\n    at Modem.buildPayload (/usr/src/app/node_modules/docker-modem/lib/modem.js:256:5)\n    at IncomingMessage.<anonymous> (/usr/src/app/node_modules/docker-modem/lib/modem.js:232:14)\n    at IncomingMessage.emit (events.js:326:22)\n    at endReadableNT (_stream_readable.js:1223:12)\n    at processTicksAndRejections (internal/process/task_queues.js:84:21)"}}
```

The supervisor image is now 4MB larger

Dockerfile.template
```diff
@@ -1,70 +1,117 @@
...
 WORKDIR /usr/src/app
 
 RUN apk add --no-cache \
 	g++ \
-	git \
 	make \
-	python \
-	curl \
-	binutils \
+	python3 \
```
cywang117, line 23: Do you know what python3 is for?

cywang117, line 23: Do you know what python(3) is for?

Dockerfile.template
```diff
@@ -1,70 +1,117 @@
 ARG ARCH=%%BALENA_ARCH%%
-ARG NODE_VERSION=12.16.2
```
cywang117, line 2: Dockerfile looking a lot more organized. Good job!

cywang117 (commented):
@pipex Testing this on my device:
```
CONTAINER ID        IMAGE                        COMMAND                  CREATED             STATUS              PORTS               NAMES
0eb36ff8a615        b5c8ce03e537                 "/bin/sh -c 'while t…"   4 minutes ago       Up 4 minutes                            romantic_buck
14fb06b54d1e        e52497a334fd                 "/bin/sh -c 'while t…"   4 minutes ago       Up 4 minutes                            affectionate_lovelace
659afe8d5d97        62a38c3dc715                 "/bin/sh -c 'while t…"   4 minutes ago       Up 4 minutes                            objective_thompson
4a7c56ceb560        livepush-supervisor-14.0.9   "/bin/sh -c 'LIVEPUS…"   4 minutes ago       Up 4 minutes                            balena_supervisor
```
On closing the livepush stream, removing the livepush container & image, and restarting the supervisor service:
```
4eee6114559b        registry2.balena-cloud.com/v2/5335fb2a14f2871a5be88f234500311f   "/usr/src/app/entry.…"   46 seconds ago      Up 45 seconds (health: starting)                       balena_supervisor
0eb36ff8a615        b5c8ce03e537                                                     "/bin/sh -c 'while t…"   8 minutes ago       Up 8 minutes                                           romantic_buck
14fb06b54d1e        e52497a334fd                                                     "/bin/sh -c 'while t…"   8 minutes ago       Up 8 minutes                                           affectionate_lovelace
659afe8d5d97        62a38c3dc715                                                     "/bin/sh -c 'while t…"   8 minutes ago       Up 8 minutes                                           objective_thompson
```

I'm guessing these intermediate containers are linked to staged images? It looks like they'd remain forever on the device due to their entrypoint being a `sleep` command. Shouldn't they be cleaned up on livepush exit?

pipex:
Regarding the `while true` containers. Those are created by livepush for intermediate stages. I'll check the balena CLI codebase, maybe there is an easy way to clean up on Ctrl+C

Dockerfile.template
```diff
@@ -1,70 +1,127 @@
...
-# to be the least intrusive. The commands start commented
-# out because the default build for balenaCI is amd64 (and
-# we can't run any sed preprocessing on it there)
-# RUN ["cross-build-start"]
+###################################################
+# Build the supervisor dependencies
+###################################################
+FROM balenalib/${ARCH}-alpine-node:12-run as build-base
+
+ARG PREFIX
```
klutchell, line 13: Should we default this to `library` for convenience?

cywang117:
@pipex You could try [cleanupIntermediateContainers](https://github.com/balena-io-modules/livepush/blob/c182d1191d66e064ce0c1107320573b5edd5f257/lib/livepush.ts#L157) on Ctrl+c?

pipex:
Yup, adding that now @cywang117

pipex:
@cywang117 Added the cleanup behavior :)

cywang117 (commented):
Nice work, looks good! The `/tmp/livepush-cache.json` is a clever solution.

pipex:
@balena-ci republish

Issue #1945, opened 2022-05-16
Defer current state reports when a report is in progress


20k-ultra:
The Supervisor checks if a report is already pending (being created/sent) before sending another report.

This is a problem because other requests to send a report will be ignored during this time. This may seem ok now because the Supervisor sends current state every N seconds. That behaviour was added just for device metrics. If we allow a user to customize device metrics interval with say 24 hours and a report is pending when a device state changes like container stops, it will take 24 hours for that update to reach the API.

If we make the report which previously got thrown away because a report was already pending use some mechanism to automatically run again when allowed than the device will update the API with the state changes within the max report frequency instead of the device metrics poll.

PR #1983, opened 2022-07-07
Allow backOff to abort throttle

20k-ultra:
The reportCurrentState function (the function we want to throttle) calculated the diff internally, if we add a throttle and the diff was empty nothing will be reported but now the function is throttled. This causes the device to not send anything for no reason than (until the throttle is cleared).

This PR solves that issue by adding AbortLimit exception to the backoff module so it knows when to stop retrying. This AbortLimit is thrown inside the report which allows the doReport function to catch these and clear any pending throttled reports. This effectively gives the reportCurrentState function a way to clear throttled reports if there is nothing to report. Because the throttler queues requests, we still get the benefit of sending a report after a currently pending report has completed which resolves https://github.com/balena-os/balena-supervisor/issues/1945


src/api-binder/report.ts
```diff
@@ -138,12 +129,21 @@ export async function startReporting() {
 		'appUpdatePollInterval',
 	])) as StateReportOpts;
 
-	let reportPending = false;
+	const withThrottle = pThrottle({
+		limit: Infinity,
```
cywang117, line 133: To report once every maxReportFrequency, shouldn't limit = 1?

20k-ultra: that makes much more sense than what I assumed it for was xD thanks

src/lib/limit.ts
```diff
@@ -0,0 +1,164 @@
...
+				// Remove any queued throttles so the next call happens right away!
+				throttledFunc.abort();
+			}
+			// This will be the exception thrown by func
+			throw e;
+		}
+	};
+}
+
+export type OnRetry = {
```
pipex, line 61: Small nit. I liked the `FailureInfo` name that we had for this type. `OnRetry` is like a name you would use for an event, not for something that is informational.

src/lib/limit.ts
```diff
@@ -0,0 +1,164 @@
+import { strict as assert } from 'assert';
+import { promisify } from 'util';
+import { TypedError } from 'typed-error';
+import pThrottle = require('p-throttle');
+
+/**
+ * Generic error thrown when something does not want to be retried/delayed.
+ * This can be things like a backoff or throttle.
+ */
+export class AbortLimit extends TypedError {}
```
pipex, line 10: How about?
```suggestion
export class AbortError extends TypedError {}
```

Issue #1978, opened 2022-06-24
Lockfiles created as directories cannot be overridden


klutchell:
This can be reproduced with the [proper-lockfile](https://www.npmjs.com/package/proper-lockfile) module in this example block: https://github.com/balena-io-playground/lockr-block

```
23.06.22 16:43:52 (-0700) Applying configuration change {"SUPERVISOR_OVERRIDE_LOCK":"1"}
23.06.22 16:43:52 (-0700) Applied configuration change {"SUPERVISOR_OVERRIDE_LOCK":"1"}
23.06.22 16:43:52 (-0700) Updates are locked, retrying in 1s. Reason: Lockfile exists for {"serviceName":"application-locks","appId":"1841113"}
23.06.22 16:43:53 (-0700) Updates are locked, retrying in 2s. Reason: Lockfile exists for {"serviceName":"application-locks","appId":"1841113"}
23.06.22 16:43:55 (-0700) Updates are locked, retrying in 4s. Reason: Lockfile exists for {"serviceName":"application-locks","appId":"1841113"}
23.06.22 16:44:00 (-0700) Updates are locked, retrying in 8s. Reason: Lockfile exists for {"serviceName":"application-locks","appId":"1841113"}
```

cywang117:
@klutchell, This is because the Supervisor lock-checking functionality relies on the lockfile being a file, not a directory. If the override config is specified, the Supervisor will attempt to `unlink` the lockfile which only works for files.

klutchell:
@pipex was looking into this, it seems only the deletion behaviour needs to be fixed.
https://www.flowdock.com/app/rulemotion/r-resinos/threads/fAJOYWix0fvz3NdiaxVmPDV9sUF

PR #1984, opened 2022-07-13
Allow directories to be used as lockfiles

pipex:
Some libraries, like [proper-lockfile](https://www.npmjs.com/package/proper-lockfile) use directories instead of files for locking. This PR allows the supervisor to be able to work with those types of locks when lock override is requested.

Closes: #1978
Change-type: patch


klutchell (commented):
lgtm!

cywang117 (approved):
LGTM

PR #1989, opened 2022-07-19
Disable event tracking

pipex:
The supervisor used to rely on specific event reporting for identifying
issues at runtime. As the platform has grown, it has become much more
difficult to get any signal from the event noise. Recently the API side
for these events has been disabled, meaning these events only
contribute to bandwidth consumption.  This commit disables the
event reporting feature of the supervisor which will be most likely
replaced by something like Sentry in the near future.

Change-type: minor


pipex:
I'll probably need to get rid of the mixpanel token references as well on the config. I'll update soon

Page- (approved):

Issue #643, opened 2018-05-04
Allowing bind-mounts for "relative paths" in a limited directory scope that the supervisor controls and can purge


shaunmulligan:
In some cases, it is useful for users to define bind-mounts in their docker compose, for example in the case where one would have an additional partition or usb drive that they need to bind-mount into a service at runtime.

The proposal would be to allow users to define relative paths , something like:
```
volumes:
     - "./my-data:/opt/my-data"
```
this would be mapped to somewhere like `/mnt/data/user-data` which the supervisor could then delete when the user requests a "purge data"

We would not allow bind-mounts to absolute paths or to parts of the root sys.

lekkas:
Fixes https://github.com/resin-io/hq/issues/1288

lekkas:
This looks like a fairly common request in both paid support and the forums - some threads I just bumped into:

- https://forums.resin.io/t/multicontainer-how-to-configure-persistent-storage/2678
- https://forums.resin.io/t/how-to-mount-sd-card-with-new-multicontainer-support/2706

Issue #1532, opened 2020-11-24
The supervisor should automount removable storage and provide it to specified containers


CameronDiver:
The supervisor should mount both removable storage which is present at boot, and any hot-plugged storage and mount it into `/media` on the OS.

Any services which specify the `io.balena.features.external-storage` label should have the `/media` directory mounted into the container (`/media` being the container path too).

Automounting should be controlled via a host configuration variable (which defaults to true) `BALENA_HOST_CONFIG_AUTOMOUNT`. This should also be exposed via the dashboard.

cc @kenna-smith @shaunmulligan @20k-ultra @szony

20k-ultra:
related https://github.com/balena-io/balena-io/issues/1888

20k-ultra:
The Supervisor is [started](
https://github.com/balena-os/meta-balena/blob/457c37c7fc652af296d045bd00df804aea0dabed/meta-balena-common/recipes-containers/resin-supervisor/resin-supervisor/start-resin-supervisor#L62) with --privileged flag so all that's needed to get udev events is to add ENV UDEV=on to the dockerfile.

20k-ultra:
No configurable toggle will be used just labels as decided in 01-14-2021 arch call.

pipex:
Regarding mounted `/dev` directory on containers by the OS https://github.com/balena-os/balena-engine/issues/125

tmigone:
I believe this should be cross-linked here: https://github.com/balena-io/balena-io/issues/1288

20k-ultra:
If the Supevisor mounts the device in `/media` and then mounts that folder into the containers, we should see if the containers are able to perform operations such as formatting because this would be a useful feature. The concern is that the device will be marked as "in use" so applications won't be able to format the drives.

PR #1993, opened 2022-08-15
Add bind-mount label

maggie0002:
A label which adds a `/bind-mount` folder into the container, bind mounted to empty directory `/mnt/data/bind-mount` on the host (creates directories if they do not exist). 

By preventing users from bind mounting to the host we have prevented some features that require bind mounts to work, such as sharing a mounted USB drive between two or more containers (https://www.balena.io/docs/learn/develop/runtime/#mounting-external-storage-media) or other mounts such as LUKs. I have recently ran in to the same limitation working on a method for encrypting a folder and sharing between containers which relies on creating a mount (some form of encryption being a heavily weighed feature: 

https://jel.ly.fish/brainstorm-topic-1-0-0-front-cnv-91qc7bh/search/pattern-mounting-encrypted-external-drives-10fd075@1.0.0; 
https://jel.ly.fish/brainstorm-topic-1-0-0-front-cnv-91qc7bh/search/support-thread-1-0-0-front-cnv-cxxuebx@1.0.0; 
https://jel.ly.fish/org-balena/loop-balena-io/pattern-users-are-asking-for-full-disk-encryption-d8b4ee91-74aa-4bfc-b2be-b6a0a7c3f9d7@1.0.0. 

This proposal is distinct from some other proposals (https://github.com/balena-os/balena-supervisor/issues/1532) which specifically target a single feature such as USB mounting. It instead aims to provide a non-prescriptive bind-mount option back to users for various use cases without needing to remove any write protections from the OS or open bind-mounts more broadly.

To date, a number of users have been trying to find their own way around the issue, using NFS servers to serve content between containers which adds overhead, reduces performance and creates friction for a feature that could be added out of the box: https://www.balena.io/blog/using-nfs-server-to-share-external-storage-between-containers-balena/

I considered adding in the ability to remove content from the bind mount directory when executing the `purge` function on the dashboard, however the purge targets named volumes and seems like the functionality is quite different. Purge also targets unused volumes, and it is tricker to identify occasions when bind-mount content is unused. This shouldn’t cause issues as any heavy data use that may need purging should be put in to named volumes anyway. 

Closes: https://github.com/balena-io/balena-io/issues/1288
Closes: https://github.com/balena-os/balena-supervisor/issues/643

Change-type: minor


maggie0002:
Linking here to another use case, NFS mounts used between containers could benefit from being able to bind mount. At present each container that needs access to the mount has to run its own NFS, that requires different setups for Ubuntu, Alpine etc. By allowing bind mounts to an empty directory, NFS mounts could potential be initiated by a container (as per good Docker practice of one service per container) and the other containers then access that mount point rather than running NFS themselves. This also makes solutions such as NFS mounts viable as 'blocks' that can be published to the hub for frictionless use. 

https://forums.balena.io/t/nfs-mounts-and-3rd-party-containers/361626/6

maggie0002:
This now seems to be working with the latest changes, although required removing the `MountFlags=slave` from the systemd service which starts balenaEngine. With this flag set to slave, it prevents balenaEngine from using the mount as a shared mount, despite it being configured ok on the OS. 

I am testing with a containerised version of UDEV that would eventually become a block if this method is implemented: https://github.com/balena-community/usb-mount. 

So far the tests have only confirmed that data can be seen and used between two containers from one USB drive. Because of the MountFlags issue I haven't done any tests on things like NFS, or transfer tests etc.

Will explore the MountFlags issue and then come back to it.

PR #1996, opened 2022-08-17
Refactor test suite to use tsconfig paths instead of relative paths

pipex:
Refactor the test suite to stop using relative paths, this is done in preparation of splitting up the test suite in unit and integration tests which will require moving files, so using aliased paths will simplify that task.

List of changes

*  Simplify test commands in package.json. This removes some scripts from `package.json` so the supervisor build lines up better with our [typescript skeleton project](https://github.com/product-os/balena-typescript-skeleton/blob/master/package.json) that includes best practices
* Move sinon to dev dependencies. It was added to prod dependencies by mistake, this fixes that issue
* Setup [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) for easier access the source. This allows to stop using relative paths in tests
* Refactor test suite to use tsconfig paths. Changes all test files to use the new aliases, this will simplify the split of tests in unit and integration later.

Change-type: patch


cywang117 (commented):
I noticed that running both `npm run test` and `npm run test:fast` with ` -- -g 'unit test description'` now log `Warning: Cannot find any files matching pattern "unit test description"` whereas running `npx mocha --config test/.mocharc.js -g 'unit test description'` which is equivalent to `npm run test:node` tests just that single unit @pipex. What could be going on here?

I would also argue for removing `test:fast` entirely with this new setup. The time saved with `test:fast` compared to `test` now is not that big, subjectively.

pipex:
`test:fast` calls `test:build && test:node`, my guess is that npm just doesn't know where to add the additional arguments after `--`. 

I do agree I don't see too much the point for `test:fast`, but I was going by https://github.com/product-os/balena-typescript-skeleton/blob/master/package.json#L32

If you just want to test a single test, the right way is to use `npm run test:node -- -g "some test"`

pipex:
Actually this is what happens with `npm run test:fast -- -g "system information"`

The `--` means "pass the arguments to the called command" , in this case it means append at the end of the `&&` 
```
> balena-supervisor@14.0.13 test:fast
> npm run test:build && npm run test:node "-g" "system information"
```

First the build happens
```
> balena-supervisor@14.0.13 test:build
> tsc --noEmit && tsc --noEmit --project tsconfig.js.json
```

Finally, the argument is passed to the `mocha` call as an extra argument, mocha, looks for file names matching `system information`, and there are none, so it tests everything 

```
> balena-supervisor@14.0.13 test:node
> mocha --config test/.mocharc.js "system information"
```

I'm updating the README, although this will change in the next PR

cywang117:
Ahh I see. Thanks for clarifying @pipex . I was about to suggest the README change but you beat me to it. As for leaving in the `test:fast`, I say take it out even if it's in the skeleton, if there's no other reason for leaving it in. But I'll leave the final decision for that up to you!

pipex:
@balena-ci I self certify!

PR #2000, opened 2022-09-01
Allow most printable ASCII chars for service label key

### LINKED ISSUE #1949: Remove restrictions on docker labels
The supervisor [restricts docker labels](https://github.com/balena-os/balena-supervisor/blob/master/src/types/basic.ts#L114) to `/^[a-zA-Z][a-zA-Z0-9\.\-]*$/` according to [docker recommendations](https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations). 

In reality, the moby engine doesn't impose restrictions on docker labels, which is why services like [traefik](https://doc.traefik.io/traefik/routing/providers/ecs/#routers) and [caddy-docker-proxy](https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/examples/standalone.yaml#L84-L96) make use of specially formatted labels to configure intra-container routing. 

Relaxing the restrictions would give an extra configuration option for users

kb2ma:
# Description

Updated the regex for valid characters for a service label to allow printable ASCII chars except space, double/single quotes, and backtick. These extended characters support apps like Traefik and caddy-docker-proxy. Characters still not supported likely are not useful as label keys and error prone.

Fixes #1949

# Type of change

patch

# How Has This Been Tested?

Updated `06-validation.spec.ts`. Also tested by POSTing all valid chars to `/v2/local/target-state`.

### DIFF
--- a/src/types/basic.ts
+++ b/src/types/basic.ts
@@ -109,9 +109,11 @@ export type VariableName = t.TypeOf<typeof VariableName>;
 
 /**
  * Valid label names are between 0 and 255 characters
- * and match /^[a-zA-Z][a-zA-Z0-9\.\-]*$/
+ * and each character is any printable ASCII character except space (0x20),
+ * double and single quotes ( " 0x22,  ' 0x27) and backtick ( ` 0x60). Rationale
+ * is to accept a character unless likely not useful and error prone.
  */
-const LABEL_NAME_REGEX = /^[a-zA-Z][a-zA-Z0-9\.\-]*$/;
+const LABEL_NAME_REGEX = /^[!#-&(-_a-~]+$/;
 
 export const LabelName = new t.Type<string, string>(
 	'LabelName',
@@ -125,7 +127,7 @@ export const LabelName = new t.Type<string, string>(
 					: t.failure(
 							s,
 							c,
-							"needs to start with a letter and may only contain alphanumeric characters plus '-' and '.'",
+							'may contain printable ASCII characters except space, single/double quotes and backtick',
 					  ),
 			),
 		),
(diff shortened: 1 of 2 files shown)


cywang117 (commented):
This LGTM! Do you have to add `Closes: <issue number` in the commit footer for bulldozer-bot to auto-close the issue on merge?

dfunckt:
Unless the Supervisor assigns labels to containers/volumes/etc. and balena-engine rejects them, it doesn't really need to validate them.

I think this in fact applies in general: data that only ever comes from the target state will necessarily be valid because tooling (eg. the CLI, builder, API, etc.) will have rejected invalid releases before they make it to the Supervisor.

kb2ma:
> Unless the Supervisor assigns labels to containers/volumes/etc. and balena-engine rejects them, it doesn't really need to validate them.

But the accepted character set on an older Supervisor version may not include characters from a newer balena-builder or balena-cli. So in a sense the Supervisor is acting as the guardian of the device. It knows best what the device should accept.

cc: @pipex

pipex:
I agree with Ken, even though there are some validations on the CLI and backend, the user may be using an incompatible CLI version, so the supervisor should still do some validations (without going too crazy). 

The supervisor also exposes the [target-state endpoint](https://www.balena.io/docs/reference/supervisor/supervisor-api/#post-v2localtarget-state) that allows to bypass the CLI entirely, so validation is needed there for sure.

dfunckt:
> But the accepted character set on an older Supervisor version may not include characters from a newer balena-builder or balena-cli.

@kb2ma @pipex from a user’s perspective this will appear as failure to deploy a release. The correct way to introduce breaking changes is via contracts. Let’s chat in JF how to go about doing that.

I wasn’t aware about that endpoint but my point still stands, validation should happen in that endpoint and ideally using the same primitives/code the CLI and builder use.

kb2ma:
@balena-ci I self-certify!

PR #2002, opened 2022-09-06
Add custom DTB support for imx8mm-var-som

acostach:
This adds support for custom device-trees for the upcoming imx8mm-var-som devkit.

Change-type: patch


cywang117 (approved):
LGTM

PR #2003, opened 2022-09-06
Remove side effects for module imports

pipex:
The supervisor uses the following pattern for async module
initialization

```typescript
// module.ts

export const initialised = (async () => {
    // do some async initialization
})();

// somewhere else
import * as module from 'module';

async function setup() {
  await module.initialised;
}
```

The above pattern means that whenever the module is imported, the
initialisation procedure will be ran, which is an anti-pattern.

This converts any instance of this pattern into a function

```typescript
export const initialised = _.once(async () => {
    // do some async initialization
});
```

And anywhere else on the code it replaces the call with a

```typescript
await module.initialised();
```

Change-type: patch

### DIFF
--- a/src/api-binder/index.ts
+++ b/src/api-binder/index.ts
@@ -547,10 +547,10 @@ async function reportInitialName(
 
 export let balenaApi: PinejsClientRequest | null = null;
 
-export const initialized = (async () => {
-	await config.initialized;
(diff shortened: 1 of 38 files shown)


cywang117 (approved):
LGTM!

PR #2004, opened 2022-09-07
Set desired es-version for downstream modules that support it

thgreasi:
Change-type: patch
See: https://jel.ly.fish/thread-9a2eafe4-0a8f-4655-896b-42466ad03d2b

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -520,9 +520,9 @@
       "dev": true
     },
     "@balena/es-version": {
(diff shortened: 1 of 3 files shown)


Page- (approved):

thgreasi:
@balena-ci retest

PR #2008, opened 2022-09-12
Update to nodejs 14

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,11 +32,11 @@
     "dbus": "^1.0.7",
     "mdns-resolver": "^1.0.0",
     "semver": "^7.3.2",
-    "sqlite3": "^4.1.1",
+    "sqlite3": "^5.0.11",
     "systeminformation": "^5.6.10"
   },
   "engines": {
-    "node": "^12.16.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -32,11 +32,11 @@
     "dbus": "^1.0.7",
     "mdns-resolver": "^1.0.0",
     "semver": "^7.3.2",
-    "sqlite3": "^4.1.1",
+    "sqlite3": "^5.0.11",
```
Page-, line 35: This is required for node >=14 but looks to cause failures on balenaCI related to it running an older docker version but looks to be fine with the `balena push` - I expect the best solution will likely be to migrate the supervisor to flowzone

pipex:
I tested this on a device and I didn't see any obvious problems while changing between fleets, pinning releases, etc. I added the extra commit, but it is just a configuration fix and unrelated to the CI issues. Once the flowzone PR merges, LGTM

pipex (approved):
LGTM!

PR #2009, opened 2022-09-12
Update blinking to 0.0.5

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


Page-:
@balena-ci I self-certify!

PR #2010, opened 2022-09-12
Switch from balenaCI to flowzone

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,20 @@
...
+  flowzone:
+    name: Flowzone
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
+    secrets:
+      FLOWZONE_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
+      GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+      GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+      BALENA_API_KEY: ${{ secrets.BALENA_API_KEY }}
+    with:
+      protect_branch: false
```
Page-, line 20: This should be removed before merging

pipex: How does this interact with the deploy-to-balena action setup here https://github.com/balena-os/balena-supervisor/blob/master/.github/workflows/balena-ci.yml?

cywang117: Maybe `balena-ci.yml` isn't needed anymore because Flowzone [pushes draft releases to balena](https://github.com/product-os/flowzone#balena) if `balena.yml` is present in project root @pipex ? I'm asking in [JF](https://jel.ly.fish/5e1dc3c9-fc6e-44fe-b60a-765e9fe21f38?event=1da613db-ea58-4d53-9f95-809900f10ea2) as well to see if there are any existing repos we can look at as an example.

pipex: I would expect that flowzone could also replace the deploy-to-balena config @cywang117, but the DTB worflow configures the fleet names and also the push to staging, I don't know if all of that is deduced by just the contract

cywang117: May need to sleep this PR until https://github.com/product-os/flowzone/issues/198 is resolved. From the (internal) JF thread I linked above, Flowzone would need to support setting the `environment` parameter as in the balena-ci.yml config, and subsequently would need a scalable way of figuring out the balena API secret env var.

cywang117: Waiting on admin access to repo to set FLOWZONE_TOKEN, so we can test the Flowzone workflow with versioning.

cywang117: balena-ci.yml was removed with latest config and the draft releases on balena cloud & staging are there and have the correct version.

package.json
```diff
@@ -19,7 +19,7 @@
     "test:env": "docker-compose -f docker-compose.test.yml -f docker-compose.dev.yml up --build; npm run compose:down",
     "test:compose": "docker-compose -f docker-compose.yml -f docker-compose.test.yml up --build --remove-orphans --exit-code-from=sut ; npm run compose:down",
     "test:base": "npm run lint && npm run test:build && npm run test:unit && npm run test:legacy",
-    "test": "npm run lint && npm run test:build && npm run test:node",
+    "test": "npm run lint && npm run test:build && (npm run test:node || true)",
```
pipex, line 22: I would not do this. Let's just get rid of the current `test` and then rename `test:base` to `test`. With that can get rid of `test:node` too, and simplify the scripts. 

You'll have to change `Dockerfile.template` to use `npm run test` and update the [README](https://github.com/balena-os/balena-supervisor#testing)

pipex: Actually let's not get rid of `test:node`, it's useful for running just the tests while developing in `test:env`

pipex:
@cywang117 can we also modify the comment at the top of `Dockerfile`?

```
# IMPORTANT: do **not** use this dockerfile
# This dockerfile only exists to allow flowzone to run
# docker-compose.test.yml but without the CI needing to do extra
# work building a throwaway image. The real build happens in the balena builder
# so for testing we just want to build up to the test stage
```

cywang117:
Done @pipex ! https://github.com/balena-os/balena-supervisor/commit/57c24488fa420a1badc9a467de7eb69523088074

package.json
pipex: Don't forget about this change before merging @cywang117  :)

package.json
```diff
@@ -24,7 +24,8 @@
     "prettify": "balena-lint -e ts -e js --typescript --fix src/ test/ typings/ build-utils/ webpack.config.js",
     "release": "tsc --project tsconfig.release.json && mv build/src/* build",
     "sync": "ts-node --files sync/sync.ts",
-    "clean": "rimraf build"
+    "clean": "rimraf build",
+    "preinstall": "./install-dbus.sh || true"
```
pipex, line 28: One last thought. Maybe the script could go in `build-utils`?

cywang117: Good shout. https://github.com/balena-os/balena-supervisor/commit/b680d4996657346a988cd8c45ac7f12a234a1662

package.json
cywang117: Done in https://github.com/balena-os/balena-supervisor/commit/6eec238ef4f3891d7dd85e58433ee78ee087b6e1!

.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: Deploy to balena-cloud
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
+    secrets:
+      FLOWZONE_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
+      GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+      GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+      BALENA_API_KEY: ${{ secrets.BALENA_TOKEN }}
+    with:
+      balena_environment: balena-cloud.com
+      protect_branch: false
```
cywang117, line 39: Reminder to whomever would like to merge this if it passes all checks -- remove both `protect_branch` lines. (They default to `true`)

package.json
```diff
@@ -15,11 +15,10 @@
     "test:unit": "mocha --config test/unit/.mocharc.js",
     "test:integration": "mocha --config test/integration/.mocharc.js",
     "test:legacy": "mocha --config test/legacy/.mocharc.js",
-    "test:node": "npm run test:unit && npm run test:integration && npm run test:legacy",
+    "test:node": "npm run lint && npm run test:build && npm run test:unit && npm run test:integration && npm run test:legacy",
```
pipex, line 18: Why add the linting here? `test:node` as I understanding is just meant as a quick test run.

pipex: I removed it and pushed the changes to re-trigger the jobs

cywang117: Good shout!

.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    types: [opened, synchronize, closed]
+    branches:
+      - "main"
+      - "master"
+
+jobs:
+  flowzone-staging:
+    name: balena-staging
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@protect
```
Page-, line 13: This needs to go back to
```suggestion
    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
```
once working and https://github.com/product-os/flowzone/pull/208 is merged

.github/workflows/flowzone.yml
```diff
@@ -0,0 +1,45 @@
...
+        balena_os/aarch64-supervisor,
+        balena_os/amd64-supervisor,
+        balena_os/armv7hf-supervisor,
+        balena_os/i386-supervisor,
+        balena_os/rpi-supervisor
+      job_name: balena-staging
+
+  flowzone:
+    name: balena-cloud
+    uses: product-os/flowzone/.github/workflows/flowzone.yml@protect
```
Page-, line 31: This needs to go back to
```suggestion
    uses: product-os/flowzone/.github/workflows/flowzone.yml@master
```
once working and https://github.com/product-os/flowzone/pull/208 is merged

Issue #2006, opened 2022-09-09
Provisioning an aarch64 device into an armv7 architecture causes the target state supervisor to be installed


pipex:
Supervisors since v13 skip the supervisor configuration from the target state as a way to avoid ending up with two supervisors (the supervisor cannot install itself yet). This is done by this function that uses the architecture from `device-type.json` to find the supervisor app uuid and service name from a set of predefined values.

https://github.com/balena-os/balena-supervisor/blob/master/src/compose/app.ts#L811

The problem is that armv7 fleets allow installing aarch64 devices, which causes the above check to fail.

PR #2012, opened 2022-09-12
Fix filtering of the supervisor app on the target state

pipex:
Under some conditions, an aarch64 device may get a reference to a armv7hf supervisor on the target state. One of the ways this can happen is if an aarch64 device is added to an armv7hf fleet and the target supervisor is set before the device fully provisions.

If that happens, the previous filtering for the supervisor app (which relied on the architecture in device-type.json) would fail and the user would end up with two supervisor containers, one running correctly and the other crash looping.

This fixes the filtering and just checks if the supervisor uuid/service name belongs to a group of known uuids.

Closes: #2006
Change-type: patch


cywang117 (approved):
LGTM

PR #2013, opened 2022-09-13
Fix withDefault type helper to work with boolean

pipex:
`withDefault` is a type helper that allows to create a type that defaults to a default value when trying to decode a nullish value. That type was not correctly working with boolean types, causing `false` values to be replaced by true. This would specifically cause issues when parsing the target state, where a `running: false` in a service would become a `running: true` due to the type decoding.

Change-type: patch

### DIFF
--- a/src/types/state.ts
+++ b/src/types/state.ts
@@ -123,17 +123,16 @@ export type DeviceState = {
 };
 
 // Return a type with a default value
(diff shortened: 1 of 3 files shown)


cywang117:
@balena-ci rebase

cywang117 (approved):
LGTM!

Issue #2014, opened 2022-09-13
Target state with a service set to `running: false` should not start the service


pipex:
Setting `running: false` in a service target state causes the supervisor to install and start the service and then kill it later. Although this gets the device to the target, there may be unintended side effects from starting the service.

PR #2015, opened 2022-09-13
Only install service if running is set to false

pipex:
The supervisor supports target state `running: false` for services. This state indicates that the service should be stopped if already running, or that the container should just be created and never started if the container does not exist. This commit fixes the latter behavior.

Although nothing in our platform currently sends this target state, this enables some potential use cases, e.g. only starting some services in manufacturing and starting the rest of the services when the device actually connects.

Change-type: patch
Closes: #2014

### DIFF
--- a/src/compose/service-manager.ts
+++ b/src/compose/service-manager.ts
@@ -298,6 +298,15 @@ export async function start(service: Service) {
 
 	try {
 		const container = await create(service);
+
+		// Exit here if the target state of the service
+		// is set to running: false
+		// QUESTION: should we split the service steps into
+		// 'install' and 'start' instead of doing this?
+		if (service.config.running === false) {
+			return container;
+		}
+
 		containerId = container.id;
 		logger.logSystemEvent(LogTypes.startService, { service });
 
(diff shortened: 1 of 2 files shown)


src/compose/service-manager.ts
```diff
@@ -310,6 +319,7 @@ export async function start(service: Service) {
 		} catch (e) {
 			// Get the statusCode from the original cause and make sure it's
 			// definitely an int for comparison reasons
+			// QUESTION: does this ever happen?
```
cywang117, line 322: You're right & I don't think this happens as these are all Docker status codes. The decode here may be too defensive with little upside

pipex: We could just remove the decode and just see if something breaks, my guess this was necessary with older engine versions

pipex: I'm doing the change

cywang117: SGTM @pipex!

pipex: Done !

pipex:
@balena-ci I self certify!

PR #2017, opened 2022-09-15
Set fail-fast to false in staging job

pipex:
Change-type: patch


pipex:
Not needed, this was meant to fix an issue with finalizing the publish to staging, but the cause for that seems to be fixed by https://github.com/product-os/flowzone/pull/210

PR #2023, opened 2022-09-20
Update to webpack 5

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
 	"compilerOptions": {
 		"baseUrl": "./",
-		"target": "ES2019",
+		"target": "ES2021",
```
Page-, line 4: This was previously blocked by webpack 4 not supporting es2021 syntaxThis was previously blocked by webpack 4 not supporting es2021 syntax

pipex (approved):
LGTM

pipex:
I already tested on a device before merging. Forgot to mention it here

PR #2029, opened 2022-09-23
Add test troubleshooting case that appears after upgrading system

cywang117:
The error may be something to the effect of `failed to add the host (vethd6242de) <=> sandbox (veth8e47b1a) pair interfaces: operation not supported`.

Change-type: patch


cywang117:
@balena-ci I self-certify!

PR #2030, opened 2022-09-26
Update dependencies to fix NPM build

pipex:
Change-type: patch


package.json
```diff
@@ -75,7 +75,9 @@
     "@types/tmp": "^0.1.0",
     "@types/webpack": "^5.28.0",
     "@types/yargs": "^15.0.12",
+    "balena-auth": "^4.1.2",
     "balena-register-device": "^8.0.0",
+    "balena-request": "^11.5.9",
```
Page-, line 80: Kinda sucks we need to add these two when we never actually use them (I think it's for balena-register-device but it uses dependency injection so we just inject minimal compatible shims for supervisor code size/memory usage)

pipex: Yeah, I wasn't too happy about it either. Will they get tree-shaken by webpack? I'm open to suggestions for alternatives

Page-: They will, webpack won't even look at them if they're not imported

Page- (approved):

PR #2031, opened 2022-09-27
Use inherited secrets for flowzone

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -11,12 +11,7 @@ jobs:
   flowzone:
     name: Flowzone
     uses: product-os/flowzone/.github/workflows/flowzone.yml@master
-    secrets:
-      FLOWZONE_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
-      GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
-      GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
-      BALENA_API_KEY: ${{ secrets.BALENA_TOKEN }}
-      CUSTOM_JOB_SECRET_1: ${{ secrets.BALENA_STAGING_TOKEN }}
+    secrets: inherit
     with:
       balena_slugs: |
         balena_os/aarch64-supervisor,
--- a/.github/actions/finalize/action.yml
+++ b/.github/actions/finalize/action.yml
@@ -15,7 +15,7 @@ runs:
   steps:
     - uses: balena-io/deploy-to-balena-action@v0.13.0
       with:
-        balena_token: ${{ fromJSON(inputs.secrets).CUSTOM_JOB_SECRET_1 }}
+        balena_token: ${{ fromJSON(inputs.secrets).BALENA_STAGING_TOKEN }}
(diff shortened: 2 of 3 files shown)


.github/workflows/flowzone.yml
```diff
@@ -11,12 +11,7 @@ jobs:
   flowzone:
     name: Flowzone
     uses: product-os/flowzone/.github/workflows/flowzone.yml@master
-    secrets:
-      FLOWZONE_TOKEN: ${{ secrets.FLOWZONE_TOKEN }}
-      GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
-      GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
-      BALENA_API_KEY: ${{ secrets.BALENA_TOKEN }}
```
Page-, line 18: This one is problematic as the name doesn't match up properly, `BALENA_TOKEN` vs `BALENA_API_KEY`

.github/workflows/flowzone.yml
```diff
@@ -11,12 +11,7 @@ jobs:
...
-      CUSTOM_JOB_SECRET_1: ${{ secrets.BALENA_STAGING_TOKEN }}
+    secrets: inherit
```
Page-, line 14: This won't work until we have gh enterprise and `product-os` is within the same enterprise

Page-:
@balena-ci I self-certify!

cywang117:
@balena-ci rebase

.github/workflows/flowzone.yml
cywang117: Checks are still failing because BALENA_TOKEN is blank from the deploy-to-balena action. I guess the 2 solutions forward to pushing this through are:

- Flowzone should pass both BALENA_API_KEY as BALENA_TOKEN to the GitHub action to conform to its interface
- The GitHub action should accept both BALENA_API_KEY and BALENA_TOKEN

cywang117: PR: https://github.com/product-os/flowzone/pull/256

Gonna push a test commit to see if this works (I except it will)

Page-: @cywang117 I think the correct answer is to add `BALENA_API_KEY` to the balena-os secrets, the problem is I don't know where the contents of `BALENA_TOKEN` for balena-os came from in order to add the extra API key - if either you or @pipex know where they came from/who added them then I can handle the rename

cywang117: Converting to draft real quick so this doesn't get auto-merged

cywang117: I'm not quite sure where, the secret isn't in the repository level I'm guessing @Page- ?

Page-: It's there at the org level @cywang117 but I cannot rename it or access the contents so I need to get the contents anew from the original source in order to add it as a "new" secret called `BALENA_API_KEY`

cywang117:
Testing https://github.com/balena-io/deploy-to-balena-action/pull/214 once that image builds

Issue #2032, opened 2022-09-27
Supervisor 14.2.3 - 14.2.5 fail to start because of missing migrations


pipex:
On start, the supervisor will just throw some minimized code and at the end of the error the lines

```                                                                                                                                                                                                                                                                                                                    
Error: Cannot find module '/usr/src/app/dist/migrations/M00008.js'
    at webpackEmptyContext (/usr/src/app/dist/app.js:2:1890126)
    at module.exports (/usr/src/app/dist/app.js:2:1888408)
    at async /usr/src/app/dist/app.js:2:1869008
    at async Promise.all (index 0)
    at async Migrator.latest (/usr/src/app/dist/app.js:2:1868932)
    at async Supervisor.init (/usr/src/app/dist/app.js:22:134589) {
  code: 'MODULE_NOT_FOUND'
}
```

The migrations are not really missing, they can be seen inside the container. This is probably due to the Knex upgrade in version 14.2.3. More investigation is needed. 

The supervisor releases have been invalidated on balena-cloud. The fix will be in 14.2.6

pipex:
Steps to replicate:

Remove the database: `rm /mnt/data/resin-data/balena-supervisor/database.sqlite`
Update to supervisor 14.2.3 or above

cywang117:
Fixed with https://github.com/balena-os/balena-supervisor/pull/2033, Supervisor v14.2.6

PR #2033, opened 2022-09-27
Fix migrations hardcode to work with knex 0.95.15 

pipex:
The latest upgrade of knex broke the webpack loader that hardcodes the
migration paths in the built code, that cause supervisors 14.2.3 to
14.2.5 (inclusive) to break anytime migrations needed to be ran (see linked issue).

This commit fixes the loader to work with the included knex version.

Relates-to: #2032
Change-type: patch


pipex:
@balena-ci I self certify!

PR #2035, opened 2022-09-30
Start state engine and API binder in parallel

pipex:
The state engine and preloading is performed before the device gets a chance to register, while this is desirable for preloaded apps, it introduces a delay on registration which is known to cause issues since the VPN is also trying to connect at the same time.

This triggers a simultaneous start of the device engine, the API binder and the supevisor API to avoid delays.

Change-type: patch


cywang117 (approved):
LGTM!

PR #2036, opened 2022-10-03
Update type dependencies

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


src/supervisor-api.ts
```diff
@@ -48,7 +33,22 @@ export class SupervisorAPI {
 		this.healthchecks = healthchecks;
 
 		this.api.disable('x-powered-by');
-		this.api.use(expressLogger);
+		this.api.use(
+			morgan(
```
Page-, line 37: Moving it in here allows for automatic type inference to work properly (ie that it's an `express.Request` rather than a plain `http.IncomingMessage`)

Page-:
@balena-ci I self-certify!

PR #2040, opened 2022-10-19
Migrate some more legacy tests to integration

pipex:
- docker-util tests
- db-format
- firewall

Change-type: patch


pipex:
@balena-ci I self certify!

PR #2044, opened 2022-10-25
Migrate core API tests to unit/integration excluding v1/v2 routes

cywang117:
This excludes route tests or refactoring. That will be done in a separate PR which will separate route handler logic from the action logic of the route. 

Also, created tests for API middleware.

The following pattern is used for supertest [per the docs](https://github.com/visionmedia/supertest#example) to avoid interacting with a test IP address or port:
```
import * as request from 'supertest';

it('should...', async () => {
  await request(app).get('/')
  ...
})
```
where `app` refers to an instance of `express.Application`, initialized using `const app = express()`. The previous pattern was similar to:
```
const request = supertest(`http://127.0.0.1:${mockedPort}`);
await request.get('/')
```

Change-type: patch


test/legacy/21-supervisor-api.spec.ts
```diff
@@ -1,288 +0,0 @@
...
-
-			await request
-				.get('/v2/applications/1/state')
-				.set('Accept', 'application/json')
-				.set('Authorization', `Bearer ${newScopedKey}`)
-				.expect(200);
-		});
-	});
-
-	describe('State change logging', () => {
```
cywang117, line 163: There was a suite for API startup & shutdown logging in the legacy tests, but I don't think the test suite is particularly useful so deleted it entirely instead of moving it to unit/integration.

cywang117:
@balena-ci rebase

cywang117:
@balena-ci I self-certify!

PR #2045, opened 2022-10-26
config/backends: Add Jetson AGX Orin Devkit to the list of DTs that support configurable device-trees

acostach:
Change-type: patch

Connects-to: https://github.com/balena-os/balena-jetson/issues/389


pipex (approved):
Looks good

PR #2047, opened 2022-10-27
Update to nodejs 18

Page-:
This also removes node-dbus as a dependency in favour or  `@balena/systemd`

Change-type: minor


cywang117:
Checking this out

package.json
```diff
@@ -32,16 +32,16 @@
...
     "got": "^12.5.3",
     "mdns-resolver": "^1.0.0",
     "semver": "^7.3.2",
     "sqlite3": "^5.0.11",
     "systeminformation": "^5.6.10"
   },
   "engines": {
-    "node": "^16.17.0",
-    "npm": "^8.15.0"
+    "node": ">=16.17.0",
```
pipex, line 43: What do you think about this @Page- ? I want to use it as a forcing function to ensure updates to latest node happen faster.

Dockerfile.template
```diff
@@ -1,12 +1,12 @@
 ARG ARCH=%%BALENA_ARCH%%
 ARG FATRW_VERSION=0.2.9
-ARG NODE="nodejs<18"
-ARG NPM="npm<9"
+ARG NODE="nodejs>16"
```
Page-, line 3: Is there not a better way to do this since this would to my understanding install nodejs 20?

package.json
```diff
@@ -32,16 +32,16 @@
   "private": true,
   "dependencies": {
     "@balena/happy-eyeballs": "0.0.6",
-    "dbus": "^1.0.7",
+    "@balena/systemd": "^0.4.0",
```
Page-, line 35: Can this update be merged prior to the nodejs bump? That way each individual PR/change can be smaller and tested in isolation in case any issues do arise

package.json
Page-: I don't think it'll help with that at all since the nodejs version used/tested will be decided by the dockerfile won't it?

Dockerfile.template
pipex: We can use ` nodejs<20`,  but the node version won't change unless we bump the alpine version.

package.json
pipex: Unfortunately no, we need at least rust v1.64 to be able to compile the module (because of dependencies), but alpine 3.17, which comes with that rust version, also comes with nodejs 18

package.json
pipex: Yeah, you have a good point. I just don't want to restrict again the Node version. If this fails in Node 20, I want to hear about it

package.json
Page-: Is the rust part not precompiled as part of the module's CI?

package.json
Page-: Yeah it's fine to have like this, it's just unlikely to be a forcing function

package.json
pipex: No, `@balena/systemd` is not a WASM module, is just some bindings to node NAPI. It has to be compiled when doing `npm install` so the build of the module matches the architecture of the host, similar to what we do with `sqlite3`

pipex:
@cywang117 check out this PR for the D-Bus stuff

src/network.ts
```diff
@@ -98,6 +98,7 @@ export const startConnectivityCheck = _.once(
 					blink.pattern.stop();
 				} else {
 					log.info('Waiting for connectivity...');
+					//
```
cywang117, line 101: Unfinished comment?

cywang117 (commented):
Nothing stood out to me through static analysis, however: is Rust now a dependency for the Supervisor to be built, tested, and developed on, due to @balena/systemd? If so, we should document it in the developer README.

I'll run some tests on-device tomorrow and update this comment with a ping.

package.json
Page-: I though you can still have prebuilds for a bunch of architectures and it'll only fallback to a local build if it doesn't find a prebuilt one that matches?

pipex: You might be able to do it, I haven’t really looked into that option. I’ll investigate a bit

pipex:
> Nothing stood out to me through static analysis, however: is Rust now a dependency for the Supervisor to be built, tested, and developed on, due to @balena/systemd? If so, we should document it in the developer README.

You are right @cywang117, I'll update the readme as well

pipex:
> we should document it in the developer README.

@cywang117 updated the README, let me know what you think

cywang117:
LGTM @pipex ! Also ran the branch on a local device and did not encounter any issues.

pipex:
Not merging yet, there is something going on with the build, the image size somehow jumped from 75MB to 370MB. I need to investigate

pipex:
> Not merging yet, there is something going on with the build, the image size somehow jumped from 75MB to 370MB. I need to investigate

Ok, I know why, [@balena/systemd](https://github.com/balena-io-modules/node-systemd/) is not deleting the `target` folder after install, which accounts for the 250MB difference in size

pipex:
@cywang117 https://github.com/balena-io-modules/node-systemd/pull/6

pipex:
I self-certify!

PR #2048, opened 2022-10-28
Separate route and action tests for Supervisor v1 API

cywang117:
This PR separates route logic (i.e. parsing input, expected response status codes) from action logic (i.e. getting device information, performing restart) in the Supervisor API codebase. It also moves the API tests over to unit/integration, and writes separate tests for route & action logic. It adds any missing tests when needed.

Continued from the core API tests PR (https://github.com/balena-os/balena-supervisor/pull/2044).

v2 refactor to be continued; partially complete here where actions between v1 & v2 are shared.

Change-type: minor


src/device-api/index.ts
```diff
@@ -59,30 +57,16 @@ export class SupervisorAPI {
...
-				if (shouldUpdateCloudKey) {
-					// report the new key to the cloud API
-					reportCurrentState({
-						api_secret: newKey,
-					});
+				const { apiKey: oldKey } = req.auth;
+				const newKey = await actions.regenerateKey(oldKey);
+				if (newKey === oldKey) {
+					// Regenerate failed, return old key
+					return res.status(500).send(oldKey);
```
cywang117, line 67: The `500` status is a new interface here to help differentiate whether a key refresh was successful or not.

src/device-api/actions.ts
```diff
@@ -64,3 +70,33 @@ export const regenerateKey = async (oldKey: string) => {
...
+		const { services } = currentState.local.apps?.[appId];
+		applicationManager.clearTargetVolatileForServices(
+			services.map((svc) => svc.imageId),
+		);
+
+		return deviceState.pausingApply(async () => {
+			await Bluebird.each(services, async (service) => {
+				await serviceManager.kill(service, { wait: true });
+				await serviceManager.start(service);
+			});
```
cywang117, line 120: We could convert this to
```
for (const service of services) {
  await kill
  await start
}
```
instead of Bluebird

test/integration/device-api/actions.spec.ts
```diff
@@ -25,10 +29,279 @@ describe('regenerates API keys', () => {
...
+		let targetState: Awaited<ReturnType<typeof generateTarget>>;
+		let containers: Docker.ContainerInspectInfo[];
+
+		before(async () => {
+			targetState = await generateTarget({
+				serviceCount,
+				serviceNames,
+			});
+
+			// Create a single-container application in local mode
```
cywang117: The "single-container" typo is fixed in a future commit

src/device-api/v1.ts
```diff
@@ -46,84 +49,50 @@ router.post('/v1/restart', (req: AuthorizedRequest, res, next) => {
...
-				})
-				.then((service2) =>
-					res.status(200).json({ containerId: service2.containerId }),
-				);
-		})
-		.catch(next);
+			// QUESTION: The old v1 stop/start endpoints return the containerId
+			// of the new service. This 'OK' response is more in line with v2 service
+			// action responses -- should we use this or keep returning the containerId?
+			return res.status(200).send('OK');
```
cywang117, line 74: 👀

src/device-api/v1.ts
```diff
@@ -46,84 +49,50 @@ router.post('/v1/restart', (req: AuthorizedRequest, res, next) => {
...
-	const appId = checkInt(req.params.appId);
-	const force = checkTruthy(req.body.force);
-	if (appId == null) {
-		return res.status(400).send('Missing app id');
-	}
-
-	return Promise.all([applicationManager.getCurrentApps(), getApp(appId)])
-		.then(([apps, targetApp]) => {
-			if (apps[appId] == null) {
-				return res.status(400).send('App not found');
```
cywang117, line 63: This became a `404 Not Found` (thrown from the underlying action as a `NotFoundError`) instead of `400 Bad Request`, which is an interface change, but 404 is the correct response that should be returned anyway.

src/device-api/v1.ts
```diff
@@ -46,84 +49,50 @@ router.post('/v1/restart', (req: AuthorizedRequest, res, next) => {
...
-			}
-			const app = apps[appId];
-			let service = app.services[0];
-			if (service == null) {
-				return res.status(400).send('No services on app');
```
cywang117, line 68: Same as above, this also became `404 Not Found` instead of `400 Bad Request` which is an interface change, but I believe is more correct.

src/device-api/v1.ts
```diff
@@ -94,29 +95,21 @@ const handleLegacyServiceAction = (action: CompositionStepAction) => {
...
-			Error: (e != null ? e.message : undefined) || e || 'Unknown error',
-		});
-	}
+const handleDeviceAction = (action: deviceState.DeviceStateStepTarget) => {
+	return async (req: AuthorizedRequest, res: Response, next: NextFunction) => {
+		const force = checkTruthy(req.body.force);
+		try {
+			await actions.executeDeviceAction({ action }, force);
+			// QUESTION: this used to return { Data: 'OK', Error: null }, okay to change to just 'OK'?
+			return res.status(202).send('OK');
```
cywang117, line 94: 👀

src/device-api/v1.ts
```diff
@@ -184,25 +182,13 @@ router.post('/v1/purge', (req: AuthorizedRequest, res, next) => {
 		.catch(next);
 });
 
-router.post('/v1/update', (req, res, next) => {
-	eventTracker.track('Update notification');
-	if (isReadyForUpdates()) {
```
cywang117, line 225: Removed the `isReadyForUpdates` check, any downsides of this?

src/device-api/actions.ts
```diff
@@ -386,3 +388,36 @@ export const updateTarget = async (force: boolean = false) => {
...
+ * Used by:
+ * - GET /v1/apps/:appId
+ */
+export const getSingleContainerApp = async (appId: number) => {
+	eventTracker.track('GET app (v1)', { appId });
+	const apps = await applicationManager.getCurrentApps();
+	const app = apps[appId];
+	const service = app?.services?.[0];
+	if (service == null) {
+		throw new NotFoundError('App not found');
```
cywang117, line 404: This used to return `400` but is now returning `404` which is more correct, but is an interface change.

src/device-api/v1.ts
```diff
@@ -218,25 +211,10 @@ router.patch('/v1/device/host-config', async (req, res) => {
...
-		);
-		res.status(200).send('OK');
-	} catch (err: any) {
-		// TODO: We should be able to throw err if it's UpdatesLockedError
-		// and the error middleware will handle it, but this doesn't work in
-		// the test environment. Fix this when fixing API tests.
-		if (err instanceof UpdatesLockedError) {
-			return res.status(423).send(err?.message ?? err);
-		}
-		res.status(503).send(err?.message ?? err ?? 'Unknown error');
```
cywang117, line 316: The error handling middleware handles these errors now, so the 503 message returned in this case will be a JSON message `{ status: 'failed', message: err?.message ?? err ?? 'Unknown error' }` instead of what's sent here.

test/integration/device-api/v1.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+		api = new deviceApi.SupervisorAPI({
+			routers: [v1.router],
+			healthchecks: [],
+			// @ts-expect-error
+		}).api;
+	});
+
+	describe('GET /v1/healthy', () => {
+		// Actions are tested elsewhere so we can stub the dependency here
+		beforeEach(() => stub(actions, 'runHealthchecks'));
```
pipex, line 40: Instead of stubbing `actions.runHealthchecks` which is an implementation detail from the point of view of the API user, you could pass stubs when initializing the API

```
api = new deviceApi.SupervisorAPI({
    routers: [v1.router],
    healthchecks: [stub().resolve(true), stub.resolve(false)],
	// @ts-expect-error
    }).api
    
await request(api).get('/v1/healthy').expect(200);
```

test/integration/device-api/v1.spec.ts
```diff
@@ -38,4 +38,17 @@ describe('device-api/v1', () => {
 			await request(api).get('/v1/healthy').expect(500);
 		});
 	});
+
+	describe('POST /v1/blink', () => {
+		// Actions are tested elsewhere so we can stub the dependency here
+		before(() => stub(actions, 'identify'));
+		after(() => (actions.identify as SinonStub).restore());
```
pipex, line 57: Again the fact that the API calls `identify` under the hood is an implementation detail, which tests should not rely on. The right way would probably be to do it passing an identification function to the API object constructor or something like that. For now this is OK since this is a major improvement on what we had before and I don't want to derail the PR. 

Just making sure to leave the comment and this will apply for similar stubs, even if I don't add a comment.

src/device-api/actions.ts
```diff
@@ -36,3 +38,29 @@ export const identify = (ms: number = DEFAULT_IDENTIFY_DURATION) => {
...
+
+		if (shouldReportUpdatedKey) {
+			deviceState.reportCurrentState({
+				api_secret: newKey,
+			});
+		}
+
+		return newKey;
+	} catch (e) {
+		log.error('Error regenerating API key', e);
```
pipex, line 90: This interface is a bit weird, if an error occurred we  should just throw right? Why do we want to return the old key? This could be misleading, making the user of the method to think that the key changed when in fact it did not.

src/device-api/index.ts
pipex: Not sure about returning data with a 500, server error just means server error right?

test/integration/device-api/v1.spec.ts
```diff
@@ -85,4 +89,87 @@ describe('device-api/v1', () => {
...
+		beforeEach(() => {
+			doRestartStub = stub(actions, 'doRestart').resolves();
+		});
+		afterEach(async () => {
+			doRestartStub.restore();
+			// Remove all scoped API keys between tests
+			await db.models('apiSecret').whereNot({ appId: 0 }).del();
+		});
+
+		it('validates data from request body', async () => {
```
pipex, line 112: Why test these methods here again? Why not test them with the full test for managing the app lifecycle? I understand maybe it is not practical for all conditions (like simulating locks or edge error cases), but I would remove any redundant code as it is just a maintenance burden.

test/integration/device-api/v2.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+		// passing it directly to supertest is easier than
+		// setting up an API listen port & timeout
+		api = new deviceApi.SupervisorAPI({
+			routers: [v2.router],
+			healthchecks: [],
+			// @ts-expect-error
+		}).api;
+	});
+
+	describe('POST /v2/applications/:appId/restart', () => {
```
pipex, line 35: Same with tests here, anything is already tested within the lifecycle tests should be removed here. I know right now you are only testing v1 tests within lifecycle, but it may be useful when you are doing the v2 tests

src/device-api/v1.ts
pipex: I'm not sure I'm understanding the question correctly. If the v1 interface returns the container id, then we need to keep it.

src/device-api/v1.ts
```diff
@@ -46,84 +49,50 @@ router.post('/v1/restart', (req: AuthorizedRequest, res, next) => {
...
-
-			// Get the service from the target state (as we do in v2)
-			// TODO: what if we want to start a service belonging to the current app?
-			const targetService = _.find(targetApp.services, {
-				serviceName: service.serviceName,
-			});
+// QUESTION: This function is quite similar to v2's handleServiceAction
+// with the exception of some response messages being different and the presence
+// of the event tracker. Should we remove these differences and just use one function
+// between v1 & v2?
```
pipex, line 46: If there is a way to reuse some of that code and just change the inputs/outputs to match the API specification, then I think that's a worthy improvement. Regarding the event tracker events, those are not really part of the public interface so we may add/remove them as necessary.

src/device-api/v2.ts
```diff
@@ -22,97 +18,82 @@ import * as images from '../compose/images';
...
+			checkString(req.body.serviceName),
+			checkTruthy(req.body.force),
+		];
 
-	return Promise.all([applicationManager.getCurrentApps(), getApp(appId)])
-		.then(([apps, targetApp]) => {
-			const app = apps[appId];
+		if (!appId) {
+			return res.status(400).json({
+				status: 'failed',
```
pipex, line 48: Maybe this needs to be a 404 as well?

src/device-api/v2.ts
```diff
@@ -22,97 +18,82 @@ import * as images from '../compose/images';
...
+		}
 
-			if (app == null) {
-				res.status(404).send(appNotFoundMessage);
-				return;
-			}
+		if (!req.auth.isScoped({ apps: [appId] })) {
+			return res.status(401).json({
+				status: 'failed',
+				message: 'Application is not available',
```
pipex, line 56: I know these messages come from before us, but maybe `Unauthorized` is the right message here?

src/device-api/v2.ts
pipex: Nevermind, I saw this is meant to be for invalid input from the tests

src/device-api/v1.ts
pipex: Answering the question. I'd argue it's not OK to change the interface. I would think it's only OK for error responses, since errors are generally not expected to contain any relevant data. For 2xx responses, we need to maintain the interface until we can go to a major version

test/integration/device-api/actions.spec.ts
```diff
@@ -608,3 +610,26 @@ describe('manages application lifecycle', () => {
 		});
 	});
 });
+
+describe('reboots or shuts down device', () => {
+	// Stub external dependencies
+	before(async () => {
+		stub(dbus, 'reboot').resolves();
```
pipex, line 672: Could you query the fake dbus API instead of stubbing the method here?

src/device-api/v1.ts
pipex: It seems like the target state module doesn't really depend on API binder so I think there should not be any issues. Still might be worth running this PR on a device and test it for a bit on real conditions.

src/device-api/actions.ts
pipex: I think this is probably OK.

src/device-api/v1.ts
pipex: I think that's a good improvement

pipex: Actually, on second though, maybe this would be considered a breaking change, since someone may have a device that is depending on the response format. The break depends on the user updating the supervisor, but still is probably not great if they update their supervisor on an old device and suddenly their app is malfunctioning.

Maybe let's open a discussion Zulip about what would be the considered breaking changes on an API like the supervisor's. What sort of changes depend on an API version change and what depend on a Supervisor version change

My guesses

- I think probably a change in an error response is probably OK (e.g. 400 ->  404)
- Maybe a change in an error message is fine (e.g. 'Unknown error' -> 'An exception occurred when trying to ...')
- A format change in error response is probably not OK, but maybe we can get away with a Supervisor version bump
- A non additive change in a 2xx response is a breaking change and should not be done outside of an API version bump

test/integration/compose/application-manager.spec.ts
```diff
@@ -1,166 +1,21 @@
...
-	);
-
-	return {
-		currentApps,
-		availableImages: images,
-		downloading,
-		containerIdsByAppId,
-	};
-}
+import {
```
pipex, line 12: Nice work refactoring these files and removing a bunch of unnecessary test files

pipex:
Great work @cywang117 !! I have minor observations in general. The only point I'm unsure about is the change in some of the [error responses](https://github.com/balena-os/balena-supervisor/pull/2048#discussion_r1056677229). I think there are some we can get away with but on others we may not want to risk breaking user devices silently.

src/device-api/index.ts
cywang117: True, it's not common behavior for a 500. The user should also already have the old key when requesting this endpoint, so I can remove this.

cywang117:
Responding to review comments & changing code in progress, will be pushing partial changes until completed. I'll request a re-review when complete.

src/device-api/index.ts
cywang117: Updated in 8410d18ce21014624c9cc14c1a64425c90f5f748

src/device-api/actions.ts
cywang117: Done in 87d2add1a2be52676ad4c8eecb641ac04ad25d3d

src/device-api/v1.ts
cywang117: I've added back the legacy interface in 524685d1bffc7edcf8db8cf5417f91bc08655f75, so that `POST /v1/apps/:appId/stop` and `POST /v1/apps/:appId/start` return `{ containerId: <CONTAINER_ID> }` as they did before.

src/device-api/actions.ts
cywang117: Reverted to original interface in 4a5c2f352f6bf25763531089f8cf023a6c2394c4, following discussion to maintain all legacy interfaces except string contents

src/device-api/v1.ts
cywang117: Reverted in 3b20224d149cfb0ae19407bd422dcf8c86f45fb3

src/device-api/v1.ts
cywang117: Reverted to original interface in 98f4ded24f1c61816f8f01ffce318c1fb0aa8c8a. This means that for endpoints that stop/start a service, v1 endpoints will throw `BadRequestError` (400) for the same code paths that v2+ endpoints throw `NotFoundError` (404). This maintains the legacy interface.

src/device-api/v1.ts
cywang117: Reverted to original interface in https://github.com/balena-os/balena-supervisor/commit/98f4ded24f1c61816f8f01ffce318c1fb0aa8c8a. This means that for endpoints that stop/start a service, v1 endpoints will throw BadRequestError (400) for the same code paths that v2+ endpoints throw NotFoundError (404). This maintains the legacy interface.

src/device-api/v1.ts
cywang117: We discussed this in a Supervisor checkin and agreed that all interfaces should stay the same, with the exception of string contents. So, error codes remain as-is even if incorrect, JSON responses retain their keys (although new keys can be added), although messages, whether as text responses or JSON values, can change. I'll revert this one case to the interface it was before to keep it consistent.

src/device-api/actions.ts
cywang117: I've changed the `regenerateKey` action to pass along the underlying `apiKey.refreshKey` error if one gets thrown, and the route `regenerate-api-key` to pass any errors to be handled by the error middleware.

test/integration/device-api/v1.spec.ts
cywang117: Done in <TODO>

test/integration/device-api/v1.spec.ts
cywang117: This is a step towards my goal, where I would like to add a PR to add an input validation middleware, after which all input validation tests can be deduplicated and tested in that middleware. In the meantime, I can remove it all if the degree of duplication is not satisfactory?

test/integration/device-api/v2.spec.ts
cywang117: Are the input validation tests the ones you're referring to? I don't test any input validation in the lifecycle tests (actions.spec.ts). As in https://github.com/balena-os/balena-supervisor/pull/2048#discussion_r1065223879, should I remove this for now since I plan to deduplicate with an input validation middleware feature after the API test migration?

src/device-api/v1.ts
cywang117: Ended up not doing this -- while the logic was the same, there are too many differences in responses sent -- v2 service actions send JSON responses, and has a code path not in v1 service actions that handles when some fields aren't present in the request body. I don't see a way of doing it that isn't overly complex.

src/device-api/v2.ts
cywang117: Good call, changing just the message doesn't break the interface. I changed all instances of JSON 401's that were returning other messages to `Unauthorized`. There were 3 instances of this in v1.ts and 4 instances in v2.ts. See https://github.com/balena-os/balena-supervisor/pull/2048/commits/c6cf6a013610a2ec8eba8149ca17f0375324bc33

test/integration/device-api/actions.spec.ts
cywang117: This is what I wanted to do originally, but the main unintended side effect is that [there is a listener event that stops the API on 'shutdown'](https://github.com/balena-os/balena-supervisor/blob/fc15ad2554a21b9e86321fc183aab09ba149e15a/src/supervisor.ts#L75). This stops the API of the `balena_supervisor` container in the integration test environment, so it would become necessary to add logic to connect to the Docker socket that manages `balena_supervisor` instead of the `dind` socket, and the spec would need to issue a restart of the `balena_supervisor` service. 

Perhaps instead of the 'shutdown' listener stopping the API, we could use something like https://www.npmjs.com/package/exit-hook to process async exit hooks? This will avoid stopping the API in the test env.

src/device-api/v1.ts
cywang117: I will check this on a device tomorrow!

test/integration/device-api/v1.spec.ts
pipex: You can leave the validation tests since you have a plan to remove them. The duplication is fairly minor as well as the lifecycle tests do not test all input scenarios.

test/integration/device-api/v2.spec.ts
pipex: As I said above, I agree you are not testing everything on the lifecycle tests so duplication right now is relatively minimal. Since you have a plan to improve I think you can keep it this way

test/integration/device-api/actions.spec.ts
pipex: Hmm, and who sends the `shutdown` event to the device state, is it a system event? If that's the case the system event should not be triggered by the fake dbus API.

cywang117: It's a system event, and I was wrong about my previous comment. I must have been mistaken from when I was implementing the reboot/shutdown action tests in a different way.

I changed the stubs for reboot/shutdown to spies, since there's no API for querying the test dbus. See 85392f2a858ef9956f211c98c38db1e69b4d1d2d

src/device-api/v1.ts
cywang117: I tested this by pinning to another release while simultaneously restarting the draft Supervisor. I haven't been able to see any issues, but there could still be a tiny window when `isReadyForUpdates` returns false and an update notification is received. Looking at the code, the window is mostly comprised of provisioning logic, so I also tested on a fresh device with the draft Supervisor preloaded, to see if changing the pinned release back & forth while the device still provisioning had any effect. I also didn't find any issues with device behavior, so I think this removal is fine.

test/integration/device-api/actions.spec.ts
pipex: I was actually thinking you could query the status of the service via the DBUS API. I'm not going to keep pushing on this point as I don't want to delay this PR any longer and this looks great anyway

docker-compose.test.yml
```diff
@@ -10,6 +10,7 @@ services:
       args:
         ARCH: ${ARCH:-amd64}
     command: [ '/wait-for-it.sh', '--', '/usr/src/app/entry.sh' ]
+    stop_grace_period: 3s
```
pipex, line 13: what is this for?

cywang117: @pipex, this shortens the period between SIGTERM & SIGKILL to speed up container shutdown while developing

cywang117:
@balena-ci I self-certify!

PR #2049, opened 2022-10-28
Decrease metrics frequency reporting

kb2ma:
# Description

Presently metrics are reported as frequently as every 10 seconds via target-state patch. This PR lowers that rate to at most every 5 minutes. Adds a new constant, `maxMetricsFrequency`. Schedules reports at that frequency, and adds a boolean parameter to `device-state.getCurrentForReport()` to exclude metrics.

With this addition, constant `maxReportFrequency` remains as a kind of backstop to avoid flooding target-state patches.

Added the comment below to `api-binder/report.startReporting()`:

```
	// Managing report frequency: We can't control the frequency of state change
	// events or the arrival time of those events relative to scheduled metrics
	// reports. So, we employ a two layer strategy to manage frequency:
	// 1) Ensure that below we send reports from the same source (event/scheduled)
	//    no sooner than the max frequency of that type.
	// 2) Rely on reportCurrentState() also to validate these frequencies because
	//    it is the single handler of requests from both sources.
```

Fixes # (issue) ???

# Type of change

patch

# How Has This Been Tested?

No testing yet; prototype


src/api-binder/report.ts
```diff
@@ -141,16 +147,40 @@ export async function startReporting() {
...
+		// If report already in progress, it's possible the state change that
+		// triggered this event might be missed, so try once more as best effort.
+		// No worries about duplicating report content, as we expect state content
+		// is diffed to validate a change has occurred. Also, metrics will be
+		// excluded from the follow-up report because maxMetricsFrequency should
+		// be a much larger interval than maxReportFrequency.
+		if (! await doReport()) {
+			await delay(constants.maxReportFrequency);
+			await doReport();
+		}
```
cywang117, line 179: I'm not sure once with best effort is enough, given that a change event is emitted for numerous events in quick succession during an update. Practically, how often does current state actually get reported on a device with this PR code?

src/api-binder/report.ts
```diff
@@ -166,5 +196,5 @@ export async function startReporting() {
 
 	// Start monitoring for changes that do not trigger deviceState events
 	// Example - device metrics
-	return recursivelyReport(constants.maxReportFrequency);
+	return recursivelyReport(constants.maxMetricsFrequency);
```
cywang117, line 199: We discussed "resetting the metrics report timer when a device-state change is reported" -- is that still the case? It looks like the recursive report will fire every maxMetricsFrequency (5min) regardless of whether a deviceState 'change' event has occurred / been reported.

src/api-binder/report.ts
cywang117: It's still better than before where we tried exactly once, and dropped the report attempt entirely if a report was in progress

kb2ma: I have revised the retry mechanism to ensure the most recent device state is reported.

src/api-binder/report.ts
kb2ma: Thanks, IMO the device has a contract with the cloud that metrics will be reported at a defined maximum frequency, and the cloud enforces this contract. However, I will revise the code to make it easy to change this decision.

kb2ma:
Reverted commit to *always* exclude metrics if previous report was less than 5 minutes. Now includes metrics if a state change occurs even if less than 5 minutes. This behavior matches the cloud API. See [this line](https://github.com/balena-io/open-balena-api/blob/691f68869155c8989504e8e5ac45cf0840d1c81e/src/features/device-state/routes/state-patch-v3.ts#L320) in open-balena-api.

src/api-binder/report.ts
```diff
@@ -141,21 +145,45 @@ export async function startReporting() {
...
-	deviceState.on('change', doReport);
+	deviceState.on('change', onStateChange);
 
 	async function recursivelyReport(delayBy: number) {
 		try {
-			// Try to send current state
-			await doReport();
+			// Follow-up when report not sent immediately on change event...
+			if (stateChangeDeferred) {
+				stateChangeDeferred = ! await doReport();
```
cywang117, line 179:
```suggestion
				stateChangeDeferred = !(await doReport());
```
I would format both instances of this as the above to be inline with the rest of the codebase.

cywang117 (commented):
@kb2ma This looks good! I didn't catch anything, perhaps @pipex will. I'd be curious about actual report frequency on a device while idle and during an update.

src/api-binder/report.ts
```diff
@@ -141,21 +144,45 @@ export async function startReporting() {
 	])) as StateReportOpts;
 
 	let reportPending = false;
-	const doReport = async () => {
+	// Reports current state if not already sending and does not exceed report
+	// frequency. Returns true if sent; otherwise false.
+	const doReport = async (): boolean => {
```
pipex, line 149:
```suggestion
	const doReport = async (): Promise<boolean> => {
```

src/api-binder/report.ts
```diff
@@ -141,21 +144,45 @@ export async function startReporting() {
...
 		if (!reportPending) {
-			reportPending = true;
-			await reportCurrentState(reportConfigs);
-			reportPending = false;
+			const sinceLastReport = performance.now() - lastReportTime;
```
pipex, line 151: Since `lastReportTime` is only used inside the `startReporting` function, I would declare it inside the function instead of as a global

src/api-binder/report.ts
```diff
@@ -141,21 +144,45 @@ export async function startReporting() {
...
 		try {
-			// Try to send current state
-			await doReport();
+			// Follow-up when report not sent immediately on change event...
+			if (stateChangeDeferred) {
+				stateChangeDeferred = !(await doReport());
+			// ... or on regular metrics schedule.
+			} else {
+				const sinceLastReport = performance.now() - lastReportTime;
+				if (sinceLastReport > constants.maxMetricsFrequency) {
```
pipex, line 182: I think we can do this comparison in a single line since the `sinceLastReport` value is not used anywhere else

```suggestion
				if (performance.now() - lastReportTime > constants.maxMetricsFrequency) {
```

src/lib/constants.ts
```diff
@@ -68,6 +68,9 @@ const constants = {
 	supervisorNetworkGateway: '10.114.104.1',
 	// How often can we report our state to the server in ms
 	maxReportFrequency: 10 * 1000,
+	// How often can we report metrics to the server in ms; mirrors server setting.
+	// Metrics are low priority, so less frequent than maxReportFrequency.
+	maxMetricsFrequency: 300 * 1000,
```
pipex, line 73: I would move this definition to the top of `report.ts`. If we ever need the value somewhere else we can move it. I think constants.ts is bloated enough

src/api-binder/report.ts
kb2ma: Funny what you learn when you compile things. ;-)

src/api-binder/report.ts
kb2ma: `reportCurrentState()` handles capturing and sending currentState, and this notion of when/what we actually capture is really import for the integrity of the report. So I would keep the logic in that function with definition of `lastReport`. `doReport()` is more of a user of `reportCurrentState()`.

src/api-binder/report.ts
kb2ma: Sounds good, and will make same change on l.151.

src/lib/constants.ts
kb2ma: Move `maxReportFrequency` as well? It also is used only in `report.ts`.

pipex: Yep, sounds good to me.

src/api-binder/report.ts
pipex: Ok, sounds good

src/api-binder/report.ts
```diff
@@ -20,6 +19,13 @@ import { shallowDiff, prune, empty } from '../lib/json';
 
 let lastReport: DeviceState = {};
 let lastReportTime: number = -Infinity;
+// Tracks if unable to report the latest state change event.
+let stateChangeDeferred: boolean = false;
+// How often can we report our state to the server in ms
+const maxReportFrequency = 10 * 1000;
```
PranavPeshwe, line 25: Going by the comments, I think the vars should be named min intervals and not max frequencies. Readers may misconstrue the semantics.
Happened to bump into this PR while looking at a support ticket. Hope I didn't present as unnecessarily pedantic.

kb2ma:
Just added a couple of commits as a result of testing. I would appreciate feedback especially on 2890140. In that case when a change arrives before it's OK to send metrics only (maxMetricsFrequency), we must verify that any changes from `deviceState.getCurrentFromReport()` include at least one change to device state, excluding metrics.

I chose the approach of removing metrics from `deviceState.getCurrentFromReport()`, exporting and separately calling `deviceState.getSysInfo()`, and appending the results to the report. The other approach would be to parse the results of `getCurrentForReport()` for presence of a non-metrics state attribute. IMO `apiBinder.reportCurrentForState()` generally needs more control over metrics reporting, and separation of metrics collection from state collection provides that fairly cleanly. It's in the direction we are taking on metrics collection overall.

src/api-binder/report.ts
kb2ma: Thanks for your sense on frequency vs. interval. I agree that interval describes the case better, but I left it at frequency to stay consistent with existing use and minimize changes.

kb2ma:
No significant changes in commits since last comment.

kb2ma:
One consequence of 280f46d is that the order of the JSON contents in the state patch may change. The `apps` dictionary, if present, now will appear before metrics entries. I don't think anyone would depend on the order of the attributes of a JSON object, but just wanted to raise the point.

kb2ma:
@pipex , would you please review the updated commit 806e8d0 for the types I used. Besides that I think this PR is ready for merge.

kb2ma:
I am able to build locally and test armv7 on my Fin with no issues.

kb2ma:
Rebased on master

src/api-binder/report.ts
```diff
@@ -70,30 +76,44 @@ async function report({ body, opts }: StateReport) {
...
+		// If metrics not yet scheduled, report must include a state change to
+		// qualify for sending.
+		const metricsScheduled =
+			performance.now() - lastReportTime > maxMetricsFrequency;
+		if (!metricsScheduled) {
+			const uuid = await config.get('uuid');
+			if (!uuid) {
+				throw new InternalInconsistencyError('No uuid found for local device');
+			}
+			const uuidMap = stateDiff[uuid] as { string: any };
```
pipex, line 103: Is this type correct? Shouldn't it be 

```suggestion
			const uuidMap = stateDiff[uuid] as { [k: string]: any };
```

Disregard if I'm misunderstanding

kb2ma: TIL, thanks.

src/api-binder/report.ts
```diff
@@ -70,30 +76,44 @@ async function report({ body, opts }: StateReport) {
...
+ * Does *not* validate time elapsed since last report.
+ */
 async function reportCurrentState(opts: StateReportOpts) {
-	// Wrap the report with fetching of state so report always has the latest state diff
 	const getStateAndReport = async () => {
-		const now = performance.now();
-		// Only try to report if enough time has elapsed since last report
-		if (now - lastReportTime >= constants.maxReportFrequency) {
-			const currentState = await deviceState.getCurrentForReport(lastReport);
-			const stateDiff = prune(shallowDiff(lastReport, currentState, 2));
```
pipex, line 80: Note that this trades off device resources for API resources, since `getCurrentForReport` incurrs a non-negligible cost. I think this is the right tradeoff to make as the API is much harder to scale, but I just wanted to point it out.

src/api-binder/report.ts
```diff
@@ -70,30 +76,44 @@ async function report({ body, opts }: StateReport) {
...
+		}
+
+		// If metrics not yet scheduled, report must include a state change to
+		// qualify for sending.
+		const metricsScheduled =
+			performance.now() - lastReportTime > maxMetricsFrequency;
+		if (!metricsScheduled) {
+			const uuid = await config.get('uuid');
+			if (!uuid) {
+				throw new InternalInconsistencyError('No uuid found for local device');
```
pipex, line 101: Could we pass uuid as an argument to this function instead? Ideally I think it could be queried once in `startReporting` and fail earlier if it's going to fail, rather than query here every 10 seconds.

pipex:
Great work on the PR @kb2ma sorry for the last minute comments, I've been jumping on and off the review process. If you can fix those then I think this is ready to merge.

src/api-binder/report.ts
kb2ma: Will do. Makes sense.

kb2ma:
No worries on the changes; I appreciate the review. 

Magically the build passed this time. There definitely is some flakiness in the checks.

src/api-binder/report.ts
pipex: Sorry, you are right, the snippet is weird. And actually I'm wrong, I was thinking `getCurrentForReport` now would be called more often on account of the removal of the `if` clause before, but I had forgot that you moved that comparison further up the stack. Disregard this then

cywang117:
I suspect that flakiness is due to our arm builders struggling. I've also seen them fail more often than succeed with the last few PRs. Re-running failed checks usually solves it 👍🏼

PR #2051, opened 2022-11-02
Update apt repositories before installing dbus

cywang117:
GitHub actions are erroring due to the package not being found; Updating the repository listings shoud fix this.

Change-type: patch


cywang117:
@balena-ci I self-certify!

PR #2052, opened 2022-11-02
Run the built supervisor as part of docker-compose tests

pipex:
This allows to test that the supervisor build actually runs and opens up the possibility of running more exhaustive API tests against a working supervisor.

Change-type: patch


cywang117:
@balena-ci rebase

test/integration/supervisor.spec.ts
```diff
@@ -0,0 +1,10 @@
+import * as request from 'supertest';
+
+const BALENA_SUPERVISOR_ADDRESS =
+	process.env.BALENA_SUPERVISOR_ADDRESS || 'http://balena-supervisor:48484';
+
+describe('supervisor app', () => {
```
cywang117, line 6: Could we consider removing `test/legacy/18-startup.spec.ts` since we have this?

pipex: yep, good call

pipex: done

cywang117 (approved):

PR #2053, opened 2022-11-07
Allow local target state to be applied in unmanaged mode

cywang117:
Allow local target state to be applied in unmanaged mode

An unmanaged Supervisor will not have a balenaApi defined, so any attempts to POST /v2/local/target-state will cause the Supervisor to crash loop. This fix allows for the Supervisor to apply local target state, which is useful for integration testing.

This was introduced in Supervisors v13+.

Change-type: patch


pipex (approved):
LGTM!

Issue #590, opened 2018-03-15
Two networks with the same name can make the supervisor unable to apply the target state (network is ambiguous)


pcarranzav:
Seen once, in a device that was updated from version 7.1.11 to 7.1.14. I'm not sure if anything particular about this device or its application might have triggered this particular situation.

Basically, there were two networks with the same name (`<appId>_default`) and this made balena complain with a `server error - 500` because a command referencing the  network by its name was ambiguous. Both networks had the `io.resin.supervised` label. I still don't see how the supervisor would create an additional network with the same name if the network already exists - unless there's some weird race condition or something like that. Needs investigation, so if this happens again please record supervisor logs and what precise actions got the device in this state.

Workaround was to delete one of the conflicting networks with `balena network rm <network ID>`.

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_cznn)

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_gtgj)

pdcastro:
```
$ journalctl -u resin-supervisor
# continuously prints the following every 30s
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at /usr/src/app/dist/app.js:573:91245
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at /usr/src/app/dist/app.js:573:91208
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at m.buildPayload (/usr/src/app/dist/app.js:573:91218)
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at IncomingMessage.<anonymous> (/usr/src/app/dist/app.js:573:90718)
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at emitNone (events.js:91:20)
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at IncomingMessage.emit (events.js:185:7)
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at endReadableNT (_stream_readable.js:974:12)
Jan 23 00:07:04 ebe09d3 resin-supervisor[930]:     at _combinedTickCallback (internal/process/next_tick.js:80:11)
[...]

PR #2054, opened 2022-11-09
Convert ensureSupervisorNetwork to native Promises, remove sys interface check

cywang117:
Convert `ensureSupervisorNetwork` to native Promises
    
Also remove system interface check from `ensureSupervisorNetwork`

Previously `ensure` was a Bluebird promise which wasn't awaited in
its composition step. This has been here for some time and may contribute
to issues with duplicate networks. The conversion to native Promises
allows `ensure` to be awaited, hopefully reducing instances of duplicate
networks.
    
Removing the system interface check for `/sys/class/net/supervisor0`
because it's superfluous given that the Engine creates the interface
with NetworkManager. It also makes testing a lot more difficult to set up
as /sys/class/net isn't a directory that can be written to for emulating
system interface creation / removal.

Additionally, improve an error method's interface (`NotFoundError`) and 
set the default `macAddressPath` from `/mnt/root/sys/class/net` to `/sys/class/net`.
As SV uses host networking, the two are identical.
    
Relates-to: https://github.com/balena-os/balena-supervisor/issues/1110
Change-type: minor

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "test:env": "ARCH=$(./build-utils/detect-arch.sh) docker-compose -f docker-compose.test.yml -f docker-compose.dev.yml up --build; npm run compose:down",
     "test:compose": "ARCH=$(./build-utils/detect-arch.sh) docker-compose -f docker-compose.yml -f docker-compose.test.yml up --build --remove-orphans --exit-code-from=sut ; npm run compose:down",
     "test": "npm run lint && npm run test:build && npm run test:unit && npm run test:legacy",
-    "compose:down": "docker-compose -f docker-compose.test.yml down",
+    "compose:down": "docker-compose -f docker-compose.test.yml down && docker volume rm $(docker volume ls -f name=balena-supervisor -q)",
     "prettify": "balena-lint -e ts -e js --fix src/ test/ typings/ build-utils/ webpack.config.js",
     "release": "tsc --project tsconfig.release.json && mv build/src/* build",
     "sync": "ts-node --files sync/sync.ts",
--- a/src/compose/composition-steps.ts
+++ b/src/compose/composition-steps.ts
@@ -276,7 +276,7 @@ export function getExecutors(app: {
 			await volumeManager.remove(step.current);
 		},
 		ensureSupervisorNetwork: async () => {
-			networkManager.ensureSupervisorNetwork();
+			await networkManager.ensureSupervisorNetwork();
 		},
 		noop: async () => {
 			/* async noop */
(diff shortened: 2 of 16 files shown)


src/compose/network-manager.ts
```diff
@@ -65,82 +64,66 @@ export async function remove(network: Network) {
...
+				Config: [
+					{
+						Subnet: subnet,
+						Gateway: gateway,
 					},
-				}),
-			);
+				],
+			},
+			CheckDuplicate: true,
```
cywang117, line 124: docker-compose enables this [by default](https://github.com/docker/compose/pull/9585), we should too, in case this prevents cases of duplicate networks as with https://github.com/balena-os/balena-supervisor/issues/1110 and https://github.com/balena-os/balena-supervisor/issues/590. 

In fact, we should probably make this the default for network creation everywhere in the Supervisor.

package.json
```diff
@@ -20,7 +20,7 @@
     "test:env": "ARCH=$(./build-utils/detect-arch.sh) docker-compose -f docker-compose.test.yml -f docker-compose.dev.yml up --build; npm run compose:down",
     "test:compose": "ARCH=$(./build-utils/detect-arch.sh) docker-compose -f docker-compose.yml -f docker-compose.test.yml up --build --remove-orphans --exit-code-from=sut ; npm run compose:down",
     "test": "npm run lint && npm run test:build && npm run test:unit && npm run test:legacy",
-    "compose:down": "docker-compose -f docker-compose.test.yml down",
+    "compose:down": "docker-compose -f docker-compose.test.yml down && docker volume rm $(docker volume ls -f name=balena-supervisor -q)",
```
pipex, line 23: Good one

src/compose/composition-steps.ts
```diff
@@ -276,7 +276,7 @@ export function getExecutors(app: {
 			await volumeManager.remove(step.current);
 		},
 		ensureSupervisorNetwork: async () => {
-			networkManager.ensureSupervisorNetwork();
+			await networkManager.ensureSupervisorNetwork();
```
pipex, line 279: I think return is enough here, but `await` works too

src/compose/network-manager.ts
pipex: Oh, TIL! Yeah, we should add it to the `Network.create` defaults

test/integration/compose/application-manager.spec.ts
```diff
@@ -853,8 +861,8 @@ describe('compose/application-manager', () => {
 	});
 
 	it('should infer that we need to create the supervisor network if it does not exist', async () => {
-		// stub the networkManager method to fail on finding the supervisor network
-		(networkManager.supervisorNetworkReady as sinon.SinonStub).resolves(false);
+		const docker = new Docker();
+		await docker.getNetwork('supervisor0').remove();
```
pipex, line 865: 🙌

pipex:
No observations @cywang117, great work! Merge when ready :)

cywang117:
@balena-ci I self-certify!

PR #2055, opened 2022-11-11
Set up dummy Supervisor image for integration tests

cywang117:
In integration test env, the Supervisor container expects an image tagged balena/$ARCH-supervisor to be able to apply target state. We can't access the SV container for stubbing the method that checks the presence of this image, so a minimal image should be pulled and tagged.

Change-type: patch


cywang117:
Closing as the setup relies on implementation details of checking the existence of the Supervisor image.

PR #2056, opened 2022-11-11
Use local socket for tests instead of dind

pipex:
While this breaks a bit the isolation of the tests, is a bit more realistic in the sense that the supervisor is running in the same engine as the rest of the containers.

Change-type: patch


pipex:
This doesn't really work, since it breaks isolation tests fail because of other docker components they weren't expecting. Ideally the supervisor itself would run inside dind so it can run in an isolated environment, it will probably be mandatory once the supervisor can update itself, because that will be the only real way to test that operation.

PR #2057, opened 2022-11-11
Enable esModuleInterop to ease the transition to ES modules

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


pipex:
No idea why the tests are failing here, I could not reproduce the following locally

```
 balena-supervisor-sut-1  |   1) lib/firewall
balena-supervisor-sut-1  |        Basic On/Off operation
balena-supervisor-sut-1  |          should confirm the `changed` event is handled:
balena-supervisor-sut-1  |      TypeError: Descriptor for property spawn is non-configurable and non-writable
balena-supervisor-sut-1  |       at assertValidPropertyDescriptor (node_modules/sinon/lib/sinon/stub.js:158:15)
balena-supervisor-sut-1  |       at Function.stub (node_modules/sinon/lib/sinon/stub.js:85:5)
balena-supervisor-sut-1  |       at stub (node_modules/sinon/lib/sinon/sandbox.js:388:37)
```

Page-:
@pipex the issue is that `esModuleInterop` enforces that a module's exports cannot be modified (since in real ESM they cannot be) and `stub(childProcess, 'spawn').callsFake(...)` is attempting to modify the `child_process` exports

cywang117:
Does that explain why the test can't be produced locally though @Page- ?

PR #2060, opened 2022-11-11
Use timers/promises for promisified setTimeout

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -22,6 +22,7 @@ import type { Service } from './service';
 import { strict as assert } from 'assert';
 
 import log from '../lib/supervisor-console';
+import { setTimeout } from 'timers/promises';
 
 interface FetchProgressEvent {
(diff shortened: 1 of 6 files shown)


Page-:
I expect the issues this PR is seeing are due to the sinon fake timers 😞

pipex:
> I expect the issues this PR is seeing are due to the sinon fake timers 😞

Yeah, I'm pretty sure that's the issue. If you keep the promisified `setTimeout` in backoff you should be able to make the tests pass.

cywang117:
I reverted backoff's `setTimeout` to the `util.promisify`'d version; we'll have to revisit SinonFakeTimers in tests to move  that particular instance over to Node 16 promisifed `setTimeout`.

cywang117 (approved):

PR #2061, opened 2022-11-11
Remove some unnecessary usage of `.reduce`

Page-:

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -1081,37 +1081,41 @@ export async function getState() {
 			.concat(stateFromServices);
 
 	// Get the list of commits for all appIds from the database
-	const commitsForApp = (
-		await Promise.all(
-			// Deduplicate appIds first
-			[...new Set(servicesToReport.map((svc) => svc.appId))].map(
-				async (appId) => ({
-					[appId]: await commitStore.getCommitForApp(appId),
-				}),
-			),
-		)
-	).reduce((commits, c) => ({ ...commits, ...c }), {});
+	const commitsForApp: Dictionary<string | undefined> = {};
+	// Deduplicate appIds first
+	await Promise.all(
+		[...new Set(servicesToReport.map((svc) => svc.appId))].map(
+			async (appId) => {
+				commitsForApp[appId] = await commitStore.getCommitForApp(appId);
+			},
+		),
+	);
 
 	// Assemble the state of apps
-	return servicesToReport.reduce(
-		(apps, { appId, appUuid, commit, serviceName, createdAt, ...svc }) => ({
-			...apps,
-			[appUuid]: {
-				...(apps[appUuid] ?? {}),
-				// Add the release_uuid if the commit has been stored in the database
-				...(commitsForApp[appId] && { release_uuid: commitsForApp[appId] }),
-				releases: {
-					...(apps[appUuid]?.releases ?? {}),
-					[commit]: {
-						...(apps[appUuid]?.releases[commit] ?? {}),
-						services: {
-							...(apps[appUuid]?.releases[commit]?.services ?? {}),
-							[serviceName]: svc,
-						},
+	const state: { [appUuid: string]: AppState } = {};
+	for (const {
+		appId,
+		appUuid,
+		commit,
+		serviceName,
+		createdAt,
+		...svc
+	} of servicesToReport) {
+		state[appUuid] = {
+			...state[appUuid],
+			// Add the release_uuid if the commit has been stored in the database
+			...(commitsForApp[appId] && { release_uuid: commitsForApp[appId] }),
+			releases: {
+				...state[appUuid]?.releases,
+				[commit]: {
+					...state[appUuid]?.releases[commit],
+					services: {
+						...state[appUuid]?.releases[commit]?.services,
+						[serviceName]: svc,
 					},
 				},
 			},
-		}),
-		{} as { [appUuid: string]: AppState },
-	);
+		};
+	}
+	return state;
 }
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -597,12 +597,15 @@ export async function inspectByName(imageName: string) {
 
 	// Run the queries in sequence, return the first one that matches or
 	// the error from the last query
-	return await [inspectByURI, inspectByReference, inspectByDigest].reduce(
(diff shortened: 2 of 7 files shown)


src/config/backends/odmdata.ts
```diff
@@ -127,16 +127,12 @@ export class Odmdata extends ConfigBackend {
 			);
 		}
 		// Find the configuration given the optionsBuffer
-		const configIndex = this.CONFIG_BYTES.reduce(
-			(currentIndex: number, _config: number, index: number) => {
-				if (
+		const configIndex = this.CONFIG_BYTES.findIndex(
```
Page-, line 130: The previous behavior was technically `findLastIndex` which requires nodejs 18 but since in practice the check can only succeed for one index `findIndex` is equivalent. Also the benefit to the `find*` calls is that they short-circuit which means that they'll stop iterating the moment a match is found vs `reduce` which will iterate them all (in this case there's only 3 elements, but as a general rule being both clearer and more performant is a win)

src/lib/system-info.ts
```diff
@@ -11,9 +11,9 @@ import { exec } from './fs-utils';
 
 export async function getCpuUsage(): Promise<number> {
 	const cpuData = await systeminformation.currentLoad();
-	const totalLoad = cpuData.cpus.reduce((load, cpuLoad) => {
-		return load + cpuLoad.load;
-	}, 0);
+	const totalLoad = _.sumBy(cpuData.cpus, ({ load }) => {
```
Page-, line 14: This one is debatable because it's using lodash but the lodash usage is pretty prevalent

src/compose/images.ts
```diff
@@ -598,12 +598,15 @@ export async function inspectByName(imageName: string) {
...
-	return await [inspectByURI, inspectByReference, inspectByDigest].reduce(
-		(promise, query) => promise.catch(() => query(imageName)),
-		Promise.reject(
-			'Promise sequence in inspectByName is broken. This is a bug.',
-		),
-	);
+	let err;
+	for (const query of [inspectByURI, inspectByReference, inspectByDigest]) {
+		try {
+			return await query(imageName);
```
Page-, line 604: This will allow for better async stack traces by using `await` and I personally find the overall reformulation clearer to understand

src/compose/application-manager.ts
```diff
@@ -1081,16 +1081,15 @@ export async function getState() {
...
-	const commitsForApp = (
-		await Promise.all(
-			// Deduplicate appIds first
-			[...new Set(servicesToReport.map((svc) => svc.appId))].map(
-				async (appId) => ({
-					[appId]: await commitStore.getCommitForApp(appId),
-				}),
-			),
-		)
-	).reduce((commits, c) => ({ ...commits, ...c }), {});
```
thgreasi, line 1093: Just to note that `.reduce((commits, c) => ({ ...commits, ...c }), {})` is `O(N^2)`

src/compose/application-manager.ts
```diff
@@ -1081,37 +1081,41 @@ export async function getState() {
...
+					services: {
+						...state[appUuid]?.releases[commit]?.services,
+						[serviceName]: svc,
 					},
 				},
 			},
-		}),
-		{} as { [appUuid: string]: AppState },
-	);
+		};
```
thgreasi, line 1118: How about:
```ts
		state[appUuid] ??= {releases: {}};
		if (commitsForApp[appId]) {
			// Add the release_uuid if the commit has been stored in the database
			state.release_uuid = commitsForApp[appId]
		}
		const {releases} = state[appUuid];
		releases[commit] ??= {services: {}};
		const services = releases[commit];
		services[serviceName]= svc;
```
?

thgreasi (commented):
LGTM

src/compose/application-manager.ts
pipex: Yeah, but the size of `getCommitForApp(appId)` and the number of apps is usually 1, so `O(N^2) = O(1)` @thgreasi

Page-: In this case yes, but as a general rule it's worth knowing and for things like services/networks/volumes it will be >1 even in the supervisor. Although it's definitely true that we come from the server side where these differences can be amplified much more easily due to the number of requests we see

src/compose/images.ts
```diff
@@ -598,12 +598,15 @@ export async function inspectByName(imageName: string) {
...
+		} catch ($err) {
+			err = $err;
+		}
+	}
+	throw err;
```
pipex, line 609: Won't this throw every time?

src/compose/application-manager.ts
pipex: I personally find the existing syntax much more readable than using specific assignments, but I agree that is subjective.

src/compose/images.ts
Page-: Nope, because `return await query(imageName);` will return and exit the function before throwing, so this will throw the final error but only if they all error (preserving previous behavior). That said this logic is generally kinda weird in both cases so I wonder if there's an even clearer way to write it, but I haven't thought of one yet

pipex: you're right, I was missing the return. I think is debatable that the last error is the one that needs to be returned. Maybe the throw error can be replaced by a 
```suggestion
	new StatusError(
	        404,
	        `Failed to find an image matching ${imageName}`,
        )
```
and just discard the other errors

src/lib/system-info.ts
cywang117: If using lodash would result in fewer computations, then the usage is fine by me!

src/compose/images.ts
Page-: Maybe, but my aim isn't to change the observed error (I have no idea what sort of knock on effect that might have)

src/compose/application-manager.ts
thgreasi: Given that destructuring inside a loop and an O(N^2) anti-pattern, I would try to avoid it everywhere to discourage ppl introducing such code later.
Even the updated version causes 5 unnecessary GCs per loop.
And on top of that we can add that TS checks are way stricter in the specific assignments, rather than in destructured objects.

Page-:
@balena-ci I self-certify!

Issue #2062, opened 2022-11-14
Create `$APPUUID_default` networks as config-only when not needed to not conflict with host addressing


cywang117:
A `$APPUUID_default` network is created with default networking for the user application, unless configured explicitly with:
```
networks:
  default:
    driver: bridge
    ipam:
      driver: default
      config:
        - subnet: ...
          gateway: ...
```
in the fleet's docker-compose. 

v13+ Supervisors move in the direction of the user-app model, so a `$SUPERVISOR-APP-UUID_default` network is created in the same way. However, because no `ipam` configs are specified, the address space of this network may collide with user networks.

Solution: as network_mode for the Supervisor app is `host`, the network can still be created for purposes of determining whether target state has been applied, but only as a config, with the `ConfigOnly` create option. The network will still be available in the Engine to query, but no networks will actually be set up in host. This change will apply for any app where all services have `network_mode: host` specified.

cywang117:
This isn't an issue that only applies to the Supervisor app, but to any user app where the Docker networking address space may conflict with the host. Even if a service has `network_mode: host` and there are no containers on any bridge networks, the Supervisor will create an `$APPUUID_default` network regardless to serve as a way to distinguish whether target state has been successfully applied to the device, afaik. 

We could pass `ConfigOnly: true` to the networks which are created for this purpose but not used so that they don't take addresses that may be in use.

PR #2063, opened 2022-11-14
Specify ipam config for Supervisor app in a rarer address space

cywang117:
See issue #2062 for a description of the issue this PR attempts to solve.

Closes: #2062
Change-type: patch


cywang117:
This doesn't have the intended effect of changing the `$APPUUID_default` network of each application to a different subnet, because even if the Supervisor application (which also applies its own target state through the state funnel) or any application has a `network_mode: host`, a `$APPUUID_default` network is created anyway with the default network configs as a way to differentiate target state from being applied / not applied.

This does present a problem if the default network configs include an address conflict with the host network.

PR #2064, opened 2022-11-15
Create default network as config-only when services have host networking

cywang117:
This eliminates chances of host-Docker address collision for apps such as the Supervisor where all services have host networking.

Context: [internal thread](https://balena.zulipchat.com/#narrow/stream/345889-loop.2Fbalena-os/topic/Supervisor.20v13.2B.20default.20Docker.20network)

Closes: #2062
Change-type: patch

### DIFF
--- a/test/unit/compose/app.spec.ts
+++ b/test/unit/compose/app.spec.ts
@@ -602,6 +602,78 @@ describe('compose/app', () => {
 			// The network should not be created again
 			expectNoStep('createNetwork', steps);
 		});
+
+		it('should create a config-only network if network_mode is host for all services', async () => {
+			const svcOne = await createService({
+				appId: 1,
+				serviceName: 'one',
+				composition: { network_mode: 'host' },
+			});
+			const svcTwo = await createService({
+				appId: 1,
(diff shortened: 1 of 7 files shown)


test/unit/compose/app.spec.ts
```diff
@@ -602,6 +602,65 @@ describe('compose/app', () => {
...
+			const steps = current.nextStepsForAppUpdate(defaultContext, target);
+
+			const [createNetworkStep] = expectSteps('createNetwork', steps);
+			expect(createNetworkStep)
+				.to.have.property('target')
+				.that.has.property('config')
+				.that.deep.includes({ configOnly: true });
+		});
+
+		it('should not create a config-only network if network_mode: host is not specified for any service', async () => {
```
pipex, line 636: Maybe you could add a third test here for the "empty release" case, to ensure that a default network is created with `configOnly: true`?

cywang117: Good shout! Will do

cywang117:
@balena-ci I self-certify!

PR #2065, opened 2022-11-16
Simplify getting images for cleanup

cywang117:
`getImagesForCleanup` used to query the Engine for the Supervisor image, which is unnecessary given that the Supervisor has access to `constants.supervisorImage`. Thus, this Engine query is removed.

The method is simplified and made more clear, and
`imageManager.isCleanupNeeded` doesn't need to be stubbed in tests.

Change-type: patch

### DIFF
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -465,71 +465,70 @@ export const save = async (image: Image): Promise<void> => {
 	await markAsSupervised(image);
 };
 
-async function getImagesForCleanup(): Promise<string[]> {
-	const images: string[] = [];
+// TODO: remove after we agree on what to do for Supervisor image cleanup after hup
+const getSupervisorRepos = (imageName: string) => {
+	const supervisorRepos = new Set<string>();
+	supervisorRepos.add(imageName);
+	// If we're on the new balena/ARCH-supervisor image, add legacy image.
+	// If the image name is legacy, the `replace` will have no effect.
+	supervisorRepos.add(imageName.replace(/^balena/, 'resin'));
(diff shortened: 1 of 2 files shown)


pipex (commented):
Great work @cywang117. LGTM when you are ready!

cywang117:
@balena-ci rebase

cywang117:
Ran some tests upgrading to this version, including on an older OS, v2.38 & v9.x Supervisor. Found that `image.RepoTags` can be falsey, so accounted for that in https://github.com/balena-os/balena-supervisor/commit/8174ea9643fb1059aad6fb05c68f42ade0579475 to prevent "RepoTags is not an iterator" errors. 

An observation about why an old Supervisor image may not be cleaned up: `constants.supervisorImage` is always passed to the Supervisor as an env var as SUPERVISOR_IMAGE, and this is equivalent to SUPERVISOR_IMAGE in `/etc/balena-supervisor/supervisor.conf` or `/tmp/update-supervisor.conf` currently. This interface is expected to change. As the image name is in the format `registry2.balena-cloud.com` with a unique hash after it per image, and older Supervisor images have a different unique hash, old Supervisor images may not be cleaned up (unless going through the official Supervisor self-serve upgrade process). 

The cleanup function was from a time when Supervisor images were tagged `balena/$ARCH-supervisor:$VERSION`, thus the cleanup works better when Supervisor images are tagged in that format.

In the future perhaps we can cleanup images by their labels instead of by their names, and Supervisor images can use a specific label.

cywang117:
@balena-ci I self-certify!

PR #2067, opened 2022-11-21
Enable node16 module resolution in tsconfig to ease the ESM transition

Page-:
This means that dynamic import statements will emit actual `import` statements rather than being translated to `require`, the benefit being that we can now import ES modules via dynamic imports

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/test/integration/config.spec.ts
+++ b/test/integration/config.spec.ts
@@ -9,6 +9,9 @@ import * as hostUtils from '~/lib/host-utils';
 import constants = require('~/lib/constants');
 import { fnSchema } from '~/src/config/functions';
 
+// Utility method to use along with `require`
+type Config = typeof import('~/src/config');
+
 describe('config', () => {
 	const configJsonPath = path.join(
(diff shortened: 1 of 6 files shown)


Page-:
A major benefit of this is it means the supervisor will now be able to use ES modules via `import()` rather than having to avoid updating to the ESM only versions of modules

test/integration/config.spec.ts
```diff
@@ -42,15 +42,15 @@ describe('config poop', () => {
 	});
 
 	it('reads and exposes values from config.json', async () => {
-		const config = await import('~/src/config');
```
Page-, line 45: The supervisor tests currently rely on removing the require cache but that will not work with actual `import()` statements hence I've had to change the code that relies specifically on `require` semantics to use `require`

pipex (approved):
I made a small change adding some type definitions to avoid repeating the `typeof import('...')`. Other than that LGTM

PR #2069, opened 2022-12-05
Stop using host journalctl

pipex:
The supervisor had to chroot into the host root in order to read the journal logs. This won't be possible anymore once the supervisor becomes an app. This commit copies the journalctl binary and necessary libraries from a debian image into the supervisor image in order to be able to use the tool on runtime.

Change-type: patch


pipex:
@balena-ci I self certify!

PR #2078, opened 2022-12-12
Fix container name creation to generate valid DNS hostnames

### LINKED ISSUE #2077: supervisor generates invalid DNS names that break reverse DNS lookups
[ramirogm] [ramirogm] Summary of issue discussed on [a forums thread](https://forums.balena.io/t/352814):

A container name longer than 63 chars is causing the issue. From [this docker guide](https://gdevillele.github.io/engine/userguide/networking/configure-dns/) _Container name configured using --name is used to discover a container within an user-defined docker network. The embedded DNS server maintains the mapping between the container name and its IP address (on the network the container is connected to)._ 

A container was created with name `zookeeper_5826140_2403300_72b853b44d0246fd789c89269db742d04a2e5ef9`, which is 66 characters long.
When doing a ping to that container, an error shows up on journalctl:
```
Dec 08 20:54:22 e874ef2 balenad[1236]: time="2022-12-08T20:54:22.376867182Z" level=error msg="[resolver] error writing resolver resp, dns: bad rdata"
```

ramirogm:
Fixes #2077

Change-type: patch
Signed-off-by: Ramiro Gonzalez <ramiro.gonzalez@balena.io>

# Description

Fix container name creation to generate valid DNS hostnames

Fixes #2077

# Type of change

Patch

# How Has This Been Tested?

Unit tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,9 +251,9 @@ If you have upgraded system packages and find that your tests are failing to ini
 
 ### DBus
 
-When developing on macOS you may need to install DBus on the development host.
+When developing on macOS you may need to install DBus on the development host if you get errors like `Package dbus-1 was not found in the pkg-config search path.`.
 
-1. `brew install dbus`
+1. `brew install dbus` or `sudo port install dbus`
 2. `npm ci`
 
 On Debian-based systems, `sudo apt install libdbus-1-dev` would be the equivalent.
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -651,7 +651,7 @@ export class Service {
 		const nameMatch = container.Name.match(/.*_(\d+)_(\d+)(?:_(.*?))?$/);
 		if (nameMatch == null) {
 			throw new InternalInconsistencyError(
-				`Expected supervised container to have name '<serviceName>_<imageId>_<releaseId>_<commit>', got: ${container.Name}`,
+				`Expected supervised container to have name '<serviceName>_<imageId>_<releaseId>(_<commit>)', got: ${container.Name}`,
 			);
 		}
 
@@ -704,8 +704,16 @@ export class Service {
 			}
 			this.config.networkMode = `container:${containerId}`;
 		}
+
+		// Format container name to keep it under 64 characters to make it a legal DNS hostname
+		const imageIdPart = ('' + this.imageId).substring(0,10);
+		const releaseIdPart = ('' + this.releaseId).substring(0,10);
+		// Shorten the service name so that it leaves room for the required imageId and releaseId parts ( see fromDockerContainer )
+		const serviceNamePart = this.serviceName?.substring(0,28);
+		// Use a commit prefix so that all containers will get the same commit suffix in normal cases where the name length is <= 63
+		const commitPrefix = this.commit?.substring(0,12)
 		return {
-			name: `${this.serviceName}_${this.imageId}_${this.releaseId}_${this.commit}`,
+			name: `${serviceNamePart}_${imageIdPart}_${releaseIdPart}_${commitPrefix}`.substring(0, 63),
 			Tty: this.config.tty,
 			Cmd: this.config.command,
 			Volumes: volumes,
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -706,14 +706,17 @@ export class Service {
 		}
 
 		// Format container name to keep it under 64 characters to make it a legal DNS hostname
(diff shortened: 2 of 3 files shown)


pipex:
@ramirogm why do you need this change? Could you describe the problem you are trying to solve? Who is trying to use the full container name as a hostname? The supervisor already configures the serviceName as an alias for all services, see https://github.com/balena-os/balena-supervisor/blob/a482174aa0c69d8fca3da83a751fd375f1b65dd9/src/compose/service.ts#L189

The commit in the container name is used in a bunch of places in the code, including to identify the releases when patching the current state so the way you are doing the transformation might not work. The Web terminal also uses the container name for accessing the services, although that is easier to change.

FWIW, I hope that we can remove the serviceId and imageId from the name in the future, making the container name much shorter.

cywang117:
@pipex This looks to be in response to https://github.com/balena-os/balena-supervisor/issues/2077, which was created from the attached JF ticket in that issue. I'll upload a docker-compose.yml with this message shortly with a local reproduction, if possible.

pipex:
Oh, thank you @cywang117, I'll reply on the issue

ramirogm:
@pipex Hi Felipe! Thanks for looking into this. I've just added more context on https://github.com/balena-os/balena-supervisor/issues/2077#issuecomment-1347613477

This is an issue we found with container names being used by the docker embedded DNS server to perform reverse DNS lookups. Some apps use that, and is also used by `ping` which is what our balena user found to reproduce the issue.

Then I went ahead and created this PR with a candidate solution, just to get this going. The main goal is to get the hostname to be a valid DNS hostname, meaning 1 <= length <= 63. I don't have much knowledge about the dependencies of other projects on that, so I chose a solution that didn't break what's here. 

The solution is basically to leave the `release` and `imageId` as they currently are, and then shorten the commit to 12 characters which should be enough as a `git commit` shortener, and then shorten the serviceName if needed. 

About the specific questions:

>Who is trying to use the full container name as a hostname? The supervisor already configures the serviceName as an alias for all services,
It would be the embedded DNS resolver.
Example:

On a test device, on a container running `ubuntu`:

it resolves names using the embedded DNS server:

```
root@e404e104187f:/# cat /etc/resolv.conf
nameserver 127.0.0.11
options timeout:15 ndots:0
```

We get the IP from a container with a long but legal name:

```
root@e404e104187f:/# nslookup zookeeperXX62
Server:         127.0.0.11
Address:        127.0.0.11#53

Non-authoritative answer:
Name:   zookeeperXX62
Address: 172.17.0.4
```

and if we do the reverse lookup we get the container name, not the alias:

```
root@e404e104187f:/# nslookup -x 172.17.0.4
4.0.17.172.in-addr.arpa name = zookeeperXX62_5841324_2408231_29796bdae4079eecb9e82d78a18c0156.ae8c6ddc272547a49531149bd2dd187f_default.
```

```
zookeeperXX62_5841324_2408231_29796bdae4079eecb9e82d78a18c0156.ae8c6ddc272547a49531149bd2dd187f_default
hostname: 62 chars
domain: 40 chars
```

If I do the same with a container whose name is 64 bytes, the embedded DNS server logs an error.

```
root@e404e104187f:/# nslookup  zookeeperxxxx64
Server:         127.0.0.11
Address:        127.0.0.11#53

Non-authoritative answer:
Name:   zookeeperxxxx64
Address: 172.17.0.3

# this hangs, and I have to CTRL-C it
root@e404e104187f:/# nslookup -x 172.17.0.3
^C
```

while on journalctl:

```
Dec 13 01:43:14 29eba2b balenad[1343]: time="2022-12-13T01:43:14.566083455Z" level=error msg="[resolver] error writing resolver resp, dns: bad data"
```

> The commit in the container name is used in a bunch of places in the code, including to identify the releases when patching the current state so the way you are doing the transformation might not work.

Ok, if we decide to shorten it to 12-chars it should be consistent. Something like https://pkg.go.dev/github.com/docker/docker/pkg/stringid#TruncateID 

@cywang117 About
>ping: bad address 'very-very-very-very-very-very-very-very-very-very-very-long_2_1_localrelease'

I think that this doesn't faithfully reproduces the issue, because the error we found is with the reverse lookup.

pipex:
Closing as this should be improved by #2136

ramirogm:
👍🏼😁Sent from my iPhoneOn 19 Jun 2023, at 19:01, Felipe Lalanne ***@***.***> wrote:﻿
Closed #2078.

PR #2079, opened 2022-12-14
patch: Update API.md

cywang117:
Clarify that only containers on a single device will be restarted following API key refresh, not the whole fleet. Each device has its own distinct API key.


cywang117:
@balena-ci I self-certify!

PR #2081, opened 2022-12-15
Fix journalctl to work with persistent logs

pipex:
v14.4.5 introduced a new way to retrieve the host logs by symlinking the log directories into the host /run/log/journal and using a local copy of the journalctl binary instead of chrooting into the host. This failed to account for the location of persistent logs when that option is turned on (/var/log/journal instead of /run/log/journal). This fixes the checks consider this case and also makes sure to remove the original symlink to avoid having nested links `/run/log/journal/journal/journal`, which was a bug in the previous implementation.

Change-type: patch

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -30,10 +30,23 @@ fi
 # NOTE: this won't be necessary once the supervisor can update
 # itself, as using the label io.balena.features.journal-logs will
 # achieve the same objective
-if [ -d /mnt/root/run/log/journal ]; then
+if { [ ! -d /run/log/journal ] || [ -L /run/log/journal ]; } && [ -s /mnt/root/etc/machine-id ]; then
+	# Only enter here if the directory does not exist or the location exists and is a symlink
+	# (note that test -d /symlink-to-dir will return true)
+
+	# Create the directory
 	mkdir -p /run/log
-	ln -sf /mnt/root/run/log/journal /run/log/journal
+
+	# Override the local machine-id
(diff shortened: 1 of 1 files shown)


entry.sh
```diff
@@ -30,10 +30,22 @@ fi
 # NOTE: this won't be necessary once the supervisor can update
 # itself, as using the label io.balena.features.journal-logs will
 # achieve the same objective
-if [ -d /mnt/root/run/log/journal ]; then
+if { [ ! -d /run/log/journal ] || [ -L /run/log/journal ]; } && [ -s /mnt/root/etc/machine-id ]; then
```
cywang117, line 33:
```suggestion
if { [ ! -d /run/log/journal ] || [ ! -L /run/log/journal ]; } && [ -s /mnt/root/etc/machine-id ]; then
```
Not a symlink --> forgot the `!`

cywang117: Otherwise LGTM!

pipex: Sorry, my comment in the line below is wrong. `-d` returns true even if the file is a symlink. If the directory exists, then only if the file is a symlink we want to enter the `if` clause. I'll fix the comment

pipex: No, my comment  comment in the following line is wrong. If the directory exists, then we only want to enter the `if` section if the file is a symlink (`-d` returns true

pipex: I updated the comment

pipex:
@balena-ci I self certify!

PR #2082, opened 2022-12-20
Create mechanism to read files from host /

pipex:
The supervisor needs to read some files from the host root since there are no alternative interfaces that the host provides. These files are

- /etc/os-release to read the OS version
- /run/openvpn/vpn_status/active to check the VPN status

While we are working to introduce alternative interfaces for reading the information provided by these files, the supervisor still needs a backwards compatible way to get that info.

This PR uses the fact that the supervisor has access to the docker socket to mount a file into a container and read the file from there.

Change-type: minor


pipex:
TODO

- [ ] Get the supervisor image id and use that image to do the reading
- [ ] Write tests
- [ ] Read os-config from root

PR #2088, opened 2023-01-09
Enable http2 when fetching target state

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/src/lib/request.ts
+++ b/src/lib/request.ts
@@ -81,6 +81,7 @@ const getRequestInstances = once(async () => {
 		got: got.extend({
 			responseType: 'json',
 			decompress: true,
+			http2: true,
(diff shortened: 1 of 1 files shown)


cywang117:
@balena-ci rebase

cywang117:
Do not merge yet, pending resolution in [this internal thread](https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/supervisor.20http2.20target.20state)

kb2ma:
Tests show that any bandwidth improvement with HTTP/2 is marginal for single requests like fetching device state. Closing for now at there is not a compelling reason to merge.

Issue #2046, opened 2022-10-26
Investigate how unsupported device configs behave on older Supervisors


cywang117:
When an unsupported `config.txt` variable is passed through target state and received by a Supervisor that doesn't explicitly support it, the behavior on the device is undefined and/or unexpected. We should look into whether the behavior on the device is recoverable. 

This issue will document these undefined behaviors to serve as a reference.

pipex:
Tested this with supervisor 8.0.0 and balenaOS 2.26.0. When given an unknown format for a config variable, the supervisor rejects the target state, at least coming from the `/v2/local/target-state` endpoint

pipex:
State validation seems to have been introduced here, with multi-app https://github.com/balena-os/balena-supervisor/commit/f77d3e15638b20230629aa5274afb18140ff0fef#diff-33390a6675cc73291f322333980c530714faa57fa99e4f443bacdcab6a8136a7R21-R28

PR #2090, opened 2023-01-11
Allow using colon character in config vars

pipex:
The Raspberry Pi config.txt file defines the use of colon to configure variables of the same name in different ports, for instance on those devices with two hdmi ports. This syntax was previously not supported by the supervisor. This change relaxes the syntax validation on config vars to allow the use of the colon character.

Relates-to: #1573, #2046
Change-type: minor


cywang117 (commented):
Nice work! Assuming it works as intended, merge when you're ready 👍🏼

pipex:
@balena-ci I self-certify!

PR #2084, opened 2022-12-27
Feature: Support `since` and `until` in supervisor journald wrapper

rkeulemans:
# Description
The currently implemented API endpoint contains a [subset](https://www.balena.io/docs/reference/supervisor/supervisor-api/#journald-logs) of the journalctl functionality. To retrieve logs within a closed time interval without manually filtering all retrieved logs, support for `since` and `until` is welcome.

Fixes #2083

# Type of change
Change-type: minor

# Notes
- The `since` timestamp was already supported at the library side in `src/lib/journald.ts`, only the argument was not passed from the request body in `src/device-api/v2.ts`. At the library side the type of the `since` timestamp is defined to be a `number`. Therefore, it is assumed that this API was designed to take a UNIX timestamp as input. Since no type checking is happening at runtime, users could theoretically also pass any other argument valid to the javascript `Date()` constructor, if this is the desired API, the documentation should be adjusted accordingly. 

# How Has This Been Tested?
Tested with a device registered in Balena connected to the local network, using [these instructions](https://github.com/balena-os/balena-supervisor#developing-using-a-production-image-or-device).

PR #2099, opened 2023-01-24
Remove custom action restriction for external contributors

cywang117:
Found that custom actions are disabled by default for external contributions, as in https://github.com/balena-os/balena-supervisor/pull/2084. This PR enables them.

Change-type: patch


cywang117:
@balena-ci I self-certify!

Issue #2100, opened 2023-01-25
Supervisor process cleanup could be more efficient


cywang117:
Currently, the Supervisor performs some shutdown operations on receiving a SIGKILL, but I don't believe it receives a SIGTERM for some reason, not necessarily because of the use of `/bin/sh`. This results in shutdown commands sent by the Engine taking 10s each time. 

We can either shorten this grace period by setting `stop_grace_period` in the production Supervisor's docker-compose.yml (this requires Supervisor-as-an-app), or capturing SIGTERM somehow and explicitly calling `process.exit(0)` after all cleanup operations have been completed.

cywang117:
Per @hraftery , for processes like `sleep` and the Supervisor that do not have an explicit handler for SIGTERM, the kernel steps in and kills them except if they're PID 1: https://raby.sh/sigterm-and-pid-1-why-does-a-container-linger-after-receiving-a-sigterm.html

Bucknalla:
Just as a note, please be sure to update the customer on this [ticket](https://jel.ly.fish/support-thread-1-0-0-front-cnv-dntur9p) when we have a resolution :)

pipex:
Extra information. The supervisor runs with pid 1, and node [should receive SIGTERM](https://stackoverflow.com/questions/65677272/what-is-the-default-behaviour-of-node-js-on-receiving-sigterm). As far as I can tell we are not catching SIGTERM in the code so further experimentation needs to happen to see why SIGTERM is not being received

hraftery:
Extra, extra information. Other services that are instantiated in the same way as the supervisor (ie. as a [shell script in a `CMD` statement in *exec* form](https://github.com/balena-os/balena-supervisor/blob/c833d016ff8143aaf77774682efbaa96a365287b/Dockerfile.template#L206)) receive `SIGTERM` just fine. MRE:

- `Dockerfile.template`

```
FROM balenalib/%%BALENA_ARCH%%-alpine
COPY go.sh /
CMD ["/go.sh"]
```

- `go.sh`

```
#!/bin/sh

trap 'echo got_term' 15; while true; do sleep 1; done
```

Pushing this application to a balena device results in a service `main`, which logs "got_term" when it receives `SIGTERM`. This happens both when opening a shell in the service and running `kill -SIGTERM 1`, or when issuing `balena stop <container id>` from HostOS.

In this way the behaviour is exactly as it is in Docker, so rules out any obvious complications introduced by balenaOS.

[...]

PR #2102, opened 2023-01-27
Add SIGTERM listener on supervisor start

pipex:
As reported by issue https://github.com/balena-os/balena-supervisor/issues/2100, the supervisor was not correctly reacting to
`SIGTERM` sent by the engine when terminating the process (for instance
before a reboot). This would lead to the supervisor requiring an
additional 10 seconds to terminate (after which the engine will send a
`SIGKILL`).

The reason for this is explained by the following info coming from Node

> Node.js was not designed to run as PID 1 which leads to unexpected behaviour when running inside of Docker. For example, a Node.js process running as PID 1 will not respond to `SIGINT` (`CTRL-C`) and similar signals. [reference](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals)

On internal testing, it was discovered that simply adding a listener for
the signal on the Node process was enough to handle the signal, even
when the process runs as PID 1.

This adds a listener for `SIGTERM` before starting the supervisor main
loop.

Closes: https://github.com/balena-os/balena-supervisor/issues/2100
Change-type: patch

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -53,7 +53,7 @@ interface DnsLookupOpts {
 		 * and on openBalena setups, this extra code will perform
 		 * the lookup instead of passing it to the built-in
 		 * function.
-		   ================================================== */
+			 ================================================== */
 		if (name && name.endsWith('.local')) {
 			// determine which resolvers to use...
 			const getResolvers = () => {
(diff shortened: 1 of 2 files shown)


cywang117 (commented):
LGTM!

hraftery:
FWIW, I cast a dissenting vote in the issue: https://github.com/balena-os/balena-supervisor/issues/2100

cywang117:
@pipex @hraftery After reading the issue discussion (great one by the way!), it looks like a SIGTERM handler is the agreement here. So once again, this LGTM 👍🏼 good work!

hraftery:
LGTM!

hraftery (approved):

pipex:
I self certify!

pipex:
@balena-ci I self-certify!

PR #2105, opened 2023-01-30
Remove dependent devices content in codebase

cywang117:
This includes:
- proxyvisor.js
- references in docs
- references device-state, api-binder, compose modules, API
- references in tests

The commit also adds a migration to remove the 4 dependent device tables from the DB.

Change-type: minor


cywang117:
Tested in the following ways:
- Pin back and forth between releases
- Move device to another app and move it back
- Dashboard restart
- Dashboard purge
- Local mode push

All looks good, database migrations went smoothly as well. Ready for review!

src/migrations/M00010.js
```diff
@@ -0,0 +1,26 @@
+// This migration removes references to dependent devices from
+// the database, as we are no longer pursuing dependent devices.
+export async function up(knex) {
+	// Remove dependent key from each image
+	if (await knex.schema.hasColumn('image', 'dependent')) {
```
pipex, line 5: I don't think you need to test that the column and table exist, the migrations are ran in order, so you can assume that the structure of migration 9 is what exists before this migration. It doesn't really hurt if you want to leave it though

pipex:
This looks good to me except for the comment on the migration. Great work! Love to get rid of unused code!

cywang117:
@balena-ci I self-certify!

PR #2120, opened 2023-02-02
Mount host partitions into container on Supervisor start

cywang117:
As the Supervisor is a privileged container, it has access to host /dev, and therefore has access
to root, boot, data, and state balenaOS partitions. This commit does the following:

- Finds the /dev partition that corresponds to each partition based on partition label
- Mounts the partitions into set mountpoints in the device
- Removes reliance on env vars and mountpoints provided by host's start-balena-supervisor script
- Simplifies host path querying by centralizing these queries through methods in lib/host-utils.ts

Since the Supervisor would no longer rely on container `run` arguments provided by the host script that starts the Supervisor container ([`start-balena-supervisor`](https://github.com/balena-os/meta-balena/blob/c005b8bf80326b05d3e7605bdcfda43433232679/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor/start-balena-supervisor)), this change moves Supervisor closer to being able to start itself (Supervisor-as-an-app).

Change-type: minor


docker-compose.test.yml
```diff
@@ -18,17 +18,16 @@ services:
     environment:
       DOCKER_HOST: tcp://docker:2375
       DBUS_SYSTEM_BUS_ADDRESS: unix:path=/run/dbus/system_bus_socket
-      # Required by migrations
-      CONFIG_MOUNT_POINT: /mnt/root/mnt/boot/config.json
-      # Read by constants to setup `bootMountpoint`
-      BOOT_MOUNTPOINT: /mnt/boot
```
cywang117, line 24: These env vars are no longer needed as they're set by `entry.sh`.

src/compose/network-manager.ts
```diff
@@ -10,16 +9,11 @@ import log from '../lib/supervisor-console';
...
-		return docker
-			.getNetwork(network.Name)
-			.inspect()
-			.then((net) => {
-				return Network.fromDockerNetwork(net);
-			});
-	});
+export async function getAll(): Promise<Network[]> {
+	const dockerNets = await getWithBothLabels();
+	return dockerNets.map((net) => Network.fromDockerNetwork(net));
```
cywang117, line 16: The `inspect()` is superfluous because `getNetwork` already returns the needed network info.

src/config/configJson.ts
```diff
@@ -87,56 +84,38 @@ export default class ConfigJsonConfigBackend {
...
 		}
-
-		if (/^(Resin OS|balenaOS)/.test(osVersion)) {
-			// In Resin OS 1.12, $BOOT_MOUNTPOINT was added and it coincides with config.json's path.
-			if (constants.bootMountPointFromEnv != null) {
-				return path.join(constants.bootMountPointFromEnv, 'config.json');
-			}
-			// Older 1.X versions have config.json here
+		if (semver.lte(osSemver, '1.12.0')) {
 			return '/mnt/conf/config.json';
```
cywang117, line 115: See QUESTION: do we need to be backwards compatible with OS versions older than 1.12?

cywang117:
After some testing, this will need some more thought. `/mnt/root` in the Supervisor, `/` on host, is an overlayfs which also mounts some things from other parts of the host, not just the rootB (active root) partition. The main thing that is mounted is host's `/tmp`, under `/mnt/root/tmp` in container. rootB includes a `/tmp` but it isn't used for inter-container communication using update locks. There is no way to access the host overlayfs in-container, so the only way to move forward while using the same location for update locks is to bind-mount `/tmp:/tmp`. Balena disallows this, and although there is a way around that, is it a good idea?

Additionally, the location from which `os-release` is read would need to become `/mnt/boot/os-release`, however we need to determine whether this will always be the current OS release (I believe so, but needs confirmation.)

There may be other files which now need to be accessed differently, but the update lock shared directory is the main issue.

cywang117:
Idea regarding update locks: What if starting from this Supervisor, update locks were mounted by the Supervisor on startup under `/dev/shm`, and it mounted `/dev/shm/<appId>/services/<service>:/tmp/balena` for each service container? From the perspective of the user services, the path is the same as before, so they can use update locks as they've done previously. The only hiccup I can see is the matter of upgrading to this Supervisor version when there are already user-created update locks in the old `/mnt/root` mount...

cywang117:
Testing this some more, another blocker is that `/mnt/root/run/openvpn/vpn_status` is only mounted under `/mnt/root`. We need to find a way to access the dbus socket without depending on `/mnt/root` before we can proceed with this PR. For OS backwards compatibility, a viable method would probably be to spin up a container with the dbus socket bind mounted into container for any dbus operations. We could also do this for `/mnt/root/tmp` instead of my previous comment (or in concert with my previous comment). Another way which would be better as an intermediate step towards the Supervisor upgrading itself, if we can pull it off, is for Supervisor startup to register balena feature labels for the Supervisor somehow. Although, I don't know how this could be accomplished.

cywang117:
After discussion, I found that `resin-rootB`  is not a substitute for the host `/mnt/root` mount (which is an overlayfs and not accessible via `/dev`). Thus, this PR will only mount data, state, and boot partitions into container to move forward. We'll utilize some other method for moving away from relying on the `/:/mnt/root` mount from the [Supervisor startup script](https://github.com/balena-os/meta-balena/blob/d21da1fc034f4e0b09a457d0ed5eda47d0a18e13/meta-balena-common/recipes-containers/balena-supervisor/balena-supervisor/start-balena-supervisor#L74).

cywang117:
I have tested this on a device and can confirm that the following actions work:

- Setting a boot config
- Pinning between releases
- Moving the device
- Dashboard restart
- Dashboard purge
- Local mode push

Ready for review!

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/sh
```
pipex, line 1: This is not really a build-util, it's a runtime util. We should create a new folder or just put it on the same level as the `entry.sh` script

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+
+use_legacy_message="Using legacy host mounts."
+
+# Get the current boot device in case there are duplicate partition labels
+# for `(balena|resin)-(boot|state|data)` found. Because Supervisor is started with
+# `--net=host`, /etc/resolv.conf is bind-mounted by the Engine into the container,
+# so the partition where resolv.conf is located is assumed to be one of the
+# partitions of the current active device. The partition identifer `pN` is removed 
+# from the end so we can compare with the base device.
+current_boot_device=$(grep /etc/resolv.conf /proc/mounts | awk '{print substr($1,1,length($1)-2)}')
```
pipex, line 14: This seems like too much of a hack for my taste. Maybe someone on the OS (cc @alexgg) has some ideas on how to do this more cleanly?

alexgg: @pipex in the OS we do something like:
```
findmnt --noheadings --canonicalize --uniq --output SOURCE /mnt/sysroot/active
```
Which will output the device the active partition is mounted from. Is that what you are looking for?

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+if [ "${current_boot_device}" = "" ]; then
+    echo "WARNING: Could not determine boot device. Please launch Supervisor as a privileged container with host networking."
```
pipex, line 16: This should probably exit with an error, there might be no point to fallback to using /mnt/root if the container is not provileged

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+    
+    count=$(echo "${device}" | grep -c "/dev")
+    if [ "${count}" = 0 ]; then
+        # If no devices with the partition label are found, use legacy mounts.
+        echo "WARNING: No partitions labeled '(resin|balena)-${partition}' found. $use_legacy_message"
+        return 1
+    fi
+
+    if [ "${count}" = 1 ]; then
+        # QUESTION: Should we check whether do_mount succeeded in mounting and inform of failure?
```
pipex, line 59: The multiple device case seems to already includes the single device case, you might be able to remove this and simplify the code

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+    # that's part of the current boot device. as this indicates a duplicate 
+    # label somewhere created by a user or an inconsistency in the system.
+    # We've been able to identify the current boot device, so use that
+    # to find the device with the correct label amongst 2+ devices.
+    found=false
+    for d in ${device}; do
+        if [ "$(echo "$d" | grep -E "^$current_boot_device")" != "" ]; then
+            echo "INFO: Found device $d on current boot device $current_boot_device, using as mount for '(resin|balena)-${partition}'."
+            do_mount "${d}" "${path}"
+            found=true
```
pipex, line 74: You can call `break` here too

build-utils/mount-partitions.sh
pipex: Yup, that's what I was looking for, thank you @alexgg!

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+    if [ ! $found ]; then
+        echo "WARNING: Multiple partitions labeled '(resin|balena)-${partition}' found but none from current boot device. $use_legacy_message"
+        return 1
+    fi
+
+    return 0
+}
+
+# Set overlayfs root mountpoint
+export ROOT_MOUNTPOINT="/mnt/root"
```
pipex, line 87: I'm still doubtful on keeping the `/mnt/root` fallback. I agree it might be useful while we ensure to have covered all border cases, but we probably should stop relying on it. If a mounting step fails, the system may have bigger problems that might even become worse by starting a supervisor.

pipex (commented):
Not finished reviewing, just added a few comments on the script

src/compose/app.ts
```diff
@@ -794,8 +795,8 @@ export class App {
 					.getNetworkGateway(constants.supervisorNetworkInterface)
 					.catch(() => '127.0.0.1'),
 				(async () => ({
-					firmware: await pathExistsOnHost('/lib/firmware'),
```
pipex, line 797: Not sure this is an improvement, I can see utility on keeping `pathExistsOnHost`

pipex: Also not entirely sure how this fits with the current commit.

build-utils/mount-partitions.sh
cywang117: This is a good idea, however doesn't work in the Supervisor container as the container only has access to the following mounts, as listed in `/proc/mounts`:
```
/usr/src/app # cat /proc/mounts | grep -v 'overlay' | grep -v 'docker' | grep -v '/mnt/root'
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev tmpfs rw,nosuid,size=65536k,mode=755 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666 0 0
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /sys/fs/cgroup tmpfs rw,nosuid,nodev,noexec,relatime,mode=755 0 0
cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,name=systemd 0 0
cgroup /sys/fs/cgroup/blkio cgroup rw,nosuid,nodev,noexec,relatime,blkio 0 0
cgroup /sys/fs/cgroup/cpu,cpuacct cgroup rw,nosuid,nodev,noexec,relatime,cpu,cpuacct 0 0
cgroup /sys/fs/cgroup/devices cgroup rw,nosuid,nodev,noexec,relatime,devices 0 0
cgroup /sys/fs/cgroup/pids cgroup rw,nosuid,nodev,noexec,relatime,pids 0 0
cgroup /sys/fs/cgroup/freezer cgroup rw,nosuid,nodev,noexec,relatime,freezer 0 0
cgroup /sys/fs/cgroup/cpuset cgroup rw,nosuid,nodev,noexec,relatime,cpuset 0 0
cgroup /sys/fs/cgroup/net_cls,net_prio cgroup rw,nosuid,nodev,noexec,relatime,net_cls,net_prio 0 0
cgroup /sys/fs/cgroup/perf_event cgroup rw,nosuid,nodev,noexec,relatime,perf_event 0 0
cgroup /sys/fs/cgroup/memory cgroup rw,nosuid,nodev,noexec,relatime,memory 0 0
mqueue /dev/mqueue mqueue rw,nosuid,nodev,noexec,relatime 0 0
shm /dev/shm tmpfs rw,nosuid,nodev,noexec,relatime,size=65536k 0 0
/dev/nvme0n1p6 /data ext4 rw,relatime 0 0
tmpfs /var/log tmpfs rw,relatime 0 0
/dev/nvme0n1p1 /boot/config.json vfat rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,errors=remount-ro 0 0
/dev/nvme0n1p6 /boot/apps.json ext4 rw,relatime 0 0
proc /mnt/fib_trie proc rw,relatime 0 0
/dev/nvme0n1p6 /etc/resolv.conf ext4 rw,relatime 0 0
/dev/nvme0n1p6 /etc/hostname ext4 rw,relatime 0 0
/dev/nvme0n1p6 /etc/hosts ext4 rw,relatime 0 0
/dev/disk/by-state/resin-state /usr/share/ca-certificates/balena ext4 rw,relatime 0 0
tmpfs /run/balena-engine.sock tmpfs rw,nosuid,nodev,mode=755 0 0
```

This is excluding `/mnt/root` which we don't want to rely on, as well as any overlay mounts resulting from docker or otherwise. Furthermore, we need to exclude `/boot` and `/data` as those are also mounts provided by the OS and shouldn't be relied on. Based on this, `/etc/resolv.conf`, `/etc/hostname`, and `/etc/hosts` look to be the only ones we can rely on to look for the current boot device.

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+    partition=$1
+    path=$2
+
+    # Get one or more devices matching label, accounting for legacy partition labels.
+    device=$(blkid | grep -E "(resin|balena)-${partition}" | awk -F':' '{print $1}')
+    
+    count=$(echo "${device}" | grep -c "/dev")
+    if [ "${count}" = 0 ]; then
+        # If no devices with the partition label are found, use legacy mounts.
+        echo "WARNING: No partitions labeled '(resin|balena)-${partition}' found. $use_legacy_message"
```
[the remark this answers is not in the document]
cywang117: I suppose we'd have bigger problems on the device if the partitions weren't available. I'll remove this check.

build-utils/mount-partitions.sh
cywang117: The eventual goal would be to remove this, but we can't merge this change today without the `/mnt/root` dependency due to needing it for update locks and checking the VPN status.

src/compose/app.ts
cywang117: I can separate this into another commit, but the idea here is that all references to host path checking (path on root/boot/data/state) go through host-utils.ts and isn't split between fs-utils.ts and host-utils.ts. I can add `pathExistsOnHost` into host-utils if that works for you?

build-utils/mount-partitions.sh
pipex: `fdisk -l` [might be](https://askubuntu.com/a/952789) an alternative? 

```
/usr/src/app # fdisk -l | grep '*  ' | cut -d' ' -f1 2>&1
/dev/mmcblk0p1
```

pipex:
Slowly working my way down the PR. I think you might need to do something about this too @cywang117 

https://github.com/balena-os/balena-supervisor/blob/master/src/lib/system-info.ts#L30-L35

build-utils/mount-partitions.sh
cywang117: Never mind, I think I know what you mean -- we don't want to fall back to /mnt/root/mnt/boot at all if the partitions aren't found, right? Just exit with error?

cywang117:
Nice catch, thanks @pipex !

build-utils/mount-partitions.sh
pipex: yeah, that’s what I was thinking of

src/compose/network-manager.ts
```diff
@@ -10,16 +9,18 @@ import log from '../lib/supervisor-console';
...
 export async function getAll(): Promise<Network[]> {
-	return getWithBothLabels().map(async (network: { Id: string }) => {
-		return docker
-			.getNetwork(network.Id)
-			.inspect()
-			.then((net) => {
-				return Network.fromDockerNetwork(net);
-			});
-	});
+	const dockerNets = await getWithBothLabels();
```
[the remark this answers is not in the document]
cywang117: You're right, I'll separate these into another PR as this change is unrelated.

build-utils/mount-partitions.sh
cywang117: Good shout, that works! Thanks 👍🏼

build-utils/mount-partitions.sh
cywang117: Moved to project root in 71cddb6a1b9d51a59a350726a80035e5afce1403

build-utils/mount-partitions.sh
cywang117: Switched to `fdisk` command in [latest force-push](https://github.com/balena-os/balena-supervisor/pull/2120/commits/71cddb6a1b9d51a59a350726a80035e5afce1403#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R11)

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+current_boot_device=$(grep /etc/resolv.conf /proc/mounts | awk '{print substr($1,1,length($1)-2)}')
+if [ "${current_boot_device}" = "" ]; then
+    echo "WARNING: Could not determine boot device. Please launch Supervisor as a privileged container with host networking."
+fi
+
+# Mounts a device to a path if it's not already mounted.
+# Usage: do_mount DEVICE MOUNT_PATH 
+do_mount() {
+    device=$1
+    path=$2
```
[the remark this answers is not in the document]
cywang117: Renamed to `mount_path` in [latest force-push](https://github.com/balena-os/balena-supervisor/pull/2120/commits/71cddb6a1b9d51a59a350726a80035e5afce1403#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R20-R21)

build-utils/mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+setup_then_mount() {
+    # If in test environment, pretend we've succeeded at mounting everything to their
+    # new mountpoints. We don't want to actually mount in a containerized test environment 
+    # where the Supervisor is probably not running on a host that has the needed partitions.
+    if [ "${TEST}" = 1 ]; then
+        return 0
+    fi
+
+    partition=$1
+    path=$2
```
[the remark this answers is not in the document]
cywang117: Renamed to `partition_label` and `target_path` in [latest force-push](https://github.com/balena-os/balena-supervisor/pull/2120/commits/71cddb6a1b9d51a59a350726a80035e5afce1403#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R43-R44)

build-utils/mount-partitions.sh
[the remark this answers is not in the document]
cywang117: Removed in latest force push.

build-utils/mount-partitions.sh
cywang117: Added `exit 1` in [latest force-push](https://github.com/balena-os/balena-supervisor/pull/2120/commits/71cddb6a1b9d51a59a350726a80035e5afce1403#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R13-R14).

build-utils/mount-partitions.sh
cywang117: Simplified in latest force-push.

build-utils/mount-partitions.sh
cywang117: Changed to remove the `found` variable, `return 0` if mount success, and `exit 1` if no mounting occurred in [latest force-push](https://github.com/balena-os/balena-supervisor/pull/2120/commits/71cddb6a1b9d51a59a350726a80035e5afce1403#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R54-R64).

build-utils/mount-partitions.sh
cywang117: Removed the fallback to ROOT_MOUNTPOINT as base path for boot, data, and state partitions.

src/compose/network-manager.ts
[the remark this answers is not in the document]
cywang117: Removed this change from this PR in latest force-push.

cywang117:
@pipex I changed the system-information check to [this](https://github.com/balena-os/balena-supervisor/pull/2120/commits/f52fff6fae1b0010da52e6bdf571bbab80c1f652#diff-bef2acc363ca1274b929672bde119e0c8079cde6017f93cb26bfb19ecddaa27fL30-R31) -- it checks for a mountpoint matching `/data` using regex instead of checking for `/mnt/data` directly, because systeminformation's `fsSize()` doesn't consistently return all the mounts. [Another user](https://github.com/sebhildebrandt/systeminformation/issues/777#issuecomment-1436729740) has noted this in the systeminformation issues repo.

src/compose/app.ts
pipex: > I can add pathExistsOnHost into host-utils if that works for you?

Yeah, that works, we'll avoid the `exists(pathOnHost(...))` as happens a lot

mount-partitions.sh
```diff
@@ -0,0 +1,80 @@
...
+# The container must be privileged for this to function correctly.
+
+use_legacy_message="Using legacy host mounts."
+
+# Get the current boot device in case there are duplicate partition labels
+# for `(balena|resin)-(boot|state|data)` found. The partition identifer `pN`
+# is removed from the end so we can compare with the base device.
+current_boot_device=$(fdisk -l | grep '*  ' | cut -d "p" -f1 2>&1)
+if [ "${current_boot_device}" = "" ] && [ "${TEST}" != 1 ]; then
+    echo "WARNING: Could not determine boot device. Please launch Supervisor as a privileged container with host networking."
```
pipex, line 13: this should probably be prefixed by `ERROR` since it terminates execution

src/lib/constants.ts
```diff
@@ -1,43 +1,53 @@
...
-const bootMountPointFromEnv = checkString(process.env.BOOT_MOUNTPOINT);
-const rootMountPoint = checkString(process.env.ROOT_MOUNTPOINT) || '/mnt/root';
-
 const supervisorNetworkInterface = 'supervisor0';
 
+// /mnt/root is the legacy root mountpoint
+const rootMountPoint = checkString(process.env.ROOT_MOUNTPOINT) || '/mnt/root';
+const withRootMount = (p: string) => path.join(rootMountPoint, p);
+const bootMountPoint =
+	checkString(process.env.BOOT_MOUNTPOINT) || withRootMount('/mnt/boot');
```
pipex, line 10: should this be `/mnt/boot` instead of `withRootMount('/mnt/boot')`

pipex, line 10: should this be `/mnt/boot` instead of `withRootMount('/mnt/boot')`

src/config/configJson.ts
pipex: Oh, also OS 1.12.0 might still have `/mnt/boot` but they mount it to `/mnt/conf` in which case the mount script should still work and this case becomes useless

entry.sh
```diff
@@ -33,7 +33,7 @@ fi
 # NOTE: this won't be necessary once the supervisor can update
 # itself, as using the label io.balena.features.journal-logs will
 # achieve the same objective
-if { [ ! -d /run/log/journal ] || [ -L /run/log/journal ]; } && [ -s "${ROOT_MOUNTPOINT}"/etc/machine-id ]; then
+if { [ ! -d /run/log/journal ] || [ -L /run/log/journal ]; } && [ -s "${STATE_MOUNTPOINT}"/machine-id ]; then
```
pipex, line 36: Is this always the case? Even after HUP? Maybe we can check with the OS (if you have not done it yet)

src/lib/constants.ts
```diff
@@ -24,13 +28,14 @@ const constants = {
 		checkString(process.env.HOST_OS_VERSION_PATH) ||
 		withRootMount('/etc/os-release'),
 	// Data paths
+	dataMountPoint,
 	databasePath:
 		checkString(process.env.DATABASE_PATH) || '/data/database.sqlite',
 	appsJsonPath:
 		process.env.APPS_JSON_PATH || withRootMount('/mnt/data/apps.json'),
```
pipex, line 35: I don't think we set `APPS_JSON_PATH` anywhere, so probably this should default to `/mnt/data/apps.json`

pipex (commented):
OK, I did the final pass. Just a couple of extra comments. Let's make sure to test it well with a couple of device types and scenarios (first boot (no database)/preloading/pre-registering), etc., to avoid any surprises. I can help you with the round of tests if you need

src/config/configJson.ts
cywang117: I don't think we need to support this - I took a look at the oldest version listed when checking `balena os versions ${slug} | tail -1` for every device in `balena devices supported`, and the oldest is `v1.16.0` for intel-nuc and some other DTs.

mount-partitions.sh
cywang117: Added this [here](https://github.com/balena-os/balena-supervisor/pull/2120/files#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R13)

src/compose/app.ts
cywang117: I added `pathExistsOnRoot` and `pathExistsOnState` to host-utils [here](https://github.com/balena-os/balena-supervisor/pull/2120/files#diff-caf13efa686c7a4bb534ffdcb897f8028fcf67898446a4db30724ec875639405R26-R42), previously the function that would check on state would get the state path first and then pass it to `pathExistsOnHost` (root). However, I didn't want to do this here because the intent is to refer to host-utils for any path-getting or path-resolving functionality.

src/config/configJson.ts
cywang117: I removed the check for OS 1.12.0 entirely [here](https://github.com/balena-os/balena-supervisor/pull/2120/files#diff-3a034722503c6a7e6a3654f6f42c046256faaca32ff3d770ba5d727efff82470L114-L129)

src/lib/constants.ts
cywang117: I updated boot, data, and state mountpoints to not use `withRootMount` as fallback [here](https://github.com/balena-os/balena-supervisor/pull/2120/files#diff-8c943ccfb5928fff63708e99eaafa93f8845c9405ed8e4b32afed23e5003025aR9-R12)

src/lib/constants.ts
cywang117: Removed the env var [here](https://github.com/balena-os/balena-supervisor/pull/2120/files#diff-8c943ccfb5928fff63708e99eaafa93f8845c9405ed8e4b32afed23e5003025aR32)

cywang117:
Tests to run (in progress):
- [ ] Check if machine-id is present in state partition after HUP
- [ ] Behavior on first boot
- [ ] Older OS (intel-nuc 2.38.3+rev5)
- [ ] Preloading user release (pin / no pin)
- [ ] Pre-registering device
- [ ] RPi3 (`armv7hf`) DT
- [ ] RPi4 (`aarch64`) DT
- [ ] Behavior on deleting DB
- [ ] OS automated tests with draft release

mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# Mounts boot, state, & data partitions from balenaOS.
+# The container must be privileged for this to function correctly.
+
+# Get the current boot block device in case there are duplicate partition labels
+# for `(balena|resin)-(boot|state|data)` found.
+# Returns sdaX, mmcblk0pX, nvme0n1pX, etc.
+current_boot_partition=$(fdisk -l | grep '*  ' | awk '/^\/dev/{sub("/dev/","",$1); print $1}')
+# Returns /dev/sda, /dev/mmcblk0, /dev/nvme0n1, etc. This is the 
+current_boot_block_device="/dev/$(basename "$(readlink -f "/sys/class/block/${current_boot_partition}"/..)")"
```
cywang117, line 11: I made this change in the latest commit because before, we were running `fdisk -l | grep '*  ' | cut -d' ' -f1 2>&1`, which would return the boot partition device, `/dev/sda1` / `/dev/nvme0n1p1` / `/dev/mmcblk0p1`, however, to see whether other partitions belong to the current block device, we need to remove the partition identifier from those devices. That identifier differs between sda, nvme, and mmcblk - it's either simply a number or `pN`. I've tested this on an `sda` block device and an `nvme0n1` block device and it works on both, so should be the more flexible way to get the current boot block device.

entry.sh
cywang117: I tested this (see checklist below) and even on old devices, before and after HUP, the machine-id is present in the state partition at this path.

mount-partitions.sh
pipex: I'm not entirely sure we should rely on the folder structure across kernels/os versions. I searched for a standard way to do this, and it looks like `lsblk` is the tool you might want to use  https://unix.stackexchange.com/a/308726

```
root@b308f87:~# lsblk -no pkname /dev/mmcblk0p1
mmcblk0
```

cywang117: Changed it to use lsblk (and install in Dockerfile) here: 
https://github.com/balena-os/balena-supervisor/pull/2120/files#diff-8a96bc9af661488f86b92785e9382315f61c548974c4bcce92c1a1deb9a129f8R8-R15

mount-partitions.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/sh
+
+# Mounts boot, state, & data partitions from balenaOS.
+# The container must be privileged for this to function correctly.
+
+# Get the current boot block device in case there are duplicate partition labels
+# for `(balena|resin)-(boot|state|data)` found.
+if [ "${TEST}" != 1 ]; then
+    current_boot_partition=$(fdisk -l | grep '*  ' | cut -d' ' -f1 2>&1)
+    export current_boot_block_device=$(lsblk -no pkname "${current_boot_partition}")
```
pipex, line 10: I don't think the export is necessary

cywang117: Fixed it in latest!

pipex:
@cywang117 no more observations, merge when ready !

cywang117:
@balena-ci I self-certify!

cywang117:
RE: Automated OS tests: Jenkins tests [pass](https://github.com/balena-os/meta-balena/pull/3088)

PR #2129, opened 2023-02-06
patch: Convert internal timestamp passed to journalctl from number to string

cywang117:
Following PR #2084, I discovered that logging/monitor.ts was passing `since` to `spawnJournalctl` as a number, whereas `journalctl` (and `spawnJournalctl`, our wrapper) only accepts dates formatted in accordance with the systemd.time spec. (`YYYY-MM-DD HH:MM:SS(.s*)`). Thus I added back in [this commit](https://github.com/balena-os/balena-supervisor/commit/cb468b298a7cb87ffbd486dfd5482d5f2e43bb14) from @rkeulemans  for internal use only.

See: https://github.com/balena-os/balena-supervisor/pull/2084
Change-type: patch

### DIFF
--- a/src/lib/journald.ts
+++ b/src/lib/journald.ts
@@ -2,6 +2,17 @@ import { ChildProcess, spawn } from 'child_process';
 
 import log from './supervisor-console';
 
+/**
+ * Given a date integer in ms, return in a format acceptable by journalctl.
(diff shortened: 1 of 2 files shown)


cywang117:
@balena-ci I self-certify!

PR #2130, opened 2023-02-09
Find and remove duplicate networks

pipex:
We have seen a few times devices with duplicated network names for some
reason. While we don't know the cause the networks get duplicates, this
can be disruptive for updates as trying to create a container referencing a duplicate
network results in a 400 error from the engine.

This commit finds and removes duplicate networks via the state engine,
this means that even if somehow a container could be referencing a
network that has been duplicated later somehow, this will remove the
container first.

While thies doesn't solve the problem of duplicate networks being
created in the first place, it will fix the state of the system to
correct the inconsistency.

Change-type: minor
Closes: https://github.com/balena-os/balena-supervisor/issues/590

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -35,8 +35,8 @@ export interface AppConstructOpts {
 	isHost?: boolean;
 
 	services: Service[];
-	volumes: Dictionary<Volume>;
-	networks: Dictionary<Network>;
+	volumes: Volume[];
+	networks: Network[];
 }
 
 export interface UpdateState {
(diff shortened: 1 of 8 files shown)


src/compose/app.ts
```diff
@@ -231,34 +236,64 @@ export class App {
...
+			current.forEach((item, index) => {
+				const hasDuplicate =
+					current.findIndex((it) => it.name === item.name) !== index;
+
+				if (
+					hasDuplicate &&
+					// Skip components that are being updated as those will need to be removed anyway
+					!toBeUpdated.includes(item.name) &&
+					// Avoid adding the component again if it has already been marked for removal
+					!toBeRemoved.includes(item.name)
```
cywang117, line 282: Potentially could avoid this if a `Set` is used instead

cywang117 (commented):
Nice work @pipex ! Merge when ready :)

pipex:
@balena-ci I self certify!

pipex:
I self-certify!

PR #2135, opened 2023-02-28
Replace BALENA-FIREWALL rule in INPUT chain instead of flushing

cywang117:
The issue with the original Supervisor implementation of the firewall is that on Supervisor start, the Supervisor flushes the INPUT chain of the filter table. This doesn't play well with services that add to the INPUT chain on startup that may start up before the Supervisor, such as certain NetworkManager connection profiles. This change only replaces the BALENA-FIREWALL rule in the INPUT chain, preserving the other rules as well as their order.

Closes: #1482
Change-type: patch


src/lib/firewall.ts
```diff
@@ -191,16 +195,20 @@ export async function applyFirewallMode(mode: string) {
...
+						if (firewallRuleExists) {
+							chain.addRule({
+								action: iptables.RuleAction.Delete,
 								target: 'BALENA-FIREWALL',
-							}),
-					),
+							});
+						}
+						chain.addRule({
+							action: iptables.RuleAction.Insert,
```
pipex, line 206: Does the rule have to go in the same position? Why?

pipex: Could we alternatively just remove the existing rule and add a new one each time, what are the drawbacks?

src/lib/iptables.ts
```diff
@@ -291,3 +303,31 @@ async function applyRules(rules: Rule | Rule[], adaptor: RuleAdaptor) {
...
 	await adaptor(processedRules);
 }
+
+class StdError extends TypedError {}
+
+export async function getRulePosition(
+	chain: string,
+	ruleMatch: string,
+): Promise<number> {
+	const cmd = `iptables -L ${chain} --line-numbers`;
```
pipex, line 313: Are the results of this command consistent across OS versions? This goes with the above comments.

src/lib/firewall.ts
cywang117: The change does remove the existing rule and add a new one. It maintains the same position to preserve iptables behavior when there are multiple rules in the INPUT chain. since the rules evaluate based on their position in the chain.

src/lib/iptables.ts
cywang117: `--line-numbers` is supported all the way back to 1.1.2, the oldest version with source code I could find from [this project archive](https://netfilter.org/projects/iptables/downloads.html). 1.1.2 was released in year 2000 so I think we're good, especially because `iptables` itself adds the line numbers to the results of the `-L ${chain}` query.

majorz, line 313: Let's add `-n` flag here after the `-L` argument, because without it sometimes I see significant slowdowns in iptables output due to reverse DNS lookup.

majorz: Mentioned [here](https://ipset.netfilter.org/iptables.man.html) as well on the `-L` docs.

src/lib/firewall.ts
majorz: I agree that we need to preserve the order, so that we do not scramble the the rules if there is some change from a container that would be affected by reordering.

src/lib/iptables.ts
majorz: The iptables output is stable indeed as it is regularly used for parsing.

majorz (changes requested):
Just one change requested about using `-n`.

src/lib/iptables.ts
cywang117: Added `-n` in [this line](https://github.com/balena-os/balena-supervisor/pull/2135/commits/01faa2cba7c902aad28e5e5f0ed2d1fb17645620#diff-fa58e353d5a5691026b53449e0a68869fbfda197dc1c2b62b1cc2e1f56303272R313).

majorz (approved):
Great, thanks!

cywang117:
The iptables stdout from the exec query (`iptables -L INPUT -n --line-numbers) should be stable across multiple device types. Raw stdout when there are 0 rules in the INPUT chain:
```
'Chain INPUT (policy ACCEPT)\n' +
  'num  target     prot opt source               destination         \n'
```
Running `stdout.split('\n').find((l) => l.includes(ruleMatch))` will return `undefined`, and therefore the function will return -1. 
For a case where the BALENA-FIREWALL rule is in the middle of the chain, the raw stdout:
```
'Chain INPUT (policy ACCEPT)\n' +
  'num  target     prot opt source               destination         \n' +
  '1               tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:22\n' +
  '2               tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:443\n' +
  '3               tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:80\n' +
  '4    BALENA-FIREWALL  all  --  0.0.0.0/0            0.0.0.0/0           \n' +
  '5               tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:81\n'
```
Running `stdout.split('\n').find((l) => l.includes(ruleMatch))` returns `'4    BALENA-FIREWALL  all  --  0.0.0.0/0            0.0.0.0/0           '`, and running `parseInt(line.split(' ')[0], 10)` on that returns `4` as a number. 

Lastly, any errors will return -1, which the calling function knows how to handle in the case that it was called here.

You can test this yourself by adding some dummy rules to the INPUT chain from the Supervisor container:
```
// Flush INPUT
iptables -F INPUT
// Append the following rules in this order
iptables -A INPUT -p tcp --dport 22
iptables -A INPUT -p tcp --dport 443
iptables -A INPUT -p tcp --dport 80
iptables -A INPUT -j BALENA-FIREWALL
iptables -A INPUT -p tcp --dport 81
```
...then livepushing the Supervisor on this branch to the device. On Supervisor start, you'll see that these arbitrary non-BALENA-FIREWALL input rules are no longer deleted. I've already tested this myself and am satisfied with the results.

cywang117:
Made a small patch to handle rule position reading of iptables & ip6tables separately, because the chain contents may differ between IPv4 and IPv6.

PR #2136, opened 2023-03-02
Remove releaseId and imageId from container names

pipex:
The release id and image id are not really necessary for reporting the current state since v3 (it has been replaced by `releaseUuid` + `imageName`) and is only kept for backwards compatibility for some supervisor API endpoints.

Until now, both the `releaseId` and `imageId` were part of the container name, leading to longer names than desired. This change removes those values from the container name and modifies queries to look for the value in the image database.

Change-type: minor
Relates-to: #2077


src/device-api/v2.ts
```diff
@@ -421,25 +424,34 @@ router.get('/v2/containerId', async (req: AuthorizedRequest, res) => {
...
 				'appId',
-				'imageId',
 				'serviceId',
+				'imageId',
 				'containerId',
 				'createdAt',
 			);
+
+			// QUESTION: what to do if we cannot find the image?
+			// this will happen in local mode or if the image database got deleted
```
pipex, line 444: This makes that endpoints behave differently on local mode vs in production, that is not ideal as it means you cannot rely on these endpoints during development. I’ll take a look at improving local mode before merging this PR

src/device-api/actions.ts
```diff
@@ -496,6 +510,10 @@ export const getLegacyDeviceState = async () => {
 		_.toPairs(state.local?.apps)[0]?.[1]?.services,
 	)[0]?.[1];
 
+	// Default to idle when no release has been installed yet
+	// or during local mode
```
pipex, line 514: This makes that endpoints behave differently on local mode vs in production, that is not ideal as it means you cannot rely on these endpoints during development.

test/integration/device-api/actions.spec.ts
```diff
@@ -96,7 +96,7 @@ describe('manages application lifecycle', () => {
 		return {
 			[appId]: {
 				name: 'localapp',
-				commit: 'localcommit',
+				commit: '10ca12e1ea5e',
```
pipex, line 99: TODO: we need to change the CLI to provide a valid commit

pipex:
@cywang117 remember to take a look at this when you get a chance

src/device-api/v2.ts
```diff
@@ -424,25 +424,34 @@ router.get('/v2/containerId', async (req: AuthorizedRequest, res) => {
...
 				'createdAt',
 			);
+
+			// QUESTION: what to do if we cannot find the image?
+			// if the image database got deleted
+			s.imageId = allImages.find(
+				(img) => img.commit === svc.commit && img.serviceName === s.serviceName,
+			)?.imageId;
+
+			return s;
```
cywang117, line 449: I don't think we need to worry about this, if the image table got deleted unexpectedly, then the action would have thrown a lot earlier during `images.getState`

cywang117 (commented):
LGTM, great work!

src/device-api/v2.ts
pipex: I don't think `images.getState` would throw, but we recreate the image database when initializing application manager so if the data from `getState` should be correct, so you are right the comment is unnecessary

pipex:
Will need to refactor this, I had not considered that, because of the git workflow, in a multi-app world you could have two apps pointing to the same release commit. Will probably need to keep the releaseId

Issue #2133, opened 2023-02-16
Do not fail pinDevice if the release doesn't exist or is not accessible


pipex:
The supervisor currently fails and keeps retrying to pin a device if the [release query fails](https://github.com/balena-os/balena-supervisor/blob/935a4fba59c82df916311cafe28b9a04da922809/src/api-binder/index.ts#L291-L309). This can be a problem if a preloaded device is also pinned to a release that gets deleted later. This will mean that the device will show as registered on the dashboard and will probably still have VPN access but will show as VPN only. 

A deleted release means that the supervisor should assume that device should default to track latest.

Related to: balena-os/balena-supervisor#1752

PR #2139, opened 2023-03-09
Skip pin device step if release was deleted

pipex:
Preloaded devices can require that the device is pinned to the preloaded release on provisioning. However if the provisioned release gets released in the future, that would lead to the device remaining in "VPN only" state forever as the provisioning process could not finish due to pinning failure.

This commit changes the behavior so if the release does not exist, the pinning step is skipped and the device follows the fleet pinning state.

Closes: #2133
Change-type: patch


cywang117 (approved):
LGTM!

PR #2140, opened 2023-03-09
Use log endpoint subdomain if it exists in config.json

cywang117:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345890-balena-io/topic/cloudflare.20proxy.20.26.20logs


pipex:
I don't have any observations. Merge when ready @cywang117

cywang117:
@balena-ci I self-certify!

PR #2141, opened 2023-03-10
Use multi-arch in dockerfile

pipex:
This is necessary since the builder no longer passes the platform flag
to the build. This would lead to dockerfiles that are mixing multi and single
arch stages to pull the wrong architecture images, particularly when
trying to build images in emulated builds (e.g. armv7hf built on aarch64).

Moving the full build to multi-arch solves this as the docker engine is
capable of chosing the right architecture from the manifest.

Relatest-to: https://github.com/balena-io/balena-builder/pull/1010
Change-type: patch

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,43 +1,37 @@
 ARG ARCH=%%BALENA_ARCH%%
 ARG FATRW_VERSION=0.2.9
-
-
-FROM arm32v6/alpine:3.16 as alpine-rpi
-FROM arm32v7/alpine:3.16 as alpine-armv7hf
-FROM arm64v8/alpine:3.16 as alpine-aarch64
-FROM amd64/alpine:3.16 as alpine-amd64
-FROM i386/alpine:3.16 as alpine-i386
-
-FROM arm32v6/alpine:3.11 as procmail-rpi
-FROM arm32v7/alpine:3.11 as procmail-armv7hf
-FROM arm64v8/alpine:3.11 as procmail-aarch64
-FROM amd64/alpine:3.11 as procmail-amd64
-FROM i386/alpine:3.11 as procmail-i386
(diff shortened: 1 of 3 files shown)


pipex:
Testing first to see the resulting image size

Dockerfile.template
```diff
@@ -1,39 +1,33 @@
...
 
-RUN apk add --no-cache \
+RUN apk add --update --no-cache \
 	g++ \
 	make \
 	python3 \
-	libgcc \
+	curl \
+	'nodejs<18' \
+	'npm<9' \
```
cywang117, line 22: Should we define the Node / NPM version at the top of the Dockerfile as an ARG for easy updating? I see it's referenced in multiple places.

cywang117: Otherwise, LGTM. Good work!

klutchell: 👍

pipex: Good idea, I'll make the change

Dockerfile.template
```diff
@@ -51,7 +45,7 @@ RUN npm ci --build-from-source --sqlite=/usr/lib
 # musl. We hack around this by copying the binary and its library
 # dependencies to the final image
 ###################################################################
-FROM balenalib/${ARCH}-debian:bullseye-run as journal
+FROM debian:bullseye-slim as journal
```
klutchell, line 48: Love a slim image :)

pipex:
With the improvements that Kyle did to the builder, and the bump of the CLI on the deploy-to-balena action (see https://github.com/balena-io/deploy-to-balena-action/pull/251) this PR is now generating the right architecture images

cc @cywang117 @klutchell

klutchell:
LGTM!

PR #2142, opened 2023-03-10
Use single-arch in dockerfile

pipex:
This is necessary since the builder no longer passes the platform flag
to the build. This would lead to dockerfiles that are mixing multi and single
arch stages to pull the wrong architecture images, particularly when
trying to build images in emulated builds (e.g. armv7hf built on aarch64).

Moving the full build to single-arch solves this as the docker engine is
capable of chosing the right architecture from the manifest. Once some
of the builder issues are fixed, we should move to #2141

Relates-to: balena-io/balena-builder#1010
Change-type: patch

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,10 +1,18 @@
 ARG ARCH=%%BALENA_ARCH%%
-
-# Used by livepush to support multi arch images in older
-# balenaOS with buggy platform support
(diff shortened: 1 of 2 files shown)


pipex:
Closing in favor of #2141 which leads to smaller image sizes

cywang117 (commented):
LGTM! Great work as usual 💪🏼

klutchell (approved):

PR #2146, opened 2023-03-23
Remove anonymous build volume from Dockerfile

cywang117:
We don't need this anonymous volume as /data is bind mounted into the container from host (legacy), and will soon be mounted by the Supervisor itself on startup.

Change-type: patch


pipex:
We should get rid of the tmpfs in docker compose as well as the `docker volume rm` in the `compose:down` script (this latter was failing more often than not)

cywang117:
Ahhh I see what you mean. That's my bad for assuming the directory name of the repo. I also just realized that you can specify `--volumes` when running `docker compose down`, so we can just use that as it's much simpler.

cywang117:
I messed this up with a rebase typo, I'm just going to close this and push the changes to another PR..

PR #2147, opened 2023-03-24
Revert "Use multi-arch in dockerfile"

pipex:
This reverts commit 00e389e5f559dd10e49cfa411784b89498c3c0eb.

Images generated using this dockerfile still don't have the right architecture. More testing is needed

Change-type: patch


cywang117 (approved):
LGTM

PR #2153, opened 2023-04-06
Refactor state engine to be able to use current state as target

pipex:
This PR contains multiple changes to allow to simplify the interface for the state engine.

As an objective, every operation that performs a change to the device or runtime application should use the target state to do it. For instance, stopping service X is equivalent to taking the current state and passing it as a target to the engine changing the desired state of service X to `running: false`

This was not possible before, because the current state was substantially different from the target state as specific information (like image names) was not available. 

This PR makes the current state returned by `deviceState.getCurrentState()` usable as a target in `applyIntermediateTarget` (this method will probably be renamed, but that's a different issue), and modifies the `doRestart` and `doPurge` actions to use the new interface.

This PR also makes some changes to application manager to make it a bit more agnostic about local mode so it works more closely to the way cloud mode does. This makes it easier to integration test (as integration tests rely on local mode)

Change-type: patch

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -40,7 +40,6 @@ export interface AppConstructOpts {
 }
 
 export interface UpdateState {
-	localMode: boolean;
 	availableImages: Image[];
 	containerIds: Dictionary<string>;
 	downloading: string[];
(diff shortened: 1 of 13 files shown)


cywang117:
You're getting some TypeScript errors that are failing the GH checks @pipex, so I'll wait to look at this.

pipex:
Thanks @cywang117, I missed that, all fixed now

cywang117 (commented):
Great work @pipex . I didn't notice anything out of place and I appreciated the explanations in the commits to better understand the changes. Assuming this has been tested on a device, LGTM!

pipex:
Found a tiny bug while testing other stuff, I'll make the change before merging

pipex:
I self certify!

PR #2155, opened 2023-04-07
backends: Add Jetson Orin NX custom device-tree support

acostach:
Change-type: patch

# Description

This PR enables custom device-tree loading support for the Jetson Orin NX device-type

# How Has This Been Tested?

This has been tested on the Orin NX device by setting the custom_fdt_file variable in /mnt/boot/extra_uEnv.txt , same as the SV sets it for the rest of the Jetson modules which support this feature.


acostach:
It's not clear to me from the Flowzone logs if it's something in this commit that's causing the Publish balena/Publish custom checks to fail

cywang117:
@acostach Sometimes the Supervisor integration tests time out in the builder, we've discussed looking at this but it's not clear the root cause. Usually re-running the failed checks causes them to pass 👍🏼

acostach:
ack, thank you @cywang117 , otherwise the PR good to go?

cywang117 (approved):
LGTM!

Issue #1576, opened 2021-01-19
A network config change in the same release as an app update can make Supervisor erroneously report device state as applied when it's not applied


20k-ultra:
I have seen instances where a container is shown on the dashboard and in a "Downloaded" state not the desired "Running" or even "Stopped".... No matter how many times the Supervisor checks the target state even after restarting, the Supervisor won't run the image that is downloaded.

When you try to manually start it with POST /v2/applications/:app-id/start-service from the dashboard (click the Play icon) a 404 is logged. That means the Supervisor's database has somehow lost track of this container or image. That depends on what value is used in the body payload since you can start an Image by serviceId or ImageId.

cywang117:
I made an easy reproduction repo here which can be used for investigating: https://github.com/balena-io-experimental/network-ipam-config-reproduction

louisburton:
If we hit this issue, is there any way to manually intervene to get the service started? A workaround that doesn't involve release switching?

cywang117:
@louisburton, I'm not quite sure. What happens if you stop & remove the containers that are on the incorrect network, restart the Supervisor?

I took a look at this issue before the holidays and ran into some confusion with it, but no worries, it's still on my list.

louisburton:
I don't believe that worked. I will double check it next time it occurs though 👍 we have some offline devices which might bunny hop the network upgrade release.

cywang117:
This should be fixed with [v14.10.8](https://github.com/balena-os/balena-supervisor/blob/master/CHANGELOG.md#v14108) cc @louisburton

PR #2161, opened 2023-04-19
Fix device state not applied when a network change happens during the update

pipex:
This fixes the issue #1576, adding tests to avoid regressions and also fixes a bug that was preventing devices from recovering from this issue. 

Relates-to: #1576

### DIFF
--- a/test/integration/state-engine.spec.ts
+++ b/test/integration/state-engine.spec.ts
@@ -134,6 +134,118 @@ describe('state engine', () => {
 		]);
 	});
 
+	// This test recovery from issue #1576, where a device running a service from the target release
+	// would not stop the service even if there were still network and container changes to be applied
+	it('always stops running services depending on a network being changed', async () => {
+		// Install part of the target release
+		await setTargetState({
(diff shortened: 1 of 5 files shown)


test/integration/state-engine.spec.ts
```diff
@@ -134,6 +134,116 @@ describe('state engine', () => {
 		]);
 	});
 
+	it('recovers from a device already running one of the target release services', async () => {
```
cywang117, line 137: Without the context of issue 1576, this unit test description is a bit vague. Perhaps could add a reference to the issue in a comment in the code?

cywang117 (commented):
Great work @pipex . LGTM

test/integration/state-engine.spec.ts
pipex: Added the comment and improved the test description (i think). Let me know what you think

cywang117: @pipex The change looks good!

pipex:
I self-certify!

Issue #2160, opened 2023-04-17
`PATCH /v1/device/host-config` should check for the hostname before changing it


lmbarros:
We had a user on support that apparently was unconditionally using the Supervisor API ([`PATCH /v1/device/host-config`](https://docs.balena.io/reference/supervisor/supervisor-api/#patch-v1devicehost-config)) to change the hostname right when their service starts. This causes the device to get into a loop (rename host, restart Engine, restart service, rename host...). This can be fixed by first using the GET endpoint to check if the hostname isn't already what they want.

But the Supervisor itself could do this check and become a no-op if the requested hostname is already the used one.

pipex:
Below is the logic that handles this. In a newer OS, the supervisor will never enter the `if` clause to restart the service. 

https://github.com/balena-os/balena-supervisor/blob/master/src/host-config.ts#L188-L203

This means the write will only happen if `config.set` is changing the value from what the supervisor knows (here is the logic of configJson handling). 
 
https://github.com/balena-os/balena-supervisor/blob/935a4fba59c82df916311cafe28b9a04da922809/src/config/configJson.ts#L58-L60

There is something else happening in the attached ticket that is not related to not checking. Somehow the hostname is not being reflected inside the containers.

pipex:
The OS removed `config-json.target` in v2.91.0 in favor of `config-json.service`, which means the supervisor is reverting to calling `balena-hostname`, which most likely is the cause of the issue. We are trying to figure out what is the best approach here

pipex:
This issue has been fixed in #2162 by removing the call to the balena-hostname service

PR #2162, opened 2023-04-20
Do not restart balena-hostname on rename

pipex:
The OS since v2.82.6 will monitor changes to config.json and restart the relevant services to apply the changes. There is no need to trigger restart of the services via the supervisor. Users on older OS versions will need to update their OS or restart the services manually as OS loses support after 2y.

Change-type: patch
Closes: #2160

### DIFF
--- a/src/host-config.ts
+++ b/src/host-config.ts
@@ -185,22 +185,10 @@ async function readHostname() {
 }
 
 async function setHostname(val: string) {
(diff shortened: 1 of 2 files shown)


klutchell:
Don't some test cases need to be updated as well?

pipex:
@klutchell good point, I had forgot about that

PR #2164, opened 2023-04-27
Run test supervisor under a different service name

klutchell:
The docker compose V2 spec no longer accepts `network_mode: bridge`,
which means we can no longer override the network configuration of
the `balena-supervisor` service for tests.

For this reason we now create a separate service to run the built
supervisor `balena-supervisor-sut` and run API tests against this
service instead of the default `balena-supervisor`.

Change-type: patch
See: https://github.com/balena-os/balena-supervisor/actions/runs/4820624634/jobs/8585407479?pr=2149


pipex:
> network_mode and networks are mutually exclusive settings

I don't see anything about this in docker-compose documentation. I think github runners are using an old version of docker-compose. This looks ok as a workaround though

docker-compose.test.yml
```diff
@@ -2,17 +2,20 @@ version: '2.3'
 
 services:
   # Override the supervisor service configuration. The tests will
-  # be run through the `sut` service
+  # be run through the `sut` service and the running superviisor will be
+  # tested through the `balena-supervisor-sut` service
   balena-supervisor:
+    command: sleep infinity
```
klutchell, line 8: We don't need to override the entrypoint for this service. As long as nothing requires it, it will not be started.

This is because Flowzone runs `docker compose up sut --exit-code-from sut` and not `docker compose up --exit-code-from sut`.

https://github.com/product-os/flowzone/blob/854998ed291f416958a6eeb9b6a21306eb5bbd66/flowzone.yml#L2037

This is the same command run by DockerHub so we are following their process.

pipex: Fixed @klutchell

klutchell: Looks good, I think we can squash the commits though, since mine didn't fix anything.

klutchell: Other than that LGTM

PR #2165, opened 2023-04-28
mount-partitions.sh: Add support for encrypted partitions

mtoman:
After a recent change enforcing all the partitions to be on the same block device, encrypted partitions are no longer being detected correctly. This is because the assumption that the parent block device is a substring of the actually mounted block device does not work for LUKS devices - the mount will either be `/dev/mapper/luks-XXX` or `/dev/dm-X` while the parent device is still e.g. `/dev/sda`.

The usual balenaOS boot partition is also split in two - `boot` and `efi`. The boot partition (mounted under `/mnt/boot`) is encrypted and the efi partition (mounted under `/mnt/efi`) is not.

This patch generalizes the detection of the parent device so that it works with both encrypted and unencrypted partitions.

### DIFF
--- a/mount-partitions.sh
+++ b/mount-partitions.sh
@@ -9,25 +9,42 @@ export ROOT_MOUNTPOINT="/mnt/root"
 # Set DBus system bus address for getting the current boot block device
 export DBUS_SYSTEM_BUS_ADDRESS="${DBUS_SYSTEM_BUS_ADDRESS:-unix:path="${ROOT_MOUNTPOINT}"/run/dbus/system_bus_socket}"
 
+# Get the block device from systemd
+# The dbus-send command below should return something like:
+# ```
+# method return time=1680132905.878117 sender=:1.0 -> destination=:1.20155 serial=245193 reply_serial=2
+#   variant       string "/dev/sda1"
+# ```
+# Usage: dbus_get_mount PARTITION
+# Partition is only the label, e.g. boot, state, data
+dbus_get_mount() {
+    part="$1"
+
+    result=$(dbus-send --system --print-reply \
+        --dest=org.freedesktop.systemd1 /org/freedesktop/systemd1/unit/mnt_2d${part}_2emount org.freedesktop.DBus.Properties.Get \
+        string:"org.freedesktop.systemd1.Mount" string:"What" | grep "string" | cut -d'"' -f2 2>&1)
+    # If the output doesn't match the /dev/* device regex, exit with an error
+    if [ "$(echo "${result}" | grep -E '^/dev/')" = "" ]; then
+        echo "ERROR: Could not determine ${part} device from dbus. Please launch Supervisor as a privileged container with DBus socket access."
+        exit 1
+    fi
+
+    echo "${result}"
+}
+
 # Get the current boot block device in case there are duplicate partition labels
 # for `(balena|resin)-(boot|state|data)` found.
 current_boot_block_device=""
 if [ "${TEST}" != 1 ]; then
-    # Get the current boot block device from systemd
-    # The dbus-send command below should return something like:
-    # ```
-    # method return time=1680132905.878117 sender=:1.0 -> destination=:1.20155 serial=245193 reply_serial=2
-    #   variant       string "/dev/sda1"
-    # ```
-    mnt_boot_mount=$(dbus-send --system --print-reply \
-        --dest=org.freedesktop.systemd1 /org/freedesktop/systemd1/unit/mnt_2dboot_2emount org.freedesktop.DBus.Properties.Get \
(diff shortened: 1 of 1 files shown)


mount-partitions.sh
```diff
@@ -106,6 +120,13 @@ DATA_MOUNTPOINT="/mnt/data"
 setup_then_mount "data" "${DATA_MOUNTPOINT}"
 export DATA_MOUNTPOINT
 
+# Set efi mountpoint if efi partition exists
+if [ "${has_efi}" = "1" ]; then
+    EFI_MOUNTPOINT="/mnt/efi"
+    setup_then_mount "efi" "${EFI_MOUNTPOINT}"
+    export EFI_MOUNTPOINT
+fi
```
cywang117, line 128: The exported vars in this file are for use in the Supervisor to find file paths. I'm not quite sure I understand what files / partitions are found in the EFI partition - are they the boot partition files that aren't encrypted?

mtoman: Yes, exactly. The boot partition on non-encrypted devices contains a combination of files necessary to actually boot the device and metadata that balenaOS uses, e.g. `config.json`, `system-connections` etc. On encrypted devices we had to split the two use-cases as we want to encrypt/protect things that contain sensitive information such as API keys in `config.json` or wifi passwords in `system-connections`. At the same time we need to keep the files necessary to boot the device (in general GRUB + encryption metadata) unencrypted.

On non-encrypted devices all the files are in the `boot` partition, mounted under `/mnt/boot`. On encrypted devices, the two partitions are mounted under `/mnt/boot` and `/mnt/efi` respectively. `/mnt/boot/EFI` is a symlink pointing to `/mnt/efi` instead of a regular directory. With that little trick the change is pretty much transparent as all the file paths are the same, encrypted or not.

I've only added the export to keep inline with the other partitions but if you think it is unnecessary I can remove it.

mtoman: For a better idea, this is what the output of `lsblk` looks like on an encrypted system:
```
# lsblk
NAME                                          MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
sda                                             8:0    0  83.8G  0 disk  
|-sda1                                          8:1    0    40M  0 part  /mnt/efi
|-sda2                                          8:2    0    42M  0 part  
| `-luks-a10745a8-24d1-46f8-b60b-6a4e3060999d 250:0    0    40M  0 crypt /mnt/boot
|-sda3                                          8:3    0  1002M  0 part  
| `-luks-b1179996-173b-4087-8667-0f6877c644c9 250:1    0  1000M  0 crypt /mnt/sysroot/active
|-sda4                                          8:4    0  1002M  0 part  
| `-luks-5f61b7ed-6c40-4085-b06a-8df02d2a0311 250:2    0  1000M  0 crypt /mnt/sysroot/inactive
|-sda5                                          8:5    0    22M  0 part  
| `-luks-463fdafb-1257-4c8f-80fd-a25faa10ec10 250:3    0    20M  0 crypt /mnt/state
...
`-sda6                                          8:6    0  81.8G  0 part  
  `-luks-1bb29fe3-130b-4abe-b0e6-bce1df5d946a 250:4    0  81.8G  0 crypt /mnt/data
...
```

cywang117: Thanks for clarifying @mtoman . Besides `/mnt/boot/config.json` and system-connections files, are there any other files in the encrypted part of the boot partition? As the non-encrypted files would be a different path (`/mnt/efi`) compared to the encrypted files, there would need to be another commit which allows the Supervisor to access those files under `/mnt/efi` as backup if the file doesn't exist under `/mnt/boot`. `device-type.json` comes to mind. We probably also want to add a test case for this. I can assist if you don't wish to write TypeScript.

mtoman: There is only GRUB, kernel and the encryption metadata:
```
root@f986c5c:/mnt/efi# find
.
./EFI
./EFI/BOOT
./EFI/BOOT/grub.cfg
./EFI/BOOT/grub.cfg.sig
./EFI/BOOT/grubenv
./EFI/BOOT/grub_extraenv
./EFI/BOOT/bootx64.efi
./balena-luks.ctx
./balena-luks.enc
./bzImage
./bzImage.sig
```

I can share support access to an encrypted device if it helps.

cywang117:
Since the necessary files are still under `/mnt/boot`, the only testing that could be done is with mount-partitions.sh, which we skip in integration tests due to running in a containerized environment. Therefore with the current test setup, there isn't a meaningful Supervisor test that can be added for this. Let's wait to see if @pipex has any comments, but otherwise LGTM!

pipex (approved):
LGTM

PR #2167, opened 2023-05-09
Add `arch.sw` to supported container requirements

pipex:
This allows a container contract in a release to include `arch.sw` as a requirement for running the container.

Change-type: minor

### DIFF
--- a/test/integration/device-state.spec.ts
+++ b/test/integration/device-state.spec.ts
@@ -25,6 +25,7 @@ describe('device-state', () => {
 		initializeContractRequirements({
 			supervisorVersion: '11.0.0',
 			deviceType: 'intel-nuc',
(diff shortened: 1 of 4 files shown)


klutchell:
lgtm

test/integration/device-state.spec.ts
```diff
@@ -406,8 +407,8 @@ describe('device-state', () => {
 												name: 'Enforce supervisor version',
 												requires: [
 													{
-														type: 'sw.supervisor',
-														version: '>=12.0.0',
+														type: 'arch.sw',
+														version: 'armv7hf',
```
klutchell, line 411: This should be `name: 'armv7hf'`  not `version:`?

PR #2170, opened 2023-05-16
Handle Engine-host race condition

cywang117:
There exists a race condition between Engine and a host resource that may not be immediately created. In this race condition, if a container's compose config depends on the existence of that host resource, such as a network interface, and the Engine tries to create & start the container before the host resource is created, the Engine will not reattempt to start the container, regardless of the restart policy. This is undesireable behavior but seems to be the behavior as implemented by Docker.

To rectify this, the Supervisor state funnel noops for a grace period of 1 minute after starting a container to see that the container's status has become 'running'. If the container exits because of the race condition, the status becomes 'exited' and the Supervisor will attempt to generate another start step. As long as the createdAt date of the container is less than a minute ago, the Supervisor will noop to keep the state funnel alive. When the createdAt date is longer than 1 minute, the Supervisor will issue a start step. (NOTE: this will not work if the container briefly enters a `Starting` state that the Supervisor catches, as the Supervisor will exit the state funnel upon reading that `Starting` state  -- it's unclear if this occurs because it's very hard to catch this state, but I don't believe the `Starting` state is ever reached if this race condition is present.)

If the container is never able to start, there was a problem in the host in the creation of the host resource, and that should be fixed at the host level.

Change-type: patch

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -4,7 +4,6 @@ import { promises as fs } from 'fs';
 import Network from './network';
 import Volume from './volume';
 import Service from './service';
-
 import * as imageManager from './images';
 import type { Image } from './images';
 import {
@@ -12,6 +11,7 @@ import {
 	generateStep,
 	CompositionStepAction,
 } from './composition-steps';
+import { isOlderThan } from './utils';
 import * as targetStateCache from '../device-state/target-state-cache';
 import * as dockerUtils from '../lib/docker-utils';
 import * as constants from '../lib/constants';
@@ -25,6 +25,8 @@ import { ServiceComposeConfig, DeviceMetadata } from './types/service';
 import { ImageInspectInfo } from 'dockerode';
 import { pathExistsOnRoot } from '../lib/host-utils';
 import { isSupervisor } from '../lib/supervisor-metadata';
+
+const SECONDS_TO_WAIT_FOR_START = 60;
 
 export interface AppConstructOpts {
 	appId: number;
@@ -184,7 +186,6 @@ export class App {
 				}),
 			);
 		}
-
 		return steps;
 	}
 
@@ -377,9 +378,22 @@ export class App {
 
 			// Only start a Service if we have never started it before and the service matches target!
 			// This is so the engine can handle the restart policy configured for the container.
+			//
+			// However, there is a certain race condition where the container's compose depends on a host
+			// resource that may not be there when the Engine starts the container, such as a port binding
+			// of 192.168.88.1:3000:3000, where 192.168.88.1 is a user-defined interface configured in system-connections
+			// and created by the host. This interface creation may not occur before the container creation.
+			// In this case, the container is created and never started, and the Engine does not attempt to restart it
+			// regardless of restart policy.
+			const requiresExplicitStart =
+				['always', 'unless-stopped'].includes(serviceTarget.config.restart) &&
+				serviceCurrent.status === 'exited' &&
+				serviceCurrent.createdAt != null &&
+				isOlderThan(serviceCurrent.createdAt, SECONDS_TO_WAIT_FOR_START);
 			return (
 				(serviceCurrent.status === 'Installing' ||
-					serviceCurrent.status === 'Installed') &&
+					serviceCurrent.status === 'Installed' ||
+					requiresExplicitStart) &&
 				isEqualExceptForRunningState(serviceCurrent, serviceTarget)
 			);
 		};
@@ -398,6 +412,26 @@ export class App {
 			return (
 				serviceTarget.config.running === false &&
 				serviceCurrent.status !== 'Stopped'
+			);
+		};
+
+		/**
+		 * Under the race condition of a container depending on a host resource that may not be there when the Engine
+		 * starts up, resulting in the Engine not restarting the container regardless of restart policy, the Supervisor
+		 * state loop needs to stay alive so that the Supervisor may issue a start step after a wait.
+		 * @param serviceCurrent
+		 */
+		const shouldWaitForStart = (
+			serviceCurrent: Service,
+			serviceTarget: Service,
+		) => {
+			return (
+				['always', 'unless-stopped'].includes(serviceTarget.config.restart) &&
+				serviceCurrent.config.running === false &&
+				serviceTarget.config.running === true &&
+				serviceCurrent.status === 'exited' &&
+				serviceCurrent.createdAt != null &&
+				!isOlderThan(serviceCurrent.createdAt, SECONDS_TO_WAIT_FOR_START)
 			);
 		};
 
@@ -425,7 +459,8 @@ export class App {
 					!isEqualExceptForRunningState(c, t) ||
 					shouldBeStarted(c, t) ||
 					shouldBeStopped(c, t) ||
-					shouldWaitForStop(c),
+					shouldWaitForStop(c) ||
+					shouldWaitForStart(c, t),
 			);
 
 		return {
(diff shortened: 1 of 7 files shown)


cywang117:
TODO: Handle the case of restart policy: 'no' (oneshot containers that have suffered from this race condition) -- this is a bit tougher to handle without resorting  to parsing the error message (which is the iconic "external connectivity" Engine error)

cywang117:
I will be self-certifying this as @pipex is on leave. Here are the details of how I tested the fix. 

First, I found a way to reproduce the condition locally for testing/development. (While we have access to user devices where this issue is manifesting, that is difficult to manage due to devices not having guaranteed uptime, and being remote as opposed to local.) I did this by creating a dummy NetworkManager interface:
```
root@nuccc:~# cat /mnt/boot/system-connections/localhost-routing 
[connection]
id=localhost-routing
uuid=de288a6a-4b53-4772-8b2a-319208f29528
type=ethernet
autoconnect=false

[ipv4]
address=192.168.88.1/24
method=manual

[ipv6]
method=ignore

[proxy]
method=none

[route-metric-1]
dst=0.0.0.0/0
gateway=127.0.0.1
metric=1
table=1000
```
This routes all requests from 192.168.88.1 to 127.0.0.1. Notice the `autoconnect=false`, indicating the interface will be created but not available until it's brought up manually or by a script...

So, I created a script to bring up the interface manually after a delay:
```
root@nuccc:~# cat /usr/src/delayed-connection.sh 
#!/bin/bash
echo 'Sleeping 120 before starting localhost-routing' | systemd-cat -t delayed-connection -p info
sleep 120
echo 'Starting localhost-routing connection' | systemd-cat -t delayed-connection -p info
nmcli c up localhost-routing
```
and added a systemd service to execute this script when starting up NetworkManager:
```
root@nuccc:~# cat /etc/systemd/system/delayed-connection.service 
[Unit]
Description=Delayed NetworkManager connection for testing Engine-host test
After=network.target
 
[Service]
ExecStart=/usr/src/delayed-connection.sh
 
[Install]
WantedBy=default.target
```
and finally enabled the service with `systemctl enable delayed-connection.service`.

What the above achieves is a dummy network interface that comes up 2 minutes after NetworkManager starts. This should leave enough room for the race condition to manifest. I then pushed a release with the following docker-compose.yml:
```
version: '2.4'
services:
  one:
    build:
      context: ./
    ports:
      - '192.168.88.1:8081:8081'
    restart: unless-stopped
    environment:
      PORT: '8081'
      SERVICE: one
  two:
    build:
      context: ./
    ports:
      - '192.168.88.1:8082:8082'
    restart: always
    environment:
      PORT: '8082'
      SERVICE: two
  four:
    build:
      context: ./
    ports:
      - '192.168.88.1:8084:8084'
    restart: 'no'
    environment:
      PORT: '8084'
      SERVICE: four
  three:
    build:
      context: ./
    ports:
      - '192.168.88.1:8083:8083'
    restart: on-failure
    environment:
      PORT: '8083'
      SERVICE: three
```
Because all four services use the same script, the environment variables avoid port conflicts and provide info when logging during startup.

Now, we'll be able to reproduce the conditions by rebooting the device. When starting the device up with the current production Supervisor, I observed that the Supervisor issued the start steps and exited the state funnel immediately even though the services never started. With the draft Supervisor (which I loaded into `/etc/balena-supervisor/supervisor.conf` so that the device would boot up with the draft Supervisor), I observed that the Supervisor would keep trying to start the services until the network interface came up.

This PR also adds some test cases to check that the proper steps are generated for the race condition. Notably, `nextStepsForAppupdate` becomes `async` due to needing to query the `State.ExitCode` of services with restart policy `on-failure` to determine whether they need to be started. This has been tested and everything still works.

cywang117:
@balena-ci I self-certify!

src/compose/app.ts
```diff
@@ -631,11 +652,56 @@ export class App {
...
+			target.config.running !== true
+		) {
+			return false;
+		}
+		let conditionsMetWithRestartPolicy = ['always', 'unless-stopped'].includes(
+			target.config.restart,
+		);
+
+		if (current.containerId != null && !conditionsMetWithRestartPolicy) {
+			const containerInspect = await inspectByDockerContainerId(
```
pipex, line 676: @cywang117 We can further improve this code to use the data in the `Service` object. This probably will mean adding some additional metadata to the service class, but that is an acceptable trade-off imho. The `App` methods were previously pure functions which had a lot of benefits and adding an `async` query at this point may have undesired effects, that in addition that this can make testing harder.

src/compose/app.ts
```diff
@@ -631,11 +652,56 @@ export class App {
...
+	): Promise<boolean> {
+		// Shortcut the Engine inspect queries if status isn't exited
+		if (
+			current.status !== 'exited' ||
+			current.config.running !== false ||
+			target.config.running !== true
+		) {
+			return false;
+		}
+		let conditionsMetWithRestartPolicy = ['always', 'unless-stopped'].includes(
```
pipex, line 671: Should `on-failure` be here too?

src/compose/app.ts
```diff
@@ -631,11 +652,56 @@ export class App {
...
+				current.containerId,
+			);
+			// If the container has previously been started but exited unsuccessfully, it needs to be started.
+			// If the container has a restart policy of 'no' and has never been started, it
+			// should also be started, however, in that case, the status of the container will
+			// be 'Installed' and the state funnel will already be trying to start it until success,
+			// so we don't need to add any additional handling.
+			if (
+				target.config.restart === 'on-failure' &&
+				containerInspect.State.ExitCode !== 0
```
pipex, line 686: To add to the point above, if we add `Status.ExitCode` to the `Service` class we can avoid using `async` in this method (and the call chain)

src/compose/utils.ts
```diff
@@ -693,3 +693,10 @@ export function dockerMountToServiceMount(
 
 	return mount as LongDefinition;
 }
+
+export function isOlderThan(currentTime: Date | null, seconds: number) {
```
pipex, line 697: I think this should be called something like `isValidDateAndOlderThan` since it does more than just compare dates.

test/integration/compose/application-manager.spec.ts
```diff
@@ -1417,4 +1417,349 @@ describe('compose/application-manager', () => {
...
+		const date61SecondsAgo = new Date();
+		date61SecondsAgo.setSeconds(date61SecondsAgo.getSeconds() - 61);
+
+		// Time 59 seconds ago
+		const date50SecondsAgo = new Date();
+		date50SecondsAgo.setSeconds(date50SecondsAgo.getSeconds() - 50);
+
+		// TODO: We need to be able to start a service with restart policy "no" if that service did not start at all due to
+		// the host resource race condition described above. However, this is harder to parse as the containers do not include
+		// the proper metadata for this. The last resort would be parsing the error message that caused the container to exit.
```
cywang117, line 1437: This is an outdated comment as the case of restart policy 'no' is already handled by the state funnel without any changes added by this PR. If a service with policy 'no' has never been started, its status is `Installed` which the Supervisor will issue a start step for. Comment to be removed in a future PR.

test/integration/compose/application-manager.spec.ts
```diff
@@ -1417,4 +1417,349 @@ describe('compose/application-manager', () => {
...
+		it('should not infer any steps for a service with a status of "exited" if restart policy is "no" or "on-failure"', async () => {
```
cywang117, line 1438: This test case needs to be updated for the case of `on-failure` with the `ExitCode` metadata changes

test/integration/compose/application-manager.spec.ts
```diff
@@ -1417,4 +1417,349 @@ describe('compose/application-manager', () => {
...
+			expect(noopStep2).to.have.property('action').that.equals('noop');
+
+			expect(nextSteps).to.have.lengthOf(0);
+		});
+
+		it('should infer a start step for a service that was created >1 min ago with status of "exited" if restart policy is "always" or "unless-stopped"', async () => {
+			// Conditions:
+			// - restart: "always" || "unless-stopped"
+			// - status: "exited"
+			// - createdAt: <= SECONDS_TO_WAIT_FOR_START ago
```
cywang117, line 1675: >SECONDS_TO_WAIT_FOR_START ago

Issue #2174, opened 2023-06-14
`/v2/applications/state` endpoint returns stale information


pipex:
This endpoint should return the state of the target release of the application, including download progress of target images for the running app, however, it currently is returning the state of the current release, meaning download progress is never updated.

Replication
- Update app to a new release wait until progress is getting reported
- Query `/v2/applications/state` 
- See `downloadProgress` remain at null

PR #2175, opened 2023-06-14
Fix /v2/applications/state endpoint

pipex:
It was returning stale information, particularly the download progress of the target release images never got updated.

Change-type: patch
Closes: #2174


cywang117 (commented):
LGTM, merge after you're done with testing!

pipex:
I self-certify!

PR #2177, opened 2023-06-16
Specify fs type when mounting partitions to prevent "Can't open blockdev" warnings

cywang117:
Tests with patch (run from within Supervisor container):
```
/usr/src/app # findmnt -n -o FSTYPE --first-only -- /mnt/boot
vfat
/usr/src/app # findmnt -n -o FSTYPE --first-only -- /mnt/data
ext4
/usr/src/app # findmnt -n -o FSTYPE --first-only -- /mnt/state
ext4
```

Change-type: patch


acostach:
lgtm

Issue #2178, opened 2023-06-20
Services are started by the Supervisor even after manual stop


cywang117:
Following changes in https://github.com/balena-os/balena-supervisor/pull/2170, the Supervisor will start services which have previously been manually stopped via Supervisor API or directly through the Engine. This is a bug and an oversight made during #2170.

Steps to reproduce:
1. Push a release with restart: always:
```
version: '2.4'

services:
  always:
    restart: 'always'
    image: alpine
    command: sleep infinity
```
2. Manually stop the service using the dashboard.
3. Trigger a manual Supervisor update check using /v1/update.

Expected behavior: service should remain in `exited` status.
Actual behavior: service starts back up.

cywang117:
After thorough discussion, we've unavoidably settled on parsing the exit message of the container, as the current implementation was too general and based on the mistaken assumption that a service status of `exited` is differentiated from a service status of `Stopped`. The former is an official service status acquirable through the container's inspect, while the latter is an internal Supervisor-only status which isn't currently set anywhere (this is another bug).

cywang117:
Closed by #2179, available with Supervisor v14.11.8

PR #2179, opened 2023-06-20
Parse container exit error message instead of status to fix Engine-host race

cywang117:
The previous implementation in #2170 of parsing the container status was too general, because it relied on the mistaken assumption that a container would have a status of `Stopped` if it was manually stopped. This turned out to be untrue, as manually stopped containers were also getting restarted by the Supervisor due to their inspect status of `exited`. With this, parsing the exit message became unavoidable as there are no other clear ways to discern a container that has been manually stopped and shouldn't be started from a container experiencing the Engine-host race condition issue (again, see #2170).

Since we're just parsing the exit error message, we don't need to worry about different behaviors amongst restart policies, as any container with the error message on exit should be started.

Change-type: patch
Closes: #2178

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -12,7 +12,6 @@ import {
 	generateStep,
 	CompositionStepAction,
 } from './composition-steps';
-import { isValidDateAndOlderThan } from './utils';
 import * as targetStateCache from '../device-state/target-state-cache';
 import { getNetworkGateway } from '../lib/docker-utils';
 import * as constants from '../lib/constants';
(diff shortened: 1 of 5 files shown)


pipex:
Do we still need this? https://github.com/balena-os/balena-supervisor/blob/c4f4af95c58928af5de2ff676ac7673b226c7e7d/src/compose/app.ts#L669 

If starting the container fails, with the error we should catch it on the next iteration anyway (maybe?)

Also, do we need to wait on the `createdAt` date or the `startedAt` date? Is not that clear to me

cywang117:
I've removed the date checks in https://github.com/balena-os/balena-supervisor/commit/71d24d6e3366b28e1466e99322d58420b5ebbb67 since we don't need to wait to start if a container has exited with the error message, as the Engine won't be trying to restart the container. This simplified the tests and bit and they still pass; will be updating this comment after I test on a device.

pipex (commented):
LGTM

pipex (approved):

cywang117:
@balena-ci I self-certify!

PR #2180, opened 2023-06-22
Remove the 'Stopped' status for services

cywang117:
It's not an official status from container inspects, and the Supervisor doesn't set it internally anywhere. It's better to remove it entirely as the method by which Supervisor sets internal service statuses is by using a global event emitter (reportNewStatus) which makes things difficult to test.

Change-type: patch


pipex:
This looks good to me @cywang117, if you have done a sanity check LGTM

cywang117:
Couldn't find any references to `Stopped` in open-balena-api or balena-api. balena-ui had [this reference](https://github.com/balena-io/balena-ui/blob/7d9c92b3722760943b9c117709aee745178ef189/src/pages/dashboard/device/device-summary/ContainerTable.tsx#L358) which suggests that both `Exited` and `Stopped` are supported. Since Supervisors are currently reporting Exited and nothing has broken on the dashboard, this is safe to merge.

PR #2190, opened 2023-08-17
Update dependencies

pipex:
Update knex and sqlite3 to latest versions. Also move non-binary dependencies outside of `dependencies` into `devDependencies`

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
   },
   "private": true,
   "dependencies": {
-    "@balena/happy-eyeballs": "0.0.6",
     "@balena/systemd": "^0.4.1",
     "got": "^12.5.3",
     "mdns-resolver": "^1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -46,6 +41,7 @@
   "devDependencies": {
     "@balena/contrato": "^0.6.0",
     "@balena/es-version": "^1.0.1",
+    "@balena/happy-eyeballs": "0.0.6",
```
cywang117, line 44: Shouldn't this remain in dependencies? See: https://balena.fibery.io/Work/Pattern/Implement-Happy-Eyeballs-Across-All-Balena-Services-(device-status-is-VPN-Only-)-2877

pipex: It is, I just moved it to "devDependencies"

pipex:
LGTM

PR #2192, opened 2023-08-29
Dump target-state to hostOS tmp dir

jaomaloy:
This change is mainly for the hostOS
to know if update locks should be ignored
when updating to a newer version.

Change-type: patch

### DIFF
--- a/src/device-state.ts
+++ b/src/device-state.ts
@@ -43,6 +43,12 @@ import type {
 	CompositionStepT,
 	CompositionStepAction,
 } from './compose/composition-steps';
+import * as fsUtils from './lib/fs-utils';
+import { pathOnRoot } from './lib/host-utils';
+
+const TARGET_STATE_CONFIG_DUMP = pathOnRoot(
+	'/tmp/balena-supervisor/target-state-config',
+);
 
 function parseTargetState(state: unknown): TargetState {
 	const res = TargetState.decode(state);
@@ -330,6 +336,11 @@ export async function setTarget(target: TargetState, localSource?: boolean) {
 	}
 
 	const localTarget = target[uuid];
+
+	await fsUtils.writeAndSyncFile(
+		TARGET_STATE_CONFIG_DUMP,
+		JSON.stringify(localTarget.config),
+	);
 
 	await usingWriteLockTarget(async () => {
 		await db.transaction(async (trx) => {


src/device-state.ts
```diff
@@ -319,6 +320,8 @@ export async function setTarget(target: TargetState, localSource?: boolean) {
 	// This will throw if target state is invalid
 	target = parseTargetState(target);
 
+	await writeAndSyncFile('/tmp/balena-supervisor/target-state', target);
```
alexgg, line 323: @jaomaloy I was under the impression the agreement was to dump only the device configuration section, not the whole target state, as that's the part the OS might be interested. Maybe listen to the arch call recording to confirm, or @pipex can chip in.

pipex: Yes, that was the agreement. You'll need to get that from the local target (see below line 335). The config property has all the configuration variables. 

You'll also need to stringify the value (the variable contains an object), in order to call `writeAndSync` as the error above suggests

pipex: Yes, let's just dump the config.

src/device-state.ts
```diff
@@ -344,6 +350,16 @@ export async function setTarget(target: TargetState, localSource?: boolean) {
 			await config.set({ targetStateSet: true }, trx);
 		});
 	});
+
+	const targetStateConfig = await fsUtils.exists(TARGET_STATE_CONFIG_DUMP);
+	if (!targetStateConfig) {
+		await fsUtils.touch(TARGET_STATE_CONFIG_DUMP);
```
pipex, line 356: What was the logic behind this? I don't think this is necessary, right?

jaomaloy: It isn't. Thanks!

src/device-state.ts
```diff
@@ -344,6 +350,11 @@ export async function setTarget(target: TargetState, localSource?: boolean) {
 			await config.set({ targetStateSet: true }, trx);
 		});
 	});
+
+	await fsUtils.writeAndSyncFile(
```
pipex, line 354: Sorry, one last request. Could you move this just above the `usingWriteLockTarget` call? That call can take a while and it's better that the configuration is dumped earlier in the process.

jaomaloy: Done

pipex:
Looks good @jaomaloy, merge when ready

jaomaloy:
lgtm

PR #2198, opened 2023-09-07
Update runtime-base image to alpine:3.18

cywang117:
We were still on 3.16, where the nodejs package was pointed at Node 16. 3.18 moves the pointer to Node 18.

Change-type: minor


pipex:
Can we make ALPINE_VERSION a docker arg, and reference it in both build and runtime stages so we don't make this error again?

cywang117:
Placing this in draft until we sort out any possible memory issues with current production Supervisor

pipex:
@cywang117 we should test this for memory usage and get it merged

cywang117:
Testing!

cywang117:
@balena-ci I self-certify!

PR #2201, opened 2023-09-29
Use `~=` to specify alpine package versions

pipex:
Alpine allows the `~=` syntax to match a part of the package version when installing. In this case we want to use it to specify node and npm major versions

Change-type: patch


pipex:
I self-certify!

PR #2202, opened 2023-09-29
Use native `structuredClone` instead of _.cloneDeep

cywang117:
Now that we're on Node 18 with 14.13.0, we can use native `structuredClone`, which memory tests have shown to have better performance than lodash's `_.cloneDeep`.


cywang117:
@balena-ci I self-certify!

PR #2204, opened 2023-10-02
Add tmpfs for /mnt/root/tmp in test env

cywang117:
This is to prevent the target state cache from getting written to host during tests.

Change-type: patch


cywang117:
@balena-ci I self-certify!

PR #2215, opened 2023-10-16
Revert "Do not expose ports from image if service network mode"

pipex:
This reverts commit 0c7bad779291e15e419166a2c66c2a21dd06aa83, as that change causes a service restart loop. The supervisor cannot distinguish between ports exposed via the `EXPOSE` directive and the docker-compose `expose` property. Because of this, in the case of `network-mode: service:<...>` the current state and target state never match, leading to a service restart loop.

Change-type: patch

### DIFF
--- a/src/compose/service.ts
+++ b/src/compose/service.ts
@@ -346,9 +346,7 @@ export class Service {
 			'Config.ExposedPorts',
 			{},
 		);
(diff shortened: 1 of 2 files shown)


pipex:
I self-certify!

PR #2216, opened 2023-10-16
Run memory healthcheck for Supervisor process

cywang117:
The healthcheck will restart the Supervisor upon memory leaks surpassing a threshold of +15Mb.

Change-type: patch

### DIFF
--- a/src/memory.ts
+++ b/src/memory.ts
@@ -1,0 +1,58 @@
+import { memoryUsage } from 'process';
+
+import * as deviceState from './device-state';
+import log from './lib/supervisor-console';
+
+export let initialMemory: number = 0;
+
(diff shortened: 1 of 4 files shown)


cywang117:
@pipex Ready for review! Moved the memory threshold to 15mb, ran additional tests and got similar results.

cywang117:
I am holding off on merging this for the time being. As memory healthchecks would result in an increase in Supervisor restarts, and some Supervisor operations on startup (i.e. current state report) are expensive, concerns were raised over the increased load this would have on the backend. Looking to improve the implementation of startup current state reporting by caching the last reported state first, before merging this.

cywang117:
@balena-ci I self-certify!

Issue #2211, opened 2023-10-12
Supervisor uses image EXPOSE directive to expose container ports


pipex:
This is different than what docker-compose does, and it can conflict with some options, for instance, for using `network_mode: service:<service name>`, that will fail with 

```
Error: Failed to apply state transition steps. (HTTP code 400) unexpected - conflicting options: port exposing and the container type network mode 
```

It is unclear yet why this configuration was made so it is not safe to remove it yet, but we might be able to add an exception for service network mode.

klutchell:
I just got caught by this too. Is there any workaround to use `network_mode: service:<service name>` while we investigate?

pipex:
What do you think about this PR #2212? It seems like a good compromise to me

pipex:
No workaround really @klutchell, there is no way to undefine the EXPOSE directive on the docker image, and docker-compose configurations are additive. Depending on what you are trying to do there may be alternatives using network-mode host, but I understand that is not desirable.

klutchell:
The workaround seems fine, but the expose instruction has been deprecated for years now and the SV shouldn't have to apply anything for containers to be reachable on the same bridge.

Perhaps it is leftover from when Docker had to take action on that?

Since then docker networks have changed so containers on the same network automatically have all ports exposed.

pipex:
The change seems to be 5 years old, to [this commit](https://github.com/balena-os/balena-supervisor/blame/80f06adb0f76aefcd82dca30b40b0ea1c68c51c6/src/compose/service.coffee#L369). 

What do you mean the expose directive is deprecated? I though it could be used with the publish all option on docker run. 

You have a point that we don't really care about the expose directive as ports need to be manually published via `docker-compose`, so it might be OK to remove it

klutchell:
https://docs.docker.com/engine/reference/builder/#expose

> The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.

Based on that, I don't think the SV should be acting on the expose instruction. But I can see that being a bigger change in case users were depending on it being published.

klutchell:
Perhaps we need it for cases of Dockerfile publish, without compose, but in that case it runs in host network mode anyway IIRC

pipex:
The supervisor doesn't publish the port defined with expose, it just exposes it for the container, is the equivalent of doing `docker run --expose` that doesn't really do much

klutchell:
[...]

PR #2217, opened 2023-10-18
Ignore `expose` service compose configuration

pipex:
The docker EXPOSE directive and corresponding docker-compose `expose`
service configuration serves as documentation/metadata that a container
listens on a certain port that may be used for service discovery but it doesn't
have any real impact on the ability for
other containers on the same network to access the exposed service via
the port. In newer engine implementations, this property may conflict
with other network configurations, and prevent the container from being
started by the docker engine (see https://github.com/balena-os/balena-supervisor/issues/2211).

This PR removes code that would manage the expose property and takes the
property out of the whitelist. A composition with the `expose` property
will result in the log message `Ignoring unsupported or unknown compose fields: expose`.

While this change should not have operational impact, it still removes
a previously supported configuration and as such there is a chance of it
being a breaking change for some applications. For this reason it is
being published as a new major version.

Change-type: major
Closes: https://github.com/balena-os/balena-supervisor/issues/2211

### DIFF
--- a/src/compose/ports.ts
+++ b/src/compose/ports.ts
@@ -15,7 +15,6 @@ export interface PortBindings {
 }
 
 export interface DockerPortOptions {
-	exposedPorts: Dictionary<{}>;
 	portBindings: PortBindings;
 }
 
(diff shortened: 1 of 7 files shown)


klutchell:
Does this change the behaviour if the `EXPOSE` instruction was provided via the Dockerfile, and is thus already in the image metadata?

pipex:
If the EXPOSE instruction is used, it will be ignored by the supervisor but used by Docker same as before. The warning for the compose happens because we are purposely ignoring a compose configuration, so if you are only adding an `expose` property in DC, you will no longer see exposed ports in the container configuration.

klutchell:
LGTM

PR #2219, opened 2023-10-24
Expose ports from port mappings on services

pipex:
PR #2217 removed the expose configuration but also caused a regresion where ports set via the `ports` configuration would no longer get exposed to the host, despite portmappings being set. This fixes that issue by exposing only those ports comming from port mappings.

Change-type: patch


pipex:
I self-certify!

PR #2220, opened 2023-10-25
Set applyInProgress to true while applying intermediate state

cywang117:
Intermediate state is utilized when executing device actions such as a volume purge. It's a type of state apply, but despite that, applyInProgress is not true.

Change-type: patch


cywang117:
I performed some manual tests on a device and did not observe any abnormal behavior

cywang117:
I self-certify!

cywang117:
@balena-ci I self-certify!

PR #2225, opened 2023-11-09
Cache last reported current state

cywang117:
Whenever the Supervisor reports current state, it diffs the current state
with its last reported current state. However, when the Supervisor starts
up, there is no last reported state, since that last report is stored in
process memory. Caching the last report in a location that survives
Supervisor restarts will reduce the current report bandwidth used on startup.

Change-type: patch

### DIFF
--- a/src/api-binder/report.ts
+++ b/src/api-binder/report.ts
@@ -2,11 +2,8 @@ import * as url from 'url';
 import * as _ from 'lodash';
 import { CoreOptions } from 'request';
 import { performance } from 'perf_hooks';
-
-import { withBackoff, OnFailureInfo } from '../lib/backoff';
-import { log } from '../lib/supervisor-console';
-import { InternalInconsistencyError, StatusError } from '../lib/errors';
-import { getRequestInstance } from '../lib/request';
+import { setTimeout } from 'timers/promises';
+import { readFile } from 'fs/promises';
 
 import { DeviceState } from '../types';
 import * as config from '../config';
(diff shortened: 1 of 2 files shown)


Page- (commented):
I think it makes more sense to cache the target state when we get a new one rather than on every supervisor exit, this helps for a few reasons:
1. If the supervisor exists unexpectedly (eg power pulled) we still have the cache
2. We only write when the cache has actually changed, and it will be at a time when we're almost certainly doing writes anyway in order to attempt to apply the target state so there shouldn't be any real worries about increased writes (since it won't be particularly significant)
3. It's probably simpler/more robust than hooking into process exiting if we don't need to

cywang117:
@Page- Thanks for the feedback. The original implementation was caching last reported current state after every report. I'd discussed this with @pipex previously and we were concerned about  the increased I/O that would involve on the device with every cache operation. Leading to the current implementation of it. I'm going to run the draft Supervisor on an RPi 0 to see if the memory impact is significant and will report back. For the record, I'm okay with either implementation.

pipex:
This looks good @cywang117  i don't really have any observations. Merge when you are ready with testing

cywang117:
@balena-ci I self-certify!

src/api-binder/report.ts
```diff
@@ -128,6 +138,43 @@ async function reportCurrentState(opts: StateReportOpts, uuid: string) {
...
+		// Touch the file, which will create it if it doesn't exist
+		await touch(CACHE_PATH);
+
+		// Get last reported current state
+		const rawStateCache = await readFile(CACHE_PATH, 'utf-8');
+		const state = JSON.parse(rawStateCache);
+
+		// Return current state cache if valid
+		if (!DeviceState.is(state)) {
+			throw new Error();
```
thgreasi, line 166: Shouldn't it delete the cache file if it's invalid rather than throwing and retrying forever?

PR #2226, opened 2023-11-15
Force remove container if updateMetadata fails

pipex:
The `updateMetadata` step renames the container to match the target release when the service doesn't change between releases. We have seen this step fail because of an engine bug that seems to relate to the engine keeping stale references after container restarts. The only way around this issue is to remove the old container and create it again. This implements that workaround during the updateMetadata step to deal with that issue.

Change-type: minor
Relates-to:  balena-os/balena-engine#261

### DIFF
--- a/src/compose/service-manager.ts
+++ b/src/compose/service-manager.ts
@@ -157,9 +157,20 @@ export async function updateMetadata(service: Service, target: Service) {
 		);
 	}
 
-	await docker.getContainer(svc.containerId).rename({
-		name: `${service.serviceName}_${target.imageId}_${target.releaseId}_${target.commit}`,
-	});
+	try {
+		await docker.getContainer(svc.containerId).rename({
+			name: `${service.serviceName}_${target.imageId}_${target.releaseId}_${target.commit}`,
+		});
+	} catch (e) {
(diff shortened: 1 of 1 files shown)


cywang117 (commented):
LGTM, merge when ready!

src/compose/service-manager.ts
```diff
@@ -157,9 +157,20 @@ export async function updateMetadata(service: Service, target: Service) {
...
+		await docker.getContainer(svc.containerId).rename({
+			name: `${service.serviceName}_${target.imageId}_${target.releaseId}_${target.commit}`,
+		});
+	} catch (e) {
+		if (isNotFoundError(e)) {
+			log.warn(
+				'Got 404 error while updating container metadata. Will attempt to recreate the container instead',
+				e,
+			);
+			return await docker.getContainer(svc.containerId).remove({ force: true });
```
klutchell, line 170: If `updateMetadata` result is a deleted container, how long until a new one is created to replace it?

pipex: Until the next iteration of the state engine, that will see that there is a missing service in the current state and will add steps to create it

pipex:
I self-certify!

Issue #2095, opened 2023-01-23
Supervisor deletes no longer defined service before finishing update


majorz:
When the `download-then-kill` update strategy is being used, if the new release removes a service, that service is killed and deleted early in the process (before the download of the new release is completed). If the update fails the device is left with a missing service.

PR #2230, opened 2023-12-15
Apply update strategies app-wide

cywang117:
The Supervisor's state engine has some edge cases where it does not respect the update strategy of the app. These edge cases relate to a release update which removes an old service and also adds a new service. This PR fixes those edge cases.

See: https://balena.fibery.io/Work/Improvement/Supervisor-improved-multi-container-updates-1329
Change-type: patch
Closes: #2095

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -15,12 +15,13 @@ import {
 import * as targetStateCache from '../device-state/target-state-cache';
 import { getNetworkGateway } from '../lib/docker-utils';
 import * as constants from '../lib/constants';
-
-import { getStepsFromStrategy } from './update-strategies';
-
-import { InternalInconsistencyError, isNotFoundError } from '../lib/errors';
+import {
+	getStepsFromStrategy,
+	getStrategyFromService,
+} from './update-strategies';
+import { isNotFoundError } from '../lib/errors';
 import * as config from '../config';
-import { checkTruthy, checkString } from '../lib/validation';
+import { checkTruthy } from '../lib/validation';
 import { ServiceComposeConfig, DeviceMetadata } from './types/service';
 import { pathExistsOnRoot } from '../lib/host-utils';
 import { isSupervisor } from '../lib/supervisor-metadata';
@@ -133,17 +134,8 @@ export class App {
 			state.containerIds,
 		);
 
-		for (const { current: svc } of removePairs) {
-			// All removes get a kill action if they're not already stopping
-			if (svc!.status !== 'Stopping') {
-				steps.push(generateStep('kill', { current: svc! }));
-			} else {
-				steps.push(generateStep('noop', {}));
-			}
-		}
-
 		// For every service which needs to be updated, update via update strategy.
-		const servicePairs = updatePairs.concat(installPairs);
+		const servicePairs = removePairs.concat(updatePairs, installPairs);
 		steps = steps.concat(
 			servicePairs
 				.map((pair) =>
@@ -511,8 +503,8 @@ export class App {
 		},
 	): Nullable<CompositionStep> {
 		if (current?.status === 'Stopping') {
-			// Theres a kill step happening already, emit a noop to ensure we stay alive while
-			// this happens
+			// There's a kill step happening already, emit a noop to ensure
+			// we stay alive while this happens
 			return generateStep('noop', {});
 		}
 		if (current?.status === 'Dead') {
@@ -521,16 +513,14 @@ export class App {
 			return;
 		}
 
-		const needsDownload = !_.some(
-			context.availableImages,
+		const needsDownload = !context.availableImages.some(
 			(image) =>
 				image.dockerImageId === target?.config.image ||
 				imageManager.isSameImage(image, { name: target?.imageName! }),
 		);
-
 		if (needsDownload && context.downloading.includes(target?.imageName!)) {
-			// The image needs to be downloaded, and it's currently downloading. We simply keep
-			// the application loop alive
+			// The image needs to be downloaded, and it's currently downloading.
+			// We simply keep the application loop alive
 			return generateStep('noop', {});
 		}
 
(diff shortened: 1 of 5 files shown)


src/compose/app.ts
```diff
@@ -133,17 +134,8 @@ export class App {
...
-			if (svc!.status !== 'Stopping') {
-				steps.push(generateStep('kill', { current: svc! }));
-			} else {
-				steps.push(generateStep('noop', {}));
-			}
-		}
-
 		// For every service which needs to be updated, update via update strategy.
-		const servicePairs = updatePairs.concat(installPairs);
+		const servicePairs = [...removePairs, ...updatePairs, ...installPairs];
```
pipex, line 138: I think `concat` is faster, and it can receive multiple arguments as well

src/compose/app.ts
```diff
@@ -546,47 +536,38 @@ export class App {
...
 			);
-
 			if (
 				!needsSpecialKill &&
+				target != null &&
 				current.isEqualConfig(target, context.containerIds)
 			) {
 				// we're only starting/stopping a service
-				return this.generateContainerStep(current, target);
+				return this.generateContainerStep(current, target!);
```
pipex, line 552: Why is the `!` necessary if you are checking that target is not null above?

src/compose/app.ts
```diff
@@ -546,47 +536,38 @@ export class App {
...
-				'kill-then-download',
-				'delete-then-download',
-				'hand-over',
-			];
-
-			if (!validStrategies.includes(strategy)) {
-				strategy = 'download-then-kill';
-			}
+			const strategy =
+				target == null
```
pipex, line 556: This assignment of variables using ternary operators is pretty messy, can you think of a way of refactoring the code so it is more readable?

test/integration/compose/application-manager.spec.ts
```diff
@@ -327,6 +329,370 @@ describe('compose/application-manager', () => {
...
+			// Service `new` in target release should be started as download has completed
+			const [startStep] = expectSteps('start', steps2, 1);
+			expect(startStep)
+				.to.have.property('target')
+				.that.deep.includes({ serviceName: 'new' });
+			// No noop steps
+			expectNoStep('noop', steps2);
+		});
+
+		it('should not infer a fetch step for target service B before current service A kill with kill-then-download', async () => {
```
pipex, line 467: why not? A and B are different services, I think it's OK to start fetching B while A is being killed. If anything I would expect that A should not be killed at all before B has been fetched

test/integration/compose/application-manager.spec.ts
```diff
@@ -327,6 +329,370 @@ describe('compose/application-manager', () => {
...
+					// deleted from the image table to avoid saveImage / removeImage steps
+					availableImages: [],
+					containerIdsByAppId: c3.containerIdsByAppId,
+				},
+			);
+			const [fetchStep] = expectSteps('fetch', steps3, 1);
+			expect(fetchStep).to.have.property('serviceName').that.equals('new');
+		});
+
+		it('should not infer a fetch step for target service B before current service A image removal with delete-then-download', async () => {
```
pipex, line 563: Again, if A and B are different services, I'd expect the opposite. We don't want to remove the old service image until we are ready with the new release images.

src/compose/app.ts
```diff
@@ -698,6 +679,27 @@ export class App {
...
+	) {
+		// Current services should be killed before target
+		// service fetch depending on update strategy
+		const servicePairsWithCurrent = servicePairs.filter(
+			(pair) => pair.current != null,
+		);
+		for (const { current, target } of servicePairsWithCurrent) {
+			// Prefer target's update strategy if target service exists
+			const strategy = getStrategyFromService(target ?? current!);
+			if (
```
pipex, line 693: I understood what you meant by this code when reading the tests, I think this is wrong as this means we cannot fetch any service image from the new until all services with (kill|delete)-then-download from the previous release have been stopped|removed. Release strategies are still per-service with this change, but we want to chose when to apply changes when it's safest for the app (meaning less downtime). With this particular code, it could be that the new images take a long time to download, meaning the app is unusable during that whole time.

test/integration/compose/application-manager.spec.ts
cywang117: The test case here and the one in the comment below are for `kill-then-download` and `delete-then-download` strategies. Does this still apply? The intent here is for `kill|delete-then-download` to apply app-wide, meaning that all service should be killed/deleted before download occurs, even for services only in target and not in current.

cywang117: Never mind, I see your last comment below which addresses this, we can continue the conversation there

src/compose/app.ts
cywang117: By this, do you mean that we should not adhere strictly to the concept of app-wide strategies, but rather make allowances that benefit performance, such as starting a fetch for a target service even when current services haven't yet been killed while using `kill|delete-then-download`?

pipex: Hmm, the term ''app wide strategies" is a bit ambiguous, we might need to look for a new term. The strategies are still per service, as they define the order of operations for updating a single service, but when updating the app, we need to combine those service operations in a way that minimizes downtime to the app, and allows us to rollback quickly if something goes wrong with the update. 

This means, suppose we have a combination of 'download-then-kill', 'kill-then-download' and 'delete-then-download' services,

1. Pull all new `download-then-kill` target images
2. Kill all `kill-then-download` and `delete-then-download` services the are being updated and remove images for the latter (in a future improvement we'll also need to kill services that depend on these)
3. Download the new images for the `kill-then-download` and `delete-then-download` services.
4. Kill remaining services to be updated and start all new services
5. Cleanup unused images

This also means, that if we need to roll back before step 5, we can do it pretty quickly (except for re-downloading the `delete-then-download`)

Let me know if this makes sense

pipex: Actually, on second thought, your PR swaps steps 1 and 2, which is OK for now. We will want to move to the above sequence when we use `depends-on` for kill, so we kill services as late as possible

pipex:
Could you add a test for this though https://github.com/balena-os/balena-supervisor/issues/2095

cywang117:
> Could you add a test for this too? #2095

That should be covered under 
https://github.com/balena-os/balena-supervisor/blob/565f3734dc792a2e2e925e1c1b41f06d50c6dc19/test/integration/compose/application-manager.spec.ts#L377

The test case defines current services A & B, and target services A & C. While C is still downloading, the test expects 2 noop steps, one for each of service A & B. When C has finished downloading, only then does the test expect the kill step.

src/compose/app.ts
cywang117: Would step 1 also apply for new target images with `kill|delete-then-download` ?

cywang117:
@pipex I made some changes following your feedback. The new changes make it so that the Supervisor fetches new target services even if the update strategy is `kill|delete-then-download` .

src/compose/app.ts
```diff
@@ -764,13 +779,30 @@ export class App {
...
 		availableImages: Image[],
+		servicePairs: Array<ChangingPair<Service>>,
+		strategy: string,
 	) {
-		// Don't kill any services before all images have been downloaded
-		return this.targetImagesReady(targetApp, availableImages);
+		// if kill|delete-then-download, don't kill any services
+		// before new images in target have been downloaded.
+		// Services are safe to kill if they are in both current and target
+		// even if the target image isn't yet ready.
```
pipex, line 788: I don't understand why this is needed. Result of `dependenciesMetForServiceKill` is not used by `getStepsForStrategy` if the strategy is `kill-then-download` or `delete-then-download` so this code would be redundant. Also this doesn't seem like the most efficient alternative considering this function is called for every service.

pipex: Oh, wait, sorry, you changed the `getStepsForStrategy` but function, but I still don't understand why

pipex (approved):
Nice work @cywang117 !

PR #2231, opened 2023-12-28
Try MDNS lookup only if regular DNS lookup fails

pipex:
This is meant to allow users to configure their device to resolve `.local` queries via dnsmasq by modifying config.json, e.g. `dnsServers": "/bob.local/172.17.0.33`.

 This would fail before as MDNS lookups would always come first

Tested, OK: https://github.com/balena-io/balena-cloud/pull/2020

(https://en.wikipedia.org/wiki/.local)

Change-type: minor

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -26,6 +26,80 @@ interface DnsLookupOpts {
 	family: number;
 }
 
+/*
+ * =====================================================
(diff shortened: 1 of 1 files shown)


cywang117 (commented):
This LGTM, merge when ready!

pipex:
I self-certify!

PR #2232, opened 2023-12-29
docs: api: start-service: specify this endpoint affects the device

alexgg:
The description incorrectly states the endpoint would start a service in the whole fleet while it only affects the device.

Change-type: patch
Relates-to: https://jel.ly.fish/support-thread-discourse-369541


cywang117 (approved):
LGTM

PR #2234, opened 2024-01-09
Update lock during state apply

cywang117:


test/unit/lib/update-lock.spec.ts
```diff
@@ -14,4 +17,52 @@ describe('lib/update-lock: unit tests', () => {
 			);
 		});
 	});
+
+	describe('getServicesLockedByAppId', () => {
+		let getLocksTaken: SinonStub;
+		beforeEach(() => {
+			getLocksTaken = stub(lockfile, 'getLocksTaken');
```
pipex, line 24: I don't like this stubbing of internal methods, I think we should avoid it from now on as it means the test is relying on internal implementation details of the function being tested. Either the environment should be setup in advance to what the function expects (you can use mocha-pod to help with that), or the function should receive `locksTaken` as argument and make it a pure function.

src/lib/process-lock.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * This module contains the functionality for locking & unlocking resources
```
pipex, line 2: Nice. Very logical improvement to move this to its own module

src/lib/process-lock.ts
```diff
@@ -0,0 +1,28 @@
...
+ * within the Supervisor Node process, useful for methods that need to acquire
+ * exclusive access to a resource across multiple ticks in the event loop, async
+ * functions for example.
+ *
+ * It is different from lockfile and update-lock modules, which handle
+ * inter-container communication via lockfiles.
+ *
+ * TODO:
+ * - Use a maintained solution such as async-lock
```
pipex, line 11: I personally also liked [BetterLock ](https://www.npmjs.com/package/better-lock) although it would be even better if we could avoid locking

src/compose/app.ts
```diff
@@ -166,17 +169,29 @@ export class App {
...
+		if (steps.length === 0) {
+			// Update commit in db if different
+			if (target.commit != null && this.commit !== target.commit) {
+				steps.push(
+					generateStep('updateCommit', {
+						target: target.commit,
+						appId: this.appId,
+					}),
+				);
+			} else if (
```
pipex: Not sure about this else. Couldn't these steps happen at the same time? Why do they need to happen in different iterations?

src/lib/update-lock.ts
```diff
@@ -82,6 +82,55 @@ async function dispose(
...
+	return [appId, serviceName];
+}
+
+export type LocksTakenMap = Map<number, string[]>;
+
+/**
+ * Return a list of services that are locked by the Supervisor under each appId.
+ */
+export function getServicesLockedByAppId(): LocksTakenMap {
+	const locksTaken = lockfile.getLocksTaken();
```
pipex, line 135: The function `getLocksTaken` returns only the locks taken during the current supervisor run. I know we delete existing locks during supervisor start, but I wonder if we need to remove that cleanup step, as user containers could take the lock during the supervisor startup and leave the user app in an inconsistent state. 

If that's the case, `getLocksTaken` should maybe return the actual state of locks on disk.

You may have already changed this in a future commit. Ignore this if that's the case

src/device-state.ts
```diff
@@ -23,6 +23,7 @@ import {
 	UpdatesLockedError,
 } from './lib/errors';
 import * as updateLock from './lib/update-lock';
+import { readLock, writeLock } from './lib/process-lock';
```
pipex, line 26: Is kind of weird that the state engine uses the same lock instance as the target state or the update lock. I may be misunderstanding how `rwlock` operates

src/lib/update-lock.ts
```diff
@@ -72,6 +76,15 @@ async function dispose(
 	}
 }
 
+/**
+ * Composition step used by Supervisor compose module.
+ * Release all locks for an appId | appUuid.
+ */
+export async function releaseLock(appId: number) {
+	const release = await writeLock(appId);
```
pipex, line 111: I think the `writeLock` function needs a better name. It took me a second to figure out this step was taking the process lock. Maybe `takeGlobalLockRW` and `takeGlobalLockRO` (instead of `readLock`)

test/integration/lib/update-lock.spec.ts
```diff
@@ -281,4 +284,117 @@ describe('lib/update-lock', () => {
...
+
+		describe('releaseLock', () => {
+			let testFs: TestFs.Enabled;
+
+			beforeEach(async () => {
+				testFs = await testfs(
+					{},
+					{ cleanup: [path.join(lockdir, '*', '*', '**.lock')] },
+				).enable();
+				// TODO: Update mocha-pod to work with creating empty directories
```
pipex, line 466: Did you try using `{}` to indicate an empty dir? e.g.

```
testfs({'/tmp/balena/myappdir': {}})
```

test/integration/lib/update-lock.spec.ts
```diff
@@ -281,4 +284,117 @@ describe('lib/update-lock', () => {
 			);
 		});
 	});
+
+	describe('composition step actions', () => {
+		const lockdir = pathOnRoot(updateLock.BASE_LOCK_DIR);
+		const serviceLockPaths = {
+			1: [
+				path.join(lockdir, '1', 'server', 'updates.lock'),
```
pipex, line 293: I know that `path.join` is technically correct, but I feel that for a testing context `\`${lockdir}/1/server/updates.lock\`` is a lot more readable. Is just a nitpick, keep it this way if you prefer it

test/integration/lib/update-lock.spec.ts
```diff
@@ -281,4 +284,117 @@ describe('lib/update-lock', () => {
...
+				}
+				// Sanity check that locks are taken & tracked by Supervisor
+				expect(lockfile.getLocksTaken()).to.deep.include.members(
+					serviceLockPaths[1],
+				);
+				// Take the write lock for appId 1
+				const release = await writeLock(1);
+				// Queue releaseLock, won't resolve until the write lock is released
+				const releaseLockPromise = updateLock.releaseLock(1);
+				// Locks should have not been released
```
pipex, line 538: I think you need some delay here. Even if the `releaseLock` call wasn't being blocked, the expectation below would still return as that line would be executed before the promise could be executed.

src/lib/update-lock.ts
```diff
@@ -82,6 +82,55 @@ async function dispose(
...
+	const parts = p.split('/');
+	if (parts.pop()?.match(/updates\.lock/) === null) {
+		return [];
+	}
+	const serviceName = parts.pop();
+	const appId = parts.pop();
+	return [appId, serviceName];
+}
+
+export type LocksTakenMap = Map<number, string[]>;
```
pipex, line 129: Looking at later commit tests, I feel that aliasing map might not be the best way to implement this as it makes the code hard to read.

test/unit/compose/app.spec.ts
```diff
@@ -1671,4 +1672,91 @@ describe('compose/app', () => {
...
+			const target = createApp({
+				services,
+				networks: [DEFAULT_NETWORK],
+				isTarget: true,
+			});
+
+			const steps = current.nextStepsForAppUpdate(
+				{
+					...defaultContext,
+					locksTaken: new Map([[1, ['server', 'client']]]),
```
pipex, line 2027: Related to my previous comment about using map for keeping track of locks. While I agree is probably the right data structure, this code is not intuitive when looking at it without having the context of what the values inside the array are. If instead this code was

```
locksTaken: new LockRegistry([{appId: 1, services: ['server', 'client']}])
```

You would immediately get a sense of what the code does. The underlying implementation may still be using Map, but the interface is more clear

pipex:
4/9 commits reviewed, will continue tomorrow @cywang117

src/compose/app.ts
cywang117: This is a good point, will change

test/unit/lib/update-lock.spec.ts
cywang117: Noted, will change

src/device-state.ts
cywang117: This is how rwlock works, by locking arbitrarily defined resources AKA strings. Device state locks the `'target'` resource:
https://github.com/balena-os/balena-supervisor/blob/d6efb0ba03142b628ce0e321dcbabc2a7cb8f6fb/src/device-state.ts#L296-L301

I agree, it's definitely confusing .

src/lib/update-lock.ts
cywang117: Makes sense!

test/unit/compose/app.spec.ts
cywang117: Makes sense, will change. I was relying on Copilot to autocomplete the locksTakenMaps in test cases so didn't have to deal with this confusion as much, but you're right

src/lib/update-lock.ts
pipex: > lockfile module stores locks it has taken in memory, meaning Supervisor-taken locks from an older Supervisor run wouldn't get picked up.

I understand that, maybe the update-lock module needs to check all locks in the filesystem during initialization?

> Why would the user containers taking the lock leave the user app in an inconsistent state? 

Is not necessarily likely, but here is a scenario. Supervisor has 2 services, service A is kill-then-download and service B is download-then-kill 

- Supervisor waits for locks and takes them
- Service A is killed and image is downloaded, supervisor crashes while downloading image for service B
- During restart the supervisor deletes the locks, which gives service B the chance to take the lock again
- Now the user app is stuck with one dead service and one service running until service B releases the lock again

cywang117: Thanks for explaining, that makes sense. I'll modify this in a separate commit.

test/unit/lib/update-lock.spec.ts
cywang117: Removed the stubs in favor of testFs [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/b5d5dd7a9837a8a939cea5d99fddc56f2d4840ac#diff-e7a2b32a616114c913967bbeb93301d1194c25d37b5c6d6393146a862d33b61fR285-R347). I wasn't able to get testfs empty directories to work consistently. I'd like to look at it more closely, but perhaps in a separate PR.

src/compose/app.ts
cywang117: Changed `updateCommit` to happen at the same time as `releaseLock` [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/02476bbc3ae61df1b42ff61f5094d723d61d9dc9#diff-857bb737cdeb3c33e2caa352769d3f96e0e8a48da6fea254eb99bc627732cec1L172-R215)

src/lib/update-lock.ts
cywang117: Updated names to be more clear [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/9d8241c9508c85012df309b275a5bdcef04b57bd#diff-d722e5c234dcc829cd6f5a4eb70cfdb9bf414361584d50bfa07fe228db711a41R22-R34)

test/integration/lib/update-lock.spec.ts
cywang117: Path names simplified [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/7b9be98a7831ff63764d7a20261881535a0b1d30#diff-e7a2b32a616114c913967bbeb93301d1194c25d37b5c6d6393146a862d33b61fR354-R365)

test/integration/lib/update-lock.spec.ts
cywang117: Added 500ms delay [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/7b9be98a7831ff63764d7a20261881535a0b1d30#diff-e7a2b32a616114c913967bbeb93301d1194c25d37b5c6d6393146a862d33b61fR446) for releaseLock and [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/fe9f940ce619cd7045cb87bba1f66b2e8f91a93d#diff-e7a2b32a616114c913967bbeb93301d1194c25d37b5c6d6393146a862d33b61fR510) for takeLock

test/unit/compose/app.spec.ts
cywang117: Updated `LocksTakenMap` to have more user-friendly instantiation args [in this commit](https://github.com/balena-os/balena-supervisor/pull/2234/commits/b5d5dd7a9837a8a939cea5d99fddc56f2d4840ac) (and added some tests)

cywang117:
I amended some commits per review thus far on the 4/9 commits, but didn't change anything else

src/lib/update-lock.ts
```diff
@@ -82,6 +82,75 @@ async function dispose(
...
+export class LocksTakenMap extends Map<number, Set<string>> {
+	constructor(lockedEntities: LockedEntity[] = []) {
+		// Construct a Map<number, Set<string>> from user-friendly input args
+		super(
+			lockedEntities.map(({ appId, services }) => [appId, new Set(services)]),
+		);
+	}
+
+	// Add one or more locked services to an appId
+	public add(appId: number, services: string | string[]): void {
```
pipex, line 144: Since you are extending `Map` I think you can override `set` here

src/lib/update-lock.ts
```diff
@@ -82,6 +82,75 @@ async function dispose(
...
+			services.forEach((s) => lockedSvcs.add(s));
+		} else {
+			this.set(appId, new Set(services));
+		}
+	}
+
+	// Return an array copy of locked services under an appId
+	public getServices(appId: number): string[] {
+		return this.has(appId) ? Array.from(this.get(appId)!) : [];
+	}
```
pipex, line 159: Since you are extending `Map` I think is best to extend `get` here, however you should be careful to return something the user can modify

```ts
public getServices(appId: number): string[] {
   let svcs = [];
   if (this.has(appId)) {
       svcs = super.get(appId);
   }
   else {
      super.set(appId, svcs);
   }
   return svcs; 
}
```

That way modifications to the returned array will also modify the reference in the Map, same as with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get

test/integration/lib/update-lock.spec.ts
```diff
@@ -365,7 +366,158 @@ describe('lib/update-lock', () => {
...
+				await lockfile.lock(serviceLockPaths[1][0]);
+				await lockfile.lock(serviceLockPaths[1][1]);
+				// Sanity check that locks are taken & tracked by Supervisor
+				expect(lockfile.getLocksTaken()).to.deep.include(
+					serviceLockPaths[1][0],
+					serviceLockPaths[1][1],
+				);
+				expect(lockfile.getLocksTaken()).to.have.length(2);
+				// Take locks using takeLock, should not error
+				const lockSpy = sinon.spy(lockfile, 'lock');
```
pipex, line 455: Can you think of a way to avoid using a spy here? Perhaps the `takeLock` function could return a list of the services for which locks have actually been taken?

src/lib/update-lock.ts
```diff
@@ -82,6 +82,75 @@ async function dispose(
...
+export function getServicesLockedByAppId(): LocksTakenMap {
+	const locksTaken = lockfile.getLocksTaken();
+	const servicesByAppId = new LocksTakenMap();
+	for (const lockTakenPath of locksTaken) {
+		const [appId, serviceName] = getIdentifiersFromPath(lockTakenPath);
+		if (!StringIdentifier.is(appId) || !DockerName.is(serviceName)) {
+			continue;
+		}
+		const numAppId = +appId;
+		servicesByAppId.add(numAppId, serviceName);
```
pipex, line 179: What if for some reason the folder only has `resin-updates.lock` and not `updates.lock`? Is the lock for the service taken in that case?

src/lib/update-lock.ts
```diff
@@ -76,6 +77,28 @@ async function dispose(
...
+	try {
+		// Filter out services that already have Supervisor-taken locks.
+		// This needs to be done after taking the appId write lock to avoid
+		// race conditions with locking.
+		const servicesWithoutLock = services.filter(
+			(svc) => !getServicesLockedByAppId().isLocked(appId, svc),
+		);
+		for (const service of servicesWithoutLock) {
+			await mkdirp(pathOnRoot(lockPath(appId, service)));
+			await lockService(appId, service);
```
pipex: I think using `lockService` is a problem here. Lock service is not "all or nothing", it could take `update.lock` but fail on `resin-update.lock`. That was solved on the [lock](https://github.com/balena-os/balena-supervisor/blob/master/src/lib/update-lock.ts#L94) function by cleaning up locks on exception, but that is not the case here

pipex: Could `takeLock` perhaps call `lock` and not receive an array of services?

src/compose/app.ts
```diff
@@ -658,10 +660,33 @@ export class App {
...
+			// meaning the service could be on new release while another service
+			// with a user-taken lock is still on old release.
+			if (
+				shouldSkipLock ||
+				locksTaken.isLocked(target.appId, target.serviceName)
+			) {
+				return generateStep('updateMetadata', { current, target });
+			}
+			// Otherwise, take lock for service first
+			return generateStep('takeLock', {
```
pipex: Could we take locks for all services after downloads are finished? Feels strange taking locks one by one since basically all locks will need to be taken: changing services, services that require updateMetadata and new services

src/compose/app.ts
```diff
@@ -607,29 +698,13 @@ export class App {
...
-	// without worrying if the step is skipped or not
-	private generateKillStep(
-		service: Service,
-		changingServices: Array<ChangingPair<Service>>,
-	): CompositionStep[] {
-		if (
-			service.status !== 'Stopping' &&
-			!_.some(
-				changingServices,
-				({ current }) => current?.serviceName === service.serviceName,
```
pipex, line 607: This would prevent emitting kill steps for a service that is already stopping. I didn't see this in the new implementation

src/compose/app.ts
```diff
@@ -476,26 +570,13 @@ export class App {
...
-					// the next state application loop
-					// FIXME: We should add to the changingServices array, as we could emit several
-					// kill steps for a service
-					steps = steps.concat(
-						dependencies.flatMap((svc) =>
-							this.generateKillStep(svc, changingServices),
-						),
-					);
-				} else {
+				if (dependentServices.length === 0) {
```
pipex, line 579: Not sure about this, wouldn't this prevent removal of a component even if there is another component that has dependent services? Example a volume A without dependencies would still wait for the removal of dependencies of a volume B. Is that the intention here?

src/compose/app.ts
```diff
@@ -513,8 +594,15 @@ export class App {
...
 			servicePairs: Array<ChangingPair<Service>>;
+			appsToLock: AppsToLockMap;
 		} & UpdateState,
 	): Nullable<CompositionStep> {
+		const currentServicesLocked = this.services.every((svc) =>
+			context.locksTaken.isLocked(svc.appId, svc.serviceName),
+		);
+		const targetServicesLocked = context.targetApp.services.every((svc) =>
+			context.locksTaken.isLocked(svc.appId, svc.serviceName),
+		);
```
pipex, line 605: Again, I'm not sure we need to distinguish between current and target services locked. The app either has its services locked or not, if there are new services, `takeLock` should preemptively take those locks

src/compose/update-strategies.ts
```diff
@@ -12,12 +13,30 @@ export interface StrategyContext {
...
+	current: Service,
+	currentServices: Service[],
+	currentServicesLocked: boolean,
+	appsToLock: AppsToLockMap,
+): Nullable<CompositionStep> {
+	if (!currentServicesLocked) {
+		currentServices.forEach((svc) =>
+			appsToLock[svc.appId].add(svc.serviceName),
+		);
+		return;
```
pipex, line 31: Why return undefined vs a `noop` step here?

src/compose/update-strategies.ts
```diff
@@ -12,12 +13,30 @@ export interface StrategyContext {
...
+	}
+	return generateStep('kill', { current });
 }
 
 export function getStepsFromStrategy(
 	strategy: string,
 	context: StrategyContext,
-): CompositionStep {
+): Nullable<CompositionStep> {
```
pipex, line 39: If it makes sense to return `noop` here I think that's better. Returning `null` or "something" makes functions signatures more complicated and means that function callers now need to add a validation step

src/compose/update-strategies.ts
```diff
@@ -12,12 +13,30 @@ export interface StrategyContext {
 	dependenciesMetForStart: boolean;
 	dependenciesMetForKill: boolean;
 	needsSpecialKill: boolean;
+	currentServices: Service[];
+	currentServicesLocked: boolean;
+	appsToLock: AppsToLockMap;
+}
+
+function generateLockThenKillStep(
```
pipex, line 21: This function is strange, as it either returns a step or updates an array and returns `null`, that is probably a sign that those things should happen in different places

src/lib/update-lock.ts
cywang117: I didn't do that as the functionality of `add` behaves different than `Map.prototype.set`: `add` adds to the list of locks that are taken for an appId, whereas `Map.prototype.set` replaces the value for a key. I'd prefer to keep it this way for that reason, but open to discussion of course.

src/lib/update-lock.ts
cywang117: I wanted to avoid direct modification and have `getServices` act as a readonly method, hence why I avoided the `get` extension here. For the purposes for which LocksTakenMap is used, there is no use case for a `get` which returns a mutable reference.

src/lib/update-lock.ts
cywang117: If both or one of either file exists, the lock is considered taken by the Supervisor, as long as the locks taken are stored in memory. 

With the discussed change to 1) remove the in-memory store of lockfiles taken, 2) move to checking the filesystem, and 3) retain Supervisor-taken lockfiles between Supervisor restarts, locks would be considered taken by the Supervisor if one or both of `(resin-)updates.lock` exists in the filesystem with the `nobody` UID. These discussed changes will be added in a separate commit (will be linked here once complete). In the meantime I added a test case [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/c83b931d8643e9f842564ba81b074e44942c819e#diff-e7a2b32a616114c913967bbeb93301d1194c25d37b5c6d6393146a862d33b61fR339) which affirms that if only `resin-updates.lock` is present for some reason, the lock is still considered taken.

test/integration/lib/update-lock.spec.ts
cywang117: I commented about this in another comment chain, but this didn't work consistently for me :( will be sure to fix in a future PR

src/lib/update-lock.ts
cywang117: > Could `takeLock` perhaps call `lock` and not receive an array of services?

I'm not sure this would work, as update-lock does not know which services should be locked under an appId. If the service directories under an appId are always assumed to be there by the update-lock module, this wouldn't be an issue because update-lock could be just read a list of services under an appId and proceed with that as the services to be locked.

However, `takeLock` can be called for a service in the target state that hasn't started up, meaning the `/tmp` bind mounts for locking haven't necessarily been created yet by the Supervisor. Therefore `takeLock` needs an array of services to know which directories to create

cywang117: I don't think `lock` would work as it only temporarily locks an app to perform an action; A `dispose` which removes all locks is performed in the same `lock` function which removes all locks regardless of whether the lock errored or not. Is that what you're referring to by lock cleanup? I don't see any other places where the lock may be cleaned up.

cywang117: I see your point though about succeeding on one of `(resin-)update.lock` but failing on the other; if `lockService` fails in this case, both lockfiles need to be removed which currently `lockService` doesn't do. This PR which touches `lockService` directly has exposed the flaws in `lockService` when it's not immediately paired with a `dispose` as is the case in `lock`. Thus, I think we can patch `lockService` to remove both lockfiles instead, what do you think?

As a side note, I mentioned that one of the 2 lockfiles failing to lock means the service should not count as locked, but if only one of the two files exists already, the service should count as locked. This is because `resin-updates.lock` is a legacy file, and while we should still support it, we shouldn't make the success of lock-taking contingent on its existence.

src/lib/update-lock.ts
pipex: I think it needs to consider the lock taken only if both files exist, as we don't know which file the container is using

src/lib/update-lock.ts
pipex: > Therefore takeLock needs an array of services to know which directories to create

You could have an earlier step to create the target directories manually and `takeLock` with just the appId can try to take all of the locks for the app at once. If one fails, it needs to dispose the locks and throw so it can try again

Alternatively, `takeLock` may receive a list of services (which is assumed to contain all services of the app) and first create the missing directories and then try to take all the locks. Again, if something fails, it throws and the state engine needs to run again

> I don't think lock would work as it only temporarily locks an app to perform an action

You are right, I forgot `lock` receives a function. The point stands, `lockService` could fail in the middle of taking the two lockfiles for a service which could be a problem (it won't be if we require both files to exist with `nobody` to consider the lock taken). As with lock, I think we need to figure out a way to take all the locks or fail as I pointed out before to avoid potential deadlock with the user app

src/compose/app.ts
cywang117: It would prevent it, but dependent services are destined for removal anyway so I don't think this is a big issue

src/compose/app.ts
```diff
@@ -132,41 +136,62 @@ export class App {
...
-					}),
-				)
-				.filter((step) => step != null) as CompositionStep[],
-		);
+		// generateStepsForService will populate appsToLock with services that
+		// need to be locked, including services that need to be removed due to
+		// network or volume changes.
+		const appsToLock: AppsToLockMap = {
+			[this.appId]: new Set<string>(),
+			[target.appId]: new Set<string>(),
```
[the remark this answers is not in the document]
cywang117: If there's never a case where appIds would be different between this & target, this will be safe to remove. I was afraid of there being some case I may have missed even if I couldn't see it in the codebase, so I left this in

src/compose/app.ts
```diff
@@ -194,10 +223,37 @@ export class App {
 		return steps;
 	}
 
+	private hasLeftoverLocks(
+		appServices: Service[],
+		locksTaken: LocksTakenMap,
+	): boolean {
+		return (
+			appServices.length > 0 &&
+			appServices.some(({ appId, serviceName }) =>
```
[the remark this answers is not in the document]
cywang117: I thought app.ts just compares app-specific changes, meaning there's only ever one app?

src/compose/app.ts
```diff
@@ -194,10 +223,37 @@ export class App {
...
 	): CompositionStep[] {
 		if (Object.keys(this.services).length > 0) {
+			// Take all locks before killing
+			if (
+				this.services.some(
+					(svc) => !state.locksTaken.isLocked(svc.appId, svc.serviceName),
+				)
+			) {
+				return [
+					generateStep('takeLock', {
```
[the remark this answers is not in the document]
cywang117: Line 186 is part of `nextStepsForAppUpdate`  which is only called for an app in target state. If an app isn't in target and is only in current, `stepsToRemoveApp` is called instead, so this is necessary.

src/compose/app.ts
cywang117: The kill step generation for dependent services got moved to `generateStepsForService`, which already accounts for `Stopping` state by emitting `noop`

src/compose/app.ts
cywang117: > Could we take locks for all services after downloads are finished? 

I don't think it's that simple, what about an app where update strategies differ between services? Locks need to be taken before kill, so the lock would need to be taken for the entire app before downloads finish if a service has `kill-then-download`, leaving the lock in place unnecessarily for services with `download-then-kill` which may still be early in their download progress

cywang117: Actually, I think https://github.com/balena-os/balena-supervisor/commit/e836028efcb284827010b0d38f99242bd9127063 already makes it so that a lock is taken for all services before a kill if `kill-then-download`. This could be undesirable for services with `download-then-kill` which are then stuck with the lock even when they haven't finished downloading yet

src/lib/update-lock.ts
cywang117: Sounds good, I think we could change takeLock steps' implementation to catch any locking failures, remove any taken locks, then throw the error again to be picked up by the state engine so it can retry

src/lib/update-lock.ts
cywang117: Removed SIGTERM lockfile removal, lockfile in-memory tracking, and tightened requirements for a service to be counted as locked here: https://github.com/balena-os/balena-supervisor/pull/2234/commits/99ddf8387317b23bfec1e5aaa0161cbf7992ec51

test/integration/lib/update-lock.spec.ts
cywang117: Removed the spy here: https://github.com/balena-os/balena-supervisor/pull/2234/commits/b56b0431f77aca675f1eb938c9be15110cc3c8cb

following your suggested solution.

src/lib/update-lock.ts
cywang117: Added the `nobody` uid & both lockfiles present restrictions in this new commit: https://github.com/balena-os/balena-supervisor/commit/99ddf8387317b23bfec1e5aaa0161cbf7992ec51

src/compose/update-strategies.ts
cywang117: I don't think we need to add to the number of loops the state funnel has to go through with a noop. `appsToLock` later gets parsed into a single takeLock step, the way for it to remain a single takeLock step and not a bunch of disjointed takeLock steps is for a map to be passed around between `generateStepsForService` for each service, which tracks which services require the lock.

src/compose/update-strategies.ts
cywang117: The alternative to this is to repeat the contents of `generateLockThenKillSteps` 3 times, once under `download-then-kill`, once under `kill-then-download`, and once under `hand-over`. I didn't want the repetition, and the functionality here is necessary: if we're not generating a kill step, we're modifying the appsToLock map to be parsed after service step generation into a takeLock step.

src/compose/app.ts
pipex: The problem is that the update would take longer, because a volume without dependencies can be immediately removed

src/compose/app.ts
pipex: If there are any services with `delete-then-download` or `kill-then-download` then all locks should be taken before executing any further steps.

src/compose/app.ts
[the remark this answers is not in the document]
pipex: As I mentioned, if moving between environment, `appUuid` for both apps could be the same but with different `appId` in which case this would make sense, although I suspect a bunch of other things would not work properly there. I also think our whole locking system presumes that apps are identified by `appId` so I think we can safely assume that both current and target apps have the same a`appId` for now

src/compose/app.ts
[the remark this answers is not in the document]
pipex: You are right, I think I misunderstood what this was doing

src/compose/update-strategies.ts
pipex: I understand. In that case I would prefer that you changed the signature to return an array of next steps, and return an empty array if no further steps are necessary. Sometimes returning a null is not a good interface imho

src/lib/update-lock.ts
pipex: That is not obvious from the class definition. Some future supervisor developer can still use the other `Map` methods to do things "the wrong way". What you could do is make the `get`method private in that case, to prevent usage.

src/compose/app.ts
cywang117: Okay, since the code already does that, we should be good to not modify any more logic around update strategies

src/compose/app.ts
```diff
@@ -302,23 +358,51 @@ export class App {
 		return outputs;
 	}
 
+	private getDependentServices<T extends Volume | Network>(
+		component: T,
+		dependencyFn: (component: T, service: Service) => boolean,
+	) {
+		return this.services.filter((s) => dependencyFn(component, s));
```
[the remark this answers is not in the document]
cywang117: This won't work if the service is in both current and target and needs to be killed to kill a network/volume it's depending on. In that case, `this.services` will not include the service after the kill, whereas `target.services` will always include the service, preventing the network/volume from being removed (since `dependentServices.length === 0` is the prerequisite to kill a network/volume.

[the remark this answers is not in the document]
cywang117: Another reason `dependentServices` should read from current services instead of target, is that they get union-ed into the set of services to be removed in `compareServices`. It doesn't make sense for target dependent services to be marked for removal since we'd still have to check if they're in current state to avoid a bad kill step.

src/compose/app.ts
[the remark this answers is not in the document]
cywang117: Before when? In `inferNextSteps`?

src/lib/update-lock.ts
cywang117: There's no way to extend a Map but modify the types of the interfaces of `get`, so I did the next best thing which was to set the `@private` annotation with a comment on `get`. [Ref](https://github.com/balena-os/balena-supervisor/pull/2234/commits/24d7a2fb63fb9a93e215e3615e5bd8b98279f107#diff-376068891dd599a79a4b6bf84d850688e14f1d5ebaf089f88518f08e40fc6f4bR126-R132)

src/lib/update-lock.ts
cywang117: I added the lockfile cleanup on lock failure in a separate commit: 48e869104620f6c8447e5abbe2af9ee40bce9605

src/compose/app.ts
```diff
@@ -132,41 +136,62 @@ export class App {
...
-			this.generateStepsForComponent(networkChanges, servicePairs, (n, svc) =>
-				svc.hasNetwork(n.name),
-			),
+			this.generateStepsForComponent(networkChanges, dependentServices),
 		);
 
+		// Generate lock steps from appsToLock
+		for (const [appId, services] of Object.entries(appsToLock)) {
+			if (services.size > 0) {
+				steps.push(
```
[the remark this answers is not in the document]
cywang117: I modified this [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/f199319d0a2ecf3dcbfc59c82f022b662d784a71#diff-857bb737cdeb3c33e2caa352769d3f96e0e8a48da6fea254eb99bc627732cec1R170) to generate takeLock steps before service/network/volume steps

src/compose/app.ts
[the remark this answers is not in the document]
cywang117: I changed this [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/f199319d0a2ecf3dcbfc59c82f022b662d784a71#diff-857bb737cdeb3c33e2caa352769d3f96e0e8a48da6fea254eb99bc627732cec1R153-R156) to only construct a set where the key is target.appId

src/compose/app.ts
cywang117: True. I modified the dependent service calculation to filter for dependent services of the current component (network/volume) only [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/f199319d0a2ecf3dcbfc59c82f022b662d784a71#diff-857bb737cdeb3c33e2caa352769d3f96e0e8a48da6fea254eb99bc627732cec1R568-R587)

src/compose/app.ts
cywang117: I combined `currentServicesLocked` and `targetServicesLocked` in commit f199319d0a2ecf3dcbfc59c82f022b662d784a71. Any lock-taking calculations will also generate steps for both current & target services now, in that same commit

src/compose/update-strategies.ts
cywang117: Done! Also in f199319d0a2ecf3dcbfc59c82f022b662d784a71

src/lib/update-lock.ts
pipex: Oh, I thought it was possible. I guess you could also create a new interface that operates on an internal map and not extend `Map`. I think that might result in cleaner code, but I leave it up to you

src/compose/app.ts
```diff
@@ -444,45 +539,50 @@ export class App {
...
-					);
-				} else {
+				// If there are any dependent services which have the volume or network,
+				// we cannot proceed to component removal.
+				const dependentServicesOfComponent = dependentServices.filter((s) => {
+					if (componentIsVolume) {
+						return this.serviceHasNetworkOrVolume(
+							s,
+							[],
+							components as Array<ChangingPair<Volume>>,
```
pipex, line 575: Why pass the full component array here? Should this be 
```suggestion
							[component],
```

src/compose/app.ts
```diff
@@ -647,38 +742,70 @@ export class App {
...
+		appsToLock: AppsToLockMap,
+		targetServices: Service[],
+		servicesLocked: boolean,
+	): CompositionStep[] {
+		// Update container metadata if service release has changed
 		if (current.commit !== target.commit) {
-			return generateStep('updateMetadata', { current, target });
+			// Only update service once locked or if legacy container
+			const shouldSkipLock = checkTruthy(
+				current.config.labels['io.balena.legacy-container'],
```
pipex, line 756: What are legacy containers? Can there still be legacy containers at this point?

src/device-api/actions.ts
```diff
@@ -279,15 +265,35 @@ export const executeServiceAction = async ({
...
-		generateStep(action, {
-			current: currentService,
-			target: targetService,
-			wait: true,
-		}),
-		force,
-	);
+	try {
+		// Take lock for service
+		await executeDeviceAction(
```
pipex, line 270: I'm not sure this should be here. We don't need to take lock on `start`, for instance. The method taking calling `executeServiceAction` should take the lock. Alternatively you could have a wrapping function `executeServiceActionWithLock` that takes the lock before calling `executeServiceAction`

pipex: What happens if the lock is taken here BTW? Will `executeDeviceActiond` throw? What error will be reported to the user then?

src/lib/lockfile.ts
```diff
@@ -1,34 +1,44 @@
...
-
-// Try to clean up any existing locks when the process exits
-process.on('exit', () => {
-	for (const lockName of getLocksTaken()) {
-		try {
-			unlockSync(lockName);
-		} catch (e) {
-			// Ignore unlocking errors
+// A file is counted as a lock by default if it ends with `.lock`.
+export const getLocksTaken = async (
```
pipex, line 15: I don't think we need to query the state of the directory each time, which is a bit wasteful. I think maybe this module could have something like an `initialize` function that receives the initial state of locks and then when new locks are taken and released the state gets updated there. Then `getLocksTaken` remains the same.

On the other hand, the `update-lock` module, also can have an initialization method that reads the directories once (that is called by application manager).

test/integration/device-api/actions.spec.ts
```diff
@@ -251,6 +265,8 @@ describe('manages application lifecycle', () => {
 		});
 
 		beforeEach(async () => {
+			// Delete all update locks
+			await updateLock.releaseLock(APP_ID);
```
pipex, line 269: You probably should delete locks afterwards as well, a bug on a test means leftover locks could be left and cause other tests to fail. Those kind of test failures are pretty hard to debug.

Alternatively, you could also use testfs, which also sets up some hooks to run after all tests.

src/compose/app.ts
cywang117: According to https://github.com/balena-os/balena-supervisor/blob/addffa3dd79ac27ba69b93606e9e4e3fa91c435b/src/compose/app.ts#L265 and https://github.com/balena-os/balena-supervisor/blob/6217546894e25c8216d305f231876fc888a6e407/src/migrations/20171129064057_multicontainer.js#L12, looks like it's a container from before we added multicontainer support. There's a possibility a Supervisor could be running a legacy app, perhaps best to keep this in

src/device-api/actions.ts
cywang117: If the lock is taken, the endpoint would return 423 Locked and `{"status":"failed","message":"Lockfile exists for { appId: $APP_ID, service: $SERVICE_NAME }"` due to the endpoint error handler which catches all 423's

src/lib/lockfile.ts
cywang117: Added back the in-memory store in https://github.com/balena-os/balena-supervisor/pull/2234/commits/d8d211cf7610008ac5754d35654e8f3f06856e05, skipped some lock tests in actions as a result, as it's not possible to manipulate the in-memory lockfile store of balena-supervisor-sut from the `sut` test process

test/integration/device-api/actions.spec.ts
cywang117: Done, used testfs

src/device-api/actions.ts
cywang117: Added lock wrapping function: https://github.com/balena-os/balena-supervisor/pull/2234/commits/c1082b6b90aa31ba256cac2f43a80b24340b25cb#diff-d953a5d18d116d746cd0b200a5d7e917714811bc4609e71fc19196941bcd7ef0R218

src/compose/app.ts
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2234/commits/10eaf8f347cc5a3536db45e9ec04fa9981be4f23#diff-857bb737cdeb3c33e2caa352769d3f96e0e8a48da6fea254eb99bc627732cec1R575), also agreed, it's not a great function to work with

src/compose/app.ts
pipex: Can legacy containers have locks? If they cannot why should we explicitely skip them? I don't want us to add code just in case

cywang117: We had update locks as far back as Supervisor v4.0.0 according to our docs, and Supervisor added multicontainer support in 7.0.0. So I think the answer to that is yes, legacy containers can have update locks. However, given how old this is and the difficulty of reproduction, if you feel strongly about removing this, that SGTM

pipex: Is not that I feel strongly about it, I'm just trying to understand the logic, in order for a user to still have legacy containers they would have to still be running the same containers from when they updated to v7, meaning no release updates from them, no OS update, and they have not yet jumped to a supervisor that recreates their containers. 

Assuming that can happen, I still don't understand why we would skip lock for those specific containers

src/lib/lockfile.ts
pipex: I added a comment on one of those tests. While you cannot easily fake the in memory lock store, locks taken by the user app should still prevent certain requests from taking place.

src/lib/update-lock.ts
```diff
@@ -82,22 +90,188 @@ async function dispose(
...
+// A wrapper function for lockfile.getLocksTaken.
+// Exported for tests only; getServicesLockedByAppId is the intended public interface.
+export const getLocksTaken = lockfile.getLocksTaken;
+
+/**
+ * Return a list of services that are locked by the Supervisor under each appId.
+ * Both `resin-updates.lock` and `updates.lock` should be present per
+ * [appId, serviceName] pair for a service to be considered locked.
+ */
+export async function getServicesLockedByAppId(): Promise<LocksTakenMap> {
```
pipex, line 219: This doesn't need to be async now

src/lib/lockfile.ts
pipex: Also, just for the sake of argument. Assuming the user app is behaving properly (i.e. not overwriting our locks), can you think of a case where the in-memory store can get out of sync with the filesystem? I want to make sure we are covering all of our bases

src/device-api/actions.ts
pipex: I don't think I expressed well on my comment, but the result is not what I meant. The `executeServiceAction` function is still taking a lock on `start` actions. I'll add some more comments on the commit

src/lib/lockfile.ts
cywang117: We discussed this last checkin (Mar 28 2024) and agreed to remove the in-memory store in case lockfile interference causes the store to get out-of-sync with the actual lock state in the filesystem.

src/lib/update-lock.ts
cywang117: Reverted the in-memory store, so the function is async again

src/compose/app.ts
cywang117: We discussed this further in our Supervisor checkin, and agreed it would be safe to remove this `skipLock` check. The chance of a device on such an old OS / Supervisor running an old release is really small, and if such a device were to appear on support, we should recommend a Supervisor / OS upgrade for them first thing anyway

cywang117:
@balena-ci I self-certify!

PR #2235, opened 2024-01-24
Supervisor Developer experience improvements

pipex:
- Add prettierrc to root folder to enable auto formatting depending on the editor configuration
- Improve test:compose and test:env commands, the supervisor container in the test environment will now re-load on code changes
- Some test fixes


pipex:
I self-certify!

PR #2238, opened 2024-02-06
Fix processing of dtoverlay/dtparams on config.txt

pipex:
DT overlays and DT params need to be consumed in the order that they
appear on the file. DT params apply to the last dtoverlay defined on the
file, or to the base overlay.

This commit updates config.txt parsing to consider this ordering, and it
also ensures global dtparams are written first so they cannot be
overriden by later overlays.

Because of the more strict parsing method, it is possible that existing
HOST_CONFIG vars do not match the interpretation of the parser. If
that's the case, the supervisor will re-apply the target state which
will cause the device to reboot.

This PR includes a mitigation for this by using a list of known base dt params, 
the list includes those params that come in the default config.txt included in the balenaOS image.
Nevertheless, a user may have added another base dtparam, in which case, the current and
target state won't match and a reboot will be needed.

Change-type: major

### DIFF
--- a/src/config/backends/config-txt.ts
+++ b/src/config/backends/config-txt.ts
@@ -31,6 +31,21 @@ type DTParam = string;
 type DTOverlays = { [name: string]: DTParam[] };
 const BASE_OVERLAY = '';
 
+function isBaseParam(dtparam: string): boolean {
+	const match = /^([^=]+)=(.*)$/.exec(dtparam);
+	let key = dtparam;
+	if (match != null) {
+		key = match[1];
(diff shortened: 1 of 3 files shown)


cywang117 (commented):
This LGTM @pipex. As you mentioned, if there is a chance of reboot if the user has additional dtparams set besides the default, it would be better for this to be a major release. Nice work!

pipex:
You are right @cywang117 I've made the PR a major

pipex:
I self-certify!

PR #2241, opened 2024-02-17
Patch config.txt backend to return array configs correctly

cywang117:
Previously, getBootConfig() of the config.txt backend was omitting array configurations such as gpio settings, thus resulting in the SV mistakenly assuming that boot config had not been applied, since gpio would not be in current config.txt config but would be in target config. This resulted in SV entering an infinite loop of attempting to apply the gpio config when it wasn't necessary.

Change-type: patch


cywang117:
I self-certify!

PR #2243, opened 2024-02-21
Patch default dtparam handling in config.txt

cywang117:
This commit completes the list of default / board-wide dtparams to include some `baudrate` and `vc` i2c params.

Change-type: patch


cywang117:
I self-certify!

PR #2246, opened 2024-02-29
Update @balena/lint to v7

pipex:
This updates balena lint to the latest version to enable eslint support
and unblock Typescript updates. This is a huge number of changes as the
linting rules are much more strict now, requiring modifiying most files
in the codebase. This commit also bumps the test dependency `rewire` as 
that was interfering with the update of balena-lint. 

Change-type: patch


pipex:
Hmmm, event logs got way too verbose with the removal of the json-mask module. I'll revisit that change

pipex:
I self-certify!

PR #2247, opened 2024-03-05
Update to typescript v5

pipex:
This also updates code to use the default import syntax instead of
`import * as` when the imported module exposes a default. This is needed
with the latest typescript version.

Change-type: patch


pipex:
I self-certify!

PR #2248, opened 2024-03-06
Update Node to v20

pipex:
This updates the supervisor runtime to latest Node LTS version. There are no breaking changes related to this bump.

Change-type: patch


pipex:
I self-certify!

PR #2254, opened 2024-03-25
Log the full error on device state report failure as it is more useful

Page-:
The message can be an empty string or similarly unhelpful, therefore logging the entire error means that we will have whatever the message may be along with the stack trace and other info that will be helpful even when the message is not

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


Page-:
Specifically this was an issue I had whilst investigating for #2253 where the error message was an empty string whereas the full error object gave far more info to know what was actually happening

pipex (approved):

PR #2256, opened 2024-03-25
Inherit Renovate settings from balena-io

klutchell:
Inherit renovate settings from [balena-io/renovate-config](https://github.com/balena-io/renovate-config/blob/master/default.json).

See: https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/Renovate.20for.20balena-supervisor


klutchell:
I self-certify!

PR #2259, opened 2024-04-03
Update docker related dependencies

pipex:
This bumps dockerode, removes resin-docker-build in favor of
@balena/compose, and updates docker-delta and docker-progress packages.

Change-type: patch


webpack.config.js
```diff
@@ -131,6 +131,10 @@ module.exports = function (env) {
 						},
 					],
 				},
+				{
+					test: /\.node$/,
+					loader: 'node-loader',
```
pipex, line 136: Without this, the webpack build fails with 

```
ERROR in ./node_modules/ssh2/lib/protocol/crypto/build/Release/sshcrypto.node 1:0
Module parse failed: Unexpected character '�' (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
(Source code omitted for this binary file)
```

`ssh2` seems to be a dependency of docker-modem 

```
balena-supervisor@16.1.10 /Users/flalanne/Development/balena/balena-supervisor/dockerode
├─┬ @balena/compose@3.2.1
│ └─┬ docker-modem@5.0.3
│   └── ssh2@1.15.0
├─┬ docker-delta@4.0.0
│ └─┬ dockerode@3.3.5
│   └─┬ docker-modem@3.0.8
│     └── ssh2@1.15.0 deduped
└─┬ mocha-pod@0.9.0
  └─┬ @balena/compose@2.3.0
    └─┬ docker-modem@3.0.6
      └── ssh2@1.15.0 deduped
```

pipex: We don't use that feature, but if we did, that file is architecture dependent, so that is an issue.

I also tried adding it to this list https://github.com/balena-os/balena-supervisor/blob/73e4f186a0635c9aa7d2d4d4114d4041c8d452be/webpack.config.js#L9-L26, but that causes an error when the module is not found

pipex:
I self-certify!

webpack.config.js
Page-: @pipex since the supervisor doesn't use docker over ssh I'd assume it to be fine but I don't know for sure either way

pipex: That is my expectation as well, I did test the built supervisor with armv7 just in case and everything seemed to work properly

PR #2263, opened 2024-04-06
Update lock state apply: patch DockerName & respect lockOverride

cywang117:
This PR contains 2 patches:

- Fix some RegEx io-ts types, this was affecting which services were passing the `DockerName` type guard. Services with dashes in their names were not passing before when they should have.
- Respect `lockOverride` when taking locks through state funnel or as an API action


cywang117:
@balena-ci I self-certify!

PR #2271, opened 2024-04-12
Add additional update lock tests for lockOverride & force

cywang117:
Change-type: patch


cywang117:
@balena-ci I self-certify!

PR #2284, opened 2024-04-17
Enable redsocks dnsu2t

cywang117:
## Release Notes
Support dnsu2t redsocks proxy config
    
Users may specify dnsu2t config by including a `dns` field
in their PATCH body:
```
{
  network: {
    proxy: {
      dns: '1.1.1.1:53'
    },
    hostname: 'balena'
  }
}
```

If `dns` is a string, IP and PORT are optional but should be in
the format [IP][:PORT] if present. For any fields that are empty,
a default IP and PORT (`8.8.8.8:53`) will be configured.

`dns` may also be a boolean. If true, defaults of `8.8.8.8:53` will be configured.
If false, the dns configuration will be removed.

If `proxy` is patched to empty, `dns` will be removed regardless of its
current or input configs, as `dns` depends on an active redsocks
proxy to function.

Change-type: minor

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -568,7 +568,8 @@ By default, with [balenaOS 2.82.6](https://github.com/balena-os/meta-balena/blob
 			"port": 8123,
 			"login": "username",
 			"password": "password",
-			"noProxy": [ "152.10.30.4", "253.1.1.0/16" ]
+			"noProxy": [ "152.10.30.4", "253.1.1.0/16" ],
+			"dns": "1.1.1.1:53"
 		},
 		"hostname": "mynewhostname",
 		"force": true
@@ -585,7 +586,9 @@ Keep in mind that, even if transparent proxy redirection will take effect immedi
 The `noProxy` setting for the proxy is an optional array of IP addresses/subnets that should not be routed through the
 proxy. Keep in mind that local/reserved subnets are already [excluded by balenaOS automatically](https://github.com/balena-os/meta-balena/blob/master/meta-balena-common/recipes-connectivity/balena-proxy-config/balena-proxy-config/balena-proxy-config).
 
-If either `proxy` or `hostname` are null or empty values (i.e. `{}` for proxy or an empty string for hostname), they will be cleared to their default values (i.e. not using a proxy, and a hostname equal to the first 7 characters of the device's uuid, respectively).
+As of v16.6.0, the Supervisor supports configuring the `dnsu2t` plugin for redsocks via the `dns` subfield under `proxy`. Only the `remote_ip` and `remote_port` fields are allowed to be modified. The input must be of type `boolean` or `string`. If boolean and `true`, the remote values will be the default, `8.8.8.8:53`. If boolean and false, the configuration will be removed. If string, it should be in the format `ADDRESS:PORT`, with `ADDRESS` and `PORT` both required. You may not configure `dnsu2t` without a redsocks proxy configuration, as `dnsu2t` depends on a working proxy to function.
+
+If any of `proxy`, `dns`, or `hostname` are nullish, falsey, or empty values (i.e. `{}` for proxy, `false` for dns, or an empty string for hostname), they will be cleared to their default settings (i.e. not using a proxy, not using dns, and a hostname equal to the first 7 characters of the device's uuid, respectively).
 
 #### Examples:
 From an app container:
@@ -636,7 +639,8 @@ Response:
 			"port":"8123",
 			"type":"socks5"
 		},
-		"hostname":"27b0fdc"
+		"hostname":"27b0fdc",
+		"dns": "1.1.1.1:53"
 	}
 }
 ```
--- a/src/host-config/index.ts
+++ b/src/host-config/index.ts
@@ -51,9 +51,15 @@ export function parse(
 		const legacyDecoded = LegacyHostConfiguration.decode(conf);
 		if (isRight(legacyDecoded)) {
 			return legacyDecoded.right;
+		} else {
+			const formattedErrorMessage = [
+				'Could not parse host config input to a valid format:',
+			]
+				.concat(Reporter.report(legacyDecoded))
+				.join('\n');
+			throw new Error(formattedErrorMessage);
 		}
 	}
-	throw new Error('Could not parse host config input to a valid format');
 }
 
 function patchProxy(
--- a/src/host-config/proxy.ts
+++ b/src/host-config/proxy.ts
@@ -4,7 +4,7 @@ import { isRight } from 'fp-ts/lib/Either';
 import Reporter from 'io-ts-reporters';
 
 import type { RedsocksConfig, HostProxyConfig, DnsInput } from './types';
-import { ProxyConfig } from './types';
+import { ProxyConfig, DnsConfig } from './types';
 import { pathOnBoot, readFromBoot, writeToBoot } from '../lib/host-utils';
 import { unlinkAll, mkdirp } from '../lib/fs-utils';
 import { isENOENT } from '../lib/errors';
@@ -89,6 +89,20 @@ export class RedsocksConf {
 			return conf;
 		}
 
+		// Extract contents of `dnsu2t {...}` using regex if exists
+		let dns: DnsConfig | null = null;
+		const rawDnsu2tBlockMatch = rawConf.match(blockRegexFor('dnsu2t'));
+		if (rawDnsu2tBlockMatch) {
+			const rawDnsu2tBlock = RedsocksConf.parseBlock(
+				rawDnsu2tBlockMatch[1],
+				disallowedProxyFields,
+			);
+			const maybeDnsConfig = DnsConfig.decode(rawDnsu2tBlock);
+			if (isRight(maybeDnsConfig)) {
+				dns = maybeDnsConfig.right;
+			}
+		}
+
 		// Extract contents of `redsocks {...}` using regex
 		const rawRedsocksBlockMatch = rawConf.match(blockRegexFor('redsocks'));
 		// No group was captured, indicating malformed config
@@ -104,6 +118,8 @@ export class RedsocksConf {
 		if (isRight(maybeProxyConfig)) {
 			conf.redsocks = {
 				...maybeProxyConfig.right,
+				// Only add dns subfield if redsocks config is valid
+				...(dns && { dns: `${dns.remote_ip}:${dns.remote_port}` }),
 			};
 			return conf;
 		} else {
--- a/src/host-config/proxy.ts
+++ b/src/host-config/proxy.ts
@@ -3,7 +3,7 @@ import path from 'path';
 import { isRight } from 'fp-ts/lib/Either';
 import Reporter from 'io-ts-reporters';
 
-import type { RedsocksConfig, HostProxyConfig } from './types';
+import type { RedsocksConfig, HostProxyConfig, DnsInput } from './types';
 import { ProxyConfig } from './types';
 import { pathOnBoot, readFromBoot, writeToBoot } from '../lib/host-utils';
 import { unlinkAll, mkdirp } from '../lib/fs-utils';
@@ -16,6 +16,9 @@ const noProxyPath = path.join(proxyBasePath, 'no_proxy');
 const redsocksConfPath = path.join(proxyBasePath, 'redsocks.conf');
 
 const disallowedProxyFields = ['local_ip', 'local_port'];
+
+const DEFAULT_REMOTE_IP = '8.8.8.8';
+const DEFAULT_REMOTE_PORT = 53;
 
 const isAuthField = (field: string): boolean =>
 	['login', 'password'].includes(field);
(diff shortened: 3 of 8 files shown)


src/device-api/v1.ts
```diff
@@ -204,6 +204,17 @@ router.patch('/v1/device/host-config', async (req, res) => {
...
+		// Validate dnsu2t config if it exists
+		const { dnsu2t } = req.body.network;
+		if (dnsu2t && typeof dnsu2t !== 'boolean' && typeof dnsu2t !== 'string') {
+			return res
+				.status(400)
+				.send(
+					'Expected dnsu2t to be a boolean or string, received: ' +
+						typeof dnsu2t,
+				);
+		}
```
cywang117, line 217: I didn't add validation for whether a dnsu2t string matches an [IP][:PORT] regex. Should we?

docs/API.md
```diff
@@ -585,7 +586,9 @@ Keep in mind that, even if transparent proxy redirection will take effect immedi
 The `noProxy` setting for the proxy is an optional array of IP addresses/subnets that should not be routed through the
 proxy. Keep in mind that local/reserved subnets are already [excluded by balenaOS automatically](https://github.com/balena-os/meta-balena/blob/master/meta-balena-common/recipes-connectivity/balena-proxy-config/balena-proxy-config/balena-proxy-config).
 
-If either `proxy` or `hostname` are null or empty values (i.e. `{}` for proxy or an empty string for hostname), they will be cleared to their default values (i.e. not using a proxy, and a hostname equal to the first 7 characters of the device's uuid, respectively).
+As of vTODO, the Supervisor supports configuring the `dnsu2t` plugin for redsocks. Only the `remote_ip` and `remote_port` fields are allowed to be modified. The input must be of type `boolean` or `string`. If boolean and `true`, the remote values will be `8.8.8.8:53`. If boolean and false, the `dnsu2t` configuration will be removed. If string, it should be in the format `[IP][:PORT]`, with current `dnsu2t` values filling in where not specified, then default values filling in where current values aren't specified.
```
cywang117, line 589: Modify this to actual Supervisor version once ready to merge

src/device-api/v1.ts
cywang117: Discussed that we aren't going to add validation as the rest of the proxy patch body isn't validated. It's up to user discretion to get their configs right

src/host-config.ts
```diff
@@ -45,41 +52,83 @@ const memoizedRegex = _.memoize(
...
+	if (dnsBlock != null) {
+		const dnsLines = dnsBlock.split('\n');
+		for (const line of dnsLines) {
+			for (const dnsField of dnsFields) {
+				const match = line.match(memoizedRegex(dnsField));
+				if (match != null) {
+					// Transform snakecase dnsField to camelCase
+					const dnsFieldCamelcase = dnsField.replace(
+						/* eslint-disable @typescript-eslint/no-shadow */
+						// This is necessary because ESLint mistakes _ here
```
pipex, line 112: The argument doesn't need to be called `_`, it just has to start with `_`, e.g. `_ignore`

pipex (changes requested):
I think the whole host-config module requires a refactor, the code is getting a bit messy with this changes, even though they follow the existing patterns. 

I think the problem is that the module is trying to mix parsing/building and conversion to and from redsocks config.

Considering just redsocks configuration for now, there is a simple conversion to/from a javascript object

```ts
interface RedsocksConfig {
  base: {
    log_debug: string;
    log_info: string;
    log: string;
    daemon: string;
    redirector: string;
  };
  redsocks: {
    local_ip: string;
    local_port: number;
    type: string;
    ip: string;
    port: number;
    login: string;
    password: string;
  };
  dnsu2t: {
    local_ip: string;
    local_port: number;
    remote_ip: string;
    remote_port: number;
  };
}
```

Conversion to/from JSON can go in its own module with `parse` and `stringify` methods. That can be unit tested for better robustness.

Having that module, processing the endpoint input means converting from 

```ts
interface HostConfig {
  type?: string;
  ip?: string;
  port?: number;
  login?: string;
  password?: string;
  dns?: boolean | string;
}
``` 

Into `RedsocksConfig` which can be done in a cleaner way as it's just object manipulation. The conversion can also be unit tested

The endpoint in the end can be written something like the following

```ts
/* validate input */
const currentConfig = RedsocksConfig.parse(readFile(rPath));
const targetConfig = patchConfig(input, current);
writeFile(RedsocksConfig.stringify(targetConfig));
``

I think refactoring will result in more testable and readable code.

**Comment by cywang117** [comment]
@pipex Addressing your last comment about a refactor, I broke that off into a separate PR here: https://github.com/balena-os/balena-supervisor/pull/2329

The separate PR maintains the same host-config interface. After that merges, I can rework the dnsu2t addition here to use the refactored host-config module.

File: src/host-config/index.ts [hunk 4]
```diff
@@ -57,75 +62,149 @@ export function parse(
...
+						([, value]) => value,
+					),
+				) as ProxyConfig;
+				if (Object.keys(definedConf).length > 0) {
+					toPatch.redsocks = definedConf;
+				} else {
+					// Mark for deletion if target proxy is empty
+					toPatch.redsocks = null;
+				}
+			}
```
pipex, line 166: Why is this chunk of code necessary now and it wasn't before?

src/host-config/index.ts
```diff
@@ -57,75 +62,149 @@ export function parse(
...
+					toPatch.dns = null;
+				}
+			}
+			if (typeof target.dns === 'string') {
+				// Convert dns string into DnsConfig object
+				const [ip, port] = target.dns.split(':');
+				toPatch.dns = {
+					remote_ip: ip || DEFAULT_REMOTE_IP,
+					remote_port: port ? parseInt(port, 10) : DEFAULT_REMOTE_PORT,
+				};
```
pipex, line 187: Perhaps we could move this to a separate function within the module for readability?

src/host-config/proxy.ts
```diff
@@ -15,6 +15,9 @@ const proxyBasePath = pathOnBoot('system-proxy');
 const noProxyPath = path.join(proxyBasePath, 'no_proxy');
 const redsocksConfPath = path.join(proxyBasePath, 'redsocks.conf');
 
+export const DEFAULT_REMOTE_IP = '8.8.8.8';
+export const DEFAULT_REMOTE_PORT = 53;
```
pipex, line 19: Why define these constants here if they are never read within this module?

docs/API.md
```diff
@@ -570,6 +570,7 @@ By default, with [balenaOS 2.82.6](https://github.com/balena-os/meta-balena/blob
 			"password": "password",
 			"noProxy": [ "152.10.30.4", "253.1.1.0/16" ]
 		},
+		"dns": "1.1.1.1:53",
```
pipex, line 573: This should be part of the proxy block no? It is the dns configuration for the proxy after all, not for the device network

pipex:
I think the `dns` property needs to go inside proxy, is what makes more sense and it will simplify a lot the code for handling proxy and dns as separate sub-object.

cywang117:
@pipex I pushed a rewrite where the `dns` field is inside `proxy`, the code is much simpler now. Thanks for your time!

src/host-config/proxy.ts
```diff
@@ -138,6 +184,32 @@ export class RedsocksConf {
 	}
 }
 
+function dnsToDnsu2t(
+	dns: unknown,
```
pipex, line 188: should this be `boolean | string` instead?

src/host-config/proxy.ts
```diff
@@ -138,6 +184,32 @@ export class RedsocksConf {
...
+): { remote_ip: string; remote_port: number } | null {
+	const dnsu2t = {
+		remote_ip: DEFAULT_REMOTE_IP,
+		remote_port: DEFAULT_REMOTE_PORT,
+	};
+
+	if (!DnsConfigFormat.is(dns)) {
```
pipex, line 195: I think validation should probably be performed in the `index` module with the rest of the validation. The endpoint should return a 400 if the input is wrong

src/host-config/proxy.ts
pipex: if we validate the input before passing it to the proxy, then this can just receive a `DnsConfig` as input

src/host-config/types.ts
```diff
@@ -47,10 +57,12 @@ export type HostProxyConfig = t.TypeOf<typeof HostProxyConfig>;
  * with host-config PATCH and provided to the user with host-config GET.
  */
 export const HostConfiguration = t.type({
-	network: t.partial({
-		proxy: HostProxyConfig,
-		hostname: t.string,
-	}),
+	network: t.exact(
```
pipex, line 60: Why change to `exact` here? This will fail validation if additional values are given no? I think the correct is to ignore additional properties.

src/host-config/types.ts
```diff
@@ -61,9 +73,18 @@ export type HostConfiguration = t.TypeOf<typeof HostConfiguration>;
...
  */
 export const LegacyHostConfiguration = t.type({
-	network: t.partial({
-		proxy: t.record(t.string, t.any),
-		hostname: t.string,
-	}),
+	network: t.exact(
+		t.partial({
+			proxy: t.intersection([
+				t.record(t.string, t.any),
```
pipex, line 79: For another PR, I think this record could be a more relaxed version of the proxy config that accepts a `string` for the type instead of a set of values

test/integration/host-config.spec.ts
```diff
@@ -415,5 +455,352 @@ describe('host-config', () => {
...
+				}` + '\n',
+		);
+	});
+
+	it('patches dnsu2t config with partial string value, using defaults where needed', async () => {
+		await patch({
+			network: {
+				proxy: {
+					// Missing ip
+					dns: ':51',
```
pipex, line 577: Do we really want to support this? I think we can just either a boolean or a `host:port` pair, this would also simplify code

test/integration/host-config.spec.ts
```diff
@@ -415,5 +455,352 @@ describe('host-config', () => {
...
+				dnsu2t {
+					remote_ip = 5.5.5.5;
+					remote_port = 55;
+					local_ip = 127.0.0.1;
+					local_port = 53;
+				}` + '\n',
+		);
+	});
+
+	it('does not add dnsu2t config when provided invalid dns input, and warns on console', async () => {
```
pipex, line 724: We don't need this test if we validate the input on the endpoint

pipex (commented):
Nice work with the PR and the comprehensive tests. I think we can still simplify a little by validating input on the host-config endpoint and returning a 400 for invalid input, that way we don't need to handle edge scenarios on the proxy module.

Also, supporting `:53` as possible input is a bit too extra :p, I think it's enough if we require `host:port` always

src/host-config/types.ts
cywang117: It actually won't fail validation, just strip additional properties from inputs. There's a test case to ensure this behavior [here](https://github.com/balena-os/balena-supervisor/pull/2284/files#diff-d3f5db585e205253dbabde1162d4c39c17be636d36c5a1354f7ae90583ff4cf3R993-R1026)

src/host-config/index.ts
```diff
@@ -57,75 +62,149 @@ export function parse(
...
+	const redsocks = {
+		...currentConf.redsocks,
+		...targetConf.redsocks,
+	};
+	patchedConf.redsocks = redsocks as ProxyConfig;
+
+	// Delete redsocks and dns configs if redsocks is not configured,
+	// because dns config depends on redsocks to work
+	if (
+		targetConf.redsocks === null ||
```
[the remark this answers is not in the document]
cywang117: Due to reworking the code to make dns a subfield of the proxy object, this part of the code is no longer here

src/host-config/index.ts
cywang117: Simplified per code review, so this chunk of code is no longer here

src/host-config/index.ts
cywang117: Moved to `function dnsToDnsu2t` [here](https://github.com/balena-os/balena-supervisor/pull/2284/commits/39efabcccc28605b2ba3b538d48cf37b6cee0dcd#diff-87da4178ce07f20bae291890a95e15b8bf7990d89aeb29ec6bfe1f77b508c3e5R171)

src/host-config/proxy.ts
cywang117: Input is validated by making the dns input type more strict [here](https://github.com/balena-os/balena-supervisor/pull/2284/commits/39efabcccc28605b2ba3b538d48cf37b6cee0dcd#diff-4a1d4485db76d595a8f99491b2dbfbde15e43e6151f3bc3f2d16edcd29129ce7R4-R30) (from `string` to regex-bounded `AddressString`), which is used when calling `hostConfig.parse` [here](https://github.com/balena-os/balena-supervisor/pull/2284/commits/a774b95c97d29fe51fe1f4c6e7a5ee435936143e#diff-aaac9f5c9b6707e5fe6448bad4487ef43f9686b75b9728aebc50aab095e0c4f4R35-R63), throwing an error message which is tested [here](https://github.com/balena-os/balena-supervisor/pull/2284/commits/a774b95c97d29fe51fe1f4c6e7a5ee435936143e#diff-d3f5db585e205253dbabde1162d4c39c17be636d36c5a1354f7ae90583ff4cf3L839-R898).

src/host-config/proxy.ts
cywang117: Removed this code in favor of input validation in `hostConfig.parse`

test/integration/host-config.spec.ts
cywang117: Made the dns input type more strict (`t.string` --> `AddressString`), which will fail `hostConfig.parse` if dns is not a string, there is no colon, or there are no characters before or after the colon. Therefore I removed all the test cases that handle missing address / port.

cywang117:
@pipex , Thanks for taking a look. I made your suggested changes of rejecting with 400 error if invalid dns string (does not match Regex of `/^.+:[0-9]+$/`), which allowed me to only handle cases of `address:ip` or boolean input.

src/host-config/types.ts
pipex: Thanks for the clarification, I was confused because I've been using Zod on another project where `exact` is the standard and you need `strict` if you want validation to fail

pipex (commented):
Almost there @cywang117, could you also add a couple of tests to check that input is validated correctly by the host-config module? Something like testing an invalid dns string or invalid ip maybe?

cywang117:
Done [here](https://github.com/balena-os/balena-supervisor/pull/2284/commits/eaa07e97a9fd2089d1f6a2f3bdff74215cdee789#diff-d3f5db585e205253dbabde1162d4c39c17be636d36c5a1354f7ae90583ff4cf3R878-R928) @pipex !

pipex (approved):
Nice work!

PR #2309, opened 2024-04-22
Gracefully handle multiple reboot/shutdown requests

cywang117:
Since HTTP's server.close() is async, there is a slim chance for two instances of /v1/reboot or /v1/shutdown to be processed. If the server is already closed when server.close() is called, the call throws ERR_SERVER_NOT_RUNNING
 which doesn't need to be surfaced to the user.

Change-type: patch

### DIFF
--- a/src/device-api/index.ts
+++ b/src/device-api/index.ts
@@ -12,10 +12,6 @@ interface SupervisorAPIConstructOpts {
 	healthchecks: Array<() => Promise<boolean>>;
 }
 
-interface SupervisorAPIStopOpts {
-	errored: boolean;
-}
-
 // API key methods
 // For better black boxing, device-api should serve as the interface
 // to the rest of the Supervisor code for accessing API key related methods.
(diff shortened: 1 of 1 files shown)


src/device-api/index.ts
```diff
@@ -92,20 +88,16 @@ export class SupervisorAPI {
 		});
 	}
 
-	public async stop(options?: SupervisorAPIStopOpts): Promise<void> {
+	public async stop(): Promise<void> {
 		if (this.server != null) {
```
pipex, line 92: Could we just make `this.server` null in order to ignore multiple calls to `stop()`, we could change the error for a warning on subsequent calls

cywang117: Sure @pipex, changed to that in https://github.com/balena-os/balena-supervisor/commit/bdc57070c873359e38c3221561cfbbd6b755813d

pipex:
Looks good

cywang117:
@balena-ci I self-certify!

PR #2311, opened 2024-04-23
Fix mdnsResolver import

pipex:
The `mdns-resolver` module does not provide a default export. Trying to use a default import notation, causes the `resolve` function to not be found, breaking MDNS resolution.

Change-type: patch


docker-compose.test.yml
```diff
@@ -40,13 +40,18 @@ services:
       MOCK_SYSTEMD_UNITS: openvpn.service avahi.socket
 
   docker:
-    image: docker:24-dind
```
pipex, line 43: This change has nothing to do with MDNS it was just overdue

cywang117 (commented):
Nice work, good implementation in testing mdns locally. Merge when ready 👍🏼

pipex:
I self-certify!

PR #2312, opened 2024-04-23
Move lib/fs-utils tests to testfs

cywang117:
This removes mock-fs as a dependency by moving fs-utils tests from unit to integration, where they use mocha-pod's testfs.

Change-type: patch


cywang117:
Closes: https://github.com/balena-os/balena-supervisor/pull/2303

cywang117:
@balena-ci I self-certify!

PR #2322, opened 2024-04-26
Update dependencies node 20 pt 3

cywang117:
- Update io-ts, io-ts-reporters, fp-ts. Upgrading `io-ts` to 2.2.21 instead of the 2.2.20 it was upgraded to here would require some `io-ts` type rewrites due to a bugfix in 2.2.21 (see https://github.com/gcanti/io-ts/issues/704) so it's not a trivial change.
- Remove unused event-stream dependency

`chai` cannot be bumped to v5, nor `husky` to v9, nor `pretty-ms` to v9, due to the Supervisor still being a CommonJS module and the aforementioned packages moving to ESM in those versions. Doing so will result in the `ERR_REQUIRE_ESM` error until we can migrate to an ESM module (`type: module` in package.json). See: https://github.com/chaijs/chai/issues/1561

Node 22 should help with ESM <--> CJS interoperability, see [docs](https://nodejs.org/api/esm.html#modules-ecmascript-modules). We can revisit some NPM bumps after a Node upgrade.

Closes: https://github.com/balena-os/balena-supervisor/pull/2321, https://github.com/balena-os/balena-supervisor/pull/2320, https://github.com/balena-os/balena-supervisor/pull/2319, https://github.com/balena-os/balena-supervisor/pull/2315, https://github.com/balena-os/balena-supervisor/pull/2290


Page-:
@cywang117 husky might be ESM but the supervisor doesn't import it anywhere, it's only used as essentially a binary

Page-:
PRed the husky update in #2323

cywang117:
@balena-ci I self-certify!

PR #2324, opened 2024-04-29
Update @types dependencies

Page-:
Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


package.json
```diff
@@ -43,37 +43,37 @@
...
     "@types/supertest": "^6.0.2",
     "@types/tar-stream": "^3.1.3",
-    "@types/webpack": "^5.28.0",
+    "@types/webpack": "^5.28.5",
     "@types/yargs": "^17.0.32",
     "balena-auth": "^6.0.1",
     "balena-register-device": "^9.0.2",
     "balena-request": "^13.3.1",
-    "blinking": "^0.0.7",
+    "blinking": "^0.0.8",
```
Page-, line 76: This is purely to fix the use of a `.ts` file instead of a `.d.ts` for types. A `.d.ts` is used as-is whereas a `.ts` will have this repo's tsconfig.json applied (which doesn't go so well..)

pipex (approved):

PR #2329, opened 2024-05-02
Refactor host config module

cywang117:
This adds better data conversion between system-proxy configuration files `redsocks.conf` and `no_proxy`, an intermediate internal data type `RedsocksConf`, and a i/o data type `HostConfiguration` which the user interacts with via `PATCH|GET /v1/device/host-config`.

Change-type: minor


src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+	// restart redsocks if it is loaded and NOT PartOf redsocks-conf.target
+	if (
+		(await dbus.servicePartOf('redsocks')).includes('redsocks-conf.target') ===
+		false
+	) {
+		await dbus.restartService('redsocks');
+	}
+}
+
+export async function readNoProxy(): Promise<string[]> {
```
pipex, line 320: I'm divided on whether the noproxy code belongs here, perhaps the module would have to be called `proxy.ts`

src/host-config/types.ts
```diff
@@ -0,0 +1,86 @@
+import * as t from 'io-ts';
+import { DoubleQuotedString, NumericIdentifier, withDefault } from '../types';
+
+/**
+ * The internal object representation of the base {...} block
+ * of redsocks.conf, part of RedsocksConfig.
+ * Mostly nonconfigurable except for log_debug (TODO)
+ */
+export const BaseConfig = t.type({
```
pipex, line 9: Using constant values inside the redsocks module makes this type unnecesary

src/host-config/types.ts
```diff
@@ -0,0 +1,86 @@
...
+			t.literal('socks4'),
+			t.literal('socks5'),
+			t.literal('http-connect'),
+			t.literal('http-relay'),
+		]),
+		ip: t.string,
+		port: NumericIdentifier,
+	}),
+	t.partial({
+		login: DoubleQuotedString,
```
pipex, line 34: Why do we need to validate that the strings are double quoted? When coming from the user, they are plain strings. When reading from redsocks.conf, we don't really need to validate them, we just need to remove the quotes if they exist, right?

src/types/basic.ts
```diff
@@ -82,6 +82,26 @@ export const NumericIdentifier = new t.Type<number, StringOrNumber>(
 );
 export type NumericIdentifier = t.TypeOf<typeof NumericIdentifier>;
 
+/**
+ * A DoubleQuotedString is a string that is enclosed in double quotes
+ */
+export const DoubleQuotedString = new t.Type<string, string>(
```
pipex, line 88: If this is only used by hostconfig, why add it here? I think we should try think of modules as independent as much as we can

test/unit/host-config.spec.ts
```diff
@@ -0,0 +1,701 @@
...
+					'Expecting DoubleQuotedString at 0.1.login but instead got: <REDACTED> (must be enclosed in double quotes)\n' +
+					'Expecting DoubleQuotedString at 0.1.password but instead got: <REDACTED> (must be enclosed in double quotes)\n' +
+					'Expecting NumericIdentifier at 1.local_port but instead got: "foo" (must be be an positive integer)',
+			);
+			(log.warn as SinonStub).resetHistory();
+			expect(conf).to.deep.equal({
+				redsocks: {
+					local_ip: '123',
+					local_port: 'foo',
+					type: 'socks6',
```
pipex, line 312: I know the API endpoint acts as a patch, but I wonder if we should use a default instead of maintaining the bad configuration

test/unit/host-config.spec.ts
```diff
@@ -0,0 +1,701 @@
...
+			expect((log.warn as SinonStub).lastCall.args[0]).to.equal(
+				'Unexpected redsocks.conf configuration detected, proxy may not behave as expected:\n' +
+					'Expecting NumericIdentifier at 0.0.port but instead got: "bar" (must be be an positive integer)\n' +
+					'Expecting one of:\n' +
+					'    "socks4"\n' +
+					'    "socks5"\n' +
+					'    "http-connect"\n' +
+					'    "http-relay"\n' +
+					'at 0.0.type but instead got: "socks6"\n' +
+					'Expecting DoubleQuotedString at 0.1.login but instead got: <REDACTED> (must be enclosed in double quotes)\n' +
```
pipex, line 303: I think we can skip the quotes validation, if it is not double quoted, we just use the value as is

test/unit/host-config.spec.ts
```diff
@@ -0,0 +1,701 @@
...
+		});
+
+		it('parses to empty config with warnings while some key-value pairs are malformed', () => {
+			// Examples of malformed key-value pairs:
+			// - no equals sign
+			// - no value
+			// - no key
+			// - no ending semicolon (will log a warning but not prevent parsing as key and value are present)
+			const redsocksConfStr = stripIndent`
+                base {
```
pipex, line 328: We don't need to parse the `base` block

pipex (changes requested):
The refactor is going in the right direction, but there is a lot that can be simplified still. There also needs to be more independence between the host config and redsocks module, there is a circular dependence now.

One thing that might help is to think of the redsocks module as an isolated component with the following features

- Patch/apply some configuration (type, ip, port, user, password), this also involves restarting the systemd service
- Read the configurable values from the existing redsocks.conf

There is no need for validation as the module can trust the type system, validation should happen only when receiving data from the outside world (the API)

The host config module has a different interface

- Receives a target config
- Splits the input, converts if necessary and applies (redsocks, noproxy, hostname)
- Reads existing config from existing sources and combines it into a hostconfig object

The API endpoint validates input, converts to the appropriate type (no more validation needed after this point) and calls HostConfig

src/host-config/redsocks.ts
cywang117: IMO no_proxy is a layer of configuration applied by the host OS on top of redsocks, so renaming to `proxy.ts` makes sense, or even having `noProxy.ts` as a separate file

test/unit/host-config.spec.ts
cywang117: For backwards compatibility, the warning would be enough, combined with the fact that the redsocks service would be error looping due to the bad value

src/host-config/index.ts
```diff
@@ -0,0 +1,149 @@
...
+					// Remove non-configurable fields
+					.filter(([key]) => !['local_ip', 'local_port'].includes(key)),
+			);
+		}
+		return proxyConf;
+	}
+}
+
+export async function patch(
+	conf: unknown,
```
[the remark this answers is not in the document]
cywang117: Addressed in abd27a80abe978d8d6883762c58f8211bb4372f5 [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/abd27a80abe978d8d6883762c58f8211bb4372f5#diff-d953a5d18d116d746cd0b200a5d7e917714811bc4609e71fc19196941bcd7ef0R449-R452) and [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/abd27a80abe978d8d6883762c58f8211bb4372f5#diff-aaac9f5c9b6707e5fe6448bad4487ef43f9686b75b9728aebc50aab095e0c4f4R92-R95)

src/host-config/index.ts
```diff
@@ -0,0 +1,149 @@
...
+	/at .+(?:login|password) but instead got/.test(line)
+		? line.replace(/but instead got: ".*"/, 'but instead got: <REDACTED>')
+		: line;
+
+export class HostConfig {
+	public static is(conf: unknown): conf is HostConfiguration {
+		// Cast this for convenience, as the checks
+		// below will only pass for HostConfiguration
+		const c = conf as HostConfiguration;
+		const hasCorrectShape =
```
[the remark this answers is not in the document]
cywang117: Addressed in COMMIT [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/abd27a80abe978d8d6883762c58f8211bb4372f5#diff-4a1d4485db76d595a8f99491b2dbfbde15e43e6151f3bc3f2d16edcd29129ce7R56-R69) and [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/abd27a80abe978d8d6883762c58f8211bb4372f5#diff-aaac9f5c9b6707e5fe6448bad4487ef43f9686b75b9728aebc50aab095e0c4f4R69-R86): define `LegacyHostConfiguration` io-ts type which has the right shape but is more permissive. If `HostConfiguration` decodes the input to left, there will be a logged warning. Then if `LegacyHostConfiguration` decodes to right, the input has the right type but may have an invalid value for one of the proxy keys `type`, `ip`, `port`, `login`, or `password`, which we should allow for backwards compatibility.

test/unit/host-config.spec.ts
cywang117: Amended all relevant commits to remove base configuration handling. `RedsocksConf.parse` now only parses the writeable subset of `redsocks {...}` block in redsocks.conf in 0ce259c1845c9137b07618aa6aafa54361771693, and `RedsocksConf.stringify` adds the base block statically in 9e89091ce35dbf010ada7e8167c3146acaf4ca3c.

test/unit/host-config.spec.ts
cywang117: Amended all relevant commits to remove double quoted string validation. `RedsocksConf.parse` now removes the double quotes from auth fields if they exist when parsing from the raw file string in 0ce259c1845c9137b07618aa6aafa54361771693, and `RedsocksConf.stringify` now adds the the double quotes if they don't exist around auth fields when stringifying RedsocksConfig to file string in 9e89091ce35dbf010ada7e8167c3146acaf4ca3c.

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+const defaultBaseConfig = {
+	log_debug: 'off',
+	log_info: 'on',
+	log: 'stderr',
+	daemon: 'off',
+	redirector: 'iptables',
+} as BaseConfig;
+
+export class RedsocksConf {
+	public static stringify(config: RedsocksConfig): string {
```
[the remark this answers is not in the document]
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/0ce259c1845c9137b07618aa6aafa54361771693#diff-4a1d4485db76d595a8f99491b2dbfbde15e43e6151f3bc3f2d16edcd29129ce7R4-R25), there's no more readable vs writeable as the local_ip and local_port values are set statically by the RedsocksConf class.

src/host-config/index.ts
```diff
@@ -0,0 +1,149 @@
...
+export async function patch(
+	conf: unknown,
+	force: boolean = false,
+): Promise<void> {
+	const apps = await applicationManager.getCurrentApps();
+	const appIds = Object.keys(apps).map((strId) => parseInt(strId, 10));
+
+	// It's possible for appIds to be an empty array, but patch shouldn't fail
+	// as it's not dependent on there being any running user applications.
+	return updateLock.lock(appIds, { force }, async () => {
```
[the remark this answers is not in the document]
cywang117: You're right, we don't need it for hostname set, so I separated it from the update lock call [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/abd27a80abe978d8d6883762c58f8211bb4372f5#diff-aaac9f5c9b6707e5fe6448bad4487ef43f9686b75b9728aebc50aab095e0c4f4R99-R101)

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		for (const block of blocks) {
+			const blockLabel = block.split(' ')[0];
+
+			// Skip unsupported blocks
+			if (
+				!supportedRedsocksConfBlocks.some(
+					(supported) => blockLabel === supported,
+				)
+			) {
+				log.warn(`Ignoring unsupported redsocks.conf block: "${blockLabel}"`);
```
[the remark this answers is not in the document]
cywang117: Removed the warning from commit 0ce259c1845c9137b07618aa6aafa54361771693

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	public static parse(rawConf: string): RedsocksConfig {
+		const conf: RedsocksConfig = {};
+		if (rawConf.length === 0) {
+			return conf;
+		}
+
+		const blocks = rawConf
+			.split('\n\n')
```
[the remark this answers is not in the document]
cywang117: Changed to regex extraction of the contents of redsocks block in 0ce259c1845c9137b07618aa6aafa54361771693

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+			blocks.push(RedsocksConf.stringifyBlock('base', config.base));
+		}
+		if (config.redsocks) {
+			blocks.push(RedsocksConf.stringifyBlock('redsocks', config.redsocks));
+		}
+
+		return blocks.length ? blocks.join('\n\n') : '';
+	}
+
+	public static parse(rawConf: string): RedsocksConfig {
```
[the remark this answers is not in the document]
cywang117: Simplified in updated commit: 0ce259c1845c9137b07618aa6aafa54361771693

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+						conf.redsocks = parsed as unknown as ReadableProxyConfig;
+					}
+				}
+			}
+		}
+
+		return conf;
+	}
+
+	public static patch(
```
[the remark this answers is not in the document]
cywang117: Renamed to `patchProxy` (as it only patches the proxy config and there's alreday `patch` in `host-config.ts` which patches both proxy and hostname. Moved it to HostConfig class in a61ac4d7d42774d5b4923b562b578bbe13406993

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+				...defaultBaseConfig,
+			};
+			// This method assumes that currentConf is a full ProxyConfig
+			// for backwards compatibility, so patchedConf.redsocks should
+			// also be a full ProxyConfig.
+			patchedConf.redsocks = {
+				...currentConf.redsocks,
+				...inputConf.redsocks,
+				// local_ip & local_port are not configurable
+				local_ip: '127.0.0.1',
```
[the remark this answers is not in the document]
cywang117: These constants are now handled only by `RedsocksConf`'s `parse` and `stringify` methods.

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		const patchedConf: RedsocksConfig = {};
+
+		// If input is empty, redsocks config should be removed
+		if (!inputConf || Object.keys(inputConf).length === 0) {
+			return patchedConf;
+		}
+
+		if (
+			inputConf.redsocks &&
+			typeof inputConf.redsocks === 'object' &&
```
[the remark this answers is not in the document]
cywang117: Not necessary, removed in a61ac4d7d42774d5b4923b562b578bbe13406993

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		label: string,
+		block: BaseConfig | ReadableProxyConfig,
+	): string {
+		const lines = Object.entries(block).map(
+			([key, value]) => `\t${key} = ${value};`,
+		);
+		return `${label} {\n${lines.join('\n')}\n}`;
+	}
+
+	private static parseBlock(
```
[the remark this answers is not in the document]
cywang117: Done in 0ce259c1845c9137b07618aa6aafa54361771693

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		block: string,
+		label: string,
+	): Record<string, string> {
+		const parsedBlock: Record<string, string> = {};
+		// Skip first and last line (`$BLOCK_LABEL {` and `}`)
```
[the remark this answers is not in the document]
cywang117: Parsing got simplified in 0ce259c1845c9137b07618aa6aafa54361771693 to use regex, so we no longer remove the first and last line

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		const lines = block.split('\n').slice(1, -1);
+		for (const line of lines) {
+			let [key, value] = line.split('=').map((s) => s.trim());
+
+			// Remove trailing semicolon
+			if (value) {
+				if (value.endsWith(';')) {
```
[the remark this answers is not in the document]
cywang117: [Updated commit](https://github.com/balena-os/balena-supervisor/pull/2329/commits/0ce259c1845c9137b07618aa6aafa54361771693#diff-87da4178ce07f20bae291890a95e15b8bf7990d89aeb29ec6bfe1f77b508c3e5R76) splits by newline and optionally semicolon, there's no semicolon validation otherwise now

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+
+	private static parseBlock(
+		block: string,
+		label: string,
+	): Record<string, string> {
+		const parsedBlock: Record<string, string> = {};
+		// Skip first and last line (`$BLOCK_LABEL {` and `}`)
+		const lines = block.split('\n').slice(1, -1);
+		for (const line of lines) {
+			let [key, value] = line.split('=').map((s) => s.trim());
```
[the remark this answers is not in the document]
cywang117: Adopted suggestion [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/0ce259c1845c9137b07618aa6aafa54361771693#diff-87da4178ce07f20bae291890a95e15b8bf7990d89aeb29ec6bfe1f77b508c3e5R76)

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+						`Missing semicolon in redsocks.conf line ${isAuthField(key) ? `"${key}"` : `"${line.trim()}"`} in "${label}" block`,
+					);
+				}
+			}
+
+			if (key && value) {
+				parsedBlock[key] = value;
+			} else {
+				// Skip malformed lines
+				log.warn(
```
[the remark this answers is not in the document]
cywang117: Removed warning in 0ce259c1845c9137b07618aa6aafa54361771693

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		return undefined;
+	}
+	const redsocksConf = RedsocksConf.parse(rawConf);
+
+	// Get and parse no_proxy
+	const noProxy = await readNoProxy();
+
+	// Build proxy object
+	const proxy = {
+		...HostConfig.fromRedsocksConfig(redsocksConf),
```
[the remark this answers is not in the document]
cywang117: I removed `toRedsocksConfig` and `fromRedsocksConfig` transformation methods after reviewing the commits again, since the patchBody.network.proxy is an object that is equal to `ProxyConfig` type after decoding to `HostConfiguration` (and loosely equal after decoding from `LegacyHostConfiguration`

src/host-config/redsocks.ts
```diff
@@ -0,0 +1,352 @@
...
+		...(noProxy.length && { noProxy }),
+	};
+
+	// Remove double quotes from auth fields
+	if (proxy.login) {
+		proxy.login = proxy.login.replace(/"/g, '');
```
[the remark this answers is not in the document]
cywang117: I changed the double quote removal / addition to be limited to `RedsocksConf.parse|stringify`

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+		const blocks: string[] = [];
+
+		if (config.redsocks) {
+			blocks.push(RedsocksConf.stringifyBlock('base', baseBlock));
+			blocks.push(
+				RedsocksConf.stringifyBlock('redsocks', config.redsocks, {
+					local_ip: '127.0.0.1',
+					local_port: 12345,
+				}),
+			);
```
pipex, line 54: Why not 
```suggestion
			blocks.push(
				RedsocksConf.stringifyBlock('redsocks', {...config.redsocks, 
					local_ip: '127.0.0.1',
					local_port: 12345,
				}),
			);
```

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+	daemon: 'off',
+	redirector: 'iptables',
+};
+
+export class RedsocksConf {
+	public static stringify(config: RedsocksConfig): string {
+		const blocks: string[] = [];
+
+		if (config.redsocks) {
+			blocks.push(RedsocksConf.stringifyBlock('base', baseBlock));
```
pipex, line 48: Love this interface, very concise and clear

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+			}
+		}
+
+		return conf;
+	}
+
+	private static stringifyBlock(
+		label: string,
+		block: Record<string, any>,
+		readonlyConfigs: Record<string, any> = {},
```
pipex, line 96: Do we need two inputs? All validation is done at this point, there should be no need to distinguish between readonly configs and the rest of the block

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+	): string {
+		const lines = Object.entries(readonlyConfigs)
+			.concat(
+				Object.entries(block).filter(
+					([key]) => !disallowedProxyFields.includes(key),
```
pipex, line 101: The type should already be validated at this point,  there should be no need to validate again

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+		const maybeProxyConfig = ProxyConfig.decode(rawRedsocksBlock);
+		if (isRight(maybeProxyConfig)) {
+			conf.redsocks = {
+				...maybeProxyConfig.right,
+			};
+		} else {
+			log.warn(Reporter.report(maybeProxyConfig).join('\n'));
+			// Allow the invalid values through for backwards compatibility
+			if (Object.keys(rawRedsocksBlock).length) {
+				conf.redsocks = rawRedsocksBlock as unknown as ProxyConfig;
```
pipex, line 86: Returning the raw field might be too permissive as it will also return extra properties that would have been removed by the `decode`, I can agree that if the proxy type (for instance) is outside the supported values we can return it anyway (as long as we validate it on the input), but not sure we should return everything that is on the redsocks block.

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+	// Build proxy object
+	const proxy = {
+		...redsocksConf.redsocks,
+		...(noProxy.length && { noProxy }),
+	};
+
+	// Assumes mandatory proxy config fields (type, ip, port) are present,
+	// even if they very well may not be. It is up to the user to ensure
+	// that all the necessary fields are present in the redsocks.conf file.
+	return proxy as HostProxyConfig;
```
pipex, line 177: Maybe a way to do this is that the returned config here allows partials while the input config does not?

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+	// restart redsocks if it is loaded and NOT PartOf redsocks-conf.target
+	if (
+		(await dbus.servicePartOf('redsocks')).includes('redsocks-conf.target') ===
+		false
+	) {
+		await dbus.restartService('redsocks');
+	}
+}
+
+export async function readNoProxy(): Promise<string[]> {
```
pipex, line 238: Do we need to export this if we are already exporting `readProxy`?

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+		}
+	} catch (e: unknown) {
+		if (!isENOENT(e)) {
+			throw e;
+		}
+		return [];
+	}
+}
+
+export async function setNoProxy(list: Nullable<string[]>) {
```
pipex, line 257: Same here, do we need to export this if we are exporting `setProxy`?

pipex (commented):
The code is a lot simpler now with the refactor. Made a lot of comments but they are mostly minor tweaks. Nice work!

I haven't gone through the tests, I'll review them later.

src/host-config/index.ts
```diff
@@ -0,0 +1,132 @@
...
+import type { RedsocksConfig, ProxyConfig } from './types';
+import { HostConfiguration, LegacyHostConfiguration } from './types';
+import { readProxy, setProxy } from './proxy';
+import * as config from '../config';
+import * as applicationManager from '../compose/application-manager';
+import { pathOnRoot } from '../lib/host-utils';
+import log from '../lib/supervisor-console';
+import * as updateLock from '../lib/update-lock';
+
+export * from './proxy';
```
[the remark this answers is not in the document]
cywang117: Exporting all the submodules is so that the rest of the Supervisor sees host-config as one submodule entity - there won't be a need to `import X from './host-config/proxy'`, for example.

src/host-config/proxy.ts
cywang117: It's exported for noProxy unit tests [here](https://github.com/balena-os/balena-supervisor/blob/787d63b27f4e0b6eb88320d79f6184bf9e338dff/test/integration/host-config.spec.ts#L103), otherwise it's not imported by any other modules.

src/device-api/v1.ts
```diff
@@ -182,19 +183,16 @@ router.patch('/v1/device/host-config', async (req, res) => {
 		// Validate proxy fields, if they exist
 		if (proxy && Object.keys(proxy).length) {
 			const blacklistedFields = Object.keys(proxy).filter((key) =>
-				disallowedHostConfigPatchFields.includes(key),
+				disallowedProxyFields.includes(key),
```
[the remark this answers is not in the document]
cywang117: SGTM 👍🏼 , removed all the validation which logs warnings from the v1 endpoint. We validate in `HostConfig.is` and log warnings there anyway.

src/host-config/proxy.ts
cywang117: I was about to say that the Supervisor allowed invalid values through before when reading the redsocks block, but the current implementation actually also excludes invalid values. Modified this [here]() (TODO link)

src/host-config/proxy.ts
```diff
@@ -0,0 +1,270 @@
...
+	log: 'stderr',
+	daemon: 'off',
+	redirector: 'iptables',
+};
+
+export class RedsocksConf {
+	public static stringify(config: RedsocksConfig): string {
+		const blocks: string[] = [];
+
+		if (config.redsocks) {
```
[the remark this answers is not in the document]
cywang117: It can! Added handling & test case for this (TODO links)

src/host-config/index.ts
```diff
@@ -0,0 +1,132 @@
...
+		const uuid = await config.get('uuid');
+		hostname = uuid?.slice(0, 7) as string;
+	}
+	// Changing the hostname on config.json will trigger
+	// the OS config-json service to restart the necessary services
+	// so the change gets reflected on containers
+	await config.set({ hostname });
+}
+
+export class HostConfig {
```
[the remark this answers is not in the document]
cywang117: > parse validates an unknown input and returns an object or throws (instead of warn)

I think we still have to log a warning instead of throwing, because we have to allow for an input to be of type LegacyHostConfiguration even if input doesn't decode to HostConfiguration for backwards compatibility.

Otherwise, SGTM, modified [here]() (TODO)

src/host-config/index.ts
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	public static is(
+		conf: unknown,
+	): conf is HostConfiguration | LegacyHostConfiguration {
+		const decoded = HostConfiguration.decode(conf);
+		if (isRight(decoded)) {
+			return true;
+		} else {
+			log.warn(
```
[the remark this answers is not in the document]
cywang117: I don't agree, we should be able to return a parsed HostConfiguration (after changing `is` --> `parse`) while also supporting returning a parsed LegacyHostConfiguration for backwards compatibility and warning about the legacy status

src/host-config/index.ts
```diff
@@ -0,0 +1,132 @@
...
+				delete patched[field];
+			}
+			patchedConf.redsocks = patched as ProxyConfig;
+		}
+		return patchedConf;
+	}
+
+	public static is(
+		conf: unknown,
+	): conf is HostConfiguration | LegacyHostConfiguration {
```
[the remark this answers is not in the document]
cywang117: We can't convert the input to HostConfiguration in all cases, since users may provide an input that does not decode to HostConfiguration:
```
t.type({
  network: t.partial({
    proxy: t.intersection([
      t.type({
        type: t.union([
          t.literal('socks4'),
          t.literal('socks5'),
          t.literal('http-connect'),
          t.literal('http-relay'),
        ]),
      }),
      t.partial({
        login: t.string,
        password: t.string,
        noProxy: t.array(t.string)
      }),
    ]),
    hostname: t.string
  })
})
```
but may decode to a shape that we historically accept (LegacyHostConfiguration):
```
t.type({
	network: t.partial({
		proxy: t.record(t.string, t.any),
		hostname: t.string,
	}),
})
```

src/host-config/index.ts
```diff
@@ -0,0 +1,132 @@
...
+		} else {
+			log.warn(
+				['Malformed host config detected, things may not behave as expected:']
+					.concat(Reporter.report(decoded))
+					.join('\n'),
+			);
+			// We haven't strictly validated user input since introducing the API,
+			// endpoint, so accept configs where values may not be of the right type
+			// but have the right shape for backwards compatibility.
+			if (LegacyHostConfiguration.is(conf)) {
```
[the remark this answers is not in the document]
cywang117: I would much prefer if we didn't accept any legacy input at all, at the cost of making this a major semver bump. The legacy input parsing makes things much more complicated and it's such a small portion of the user base. Even if we didn't support legacy inputs anymore, users will still get feedback that their input was not acceptable as this PR adds console warnings for each line that's not of the right type. 

I'm all for removing any legacy handling if you are!

src/host-config/proxy.ts
cywang117: The input config as in `setProxy`? 

`readProxy` just reads and returns whatever is from redsocks.conf, which could be anything as the user could easily preload an invalid file onto the device. I don't think there is much benefit to differentiating `Partial<HostProxyConfig>` vs `HostProxyConfig` in the return type, because we wouldn't do anything differently if the return was a partial -- if `type`, `ip`, or `port` aren't present, we can't do anything about that other than maybe log it. I really think it's the user's responsibility here to make sure any configuration in redsocks.conf is valid if preloaded.

src/host-config/proxy.ts
cywang117: It's exported for unit tests only

src/device-api/actions.ts
```diff
@@ -445,14 +446,8 @@ export const getHostConfig = async () => {
...
-export const patchHostConfig = async (
-	conf: Parameters<typeof hostConfig.patch>[0],
-	force: boolean,
-) => {
-	// If hostname is an empty string, return first 7 digits of device uuid
-	if (conf.network?.hostname === '') {
-		const uuid = await config.get('uuid');
-		conf.network.hostname = uuid?.slice(0, 7);
+export const patchHostConfig = async (conf: unknown, force: boolean) => {
+	if (HostConfig.is(conf)) {
```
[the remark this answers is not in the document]
cywang117: Good catch, I modified to throw a `400 Bad Request` error if host-config cannot parse the input [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/5dbaaeb7b7fd7aea50bd38b8793cb1c3e44be810#diff-d953a5d18d116d746cd0b200a5d7e917714811bc4609e71fc19196941bcd7ef0L450-R455), and return that in the v1 endpoint [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/5dbaaeb7b7fd7aea50bd38b8793cb1c3e44be810#diff-c75126a71bc2c9694504d800b8c8340e2b17cdf8ac7cd2128770d32cdb08ce89R190-R195).

src/host-config/proxy.ts
cywang117: Adopted this suggestion [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/70b037651d2a4a8ba6c86aa14199036b3d470e5c#diff-87da4178ce07f20bae291890a95e15b8bf7990d89aeb29ec6bfe1f77b508c3e5R45-R49), good call

src/host-config/proxy.ts
cywang117: Removed readonlyConfigs in 70b037651d2a4a8ba6c86aa14199036b3d470e5c

src/host-config/proxy.ts
cywang117: Removed the validation [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/70b037651d2a4a8ba6c86aa14199036b3d470e5c#diff-87da4178ce07f20bae291890a95e15b8bf7990d89aeb29ec6bfe1f77b508c3e5R97-R110)

src/host-config/index.ts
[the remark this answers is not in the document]
pipex: But why does the rest of the supervisor need direct access to the proxy module? The interface to interact with the proxy is this module

src/host-config/index.ts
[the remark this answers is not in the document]
pipex: I don't understand, the LegacyHostConfiguration accepts pretty much any input? What would happen if an input was unused or invalid in that case? It doesn't seem compatible to me that we have one branch that properly validates and one that accepts anything. We can discuss about it in our checkin

src/host-config/index.ts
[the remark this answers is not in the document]
pipex: Oh, hadn't seen this comment when I made my comment [here](https://github.com/balena-os/balena-supervisor/pull/2329#discussion_r1657683978). We can talk later about what happen with legacy input, I hope maybe we can coerce the legacy type into an accepted type somehow

src/host-config/proxy.ts
```diff
@@ -0,0 +1,275 @@
...
+export async function readProxy(): Promise<HostProxyConfig | undefined> {
+	// Get and parse redsocks.conf
+	let rawConf: string | undefined;
+	try {
+		rawConf = await readFromBoot(redsocksConfPath, 'utf-8');
+	} catch (e: unknown) {
+		if (!isENOENT(e)) {
+			throw e;
+		}
+		return undefined;
```
pipex, line 166: Why return undefined here instead of an empty object `{}` which is also a valid host proxy config?

src/host-config/index.ts
```diff
@@ -0,0 +1,128 @@
...
+		await setHostname(conf.network.hostname);
+	}
+
+	if (conf.network.proxy != null) {
+		const targetConf = conf.network.proxy;
+		// It's possible for appIds to be an empty array, but patch shouldn't fail
+		// as it's not dependent on there being any running user applications.
+		return updateLock.lock(appIds, { force }, async () => {
+			const { noProxy: _ignore, ...currentConf } =
+				(await readProxy()) || ({} as any);
```
pipex, line 105: This `||` is not necessary if `readProxy` can return an empty object

src/host-config/proxy.ts
```diff
@@ -0,0 +1,275 @@
...
+		return;
+	}
+
+	// Write to redsocks.conf
+	const toWrite = RedsocksConf.stringify(conf);
+	if (toWrite) {
+		await writeToBoot(redsocksConfPath, toWrite);
+	} else {
+		await unlinkAll(redsocksConfPath);
+	}
```
pipex, line 212: This simplification might work?

```suggestion
	// Write to redsocks.conf
	const toWrite = RedsocksConf.stringify(conf);
	if (toWrite) {
		await writeToBoot(redsocksConfPath, toWrite);
	} else if (!noProxyChanged) {
		await unlinkAll(redsocksConfPath);
	}
```

src/host-config/proxy.ts
```diff
@@ -0,0 +1,275 @@
...
+					local_port: 12345,
+				}),
+			);
+		}
+
+		return blocks.length ? blocks.join('\n') : '';
+	}
+
+	public static parse(rawConf: string): RedsocksConfig {
+		const conf: RedsocksConfig = {};
```
pipex, line 65: Is good practice to trim strings in case they are just whitespace

```suggestion
		const conf: RedsocksConfig = {};
		rawConf = rawConf.trim();
```

src/host-config/proxy.ts
```diff
@@ -0,0 +1,275 @@
...
+	['type', 'ip', 'port', 'login', 'password'].includes(field);
+
+const isAuthField = (field: string): boolean =>
+	['login', 'password'].includes(field);
+
+// ? is a lazy operator, so only the contents up until the first `}(?=\s|$)` is matched.
+// (?=\s|$) indicates that `}` must be followed by a whitespace or end of file to match,
+// in case there are user fields with brackets such as login or password fields.
+const blockRegexFor = (blockLabel: string) =>
+	new RegExp(`${blockLabel}\\s?{([\\s\\S]+?)}(?=\\s|$)`);
```
pipex, line 36: is `[\\s\\S]+` equivalent to `.+`? If not it might be better to make explicit what you are excluding, e.g. if you don't want numbers, you can do `[^0-9]`

pipex: I think that `(?=\\s|$)` can be `\s*$` (i.e. any whitespace character or the end of the line)

pipex: Overall the simpler version might be 
```suggestion
	new RegExp(`${blockLabel}\\s*{(.+?)}\\s*$)`);
```

But I may be missing something

src/host-config/proxy.ts
```diff
@@ -0,0 +1,275 @@
...
+	public static parse(rawConf: string): RedsocksConfig {
+		const conf: RedsocksConfig = {};
+		if (rawConf.length === 0) {
+			return conf;
+		}
+
+		// Extract contents of `redsocks {...}` using regex
+		const rawRedsocksBlockMatch = rawConf.match(blockRegexFor('redsocks'));
+		// If match length shorter than 2, no group was captured, indicating malformed config
+		if (!rawRedsocksBlockMatch || rawRedsocksBlockMatch.length < 2) {
```
pipex, line 73: Comparing the length of the match might be unnecessary, if there are no matches the `match` function will always return null IIRC?

src/host-config/proxy.ts
```diff
@@ -0,0 +1,275 @@
...
+			// backwards compatibility, while excluding unknown fields.
+			const maybeInvalidProxyConfig: Record<string, string> = {};
+			for (const key in rawRedsocksBlock) {
+				if (isAllowedProxyField(key)) {
+					maybeInvalidProxyConfig[key] = rawRedsocksBlock[key];
+				}
+			}
+
+			if (Object.keys(maybeInvalidProxyConfig).length) {
+				conf.redsocks = maybeInvalidProxyConfig as unknown as ProxyConfig;
```
pipex, line 98: I feel like this violates the contract defined by the function signature. The function declares that it returns a `RedSocks` config but it could technically return an object with a missing `type` or `ip`. We could either return a default value if `.conf` does not include a property or return an empty config if none of the required values are defined.

src/host-config/proxy.ts
pipex: This relates to this [other comment](https://github.com/balena-os/balena-supervisor/pull/2329/files#r1657733759) about the contract of the function signature. Maybe we should be stricter and say, "if there is a field in the required list missing then it is the same as if the configuration is empty"

src/host-config/index.ts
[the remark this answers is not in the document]
cywang117: We discussed and agreed to keep this for now for backwards compatibility, and remove it in the next version of the Supervisor API.

src/host-config/index.ts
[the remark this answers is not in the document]
cywang117: See comment I made in another comment chain [here](https://github.com/balena-os/balena-supervisor/pull/2329#discussion_r1661723600)

TL;DR We discussed and agreed to keep the 2 types (HostConfiguration / LegacyHostConfiguration) for now and remove LegacyHostConfiguration in the next major API bump

src/host-config/proxy.ts
cywang117: I think this is more a syntactic preference and makes the syntax in your [other](https://github.com/balena-os/balena-supervisor/pull/2329#discussion_r1657701558) comment better as well, but makes other parts of the host-config module more complicated as a result. For example, with `hostConfig.get`:
```
export async function get(): Promise<HostConfiguration> {
	const proxy = await readProxy();
	return {
		network: {
			hostname: await readHostname(),
			// Only return proxy if readProxy is not undefined
			...(proxy && { proxy }),
		},
	};
}
```
The return signature expects `HostConfiguration` which is
```
{
  network: t.partial({
    proxy: HostProxyConfig,
    hostname: t.string
  })
}
```
which means that if `readProxy` resolves `HostProxyConfig | Record<string, never>` instead of `HostProxyConfig | undefined`, even `...(Object.keys(proxy).length && { proxy })` is not enough for TypeScript not to complain as there is insufficient overlap between `undefined` and `Record<string, never>`, proxy must be explicitly cast to `HostConfiguration['network']['proxy']` for the return type to be satisfied. I don't think it's worth it for an internal method.

src/host-config/proxy.ts
cywang117: Regarding the suggestion of `\\s?` --> `\\s*`: I think it's better to keep as `\\s?` since it's more strict with its rule of "either one whitespace or nothing" as opposed to the rule of "0 or more whitespaces".

Regarding the suggestion of changing the capture group from `[\\s\\S]` --> `.`, I tried that before settling on `\\s\\S`, as `.` didn't pass for [this](https://github.com/balena-os/balena-supervisor/blob/91d090355559c3338287ac54d63a50cb10980264/test/unit/host-config.spec.ts#L130-L165) test case. I don't know the particulars of how JS applies different types of regex under the hood, but `\\s\\S` worked where `.` didn't, so something must be different.

Regarding suggestion `(?=\\s|$)` --> `\\s*$`, I also tried that initially but it also didn't pass the same test case as above, capturing `test = test` from the `dnstc` block following `redsocks`.

Feel free to play around with it yourself [here](regexr.com/82bmq).

src/host-config/proxy.ts
cywang117: I see what you mean here. Between the two suggestions, maybe warning in the console and returning nothing is better rather than returning a default value. However, given there is already a check which filters for accepted `redsocks` block keys above this `as unknown as ProxyConfig` cast:
```
for (const key in rawRedsocksBlock) {
  if (isAllowedProxyField(key)) {
    maybeInvalidProxyConfig[key] = rawRedsocksBlock[key];
  }
}
```
I think it's okay and maybe even necessary for backwards compatibility, although unpalatable -- and I do agree that it's unpalatable.

cywang117:
@pipex Thanks for your continued help with this code review. I think we're almost there! Here's my latest revisions which addresses your comments. I've also removed any exports that are unnecessary in a separate commit, 16d8a59f9c185fcef1a08eab5fb6f156b4763d6b

src/host-config/proxy.ts
cywang117: Removed unnecessary exports in separate commit 16d8a59f9c185fcef1a08eab5fb6f156b4763d6b

src/host-config/proxy.ts
cywang117: Removed unnecessary exports in separate commit https://github.com/balena-os/balena-supervisor/commit/16d8a59f9c185fcef1a08eab5fb6f156b4763d6b

src/host-config/index.ts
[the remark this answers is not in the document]
cywang117: Removed unnecessary exports in separate commit https://github.com/balena-os/balena-supervisor/commit/16d8a59f9c185fcef1a08eab5fb6f156b4763d6b

src/host-config/proxy.ts
pipex: I was wrong before, `{}` is not a valid `HostProxyConfig` so returning `undefined` is right. I guess that's what you meant by returning `Record<string, never>`

src/host-config/index.ts
pipex: I was wrong before about `readProxy` being able to return an empty object, but I don't t think we should use `as any` here as it disables all type checking potentially masking bugs. The right way might just be to use an if 

```ts
const proxyConf = await readProxy();
let currentConf: ProxyConfig | undefined = undefined;
if (proxyConf) {
	delete proxyConf.noProxy;
	currentConf = proxyConf;
}
```

src/host-config/proxy.ts
pipex: > Regarding the suggestion of \\s? --> \\s*: I think it's better to keep as \\s? since it's more strict with its rule of "either one whitespace or nothing" as opposed to the rule of "0 or more whitespaces".

I don't think it is necessarily to be more strict here. An extra space is easy to miss so we should be more permissive there

> Regarding the suggestion of changing the capture group from [\\s\\S] --> ., 

You are right, the dot does not match new lines without the [dotAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) property, the code below works, but I think what you have is fine

```
new RegExp(`${blockLabel}\\s+{(.+?)}(?=\\s|$)`, 's');
```

> Regarding suggestion (?=\\s|$) --> \\s*$

You are right, that doesn't work. You are right is probably related to the lazy lookahead

Let's leave the code as is

src/host-config/proxy.ts
pipex: I don't think it's just ugly, I think something like this is potentially dangerous. When you have a function signature 

```
public static parse(rawConf: string): RedsocksConfig
```

You are ensuring whoever uses the function that the result is a valid object of type `RedsocksConfig`, which is not true with this implementation.

I think if the configuration on disk is invalid we should just ignore it and treat it the same as if there was no configuration at all.

Regarding backwards compatibility, I don't think this breaks it. If `redsocks.conf` is broken (e.g. it is missing a required field), then the device proxy would not be functional anyway, so there is no difference if we treat the invalid configuration as no configuration

pipex:
Almost there, I think we should address these last two points before merging

src/host-config/proxy.ts
cywang117: Noted and understood, thanks for explaining. I've applied the change suggested [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/1e224be0cd167d7f037a3d1cb4b260b6ff36d458#diff-87da4178ce07f20bae291890a95e15b8bf7990d89aeb29ec6bfe1f77b508c3e5R56-R63) of returning no configuration if a valid config cannot be parsed from redsocks.conf.

src/host-config/index.ts
cywang117: I applied the suggested change [here](https://github.com/balena-os/balena-supervisor/pull/2329/commits/53f5641ef1cd2e5374f55a50a601fb80d4872511#diff-aaac9f5c9b6707e5fe6448bad4487ef43f9686b75b9728aebc50aab095e0c4f4R104-R109).

cywang117:
@pipex Last two points addressed!

pipex (approved):
Nice work!

Issue #2257, opened 2024-03-25
An interrupted service restart may leave services in the `exited` state


pipex:
A service restart via the supervisor API endpoints 

- [/v1/restart](https://docs.balena.io/reference/supervisor/supervisor-api/#post-v1restart)
- [/v2/applications/:appId/restart-service](https://docs.balena.io/reference/supervisor/supervisor-api/#post-v2applicationsappidrestart-service) 

performs the following actions

1. stop the container
2. remove the container
3. create the container with the same configuration as before 
4. start the newly created container

Because of the changes in https://github.com/balena-os/balena-supervisor/pull/1926 if the supervisor gets stopped between steps 1 and 2, on restart, the service won't get restarted by the supervisor when applying the target state, and it won't be started by the engine since a `stop` request already took place

This might be solved by combining steps 1 and 2 by doing a [remove](https://docs.docker.com/engine/api/v1.40/#tag/Container/operation/ContainerDelete) with `force: true` when killing a container

PR #2330, opened 2024-05-08
Uninstall containers using force remove

pipex:
When uninstalling a container for a restart from the API on when moving between releases, this prefers to force remove the container rather than stop then remove. This has the same effect in the end, but force remove is an atomic call, meaning if the supervisor is killed between the stop and remove, the engine will proceed with the operation and the supervisor will know to recover from the target state after the restart.

Change-type: patch
Relates-to: #2257

### DIFF
--- a/src/compose/composition-steps.ts
+++ b/src/compose/composition-steps.ts
@@ -14,9 +14,6 @@ import type { DeviceLegacyReport } from '../types/state';
 interface CompositionStepArgs {
 	stop: {
 		current: Service;
-		options?: {
-			wait?: boolean;
-		};
 	};
 	kill: {
 		current: Service;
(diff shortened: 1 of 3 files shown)


pipex:
This is not the right solution, force remove uses SIGKILL instead of sigterm

```
docker rm --help                                                                                                           dynamic-sensors

Usage:  docker rm [OPTIONS] CONTAINER [CONTAINER...]

Remove one or more containers

Aliases:
  docker container rm, docker container remove, docker rm

Options:
  -f, --force     Force the removal of a running container (uses SIGKILL)
  -l, --link      Remove the specified link
  -v, --volumes   Remove anonymous volumes associated with the container
```

PR #2334, opened 2024-05-15
Refactor code to reduce circular dependencies between modules

pipex:
This reduces circular dependencies from 250 to 3, also improving the code architecture. The last 3 dependencies will be done as part of a separate refactor as they are relate to the core supervisor architecture
```
  warn no-circular: src/device-state.ts →
      src/device-state/preload.ts →
      src/device-state.ts
  warn no-circular: src/config/index.ts →
      src/lib/register-device.ts →
      src/lib/request.ts →
      src/config/index.ts
  warn no-circular: src/config/functions.ts →
      src/config/index.ts →
      src/config/functions.ts
```

Before
Screen Shot 2024-05-15 at 11 50 36

After
Screen Shot 2024-05-16 at 11 37 26

These diagrams have been generated with [dependency-cruiser](https://github.com/sverweij/dependency-cruiser/tree/4a5dcd095e4d6e69411e37efd1a75e79956bf1a4)

```
# Generate list of errors
npx depcruise src --include-only "^src" --output-type err

# Generate architecture diagram (requires graphviz)
 npx depcruise src --include-only "^src" --output-type archi | dot -T svg > diagram.svg
```


pipex:
I self-certify!

PR #2340, opened 2024-05-28
Refactor device state module

pipex:
This is a simple refactor to organize code within the device state module in a way that it removes the circular dependency between preloading and the module entrypoint.

This moves `target-state.ts` outside the device-state module and into `api-binder/poll.ts` a new `device-state/target-state.ts` is created to handle validating storing target state.

Eventually interaction with the database will be moved outside of the device-state module once the interface has been polished


pipex:
I self-certify!

PR #2342, opened 2024-06-05
Report app release update status to the API

pipex:
This PR adds a new `update_status` field when reporting the release state to allow users to identify and filter devices that are stuck between releases.

This PR requires API support of the new `update_status` and `last_update_status_event` fields on the device entity, implemented in balena-io/open-balena-api#1692

Change-type: minor

## Release Notes

The Supervisor now reports the status of a release update/install in a new `update_status` field as part of the state report. This field will be used by the backend to populate the homonymous field on the [Device resource](https://docs.balena.io/reference/api/resources/device/) on the API and allow users to filter the device list based on its value.

This field may have the following values

- `rejected` means that the target release has required services that do not meet the [necessary requirements](https://docs.balena.io/learn/develop/container-contracts/#container-contracts) to run. 
- `aborted` means that the supervisor has given up on trying to install a new release. There are currently no conditions under which the supervisor returns this state, but future work could allow the supervisor to abort an update, for instance, after a number of failed download tries. 
- `downloading` indicates that the release images are being downloaded 
- `downloaded` indicates that the release images have been downloaded and the supervisor is waiting for locks to apply changes (depending on the [update strategy](https://docs.balena.io/learn/deploy/release-strategy/update-strategies/#controlling-the-update-strategy))
- `applying changes` means that the supervisor is performing actions to install the new release, i.e. killing old containers and making volumes/network/container changes.
- `done` means that the target has been reached and no more changes are required.  

On the API side, users will be able to identify update issues by looking at the `update_status` field and the `last_update_status_event`. A device with update status different than `done` indicates something is preventing the supervisor from downloading/installing the new release.


src/lib/contracts.ts
```diff
@@ -51,13 +94,30 @@ function isValidRequirementType(
 	return requirement in requirementVersions;
 }
 
+function partition<T>(arr: T[], predicate: (t: T) => boolean) {
```
cywang117, line 97: `_.partition` has the same functionality as this, are we trying to stay away from lodash though?

pipex: yes, ideally we should phase out lodash usage

src/types/state.ts
```diff
@@ -182,6 +198,8 @@ const DeviceReport = t.partial({
 	cpu_usage: t.number,
 	cpu_id: t.string,
 	is_undervolted: t.boolean,
+	// These are for internal reporting only, they are not sent
+	// to the API
 	update_failed: t.boolean,
 	update_pending: t.boolean,
 	update_downloaded: t.boolean,
```
cywang117, line 206: Since we do omit these fields in `getCurrentForReport`, perhaps we can remove some code that still attempts to report these fields: [[1]](https://github.com/balena-os/balena-supervisor/blob/45d1f1f964e777b5168b089858af0cec33f98c68/src/device-state/index.ts#L238-L240) [[2]](https://github.com/balena-os/balena-supervisor/blob/45d1f1f964e777b5168b089858af0cec33f98c68/src/device-state/index.ts#L536) [[3]](https://github.com/balena-os/balena-supervisor/blob/45d1f1f964e777b5168b089858af0cec33f98c68/src/device-state/index.ts#L632-L636)

However, good to save for another PR

src/types/state.ts
```diff
@@ -142,9 +153,14 @@ const ServiceState = t.intersection([
...
-const ReleaseState = t.type({
-	services: t.record(DockerName, ServiceState),
-});
+const ReleaseState = t.intersection([
+	t.type({
+		services: t.record(DockerName, ServiceState),
+	}),
+	t.partial({
+		update_status: UpdateStatus,
+	}),
```
cywang117, line 163: Why `t.partial`? Are there apps in this release that won't have `update_status`? I guess the host apps wouldn't, right?

src/compose/application-manager.ts
```diff
@@ -549,49 +545,43 @@ export async function setTarget(
...
+				services: unmetServices.map(({ serviceName }) => serviceName),
+			};
 
-				if (unmetAndOptional.length !== 0) {
-					return reportOptionalContainers(unmetAndOptional);
-				}
+			// Remove the invalid app from the list
+			delete filteredApps[appUuid];
+		} else {
+			// valid is true, but we could still be missing
```
cywang117, line 568: `valid is true` --> `App is valid`

src/types/state.ts
```diff
@@ -131,6 +131,17 @@ export const UUID = ShortString;
 /** ***************
  * Current state *
  *****************/
+
+const UpdateStatus = t.union([
+	t.literal('rejected'),
+	t.literal('downloading'),
+	t.literal('downloaded'),
+	t.literal('applying changes'),
```
cywang117, line 140: Possible to simplify this string to `applying`?

src/compose/application-manager.ts
```diff
@@ -1004,21 +1004,55 @@ export async function getState() {
...
+		};
+
+		const svcs = releases[commit].services;
+		svcs[serviceName] = svc;
+
+		// The update_status precedence order is as follows
+		// - aborted
+		// - downloading
+		// - downloaded
+		// - applying changes
```
cywang117, line 1041: See previous comment about `applying changes` --> `applying`

test/integration/compose/application-manager.spec.ts
```diff
@@ -2610,6 +2615,7 @@ describe('compose/application-manager', () => {
 									status: 'Awaiting handover',
 								},
 							},
+							update_status: 'applying',
```
cywang117, line 2618: Status here is `applying` which is not part of the UpdateStatus enum

src/compose/application-manager.ts
```diff
@@ -517,34 +528,27 @@ export async function setTarget(
...
-	// We look at the container contracts here, as if we
-	// cannot run the release, we don't want it to be added
-	// to the database, overwriting the current release. This
-	// is because if we just reject the release, but leave it
-	// in the db, if for any reason the current state stops
-	// running, we won't restart it, leaving the device
-	// useless - The exception to this rule is when the only
-	// failing services are marked as optional, then we
-	// filter those out and add the target state to the database
+	// We look at the container contracts here, apps failing contracts
```
cywang117, line 546: `apps failing contracts` --> `apps with failing contract requirements`

test/integration/device-state.spec.ts
```diff
@@ -424,7 +425,12 @@ describe('device-state', () => {
 					},
 				},
 			} as TargetState),
-		).to.be.rejected;
+		).to.not.be.rejected;
+		const targetState = await deviceState.getTarget();
+		const app = targetState.local.apps[1234];
+		expect(app).to.have.property('appName').that.equals('superapp');
+		expect(app).to.have.property('commit').that.equals('one');
+		expect(app).to.have.property('isRejected').that.is.true;
```
cywang117, line 433: What do you think of adding the test cast `accepts target state with valid contract and resets isRejected to false`?

cywang117 (commented):
Nice work!

test/integration/device-state.spec.ts
pipex: Yes, good idea

src/types/state.ts
pipex: Actually I don't know. We don't report the status for host apps yet, but I assume it will work in a similar way. I'll probably make this required

test/integration/device-state.spec.ts
pipex: Done

src/types/state.ts
pipex: This is what the server expects. We try to keep messaging consistent between api/sdk/dashboard

test/integration/device-state.spec.ts
pipex: I ended up merging the two test cases to remove duplication

src/lib/contracts.ts
```diff
@@ -198,67 +255,37 @@ export function validateContract(contract: unknown): boolean {
...
-								contract,
-								optional: checkTruthy(
-									service.labels?.['io.balena.features.optional'],
-								),
-							};
-						} catch (e: any) {
-							throw new ContractValidationError(serviceName, e.message);
+): Dictionary<AppContractResult> {
+	return Object.fromEntries(
+		Object.entries(apps).flatMap(([appUuid, app]) =>
```
Page-, line 263: I think this can just be a normal `map`:
```suggestion
		Object.entries(apps).map(([appUuid, app]) =>
```
which if so would be much better to make the intent clearer. I did try to spot if the flat map was actually required and it don't think it is, so that has either added confusion or this section is confusing enough about why a `flatMap` is required that I think clarifying it would be very beneficial

src/lib/contracts.ts
```diff
@@ -198,67 +255,37 @@ export function validateContract(contract: unknown): boolean {
...
-								serviceName,
-								contract,
-								optional: checkTruthy(
-									service.labels?.['io.balena.features.optional'],
-								),
-							};
-						} catch (e: any) {
-							throw new ContractValidationError(serviceName, e.message);
+): Dictionary<AppContractResult> {
+	return Object.fromEntries(
```
Page-, line 262: I would probably avoid the `fromEntries` and instead do
```ts
const result = {};
for (const [appUuid, app] of Object.entries(apps)) {
	...
	result[appUuid] = appContractResult;
}
return result;
```

It will reduce the nesting a bit, make it clearer that the end goal is to populate a `{ [appUuid: string]: ContractResult }` type object, and also have a marginal benefit in reduced allocations (it won't allocate an array of arrays to then convert to an object)

src/lib/contracts.ts
```diff
@@ -51,13 +94,30 @@ function isValidRequirementType(
...
+
 export function containerContractsFulfilled(
-	serviceContracts: ServiceContracts,
-): ApplicationContractResult {
-	const containers = _(serviceContracts).map('contract').compact().value();
+	servicesWithContract: ServiceWithContract[],
+): AppContractResult {
+	const containers = servicesWithContract
+		.map(({ contract }) => contract)
+		.filter((c) => !!c) as ContractObject[];
```
Page-, line 116: Can you make this more explicit as either
```suggestion
		.filter((c) => c != null) as ContractObject[];
```
or
```suggestion
		.filter((c) => c === true) as ContractObject[];
```
or whatever the true intent is. If you do that I think you'd also be able to drop the `as ContractObject[]` since latest typescript can infer the type guard typing that I mentioned above. I expect the true final form to be:
```suggestion
		.filter((c) => c != null);
```

Page-, line 116: As a general rule it's better to do `satisfies X` rather than `as X`, or `satisfies X as X` so that it first checks that the inferred typing is compatible with the expected typing before doing the cast and increase the type safety

src/lib/contracts.ts
```diff
@@ -51,13 +94,30 @@ function isValidRequirementType(
...
+		({ pass, fail }, elem) => {
+			if (predicate(elem)) {
+				pass.push(elem);
+			} else {
+				fail.push(elem);
+			}
+			return { pass, fail };
+		},
+		{ pass: [] as T[], fail: [] as T[] },
+	);
```
Page-, line 108: This can be simplified and made (marginally) more performant via
```suggestion
	const result = { pass: [] as T[], fail: [] as T[] };
	for (const elem of arr) {
		if (predicate(elem)) {
			pass.push(elem);
		} else {
			fail.push(elem);
		}
	}
	return result;
```
Although I would probably say it makes even more sense to just use the lodash partition method since lodash is already used in a bunch of places so you won't get the size benefit from removing lodash altogether for the sake of rewriting this one function (but using native versions definitely makes sense imo where they're possible, until a final push to do a full removal).

And if you wanted to code golf (don't do this) then you could do `(predicate(elem) ? pass : fail).push(elem)`

src/device-state/index.ts
```diff
@@ -364,18 +359,34 @@ export async function getCurrentForReport(
...
+				}
+
+				// Add the rejected release to apps for report
+				app.releases[r.releaseUuid] = {
+					update_status: 'rejected',
+					services: {},
+				};
+			}
+
+			return [[appUuid, app]];
```
Page-, line 387: This is another case where I think doing `appsForReport[appUuid] = app` would be far clearer as to the actual intent rather than the current unintentional obfuscation that's happening vs returning nested arrays

src/device-state/db-format.ts
```diff
@@ -92,67 +88,77 @@ export async function setApps(
...
+		);
+
+		const releases = releaseUuid
+			? {
+					[releaseUuid]: {
+						id: releaseId,
+						services,
+						networks: JSON.parse(app.networks),
+						volumes: JSON.parse(app.volumes),
+					} as TargetRelease,
```
Page-, line 139: As mentioned elsewhere `satisfies` will increase the type safety
```suggestion
					} satisfies TargetRelease,
```
but in this case it's quite possible you'll want the
```suggestion
					} satisfies TargetRelease as TargetRelease,
```
version to keep the wider typing rather than the narrowed version that `satisfies` alone gives

src/device-state/db-format.ts
```diff
@@ -92,67 +88,77 @@ export async function setApps(
...
-						releases,
-					},
-				];
-			},
-		)
-		.reduce((apps, [uuid, app]) => ({ ...apps, [uuid]: app }), {});
+						id: serviceId,
+						image_id: imageId,
+						...service,
+					} as TargetService,
```
Page-, line 127:
```suggestion
					} satisfies TargetService,
```

src/device-state/db-format.ts
```diff
@@ -92,67 +88,77 @@ export async function setApps(
...
+			(JSON.parse(app.services) as DatabaseService[]).map(
+				({
+					serviceName,
+					serviceId,
+					imageId,
+					// Ignore these fields
+					appId: _appId,
+					appUuid: _appUuid,
+					commit: _commit,
+					releaseId: _releaseId,
```
Page-, line 118: In typescript an `_` prefix has a very specific meaning which is to ignore an unused variable, that is not the case for these so I'd like to see some other prefix and the standard I tend to use as a placeholder prefix is `$`:
```suggestion
					appId: $appId,
					appUuid: $appUuid,
					commit: $commit,
					releaseId: $releaseId,
```

src/device-state/db-format.ts
```diff
@@ -40,10 +33,12 @@ export async function getApps(): Promise<InstancedAppState> {
 export async function setApps(
 	apps: TargetApps,
 	source: string,
+	rejectedApps = [] as UUID[],
```
Page-, line 36: I think this can be
```suggestion
	rejectedApps: UUID[] = [],
```

src/compose/application-manager.ts
```diff
@@ -999,21 +1004,55 @@ export async function getState() {
...
+			(svc.download_progress === 100 || svc.status === 'Downloaded')
+		) {
+			releases[commit].update_status = 'downloaded';
+		} else if (
+			// The `applying changes` state has lower precedence over the aborted/downloading/downloaded
+			// state
+			!['aborted', 'downloading', 'downloaded'].includes(
+				releases[commit].update_status!,
+			) &&
+			['installing', 'installed', `stopping`, 'awaiting handover'].includes(
```
Page-, line 1047: Nit:
```suggestion
			['installing', 'installed', 'stopping', 'awaiting handover'].includes(
```

src/compose/application-manager.ts
```diff
@@ -999,21 +1004,55 @@ export async function getState() {
...
 		};
+
+		const releases = app.releases;
+		releases[commit] = releases[commit] || {
+			update_status: 'done',
+			services: {},
+		};
+
+		const svcs = releases[commit].services;
+		svcs[serviceName] = svc;
```
Page-, line 1020:
```suggestion
		releases[commit].services[serviceName] = svc;
```

src/compose/application-manager.ts
```diff
@@ -999,21 +1004,55 @@ export async function getState() {
...
-						...state[appUuid]?.releases[commit]?.services,
-						[serviceName]: svc,
-					},
-				},
-			},
+			releases: {},
 		};
+
+		const releases = app.releases;
+		releases[commit] = releases[commit] || {
```
Page-, line 1014: Should this be
```suggestion
		releases[commit] = releases[commit] ?? {
```
?

src/compose/application-manager.ts
```diff
@@ -999,21 +1004,55 @@ export async function getState() {
 		createdAt,
 		...svc
 	} of servicesToReport) {
-		state[appUuid] = {
-			...state[appUuid],
+		const app = state[appUuid] || {
```
Page-, line 1007: Should this be
```suggestion
		const app = state[appUuid] ?? {
```
?

(also if you update to the latest balena-lint it'll ask this question automatically, but I find it very important to distinguish between when `||` is the intent and `??` is the intent as they mean different things and knowing the true intent of the code really helps at least me to understand it - and it's far easier to know that intent at the time of writing rather than trying to work it out later because if you make a mistake then it'll be a real problem)

src/compose/application-manager.ts
```diff
@@ -788,12 +786,19 @@ function reportOptionalContainers(serviceNames: string[]) {
...
+function reportRejectedReleases(violators: contracts.ContractViolators) {
+	const appStrings = Object.values(violators).map(
+		({ appName, services }) =>
+			`${appName}: Services with unmet requirements: ${services.join(', ')}`,
 	);
+	const message = `Some releases were rejected due to having unmet requirements:\n  ${appStrings.join(
+		'\n  ',
+	)}`;
+	log.error(message);
+	return logger.logSystemMessage(message, { error: true });
```
Page-, line 801: This seems like it'd be a `void` return that might be a leftover from coffeescript? I'm expecting the `return` can just be removed:
```suggestion
	logger.logSystemMessage(message, { error: true });
```
?

src/lib/contracts.ts
pipex: You are right, I think an earlier version could return an empty array for some conditions on the map of `app.releases` which is why I used `flatMap`

pipex: Actually, no, it used a `flatMap` because the mapping over releases also returns a pair. I'll use a `for` loop as you suggested as that makes it clearer

src/lib/contracts.ts
Page-: Yeah, basically I wouldn't worry about these truly useful functions until at a minimum the functions with native equivalents have all moved to those equivalent (`map`, `forEach`, `includes`, etc)

src/device-state/db-format.ts
pipex: But these variables are unused, that's why I used the prefix

src/lib/contracts.ts
```diff
@@ -198,67 +241,45 @@ export function validateContract(contract: unknown): boolean {
...
-							throw new ContractValidationError(serviceName, e.message);
+): Dictionary<AppContractResult> {
+	const result: Dictionary<AppContractResult> = {};
+
+	for (const [appUuid, app] of Object.entries(apps)) {
+		// While app.releases is an object, we expect a target to only
+		// contain a single release per app so we use the first result of
+		// the map call
+		const [appContractResult] = Object.entries(app.releases).map(
+			([commit, release]) => {
```
Page-, line 255: The code seems to imply there's exactly 0 or 1 entries, so this could be more like
```suggestion
		if (app.releases.length === 0) {
			return;
		}
		const [commit, release] = app.releases[0]
```
(and could optionally have a check if it's >1 and fail in that case. As it stands it looks like it implies it can handle arbitrary lengths due to the `.map` when in practice it really doesn't expect that

src/device-state/db-format.ts
Page-: Right, I didn't spot that this was a destructuring, my bad

src/lib/contracts.ts
pipex: Yep, the code is cleaner in this way. Done

pipex:
I self-certify!

PR #2350, opened 2024-07-04
Remove custom typings for docker-delta

pipex:
This are no longer needed since docker-delta migrated to typescript in v4 and the custom typings were actually masking some bugs

This PR fixes the issue where the supervisor would report 

```
Failed to download image 'registry2.balena-cloud.com/v2/120f8fdcef33853bcec70149ab283a55@sha256:a097a71c8ac1d38bab3f089c86adb0aa82d5906045a00117f1f54d1324b43609' due to 'Right-hand side of 'instanceof' is not an object'
```

If it fell into this catch block
https://github.com/balena-os/balena-supervisor/blob/ae823fea181e2b1680940538b04b5af37c1b3229/src/lib/docker-utils.ts#L234-L242

Change-type: patch


pipex:
This still requires a new version of docker-delta exporting `OutOfSyncError`

pipex:
I self-certify!

PR #2355, opened 2024-07-16
Cleanup images after state-engine tests

pipex:
Tests on GitHub started failing recently because of leftover images from the state engine test suite. This fixes that issue to allow tests to pass.

Change-type: patch


pipex:
I self-certify!

Issue #2193, opened 2023-08-29
Evidence of possible memory leak 


pipex:
On devices that have been running a long time, baseline memory usage of the supervisor starts to increase. 

Here is for a device with 100 days uptime

```
root@d6fa190:~# balena inspect balena_supervisor | jq  .[].State.StartedAt
"2023-06-22T12:52:58.569052909Z"
root@d6fa190:~# ps lw $(balena inspect balena_supervisor | jq  .[].State.Pid)
F   UID     PID    PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY        TIME COMMAND
4     0  826430  826414  20   0 568564 321544 do_epo Ssl ?        3210:59 node /usr/src/app/dist/app.js
```

Restarting the supervisor resets the memory usage to a baseline
```
root@f2bd8c7:~# ps lw $(balena inspect balena_supervisor | jq  .[].State.Pid)
F   UID     PID    PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY        TIME COMMAND
4     0 4136240 4136223  20   0 329560 85604 do_epo Ssl  ?          0:08 node /usr/src/app/dist/app.js
```

This is consistent with a memory leak.

We need some profiling to establish where the leak is happening, [memwatch](https://www.npmjs.com/package/memwatch) seems like it could potentially help but there are other ways to do this.

PR #2360, opened 2024-07-31
Memory usage improvements

pipex:
We've had reports and observations of the supervisor slowly increasing memory usage over time providing evidence of a memory leak. We recently uncovered a good candidate for the leak. The supervisor needs to perform long running recurring operations during its runtime (e.g. polling target state) which would be implemented using recursive async calls.

On testing we have seen these calls slowly leaking memory as they become basically infinite promise chains. 

This PR changes the pattern of recursion to break the promise chain which should fix at least one source for the leak.

Relates-to: #2193

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "balena-auth": "^6.0.1",
         "balena-register-device": "^9.0.2",
         "balena-request": "^13.3.1",
-        "blinking": "^0.0.8",
+        "blinking": "^1.0.1",
         "bluebird": "^3.7.2",
         "chai": "^4.3.4",
         "chai-as-promised": "^7.1.1",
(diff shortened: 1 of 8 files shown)


pipex:
I self-certify!

rmorgans:
I think this breaks the supervisor on rasb pi 5's 

Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:   return new FileHandle(await PromisePrototypeThen(
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:                         ^
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]: Error: ENOENT: no such file or directory, open '/sys/class/leds/led0/brightness'
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:     at async open (node:internal/fs/promises:639:25)
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:     at async Object.writeFile (node:internal/fs/promises:1219:14)
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:     at async doBlink (/usr/src/app/dist/app.js:10:367606)
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:     at async start (/usr/src/app/dist/app.js:10:367843) {
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:   errno: -2,
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:   code: 'ENOENT',
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:   syscall: 'open',
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]:   path: '/sys/class/leds/led0/brightness'
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]: }
Aug 07 04:19:50 4c7673d balena-supervisor[1897843]: Node.js v20.15.1
Aug 07 04:19:50 4c7673d systemd[1]: balena-supervisor.service: Main process exited, code=exited, status=1/FAILURE
Aug 07 04:19:50 4c7673d systemd[1]: balena-supervisor.service: Failed with result 'exit-code'.

pipex:
Thank you for reporting @rmorgans, it seems that these errors were swallowed by the blinking module before v1 and now we need to add error handling for this on the supervisor. I'm working on a fix that should be out soon. In the meantime, if you set `LED_FILE=/dev/null` in `/etc/balena-supervisor/supervisor.conf` that should avoid the crash

pipex:
@rmorgans This should fix the issue https://github.com/balena-os/balena-supervisor/pull/2362

PR #2361, opened 2024-08-07
Do not write `noProxy` to redsocks.conf

pipex:
This fixes a regression introduced by the refactor in #2329 where `noProxy` was being included in the data added to redsocks.conf.

Change-type: patch


pipex:
I self-certify!

PR #2362, opened 2024-08-07
Verify that LED_FILE exists on blinking setup

pipex:
Before v1, the blinking module would not throw when the passed led file does not exist. This change checks for file existence and defaults to `/dev/null` otherwise

Change-type: patch


pipex:
I self-certify!

PR #2364, opened 2024-08-21
Avoid unnecessary config calls during startup

cywang117:
- Avoid unnecessary config calls during Supervisor startup
- Fix FromAsCasing Dockerfile warning

Change-type: patch


cywang117:
Tested on device and works fine

cywang117:
@balena-ci I self-certify!

PR #2365, opened 2024-08-22
Revert PR #2364

cywang117:
Reverts #2364


cywang117:
@balena-ci I self-certify!

PR #2366, opened 2024-08-26
Add unit test for usingInferStepsLock

cywang117:
This unit test checks that device-state's `usingInferStepsLock` runs calls at the same depth sequentially.

Change-type: patch


cywang117:
@balena-ci I self-certify!

PR #2373, opened 2024-09-09
Add support for init field from compose

cywang117:
Init supports boolean values, and is not included in the config when not defined.

Tested on device and verified to be working with releases where init field is as follows:
- `true`
- `false`
- `"true"` (normalized to `true` by compose)
- `"false"` (normalized to `false`)
- not defined

Change-type: patch


pipex (approved):
Nice one

PR #2375, opened 2024-09-24
Add NXP support to balenaOS secure boot

alexgg:
On NXP iMX devices the partitions are not encrypted with LUKS but with the lower level dm-crypt subsystem.

Adapt the partition mount script to use dmsetup which works for both LUKS and dm-crypt encrypted partitions.

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


alexgg:
Pending tests in https://github.com/balena-os/meta-balena/pull/3541

alexgg:
Passed tests:

* [Yocto / Test (hup, balena-cloud.com, qemu, self-hosted, X64, kvm)]( https://github.com/balena-os/meta-balena/actions/runs/11476598260/job/31945399971?pr=3541)
*[ Yocto / Test (os, balena-cloud.com, qemu, self-hosted, X64, kvm)](https://github.com/balena-os/meta-balena/actions/runs/11476598225/job/31945364632?pr=3541)
* [Yocto / Test (os, balena-cloud.com, qemu, self-hosted, X64, kvm)](https://github.com/balena-os/meta-balena/actions/runs/11476598225/job/31945364632?pr=3541)

alexgg:
Secure boots tests on balena-generic:
https://github.com/balena-os/balena-generic/actions/runs/11482941263

mount-partitions.sh
```diff
@@ -30,14 +30,58 @@ dbus_get_mount() {
...
+				# Does not start with mapper/
+				# Find the corresponding DM device to the partition
+				_dm_part=$(lsblk -nlo kname "/dev/${_dm_part}" | grep dm)
+				if [ -z "${_dm_part}" ]; then
+					# No corresponding DM device, no dm-crypt in use
+					return 1
+				fi
+			fi
+		fi
+		# _dm_part is a DM device, either dm* or mapper/*
```
pipex, line 56: This point will be reached either if 

- `_part` starts with `dm-`
- `_part` starts with `mapper/`
-  `lsblk -nlo kname "/dev/${_dm_part}" | grep dm` returns something

Is that the logic that is intended? (I'm not familiar with the partition layout in NXP devices)

alexgg: @pipex if the device starts with `dm-` or `mapper` it is a DM device, if not, if falls back to listing the kernel device name and makes sure it does not contain `dm` before discarding a DM device.
If it is a DM device, then it will use its name to filter the `dmsetup` list of dm-crypt devices.
In all, it tries hard to identify non-dm devices and confirm encrypted dm devices.

mount-partitions.sh
```diff
@@ -30,14 +30,58 @@ dbus_get_mount() {
...
+# 0: The partition device is encrypted
+# 1: The partition device is not encrypted
+is_part_encrypted() {
+	_part="${1#/dev/}"
+	_dm_part="${_part}"
+	if command -v dmsetup > /dev/null; then
+		if [ "${_part#dm-}" = "${_part}" ]; then
+			# Does not start with dm-
+			if [ "${_part#mapper/}" = "${_part}" ]; then
+				# Does not start with mapper/
```
pipex, line 47:
```suggestion
		if [ "${_part#dm-}" = "${_part}" ] || [ "${_part#mapper/}" = "${_part}" ]; then
			# does not start with `dm-` or `mapper/`
```

mount-partitions.sh
pipex: I see, I added one suggestion above, otherwise it looks good to me. I cannot, however provide any suggestions relating to the used tools or the NXP context

mount-partitions.sh
alexgg: @pipex I think you meant to use `&&`:
```
if [ "${_part#dm-}" = "${_part}" ] && [ "${_part#mapper/}" = "${_part}" ]
```
I preferred the two lines as it is more legible and the `&&` is implicit.

klutchell:
LGTM

PR #2379, opened 2024-10-16
Support jetson power fan configs

cywang117:
Support `HOST_CONFIG_power_mode` and `HOST_CONFIG_fan_profile` with new `PowerFanConfig` backend, initially for Jetpack 6 devices. The configs are received in target state and written to config.json, where os services will pick up the changes and apply the configs.

This PR combined with https://github.com/balena-os/balena-supervisor/pull/2382 constitutes the complete power fan config feature.

Change-type: minor

### DIFF
--- a/src/config/backends/config-txt.ts
+++ b/src/config/backends/config-txt.ts
@@ -70,14 +70,15 @@ function isBaseParam(dtparam: string): boolean {
  * 	- {BALENA|RESIN}_HOST_CONFIG_gpio = value | "value" | "value1","value2"
  */
 export class ConfigTxt extends ConfigBackend {
-	private static bootConfigVarPrefix = `${constants.hostConfigVarPrefix}CONFIG_`;
-	private static bootConfigPath = hostUtils.pathOnBoot('config.txt');
-
-	public static bootConfigVarRegex = new RegExp(
-		'(?:' + _.escapeRegExp(ConfigTxt.bootConfigVarPrefix) + ')(.+)',
+	private static PREFIX = `${constants.hostConfigVarPrefix}CONFIG_`;
+	private static PATH = hostUtils.pathOnBoot('config.txt');
+	private static REGEX = new RegExp(
+		'(?:' + _.escapeRegExp(ConfigTxt.PREFIX) + ')(.+)',
 	);
-
-	private static forbiddenConfigKeys = [
+	// These keys are not config.txt keys and are managed by the power-fan backend.
+	private static UNSUPPORTED_KEYS = ['power_mode', 'fan_profile'];
+	// These keys are config.txt keys, but are not mutable by the Supervisor.
+	private static FORBIDDEN_KEYS = [
 		'disable_commandline_tags',
 		'cmdline',
 		'kernel',
@@ -89,7 +90,7 @@ export class ConfigTxt extends ConfigBackend {
 		'device_tree_address',
 		'init_emmc_clock',
 		'avoid_safe_mode',
-	];
+	].concat(ConfigTxt.UNSUPPORTED_KEYS);
 
 	public async matches(deviceType: string): Promise<boolean> {
 		return (
@@ -109,11 +110,8 @@ export class ConfigTxt extends ConfigBackend {
 	public async getBootConfig(): Promise<ConfigOptions> {
 		let configContents = '';
 
-		if (await exists(ConfigTxt.bootConfigPath)) {
-			configContents = await hostUtils.readFromBoot(
-				ConfigTxt.bootConfigPath,
-				'utf-8',
-			);
+		if (await exists(ConfigTxt.PATH)) {
+			configContents = await hostUtils.readFromBoot(ConfigTxt.PATH, 'utf-8');
 		} else {
 			return {};
 		}
@@ -227,19 +225,29 @@ export class ConfigTxt extends ConfigBackend {
 		}
 
 		const confStr = `${confStatements.join('\n')}\n`;
-		await hostUtils.writeToBoot(ConfigTxt.bootConfigPath, confStr);
+		await hostUtils.writeToBoot(ConfigTxt.PATH, confStr);
+	}
+
+	public static stripPrefix(name: string): string {
+		if (!name.startsWith(ConfigTxt.PREFIX)) {
+			return name;
+		}
+		return name.substring(ConfigTxt.PREFIX.length);
 	}
 
 	public isSupportedConfig(configName: string): boolean {
-		return !ConfigTxt.forbiddenConfigKeys.includes(configName);
+		return !ConfigTxt.FORBIDDEN_KEYS.includes(configName);
 	}
 
 	public isBootConfigVar(envVar: string): boolean {
-		return envVar.startsWith(ConfigTxt.bootConfigVarPrefix);
+		return (
+			envVar.startsWith(ConfigTxt.PREFIX) &&
+			!ConfigTxt.UNSUPPORTED_KEYS.includes(ConfigTxt.stripPrefix(envVar))
+		);
 	}
 
 	public processConfigVarName(envVar: string): string {
-		return envVar.replace(ConfigTxt.bootConfigVarRegex, '$1');
+		return envVar.replace(ConfigTxt.REGEX, '$1');
 	}
 
 	public processConfigVarValue(key: string, value: string): string | string[] {
@@ -254,7 +262,7 @@ export class ConfigTxt extends ConfigBackend {
 	}
 
 	public createConfigVarName(configName: string): string {
-		return ConfigTxt.bootConfigVarPrefix + configName;
+		return ConfigTxt.PREFIX + configName;
 	}
 
 	// Ensure that the balena-fin overlay is defined in the target configuration
--- a/src/config/backends/power-fan.ts
+++ b/src/config/backends/power-fan.ts
@@ -1,0 +1,159 @@
+import { isRight } from 'fp-ts/lib/Either';
+import Reporter from 'io-ts-reporters';
+import * as t from 'io-ts';
+
+import { ConfigBackend } from './backend';
+import type { ConfigOptions } from './backend';
+import { schemaTypes } from '../schema-type';
+import log from '../../lib/supervisor-console';
+import * as constants from '../../lib/constants';
+
+type ConfigJsonBackend = {
+	get: (key: 'os') => Promise<unknown>;
+	set: (opts: { os: Record<string, any> }) => Promise<void>;
+};
+
+/**
+ * A backend to handle Jetson power and fan control
+ *
+ * Supports:
+ * 	- {BALENA|RESIN}_HOST_CONFIG_power_mode = "low" | "mid" | "high" | "default" |"$MODE_ID"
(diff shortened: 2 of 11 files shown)


src/config/backends/power-fan.ts
```diff
@@ -0,0 +1,183 @@
...
+
+	public async isRebootRequired(opts: ConfigOptions): Promise<boolean> {
+		const supportedOpts = _.pickBy(
+			_.mapKeys(opts, (_value, key) => PowerFanConfig.stripPrefix(key)),
+			(_value, key) => this.isSupportedConfig(key),
+		);
+		const current = await this.getBootConfig();
+		// A reboot is only required if the power mode is changing
+		return current.power_mode !== supportedOpts.power_mode;
+	}
```
pipex, line 170: Why was it decided to bother with this? Before, all host config changes would require a reboot. Even if it is not necessary for the fan profile, it doesn't seem worth it the extra complexity just to avoid a reboot in that case.

src/config/configJson.ts
```diff
@@ -12,17 +12,23 @@ export default class ConfigJsonConfigBackend {
...
+	 */
 	private readonly configPath?: string;
 
 	private cache: { [key: string]: unknown } = {};
 
-	private readonly init = _.once(async () =>
-		Object.assign(this.cache, await this.read()),
-	);
+	private readonly init = _.once(async () => {
+		if ((await osRelease.getOSVersion(constants.hostOSVersionPath)) == null) {
```
pipex, line 23: Why move this check here? Getting the OS version was only there to know where to look for config.json no?

pipex: Actually, I'm not sure of the point of the check

src/config/backends/config-txt.ts
```diff
@@ -231,11 +232,21 @@ export class ConfigTxt extends ConfigBackend {
 	}
 
 	public isSupportedConfig(configName: string): boolean {
-		return !ConfigTxt.forbiddenConfigKeys.includes(configName);
+		return (
+			!ConfigTxt.forbiddenConfigKeys.includes(configName) &&
+			// power_mode and fan_profile are managed by the power-fan backend, so
+			// need to be excluded here as the config var name prefix is the same.
+			!isPowerFanSupportedConfig(configName)
```
pipex, line 239: This breaks the independence between backends, why use the same prefix `HOST_CONFIG` for these new configurations rather than use something different like `HOST_OS` which is the way we would distinguish configurations before?

pipex: Why not add it to `forbiddenConfigKeys` instead?

src/config/backends/power-fan.ts
```diff
@@ -0,0 +1,183 @@
...
+import * as t from 'io-ts';
+import * as _ from 'lodash';
+
+import { ConfigBackend } from './backend';
+import type { ConfigOptions } from './backend';
+import { schemaTypes } from '../schema-type';
+import log from '../../lib/supervisor-console';
+import * as constants from '../../lib/constants';
+
+const isNullOrUndefined = (v: unknown): v is null | undefined =>
```
pipex, line 12: `x == null` already tests whether x is null or undefined

src/config/backends/power-fan.ts
```diff
@@ -0,0 +1,183 @@
...
+		if (isRight(powerFanConfig)) {
+			const conf = powerFanConfig.right;
+			return {
+				...(!isNullOrUndefined(conf.power?.mode) && {
+					power_mode: conf.power.mode,
+				}),
+				...(!isNullOrUndefined(conf.fan?.profile) && {
+					fan_profile: conf.fan.profile,
+				}),
+			};
```
pipex, line 100: This is not super readable, you could probably do something like
```suggestion
			return {
				...(conf.power?.mode != null && {
					power_mode: conf.power.mode,
				}),
				...(conf.fan?.profile != null && {
					fan_profile: conf.fan.profile,
				}),
			};
```

src/config/backends/power-fan.ts
```diff
@@ -0,0 +1,183 @@
...
+		}
+		// Current config could be undefined if there's no os key in config.json, so default to empty object
+		const currentConf = maybeCurrentConf.right ?? {};
+
+		// Filter out unsupported options in target config
+		const supportedOpts = Object.fromEntries(
+			Object.entries(opts).filter(([key]) => this.isSupportedConfig(key)),
+		) as { power_mode?: string; fan_profile?: string };
+
+		const targetConf = structuredClone(currentConf);
```
pipex, line 133: Why is this clone needed?

src/config/backends/power-fan.ts
```diff
@@ -0,0 +1,183 @@
...
+
+		// Update or delete power mode
+		if ('power_mode' in supportedOpts) {
```
pipex, line 136: Seems like supported opts is unnecessary since you already have this check
```suggestion
		if ('power_mode' in opts) {
```

src/config/configJson.ts
```diff
@@ -37,14 +43,10 @@ export default class ConfigJsonConfigBackend {
 		await Bluebird.using(this.writeLockConfigJson(), async () => {
 			let changed = false;
 			_.forOwn(keyVals, (value, key: T) => {
-				if (this.cache[key] !== value) {
+				if (this.schema[key] != null && !_.isEqual(this.cache[key], value)) {
```
pipex, line 46: Could you comment on the reason to move this clause here?

```
this.schema[key] != null
```

pipex (commented):
Seems generally OK, just some questions that relate to the discussion on the project

src/config/configJson.ts
cywang117: I'm not sure either but was iffy about removing it. Looking at the section though, the check could be the determine that config.json exists when initializing `ConfigJsonConfigBackend`, because OS version should always be present in config.json on first boot. Since that's the case, we can remove this check and just allow `await this.read()` to fail if config.json doesn't exist.

src/config/configJson.ts
cywang117: This offers better protection against changing values in config.json that aren't part of the Supervisor's config.json schema. If the non-null schema check is in the inner `if` branch, there is the possibility of changing the value for a key (without deleting it) even if the key is not present in the config.json schema. 

Although, technically, the type signature for the class setter already restricts possible keys to those in the schema. Should we remove this check altogether?

src/config/backends/config-txt.ts
cywang117: Not sure about `HOST_OS`, the only occurrence of that I can see is the `HOST_OS_VERSION` env var. I'll move to adding to `forbiddenConfigKeys` instead.

cywang117: With latest iteration https://github.com/balena-os/balena-supervisor/pull/2379/commits/52f3258ffeb7f4dea1d5cb1d3f97bc250bff2bf1, forbiddenConfigKeys is now separated into forbidden and unsupported keys:
```
// These keys are not config.txt keys and are managed by the power-fan backend.
private static UNSUPPORTED_KEYS = ['power_mode', 'fan_profile'];
// These keys are config.txt keys, but are not mutable by the Supervisor.
private static FORBIDDEN_KEYS = [
	'disable_commandline_tags',
	'cmdline',
	'kernel',
	'kernel_address',
	'kernel_old',
	'ramfsfile',
	'ramfsaddr',
	'initramfs',
	'device_tree_address',
	'init_emmc_clock',
	'avoid_safe_mode',
].concat(ConfigTxt.UNSUPPORTED_KEYS);
```

The concept of unsupported keys is used by `isBootConfigVar` to filter out non-config.txt keys, but keep in forbidden config.txt keys such as initiramfs. When `isBootConfigVar` is called by `envToBootConfig` in `bootConfigChangeRequired`, this will allow `bootConfigChangeRequired` to provide the proper feedback to the user that they're attempting to set a forbidden (but not unsupported) key.

This still isn't ideal IMO, I think `isSupportedConfig`/`isBootConfigVar` are a bit redundant in terms of functionality. The config module could be simplified in this regard, but I would save this for another PR.

src/config/backends/power-fan.ts
cywang117: You're right, removed this in https://github.com/balena-os/balena-supervisor/pull/2379/commits/52f3258ffeb7f4dea1d5cb1d3f97bc250bff2bf1

src/config/backends/power-fan.ts
cywang117: Modified in https://github.com/balena-os/balena-supervisor/pull/2379/commits/52f3258ffeb7f4dea1d5cb1d3f97bc250bff2bf1

src/config/backends/power-fan.ts
cywang117: We have discussed removing the cache for ConfigJsonConfigBackend, I would consider the above behavior an example of why the cache is subpar, but I would save this for another PR.

src/config/backends/power-fan.ts
cywang117: True, modified in https://github.com/balena-os/balena-supervisor/pull/2379/commits/52f3258ffeb7f4dea1d5cb1d3f97bc250bff2bf1

src/config/backends/power-fan.ts
pipex: The fact that this module needs to be aware of internal implementation details of the ConfigJson backend is a red flag. I suspect in that case the structured clone should go on the ConfigJson backend instead of here. 

I don't think this is an issue with the cache, which exists for a reason, is just a flawed interface.

src/config/backends/power-fan.ts
pipex: Why does it cause storage wear? How often do we expect users to be changing the fan control speed? 

I agree that there are use cases for changing configurations without a reboot. 

I don't think the config variable interface is great for that. Config variables require an API call and a target state pull before the changes are applied, is not an interface made for runtime requirements.

If there is need for such an interface then we should work on that rather than trying to fit the config var interface for a specific use case

src/config/configJson.ts
pipex: > Although, it's also fine if we remove this check, since the type signature for the class setter already restricts possible keys to those in the schema. Should we remove this check altogether?

I think let's keep it for now. I don't trust the codebase to not be using `as any` or failing to check for nulls on an array element (see [strictNullChecks](https://www.typescriptlang.org/tsconfig/#strictNullChecks))

src/config/backends/power-fan.ts
cywang117: Agreed, I think this feature has exposed some limits of the way we handle configs on top of what we were aware of before.

src/config/backends/power-fan.ts
cywang117: Modified structuredClone to be called in `configJson.get` [here](https://github.com/balena-os/balena-supervisor/pull/2379/commits/828bd22ba053aa08ca2bde8e72b3aa7ed82aae02#diff-3a034722503c6a7e6a3654f6f42c046256faaca32ff3d770ba5d727efff82470R60-R61)

pipex (approved):

PR #2381, opened 2024-10-24
Add support for `io.balena.update.requires-reboot` label

pipex:
This label can be used by user services to indicate that a reboot is required after the install of a service in order to fully apply an update. In particular, this will be used by hostOS updates when we add this feature on the supervisor.

Change-type: minor
Depends-on: #2389

## Release notes

This release adds support for a new docker-compose label: `io.balena.update.requires-reboot`. When the label is present on a service, it tells the supervisor that a system reboot should be performed before starting the service for the first time. 

When a target release is received that contains services with this label, the supervisor will
- pull all target images
- create and start containers for services without the label (in dependency order).
- create containers for services with the label, but skip start
- wait until locks are released to perform a system reboot

Note that the supervisor cannot distinguish between a first time install of the service or the container being re-created. This means that actions that re-create services will also require a reboot. Ex: restart service, purge data, disable local-mode will all re-create containers and hence require a reboot.

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -654,6 +654,7 @@ class AppImpl implements App {
 				context.targetApp,
 				needsDownload,
 				servicesLocked,
+				context.rebootBreadcrumbSet,
 				context.appsToLock,
 				context.availableImages,
 				context.networkPairs,
@@ -682,6 +683,8 @@ class AppImpl implements App {
 					context.appsToLock,
 					context.targetApp.services,
 					servicesLocked,
+					context.rebootBreadcrumbSet,
+					context.bootTime,
 				);
 			}
 
@@ -761,6 +764,8 @@ class AppImpl implements App {
 		appsToLock: AppsToLockMap,
 		targetServices: Service[],
 		servicesLocked: boolean,
+		rebootBreadcrumbSet: boolean,
+		bootTime: Date,
 	): CompositionStep[] {
 		// Update container metadata if service release has changed
 		if (current.commit !== target.commit) {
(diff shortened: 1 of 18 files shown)


src/compose/app.ts
```diff
@@ -774,16 +778,36 @@ class AppImpl implements App {
...
-					appsToLock[target.appId].add(s.serviceName);
-				});
-				return [];
-			}
 			if (target.config.running) {
+				// if the container has a reboot
+				// required label and the boot time is before the creation time, then
+				// return a 'noop' to ensure a reboot happens before starting the container
+				const requiresReboot =
+					target.config.labels?.['io.balena.update.requires-reboot'] !== null &&
```
cywang117, line 786:
```suggestion
					target.config.labels?.['io.balena.update.requires-reboot'] === 'true' &&
```
Should we be more restrictive with the value of the label? Or at least, not return true for `undefined`:
```suggestion
					target.config.labels?.['io.balena.update.requires-reboot'] != null &&
```

src/device-api/actions.ts
```diff
@@ -151,9 +151,6 @@ export const doPurge = async (appId: number, force: boolean = false) => {
 			keepVolumes: false,
 			force,
 		});
-		// Restore user app after purge
-		currentState.local.apps[appId] = app;
-		await deviceState.applyIntermediateTarget(currentState);
 		logger.logSystemMessage('Purged data', { appId }, 'Purge data success');
```
cywang117, line 154: Optional change, but the message wouldn't technically be accurate as a purge has not yet occurred yeah?

cywang117 (commented):
Looking good on initial pass. Will do another pass on Mon

src/device-state/device-config.ts
```diff
@@ -396,6 +402,7 @@ function getConfigSteps(
 	target: Dictionary<string>,
 ) {
 	const configChanges: Dictionary<string> = {};
+	const rebootingChanges: Dictionary<string> = {};
```
cywang117, line 405: I'm a bit confused by the reboot changes in the device-config module. If a reboot breadcrumb step is being added where there was none before, how were device-config changes resulting in reboot before?

src/device-state/device-config.ts
```diff
@@ -524,7 +534,12 @@ async function getBackendSteps(
 	return [
 		// All backend steps require a reboot
 		...(steps.length > 0
-			? [{ action: 'setRebootBreadcrumb' } as ConfigStep]
+			? [
+					{
+						action: 'setRebootBreadcrumb',
+						target: { bootConfigChanged: 'true' },
```
cywang117, line 540: Why a string boolean?

cywang117 (commented):
Did another pass -- the changes make sense. Great work 👍🏼

src/device-api/actions.ts
pipex: The purge happens when the `applyIntermediate` call on 150 finishes, because that's what removes the app. The call on finalize re-installs the app

src/compose/app.ts
pipex: You are right, we use `checkTruthy` for the other lablels so it makes sense to be consistent

https://github.com/balena-os/balena-supervisor/blob/81b307510d758c9390836acd2d8547e1263eaf35/src/compose/utils.ts#L413

src/device-state/device-config.ts
pipex: I think is just because of the existing types of the actions on device-config, I'll check if it makes sense to refactor

pipex: I removed the target, the message won't make a lot of change anyway

cywang117 (approved):
Great work!

PR #2382, opened 2024-10-31
Stream host logs from os-power-mode.service and os-fan-profile.service

cywang117:
Add ability to stream logs from host services to cloud

Add `os-power-mode.service` and `os-fan-profile.service` which report status
from applying power mode and fan profile configs as read from config.json.
The Supervisor will set these configs in config.json for these host services
to pick up and apply in #2379.

This feature is separate enough in functionality that it's fine to be merged separately, but the power mode / fan profile feature is not considered complete on the Supervisor end until #2379 is merged. meta-balena upgrades that wish to include the feature in its entirety should upgrade to the Supervisor version from #2379.

This feature has been tested on a jetson draft OS image that includes `os-power-mode.service` and `os-fan-profile.service`, by editing config.json and observing the dashboard for correct streaming of logs from these services.

Relates-to: https://github.com/balena-os/balena-supervisor/pull/2379
See: https://github.com/balena-io/open-balena-api/pull/1792
See: https://github.com/balena-os/balena-jetson-orin/pull/513 

Change-type: minor


cywang117:
Screenshot 2024-11-07 at 17 32 57

cywang117:
@pipex @acostach This feature is ready for review and does stream both os-power-mode and os-fan-profile logs now! Apologies for the noise before, I was experimenting with different methods to get the os-power-mode logs to stream, but now the feature is ready for review. Above is a screenshot of how it looks on the dashboard. 

os-power-mode logs weren't streaming because on Supervisor startup, the Supervisor logs a `Supervisor starting` message with a later timestamp than the os-power-mode logs. I've observed that logs earlier than the timestamp don't go through. The solution here is to stream the `Supervisor starting` log after any host logs that may occur early during device startup, such as os-power-mode.

The log-streaming implementation is getting more complex with this PR however. We should strongly consider this support for streaming of host logs in the Supervisor as an intermediate step towards a dedicated log-streaming service on host.

acostach:
Tested on AGX64 and Rockpi 4B, image 758ceb59f0a928fb4408987a6da1252c

pipex:
In the screenshot above, why is the unit name repeated @cywang117 ? Is one of the values already on the journal?

src/logging/monitor.ts
```diff
@@ -113,12 +166,45 @@ class LogMonitor {
 			JOURNALCTL_ERROR_RETRY_DELAY_MAX,
 		);
 		log.debug(
-			`Spawning another process to watch balena.service logs in ${
-				wait / 1000
-			}s`,
+			`Spawning another process to watch host service logs in ${wait / 1000}s`,
```
pipex, line 169:
```suggestion
			`Spawning another process to watch journal logs in ${wait / 1000}s`,
```

src/logging/monitor.ts
```diff
@@ -51,37 +56,75 @@ async function* splitStream(chunkIterable: AsyncIterable<any>) {
...
 		};
 	} = {};
 	private setupAttempts = 0;
 
-	// Only stream logs since the start of the supervisor
+	// By default, only stream logs since the start of the Supervisor process
 	private lastSentTimestamp = Date.now() - performance.now();
 
-	public async start(): Promise<void> {
+	public async start(logSystemMessage: LogSystemMessage): Promise<void> {
```
pipex, line 113: I don't understand this new interface. The `start` function receives a function type that it calls if it sees the "Supervisor starting up" message. The `logSystemMessage` function then calls `log` on this same object.

I don't think the interface is very clean, it gives additional responsibilities to the monitor interface and introduces circular dependencies.

src/logging/monitor.ts
```diff
@@ -96,15 +139,25 @@ class LogMonitor {
 							self.containers[row.CONTAINER_ID_FULL]
 						) {
 							await self.handleRow(row);
+						} else if (self.HOST_SERVICES.includes(row._SYSTEMD_UNIT)) {
+							await self.handleHostServiceRow(row, logSystemMessage);
+						} else if (isSupervisorStartupMessage(row)) {
+							// Log Supervisor startup message, but in chronological order,
+							// allowing any earlier timestamped logs to be processed first.
```
pipex, line 146: Did you get to the bottom of the problem? How is it that before we could include logs from before the supervisor start when we stored the logs on the database? Is it something on the backend that changed that does not allow sending logs out of order?

I think we can test this by sending some logs with `curl` to the backend and playing with timestamps

src/logging/monitor.ts
```diff
@@ -113,12 +166,45 @@ class LogMonitor {
...
+			const supervisorContainer = await getSupervisorContainer();
+			if (supervisorContainer !== null) {
+				const finishedAt = supervisorContainer.State.FinishedAt;
+				const finishedAtDate = new Date(finishedAt).getTime();
+				// When a container has never exited with any exit code,
+				// the FinishedAt timestamp is "0001-01-01T00:00:00Z". Any
+				// timestamp below 0 in ms value is invalid, so use the process
+				// start time instead.
+				if (finishedAtDate > 0) {
+					this.lastSentTimestamp = finishedAtDate;
```
pipex, line 189: Are you sure this works? What happens if the underlying `journalctl` process dies? If I'm reading this correctly the `start` method will be called again and it will repeat logs from before the supervisor started, right?

src/logging/monitor.ts
pipex: Ok, I understand why we need to pass the `logSystemMessage` to monitor. Is because there are only container hooks, but not a `system` hook. 

I propose these changes
- Reuse the MonitorHook interface as the system hook, do not use the `LogSystemMessage` type, that function signature is very hard to understand and we should change it anyway. This gets rid of the circular dependency
- Do not pass the hook on start. Add an `attachSystemLogger` (or something like that) to the object interface. That mimics what already exists for container hook. Passing the hook on the `start` function does not make a lot of sense

I still don't like that this module now also has to report supervisor start. I would try first to get to the bottom of why we have issues with the out of order messages. If that's indeed an issue, then perhaps we could defer system messages to after start or skip the "supervisor" start message altogether if there are previous logs.

I also don't like depending on the "Supervisor starting up" message being on the logs as it creates a hidden interface

cywang117:
Screenshot 2024-11-18 at 18 12 03

src/logging/monitor.ts
cywang117: Resolved [here](https://github.com/balena-os/balena-supervisor/pull/2382/commits/a0fb3d8ddece08b40ad89308273db37dd33d2447#diff-ba81f47a86a6a2d1ec1033fcc6d92c30d75d3557278a367721609771c55f4662R116)

src/logging/monitor.ts
cywang117: I ran some additional tests after making some changes to the code per your review @pipex , and found that log-streaming of timestamps earlier than the `Supervisor starting` log line is now working. The changes I made were to `attachSystemHook` within the `initialized` method of `src/logger.ts` (now merged into `src/logging/index.ts`, henceforth referenced as Logger), and to use the generic `log` method provided by the Logger instead of `logSystemMessage`. I gather that this is what made a difference, however I am only certain in the fact that the log-streaming works now, and not much else.

Therefore, the changes no longer depend on `Supervisor starting` log line being in the logs and no longer have to parse that. The hook is also not passed on `logMonitor.start`, instead passed with logMonitor.attachSystemHook` which is called within `logger.initialized`. This simplifies the code from the previous iteration.

src/logging/monitor.ts
cywang117: My deduction of the behavior I observed being due to the backend not accepting out-of-order logs was an incorrect deduction. I suspect now that there was something weird in the way I set up the system log-streaming, or the use of `logSystemMessage` when calling `logMonitor.start`.

src/logging/monitor.ts
cywang117: Good catch, I modified the `lastSentTimestamp` calculation logic to only return the last FinishedAt timestamp of the Supervisor if the lastSentTimestamp is null (i.e. this is the first time logMonitor has been started). Changes [here](https://github.com/balena-os/balena-supervisor/pull/2382/commits/2ffac417ba363671da98546360fe6495b6fa1e28#diff-ba81f47a86a6a2d1ec1033fcc6d92c30d75d3557278a367721609771c55f4662R156-R199)

src/logging/monitor.ts
```diff
@@ -51,37 +56,75 @@ async function* splitStream(chunkIterable: AsyncIterable<any>) {
...
+		let container: ContainerInspectInfo | null = null;
+		try {
+			container = await docker.getContainer('balena_supervisor').inspect();
+		} catch {
+			// Ignore errors getting the supervisor container
+		}
+
+		try {
+			container = await docker.getContainer('resin_supervisor').inspect();
+		} catch {
```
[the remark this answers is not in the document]
cywang117: Updated per suggestion [here](https://github.com/balena-os/balena-supervisor/pull/2382/commits/2ffac417ba363671da98546360fe6495b6fa1e28#diff-ba81f47a86a6a2d1ec1033fcc6d92c30d75d3557278a367721609771c55f4662R62)

src/logging/monitor.ts
```diff
@@ -1,8 +1,12 @@
 import { pipeline } from 'stream/promises';
+import { setTimeout } from 'timers/promises';
 
+import type { SpawnJournalctlOpts } from '../lib/journald';
 import { spawnJournalctl, toJournalDate } from '../lib/journald';
 import log from '../lib/supervisor-console';
-import { setTimeout } from 'timers/promises';
+import type { LogSystemMessage } from '../logger';
```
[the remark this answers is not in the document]
cywang117: Removed circular dependency by using the `MonitorHook` type per suggestion

src/logging/monitor.ts
```diff
@@ -144,6 +153,50 @@ class LogMonitor {
...
+				// - finishedAtDate is greater than 0 (i.e. the supervisor container has exited at least once)
+				// - lastSentTimestamp is null (i.e. this is the first time we've started the monitor)
+				//   - This prevents the case of the logs getting streamed from State.FinishedAt for
+				//     for subsequent monitor.start() calls due to the underlying journalctl process dying.
+				if (finishedAtDate > 0 && this.lastSentTimestamp == null) {
+					this.lastSentTimestamp = finishedAtDate;
+				}
+			}
+		} catch {
+			// Ignore errors getting the supervisor container
```
pipex, line 178: Is this catch necessary since you are catching errors in the `getSupervisorContainer` function?

src/logging/monitor.ts
```diff
@@ -1,14 +1,18 @@
 import { pipeline } from 'stream/promises';
+import { setTimeout } from 'timers/promises';
 
 import { spawnJournalctl, toJournalDate } from '../lib/journald';
 import log from '../lib/supervisor-console';
-import { setTimeout } from 'timers/promises';
 
-export type MonitorHook = (message: {
+type MonitorHookMessage = {
```
pipex, line 10: You could probably extract `LogMessage` from log-backend into its own module and use that here and in the backends

pipex (commented):
Only minor comments @cywang117. Great work

src/logging/monitor.ts
```diff
@@ -135,6 +160,16 @@ class LogMonitor {
 		delete this.containers[containerId];
 	}
 
+	public attachSystemLogger(hook: SystemMonitorHook) {
+		if (this.systemHook == null) {
+			this.systemHook = hook;
+		}
+	}
+
+	public detachSystemLogger() {
```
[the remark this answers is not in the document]
cywang117: You're right, removed this

src/logging/monitor.ts
```diff
@@ -18,6 +22,7 @@ interface JournalRow {
 	MESSAGE: string | number[];
 	PRIORITY: string;
 	__REALTIME_TIMESTAMP: string;
+	_SYSTEMD_UNIT?: string;
```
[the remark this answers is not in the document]
cywang117: True, checking the [journal fields docs](https://www.freedesktop.org/software/systemd/man/latest/systemd.journal-fields.html), you're right that `_SYSTEMD_UNIT` is not optional.

src/logging/monitor.ts
```diff
@@ -157,6 +192,23 @@ class LogMonitor {
...
+	private async handleHostServiceRow(
+		row: JournalRow & { _SYSTEMD_UNIT: string },
+	) {
+		const message = messageFieldToString(row.MESSAGE);
+		if (message == null) {
+			return;
+		}
+		const isStdErr = row.PRIORITY === '3';
+		const timestamp = Math.floor(Number(row.__REALTIME_TIMESTAMP) / 1000); // microseconds to milliseconds
+		void this.systemHook?.({
```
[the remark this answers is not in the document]
cywang117: Done 👍🏼

src/logging/monitor.ts
```diff
@@ -144,6 +153,50 @@ class LogMonitor {
...
+			// On SV start, journalctl is spawned with a timestamp to only
+			// get logs since the last Supervisor State.FinishedAt. This will catch any
+			// host and container logs generated while the Supervisor was not running.
+			const supervisorContainer = await getSupervisorContainer();
+			if (supervisorContainer !== null) {
+				const finishedAt = supervisorContainer.State.FinishedAt;
+				const finishedAtDate = new Date(finishedAt).getTime();
+				// When a container has never exited with any exit code,
+				// the FinishedAt timestamp is "0001-01-01T00:00:00Z". Any
+				// timestamp below 0 in ms value is invalid.
```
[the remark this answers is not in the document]
cywang117: Noted and modified the comment to indicate this 👍🏼

src/logging/monitor.ts
cywang117: True, removed this try/catch

src/logging/monitor.ts
cywang117: Added [src/logging/types.ts](https://github.com/balena-os/balena-supervisor/pull/2382/commits/ebfcdc74ff396fd06bc5887687d0cc66bd8d7ae0#diff-b7efb7d611599542fae83eb6acd5685ad5a3af21ae51edfb220e87053497158fR1) which defines `BaseLogMessage`, `SystemLogMessage`, and `LogMessage` which can be either system or container log messages. Then [redefined](https://github.com/balena-os/balena-supervisor/pull/2382/commits/ebfcdc74ff396fd06bc5887687d0cc66bd8d7ae0#diff-ba81f47a86a6a2d1ec1033fcc6d92c30d75d3557278a367721609771c55f4662R8-R9) `MonitorHook` and `SystemMonitorHook` to refer to the relevant `LogMessage` type from types.ts.

src/logging/index.ts
```diff
@@ -1,5 +1,229 @@
...
+	void log(msgObj);
+	if (track) {
+		eventTracker.track(
+			eventName != null ? eventName : message,
+			eventObj != null ? eventObj : {},
+		);
+	}
+}
+
+export function lock(containerId: string): Bluebird.Disposer<() => void> {
```
pipex, line 134: I think this function doesn't need to be exported

src/logging/monitor.ts
```diff
@@ -55,11 +54,26 @@ async function* splitStream(chunkIterable: AsyncIterable<any>) {
  * Streams logs from journalctl and calls container hooks when a record is received matching container id
  */
 class LogMonitor {
+	// Additional host services we want to stream the logs for
+	private HOST_SERVICES = [
```
pipex, line 74: Not pushing too much for this, but I think this would be better as a constant in the module. That would avoid the need for using `this.` to reference it

pipex (commented):
Great work @cywang117. A couple of tiny last minute comments. This is pretty much ready to merge otherwise

pipex (approved):
LGTM

PR #2383, opened 2024-11-04
Delete apps not in target from db by appUuid instead of appId

cywang117:
Resolve an issue in balenaMachine instances that were installed at <v14.1.0, in which a Supervisor app with random UUID is kept in the target db due to its appId being the same, even after the BM instance has upgraded to v14.1.0 which patches the correct reserved Supervisor app UUIDs in. This results in two Supervisors running on devices under the BM instance which persists after BM upgrade.

See: https://balena.fibery.io/search/T7ozi#Inputs/Pattern/Two-supervisors-are-running-on-device-3370
Change-type: patch


pipex (approved):
Nice work

PR #2388, opened 2024-11-14
Use updateLock.takeLock for all locking actions

cywang117:
Take all locks through updateLock.takeLock

updateLock.lock disposes all locks as a final step regardless of
locking success. It also tries to take the lock for every service
under an appId, even if Supervisor locks are already taken for that
service from another context (such as from updateLock.takeLock
which is a state engine step). This latter case can result in
UpdatesLockedError where both the lock already present and the
locking attempt originate from the Supervisor, which is undesireable
as this will lead to updateLock.lock disposing all locks

The solution is this is for any lock-taking mechanism to go through
updateLock.takeLock, which only takes Supervisor locks where not already
taken by the Supervisor. This commit splits out the lock-taking functionality
into a separate function, updateLock.takeLockFor, which takes the locks and
returns the process lock release function, allowing the callee to control
when process locks should be released. This is necessary as updateLock.lock
can call takeLockFor for multiple appIds, and process locks should only
be released after all resources in scope are locked.

Also create a test case for this behavior

Change-type: patch

### DIFF
--- a/src/lib/update-lock.ts
+++ b/src/lib/update-lock.ts
@@ -81,29 +81,28 @@ async function dispose(
 }
 
 /**
- * Composition step used by Supervisor compose module.
- * Take all locks for an appId | appUuid, creating directories if they don't exist.
+ * Take process lock for an appId, then take all service locks of an appId.
+ * Override existing service locks if lockOverride is true.
+ * Return a list of services that were locked and a function that releases the process lock.
+ * Meant to only be used internally as it does not release the process lock if
+ * the lock succeeds.
  */
-export async function takeLock(
+async function takeLockFor(
 	appId: number,
 	services: string[],
 	force: boolean = false,
-) {
-	logger.logSystemEvent(logTypes.takeLock, {
-		appId,
-		services,
-		force,
-	});
-
+): Promise<{ actuallyLocked: string[]; release: () => void }> {
 	const release = await takeGlobalLockRW(appId);
 
-	let lockOverride: boolean;
-	try {
-		lockOverride = await config.get('lockOverride');
-	} catch (err: any) {
-		throw new InternalInconsistencyError(
-			`Error getting lockOverride config value: ${err?.message ?? err}`,
-		);
+	let lockOverride = force;
+	if (!lockOverride) {
+		try {
+			lockOverride = await config.get('lockOverride');
+		} catch (err: any) {
+			throw new InternalInconsistencyError(
+				`Error getting lockOverride config value: ${err?.message ?? err}`,
+			);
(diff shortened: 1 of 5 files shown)


src/lib/update-lock.ts
```diff
@@ -81,29 +81,28 @@ async function dispose(
...
-	try {
-		lockOverride = await config.get('lockOverride');
-	} catch (err: any) {
-		throw new InternalInconsistencyError(
-			`Error getting lockOverride config value: ${err?.message ?? err}`,
-		);
+	let lockOverride = force;
+	if (!lockOverride) {
+		try {
+			lockOverride = await config.get('lockOverride');
```
pipex, line 100: I know this was there before. But when does `config.get` throw? I feel like we don't take this precaution elsewhere

pipex: We can leave this as is for now

test/integration/device-state.spec.ts
```diff
@@ -616,19 +627,183 @@ describe('device-state', () => {
...
+			};
+
+			// Take lock and execute the function, which will keep the process lock
+			// until the function is resolved
+			const lockPromise = updateLock.lock(123, { force: false }, waitToResolve);
+			await setTimeout(500);
+			expect(inspect(lockPromise)).to.include('pending');
+
+			// While the Supervisor lock is taken in another context, reboot & shutdown
+			// should wait until locks are freed, so the shutdown promise is pending
```
pipex, line 781: This is not correct, if the supervisor already has the lock that means the reboot/shutdown can happen without waiting. If we wait to release the lock, that means the lock can be taken by the user app before we have a chance to reboot.

However, it is true that each `lock(appId)` call should be executed entirely before another can be processed, which is why the tests passes

test/integration/lib/update-lock.spec.ts
```diff
@@ -293,6 +296,27 @@ describe('lib/update-lock', () => {
...
+			// taken should not be taken again by the Supervisor.
+			await updateLock
+				.lock(testAppId, { force: false }, () => Promise.resolve())
+				.catch(async (err) => {
+					expect(
+						await updateLock.getLocksTaken(),
+						'Locks taken in initially should still be held',
+					).to.have.length(1);
+					expect.fail(`update lock should not throw error but got: ${err}`);
+				});
```
pipex, line 315: I think would be enough

```suggestion
			await expect(updateLock
				.lock(testAppId, { force: false }, () => Promise.resolve())).to.not.be.rejected;
```

pipex:
Superseded by #2389

PR #2389, opened 2024-11-19
Refactor update-locks implementation

pipex:
The refactor simplifies the implementation and ensures that locks per app can only be held by one supervisor task at the time.

Change-type: patch


src/lib/update-lock.ts
cywang117, line 1: The `lockApp` function is quite long, possible to split for readability?

test/integration/lib/update-lock.spec.ts
cywang117, line 1: Missing test case (though might be unnecessary): What happens if you call `lockable.lock` multiple times?

cywang117, line 1: Another useful test would be how the system recovers from partial lock failure, i.e. some locks were taken but others failed for whatever reason

src/compose/app.ts
```diff
@@ -200,19 +200,14 @@ class AppImpl implements App {
...
+			// If the app still has a lock, release it
+			if (state.lock != null) {
 				// Release locks for all services before settling state
 				steps.push(
 					generateStep('releaseLock', {
 						appId: target.appId,
+						lock: state.lock,
 					}),
 				);
 			}
```
cywang117, line 213: What happens in a case where the Supervisor dies suddenly when there are no steps expected in the state funnel? If there are no state steps, no `takeLock` steps are ever generated as `appsToLock` is empty for the current appId. So, `state.lock` would be null and no `releaseLock` step would ever be generated. Is this a possible condition on a device?

cywang117 (commented):
Overall much cleaner interface, nice work!

test/integration/lib/update-lock.spec.ts
pipex: This test sort of does that, is that what you mean?

https://github.com/balena-os/balena-supervisor/blob/797e4e1c8e9d70833a9d9405bf8201c34d9b92c9/test/integration/lib/update-lock.spec.ts#L137-L156

pipex: Yes, I'll add it

src/compose/app.ts
pipex: You are right, the `releaseLock` step needs to run every time the state settles and remove any supervisor owned locks even if there is a lock being held. I'll make that change

cywang117 (approved):
Great work!

PR #2393, opened 2024-12-05
Wait for service dependencies to be running

pipex:
This fixes a regression where dependencies would only be started in order and would start the dependent service if its dependency had been started at some point in the past, regardless of the running condition.

This makes the behavior more consistent with docker compose where the [dependency needs to be
running or healthy](https://github.com/docker/compose/blob/69a83d1303a103d82b05d512baf273244b4dbd94/pkg/compose/convergence.go#L441) for the service to be started.

Change-type: patch

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -813,24 +813,28 @@ class AppImpl implements App {
 					serviceName: target.serviceName,
 				}),
 			];
-		} else if (
-			target != null &&
-			this.dependenciesMetForServiceStart(
-				target,
-				targetApp,
-				availableImages,
-				networkPairs,
-				volumePairs,
-				servicePairs,
-			)
-		) {
-			if (!servicesLocked) {
-				this.services
-					.concat(targetApp.services)
-					.forEach((svc) => appsToLock[target.appId].add(svc.serviceName));
-				return [];
+		} else if (target != null) {
+			if (
+				this.dependenciesMetForServiceStart(
+					target,
+					targetApp,
+					availableImages,
+					networkPairs,
+					volumePairs,
+					servicePairs,
+				)
+			) {
+				if (!servicesLocked) {
+					this.services
+						.concat(targetApp.services)
+						.forEach((svc) => appsToLock[target.appId].add(svc.serviceName));
+					return [];
+				}
+				return [generateStep('start', { target })];
+			} else {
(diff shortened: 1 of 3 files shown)


src/compose/app.ts
```diff
@@ -881,7 +885,7 @@ class AppImpl implements App {
 		// different to a dependency which is in the servicePairs below, as these
 		// are services which are changing). We could have a dependency which is
 		// starting up, but is not yet running.
-		const depInstallingButNotRunning = _.some(targetApp.services, (svc) => {
+		const depInstallingButNotRunning = _.some(this.services, (svc) => {
```
pipex, line 888: Bug fix. Services in the target app are generally configured a `running: true` so dependencies would get started without waiting for the status of the actual container

src/compose/app.ts
```diff
@@ -813,24 +813,28 @@ class AppImpl implements App {
...
+			) {
+				if (!servicesLocked) {
+					this.services
+						.concat(targetApp.services)
+						.forEach((svc) => appsToLock[target.appId].add(svc.serviceName));
+					return [];
+				}
+				return [generateStep('start', { target })];
+			} else {
+				// Wait for dependencies to be started
```
pipex, line 835: We need this `noop` here to avoid settling the state too early

If the dependency requires a reboot, this will prevent the release of locks :(

test/unit/compose/app.spec.ts
```diff
@@ -1294,29 +1293,31 @@ describe('compose/app', () => {
 				.to.deep.include({ serviceName: 'main' });
 		});
 
-		it('should not try to start a container which has exited and has restart policy of no', async () => {
+		it('should not try to start a container which has exited', async () => {
```
cywang117, line 1296: Maybe I'm misunderstanding this test change, but wouldn't we want to restart a container that has exited with an error code if the restart policy is `always` or `on-failure`? Doesn't this test case go against that?

pipex: The supervisor never restarts a container that has already been started before. The restart policy is handled by the engine. The only exception is for the engine/hostOS race bug https://github.com/balena-os/balena-supervisor/blob/81b307510d758c9390836acd2d8547e1263eaf35/src/compose/app.ts#L747

PR #2396, opened 2025-01-07
Switch to image pull if delta failure

cywang117:
- Revert to regular pull immediately on HTTP 4xx delta server response when initially requesting the delta resource.
- Retry delta pull DELTA_APPLY_RETRY_COUNT (3) times when there's a delta failure during delta apply, such as in the case of https://github.com/balena-os/balena-engine/issues/244. After retry count, fall back to regular pull. Delta apply occurs with image pull, after successfully requesting the delta resource from the delta server. Errors with delta apply occur before or after image pull, see cases which raise the relevant Engine error [here](https://github.com/balena-os/balena-engine/blob/master/distribution/pull_v2.go#L43)

Change-type: patch

### DIFF
--- a/src/lib/docker-utils.ts
+++ b/src/lib/docker-utils.ts
@@ -1,23 +1,23 @@
-import type { ProgressCallback } from 'docker-progress';
 import { DockerProgress } from 'docker-progress';
+import type { ProgressCallback } from 'docker-progress';
 import Dockerode from 'dockerode';
 import _ from 'lodash';
 import memoizee from 'memoizee';
-
 import { applyDelta, OutOfSyncError } from 'docker-delta';
 
-import type { SchemaReturn } from '../config/schema-type';
+import log from './supervisor-console';
 import { envArrayToObject } from './conversions';
+import * as request from './request';
 import {
 	DeltaStillProcessingError,
 	ImageAuthenticationError,
 	InvalidNetGatewayError,
 	DeltaServerError,
+	DeltaApplyError,
+	isStatusError,
 } from './errors';
-import * as request from './request';
 import type { EnvVarObject } from '../types';
-
-import log from './supervisor-console';
+import type { SchemaReturn } from '../config/schema-type';
 
 export type FetchOptions = SchemaReturn<'fetchOptions'>;
 export type DeltaFetchOptions = FetchOptions & {
@@ -41,6 +41,18 @@ type ImageNameParts = {
 // How long do we keep a delta token before invalidating it
 // (10 mins)
 const DELTA_TOKEN_TIMEOUT = 10 * 60 * 1000;
+
+// How many times to retry a v3 delta apply before falling back to a regular pull.
+// A delta is applied to the base image when pulling, so a failure could be due to
+// "layers from manifest don't match image configuration", which can occur before
+// or after downloading delta image layers.
+//
+// Other causes of failure have not been documented as clearly as "layers from manifest"
+// but could manifest as well, though unclear if they occur before, after, or during
+// downloading delta image layers.
+//
+// See: https://github.com/balena-os/balena-engine/blob/master/distribution/pull_v2.go#L43
+const DELTA_APPLY_RETRY_COUNT = 3;
 
 export const docker = new Dockerode();
 export const dockerProgress = new DockerProgress({
@@ -140,7 +152,7 @@ export async function fetchDeltaWithProgress(
 	}
 
 	// Since the supevisor never calls this function with a source anymore,
-	// this should never happen, but w ehandle it anyway
+	// this should never happen, but we handle it anyway
 	if (deltaOpts.deltaSource == null) {
 		logFn('Falling back to regular pull due to lack of a delta source');
 		return fetchImageWithProgress(imgDest, deltaOpts, onProgress);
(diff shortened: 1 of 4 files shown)


src/lib/docker-utils.ts
```diff
@@ -225,24 +238,53 @@ export async function fetchDeltaWithProgress(
...
+							lastError = e;
+							logFn(
+								`Delta apply failed, retrying (${tryCount}/${DELTA_APPLY_RETRY_COUNT})...`,
+							);
+						}
+					}
+					if (lastError) {
+						const { statusCode, statusMessage } = lastError as StatusError;
+						throw new DeltaApplyError(statusCode ?? 404, statusMessage);
+					}
```
pipex, line 259: Is better to be strict with typings. Also `for` is preferable here
```suggestion
                                        // Try to apply delta DELTA_APPLY_RETRY_COUNT times, then throw DeltaApplyError
					let lastError: StatusError | undefined = undefined;
					for (
						let tryCount = 0;
						tryCount < DELTA_APPLY_RETRY_COUNT;
						tryCount++
					) {
						try {
							id = await applyBalenaDelta(name, token, onProgress, logFn);
							break;
						} catch (e) {
							if (!isStatusError(e)) {
								throw e;
							}
							lastError = e;
							logFn(
								`Delta apply failed, retrying (${tryCount + 1}/${DELTA_APPLY_RETRY_COUNT})...`,
							);
						}
					}
					if (lastError) {
						const { statusCode, statusMessage } = lastError;
						throw new DeltaApplyError(statusCode, statusMessage);
					}
```

src/lib/docker-utils.ts
```diff
@@ -225,24 +238,53 @@ export async function fetchDeltaWithProgress(
...
 				throw new Error(`Unsupported delta version: ${deltaOpts.deltaVersion}`);
 		}
 	} catch (e) {
+		// Log appropriate message based on error type
 		if (e instanceof OutOfSyncError) {
 			logFn('Falling back to regular pull due to delta out of sync error');
-			return await fetchImageWithProgress(imgDest, deltaOpts, onProgress);
+		} else if (e instanceof DeltaServerError) {
+			logFn(
+				`Falling back to regular pull due to delta server error (HTTP code ${e.statusCode})${e.statusMessage ? `: ${e.statusMessage}` : ''}`,
```
pipex, line 271:
```suggestion
				`Falling back to regular pull due to delta server error (${e.statusCode})${e.statusMessage ? `: ${e.statusMessage}` : ''}`,
```

cywang117:
Changed made @pipex , thank you!

pipex (approved):

src/lib/docker-utils.ts
```diff
@@ -225,24 +239,58 @@ export async function fetchDeltaWithProgress(
 							`Got an error when parsing delta server response for v3 delta: ${e}`,
 						);
 					}
-					id = await applyBalenaDelta(name, token, onProgress, logFn);
+					// Try to apply delta DELTA_APPLY_RETRY_COUNT times, then throw DeltaApplyError
+					let lastError: StatusError | undefined = undefined;
+					for (
```
pipex, line 244: Actually, I thought about it a bit more. There is already retry behavior builtin on the state engine. This is adding another retry loop on top of the existing one. The counter should live somewhere in memory and be increased when a new pull is retried and revert back to regular pull when the count has reached `DELTA_APPLY_RETRY_COUNT`

That code can probably live along the `imageFetchFailures` counter we already have here https://github.com/balena-os/balena-supervisor/blob/dad6eeea73d03592eb4a665379bed6b1335b8df5/src/compose/images.ts#L172

pipex (changes requested):
I had approved but I had some additional thoughts this morning. See https://github.com/balena-os/balena-supervisor/pull/2396/files#r1914822271

src/lib/docker-utils.ts
```diff
@@ -225,24 +239,58 @@ export async function fetchDeltaWithProgress(
...
+						tryCount++
+					) {
+						try {
+							id = await applyBalenaDelta(name, token, onProgress, logFn);
+							break;
+						} catch (e) {
+							if (!isStatusError(e)) {
+								throw e;
+							}
+							lastError = e;
```
pipex, line 256: This will also revert to regular pulls if the delta pull fails for any reason, including a network timeout. Do we know if there is something particular in the status returned by the engine when the 'layers from manifest...` error happens?

cywang117: Short of running a regex match on the error message, I don't see how we would differentiate a `layers from manifest` error from a network issue (`context exceeded` / `Client.Timeout` / etc), as neither raise an HTTP status code. See: https://balena.fibery.io/Inputs/Pattern/Image-download-fails-with-layers-from-manifest-don't-match-image-configuration-3313

cywang117: > We might need to take a look at the engine code to check for a list of possible HTTP responses when a pull happens

I've taken a look already when making this PR, the Engine doesn't list any specifics when raising an `errRootFSMismatch` / `layers from manifest` error. The best way I can think of is to reproduce this specific issue and observe the HTTP codes, or do so next time a user is facing this issue, however we don't know how to reproduce. 

My assumption that an `errRootFSMismatch` would raise a 4xx error whereas a network issue should be propagated by the Engine as a 5xx error, but we'd be guessing, even with the link to the docs which doesn't cover this case specifically -- deltas are only part of our Engine fork, after all.

pipex: This is a similar error that happens on delta pull and it can be replicated https://balena.fibery.io/favorites/Inputs/Pattern/2172

cywang117: Thanks @pipex . I used this example to make the Supervisor log a more descriptive error. There isn't an HTTP status, the error is thrown from `applyBalenaDelta`, which is a function in docker-progress. Looks like a registry stream error should get propagated from docker-progress [here](https://github.com/balena-io-modules/docker-progress/blob/master/lib/index.ts#L93), but it wraps the error from the registry stream in a generic error before propagating the error, which I believe swallows any other props the error might have had. For example:

```
export class StatusCodeError extends TypedError {
    public statusCode: number;
    constructor(message: string, statusCode: number) {
        super(message);
        this.name = "MyError";
        this.statusCode = statusCode;
    }
}

const error = new StatusCodeError("Not found", 404);
const wrapped = new Error(error as any);

console.log({
    message: wrapped.message,
    statusCode: (wrapped as any).statusCode, // statusCode was not added to the error
    stack: wrapped.stack
});
```

I'm going to PR docker-progress once i get access to propagate the error directly if it's an instanceof NodeJS Error. Maybe that'll reveal more info, maybe that won't. Will report back

cywang117: See: https://github.com/balena-io-modules/docker-progress/pull/69 

When testing, ultimately this didn't make a difference. It seems that delta apply errors don't have a status code at all.

That in mind, I modified the PR to not fall back to regular pull on network errors. Network errors can be distinguished during the delta pulling process because they will have a `500` status error. I tested a variety of scenarios:

- Network timeout
- Auth token not being valid anymore during pull
- HTTP/HTTPs mismatch when pulling

All of these resulted in `500 server error`.

src/lib/docker-utils.ts
pipex: Since this re-try should not pull anything again but try to apply the already pulled delta, then I think it's OK to leave the retry behavior here.

pipex (approved):
Nice work!

PR #2404, opened 2025-02-26
Polling improvements

pipex:
This PR fixes multiple issues with the target state poll.

- Fixes the polling healthcheck by resetting the polling timer only on a successful poll
- It logs errors that happen during connection establishment
- Temporarily switch to using the default timeout of 59s for the GOT instance when performing the poll

Change-type: patch

### DIFF
--- a/src/api-binder/poll.ts
+++ b/src/api-binder/poll.ts
@@ -87,7 +87,8 @@ const emitTargetState = (
  * We set a value rather then being undeclared because having it undefined
  * adds more overhead to dealing with this value without any benefits.
  */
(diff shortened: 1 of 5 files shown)


src/api-binder/poll.ts
```diff
@@ -188,7 +186,11 @@ const poll = async (
 		await update();
 		// Reset fetchErrors because we successfuly updated
 		fetchErrors = 0;
-	} catch {
+		lastFetch = process.hrtime();
```
Page-, line 189: It's probably worth renaming to
```suggestion
		lastSuccessfulFetch = process.hrtime();
```
to make the semantics clear

Page- (approved):

PR #2405, opened 2025-02-27
Decrease balenaCloud api request timeout from 15m to 59s

Page-:
This was mistakenly increased due to confusion between the timeout for requests to the supervisor's api vs the timeout for requests from the supervisor to the balenaCloud api. This separates the two configs and documents the difference between the timeouts whilst also decreasing the timeout for balenaCloud api requests to the correct/expected value

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


src/config/schema-type.ts
```diff
@@ -118,6 +121,13 @@ export const schemaTypes = {
 		type: PermissiveBoolean,
 		default: false,
 	},
+	/**
+	 * The timeout for requests to the balenaCloud api
+	 */
+	apiRequestTimeout: {
+		type: PermissiveNumber,
+		default: 59000,
+	},
```
pipex, line 130: You added the configuration here, but not on `config/schema.ts` which means the value is not really configurable. Given that it should perhaps be set on `lib/constants.ts` which is a simpler change?

pipex: Alternatively we can make it configurable, perhaps via config.json or a config variable

Page-: Yeah I'm thinking a config var could be good, not that we intend to use it but I've never been upset in the past when the API is able to modify supervisor behavior because it gives us more options for dealing with older supervisors. Ie being able to 1. fix in latest supervisor, 2. "backport"/workaround via api -> supervisor config vars has always been useful

Page-: Added

pipex:
@Page- can you add a line in the test linked below?

```ts
SUPERVISOR_API_REQUEST_TIMEOUT: '59000'
``` 

https://github.com/balena-os/balena-supervisor/blob/8c8b730a003118097f0cf2090a4c8d38d58ba74e/test/integration/device-config.spec.ts#L82

pipex (approved):

PR #2407, opened 2025-03-11
Reduce happy-eyeballs attempt timeout to 4.5s

pipex:
We have observed the previous 5s timeout clashing with the default socket timeout (also 5s) resulting in the socket being closed before even one attempt was finished. On testing any value less than 5s worked but we are reducing to 4.5s that should still be more than enough for most networks (the Node default is 250ms).

Change-type: patch

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -3,8 +3,8 @@ import { set } from '@balena/es-version';
 set('es2022');
 
 import { setDefaultAutoSelectFamilyAttemptTimeout } from 'net';
-// Increase the timeout for the happy eyeballs algorithm to 5000ms to avoid issues on slower networks
-setDefaultAutoSelectFamilyAttemptTimeout(5000);
+// Increase the timeout for the happy eyeballs algorithm to 4500ms to avoid issues on slower networks
+setDefaultAutoSelectFamilyAttemptTimeout(4500);
 
 // Setup MDNS resolution
 import './mdns';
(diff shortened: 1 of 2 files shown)


Page- (commented):
I'd prefer to keep it closer to 5s just to minimize issues on weird/slow networks since we are more interested in a working response rather than minimizing latency, however I am happy to approve if you disagree

pipex:
We can do 4.5s just to get some distance from the 5s cutoff?

src/app.ts
```diff
@@ -3,8 +3,8 @@ import { set } from '@balena/es-version';
 set('es2022');
 
 import { setDefaultAutoSelectFamilyAttemptTimeout } from 'net';
-// Increase the timeout for the happy eyeballs algorithm to 5000ms to avoid issues on slower networks
-setDefaultAutoSelectFamilyAttemptTimeout(5000);
+// Increase the timeout for the happy eyeballs algorithm to 3000ms to avoid issues on slower networks
+setDefaultAutoSelectFamilyAttemptTimeout(3000);
```
Page-, line 7: @pipex you seem to have updated the PR title to 4.5s but this is still 3s

pipex: Oops, forgot to push

pipex:
Superseded by https://github.com/balena-os/balena-supervisor/pull/2408

PR #2415, opened 2025-04-01
Fix search for app leftover locks

pipex:
The leftover locks search was creating an array rather than an object keyed by the appId. This could affect the lock cleanup and make leftover locks from one app affect the install of the app in local mode.

Change-type: patch


pipex:
Context https://balena.zendesk.com/agent/tickets/4848

cywang117 (approved):

PR #2416, opened 2025-05-08
Simplify contract validation module

pipex:
Use `satisfiesChildContract` instead of Blueprints as the previous implementation did.

Change-type: patch


pipex:
This should make it simpler to add new contract requirements in the future

cywang117 (approved):
Nice work, LGTM!

PR #2417, opened 2025-05-14
Support target state apply cancellation

cywang117:
## Description

Support target state apply cancellation
    
The current target state apply is cancelled when either:
    - /v1/update is called with cancel: true
    - A different target state is received from the cloud (with a non-304 status)
    
Following apply cancellation, a target state apply is re-triggered. This ensures
that the user can force a device out of a dead-locked situation where a long-running
task such as an image fetch fails to cede control back to the Supervisor, which is
the behavior observed in an Engine bug with infinite pull retries with a bad network. This
bug has been observed to affect users on support.

This Supervisor change is a way to work around the above described Engine bug, but a proper fix
should come from the Engine.

## Type of change

Minor - but should it be major?

## How Has This Been Tested?

Tested on-device by changing a config var during an image pull, as well as calling Supervisor API's POST /v1/update with cancel: true. Both resulted in pull cancellation and retry.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -94,10 +94,11 @@ Triggers a check for the target state of configurations and app services. Option
 Responds with an empty 204 (No Content) response.
 
 #### Request body
-Can be a JSON object with a `force` property. If this property is true, the update lock will be overridden.
+Can be a JSON object with `force` and `cancel` properties. If `force` is true, update locks will be overridden by the Supervisor. If `cancel` is true, the Supervisor will abort its current state apply operation to apply the most up-to-date target state. This is useful for interrupting a long operation (e.g. image pull).
 ```json
 {
-	"force": true
+	"force": true,
+	"cancel": true
 }
 ```
 
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "copy-webpack-plugin": "^12.0.0",
     "deep-object-diff": "1.1.0",
     "docker-delta": "^4.1.0",
-    "docker-progress": "^5.2.4",
+    "docker-progress": "^5.3.1",
     "dockerode": "^4.0.2",
     "duration-js": "^4.0.0",
     "express": "^4.21.2",
--- a/src/api-binder/poll.ts
+++ b/src/api-binder/poll.ts
@@ -20,8 +20,13 @@ interface TargetStateEvents {
 		targetState: TargetState,
 		force: boolean,
 		isFromApi: boolean,
+		cancel: boolean,
 	) => void;
-	'target-state-apply': (force: boolean, isFromApi: boolean) => void;
+	'target-state-apply': (
+		force: boolean,
+		isFromApi: boolean,
+		cancel: boolean,
+	) => void;
 }
 export const emitter: StrictEventEmitter<EventEmitter, TargetStateEvents> =
 	new EventEmitter();
@@ -50,12 +55,14 @@ let appUpdatePollInterval: number;
  * @param cachedResponse the response to emit
  * @param force Emitted with the 'target-state-update' event update as necessary
  * @param isFromApi Emitted with the 'target-state-update' event update as necessary
+ * @param cancel Whether to cancel the current apply operation
  * @return true if the response has been emitted or false otherwise
  */
 const emitTargetState = (
 	cachedResponse: CachedResponse,
 	force = false,
 	isFromApi = false,
+	cancel = false,
 ): boolean => {
 	if (
 		!cachedResponse.emitted &&
@@ -67,6 +74,7 @@ const emitTargetState = (
 			structuredClone(cachedResponse.body),
 			force,
 			isFromApi,
+			cancel,
 		);
 		return true;
 	} else if (
@@ -75,7 +83,7 @@ const emitTargetState = (
 		emitter.listenerCount('target-state-apply') > 0
 	) {
 		// CachedResponse has been emitted but a client triggered the check so emit an apply
-		emitter.emit('target-state-apply', force, isFromApi);
+		emitter.emit('target-state-apply', force, isFromApi, cancel);
 		return true;
 	}
 	// Return the same emitted value but normalized to be a boolean
@@ -95,11 +103,13 @@ export let lastSuccessfulFetch: ReturnType<typeof process.hrtime> =
  * Attempts to update the target state
  * @param force Emitted with the 'target-state-update' event update as necessary
  * @param isFromApi Emitted with the 'target-state-update' event update as necessary
+ * @param cancel Whether to cancel the current apply operation
  */
 export const update = async (
 	// TODO: Is there a better way than passing these params here just to emit them if there is an update?
 	force = false,
 	isFromApi = false,
+	cancel = false,
 ): Promise<void> => {
 	await config.initialized();
 	return Bluebird.using(lockGetTarget(), async () => {
@@ -147,7 +157,7 @@ export const update = async (
 		if (statusCode === 304 && cache?.etag != null) {
 			// There's no change so no need to update the cache
 			// only emit the target state if it hasn't been emitted yet
-			cache.emitted = emitTargetState(cache, force, isFromApi);
+			cache.emitted = emitTargetState(cache, force, isFromApi, cancel);
 			return;
 		}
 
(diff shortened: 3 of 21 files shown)


src/api-binder/poll.ts
```diff
@@ -161,8 +171,9 @@ export const update = async (
 			body: body as any,
 		};
 
-		// Emit the target state and update the cache
-		cache.emitted = emitTargetState(cache, force, isFromApi);
+		// Emit the target state and update the cache, cancelling any
+		// apply operations that may be in progress
+		cache.emitted = emitTargetState(cache, force, isFromApi, true);
```
cywang117, line 176: I'm worried about a potential worsening of UX if the user sets a service var or config var during an image fetch. With this change as it stands right now, a variable change would equal a target state change, which would cause any current image fetches to be aborted.

src/device-state/index.ts
```diff
@@ -790,9 +815,20 @@ export function triggerApplyTarget({
 			}
 			lastApplyStart = process.hrtime();
 			log.info('Applying target state');
-			return applyTarget({ force, initial });
+			if (!abortController) {
```
pipex, line 827: Can we make it so abortController is never null? I don't like these Inconsistency errors if we can avoid them

src/compose/images.ts
```diff
@@ -196,6 +196,14 @@ export async function triggerFetch(
 		});
 	};
 
+	// Remove image task on fetch abort to prevent noop loop
```
pipex, line 199: I don't understand the comment. Why would this cause a noop loop?

cywang117: If the AbortSignal is triggered, ImageTask's `onFinish` wasn't being called, so the state engine assumed that the ImageTask was still running, and would generate `noop` steps while waiting for the image to finish downloading. In reality though, the image was no longer being downloaded, so the state engine would wait forever.

src/device-state/index.ts
cywang117: Is this still the case with https://github.com/balena-io-modules/docker-progress/pull/76?

pipex: I mean making sure the variable in the supervisor is never null, i.e. assigning a value to it when the variable is created, so we don't have to keep checking it

src/api-binder/poll.ts
pipex: I think there is benefit of keeping the target state as a single entity as there may be interactions between config vars and updates (e.g. deltas). But I see your point and we can perhaps monitor any increase issues on support or other channels.

package.json
```diff
@@ -84,7 +84,7 @@
     "copy-webpack-plugin": "^12.0.0",
     "deep-object-diff": "1.1.0",
     "docker-delta": "^4.1.0",
-    "docker-progress": "^5.2.4",
+    "docker-progress": "^5.3.0",
```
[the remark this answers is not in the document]
cywang117: Done! https://github.com/balena-os/balena-supervisor/pull/2417/commits/4558b5a20b3f73f8185bfd364995485915bf7320

docs/API.md
```diff
@@ -94,10 +94,11 @@ Triggers a check for the target state of configurations and app services. Option
 Responds with an empty 204 (No Content) response.
 
 #### Request body
-Can be a JSON object with a `force` property. If this property is true, the update lock will be overridden.
+Can be a JSON object with `force` and `cancel` properties. If `force` is true, update locks will be overridden by the Supervisor. If `cancel` is true, the Supervisor will abort its current state apply operation to apply the most up-to-date target state. This is useful for bailing out of an image pull that's hanging in order to retry it, for example.
```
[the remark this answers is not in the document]
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2417/commits/1eb191ab126b11d3d78ba473af8cc4dcd56a569e#diff-b57590968a12cee85a37c1b91d8cc7092cd8b68e50b242c39a178121db82a797R97)

pipex (commented):
The code looks good to me, have you been able to observe the cancellation during your tests?

docs/API.md
```diff
@@ -94,10 +94,11 @@ Triggers a check for the target state of configurations and app services. Option
 Responds with an empty 204 (No Content) response.
 
 #### Request body
-Can be a JSON object with a `force` property. If this property is true, the update lock will be overridden.
+Can be a JSON object with `force` and `cancel` properties. If `force` is true, update locks will be overridden by the Supervisor. If `cancel` is true, the Supervisor will abort its current state apply operation to apply the most up-to-date target state. This is useful for interrupting a long operation (e.g. image pull), for example.
```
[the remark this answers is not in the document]
cywang117: You're right, removed the `for example` 👍🏼

cywang117:
@pipex RE: manual tests, yes, I observed the cancellation taking effect when pinning back & forth between two releases with large images. I also observed the same behavior when calling /v1/update with `{"cancel": true}`. Debug logs look like the following:
```
May 27 21:52:42 nuc balena-supervisor[2386898]: [event]   Event: Update notification {}
May 27 21:52:42 nuc balena-supervisor[2386898]: [api]     POST /v1/update 204 - 3.640 ms
May 27 21:52:42 nuc balena-supervisor[2386898]: [info]    Applying target state
May 27 21:52:43 nuc balena-supervisor[2386898]: [event]   Event: Docker image download {"image":{"name":"registry2.balena-cloud.com/v2/82fa968149db9e7036f731287125c30f@sha256:11691e035a3651d25a87116b4f6adc113a27a29d8f5a6a583f8569e0ee5ff897","appId":2239883,"appUuid":"254f9908bb3c48a494a3ea1f9bf35f0f","serviceId":3182081,"serviceName":"pytorch","imageId":11331794,"releaseId":3457735,"commit":"297ae7415f65cb932f5a8976fdbdfd9d"}}
May 27 21:52:44 nuc balena-supervisor[2386898]: [info]    Reported current state to the cloud
May 27 21:52:54 nuc balena-supervisor[2386898]: [info]    Reported current state to the cloud
May 27 21:53:05 nuc balena-supervisor[2386898]: [info]    Reported current state to the cloud
May 27 21:53:15 nuc balena-supervisor[2386898]: [info]    Reported current state to the cloud
May 27 21:53:23 nuc balena-supervisor[2386898]: [event]   Event: Update notification {}
May 27 21:53:23 nuc balena-supervisor[2386898]: [api]     POST /v1/update 204 - 3.126 ms
May 27 21:53:24 nuc balena-supervisor[2386898]: [debug]   Target state changed, aborting current apply
May 27 21:53:24 nuc balena-supervisor[2386898]: [debug]   Triggering re-apply after aborting current apply
May 27 21:53:24 nuc balena-supervisor[2386898]: [info]    Skipping applyTarget because of a cancellation
May 27 21:53:24 nuc balena-supervisor[2386898]: [info]    Applying target state
May 27 21:53:24 nuc balena-supervisor[2386898]: [debug]   Finished applying target state
```

Also confirmed that pinning back & forth without this change results in the state engine waiting until the current image download completes before applying the newest target state.

I noticed that the debug message between a target state change causing an abort vs /v1/update `"cancel": true` causing an abort is the same, which is confusing, so I clarified the console debug messages to `Aborting target state apply` before abort, and `Triggering re-apply after aborting target state apply` after abort and before `triggerApplyTarget` in cfcae7785927eeb7a70b16093d9396a2b3fca5e9. See [here](https://github.com/balena-os/balena-supervisor/commit/cfcae7785927eeb7a70b16093d9396a2b3fca5e9#diff-eb0ba0c7dd05e2a409c254c3d47d96333fd3d3266a9d30103fb37315d2dfc299R773)

pipex:
I don't understand, the [Aborting state apply message](https://github.com/balena-os/balena-supervisor/commit/cfcae7785927eeb7a70b16093d9396a2b3fca5e9#diff-eb0ba0c7dd05e2a409c254c3d47d96333fd3d3266a9d30103fb37315d2dfc299R773) happens no matter if the target state changed or the `/v1/update` endpoint was called with `cancel` no?

Regarding the logs you shared, this is a nit, but I found them a bit redundant no? Is there a way to make them simpler??

```
May 27 21:53:23 nuc balena-supervisor[2386898]: [api]     POST /v1/update 204 - 3.126 ms
May 27 21:53:24 nuc balena-supervisor[2386898]: [debug]   Target state changed, aborting current apply
May 27 21:53:24 nuc balena-supervisor[2386898]: [debug]   Triggering re-apply after aborting current apply
May 27 21:53:24 nuc balena-supervisor[2386898]: [info]    Skipping applyTarget because of a cancellation
```

pipex:
Yeah, let's remove line 3. I would also remove line 4 because it doesn't make any sense, although you are right it is part of a separate mechanism.

PR #2418, opened 2025-05-15
Remove unsupported fields from contract requirements

pipex:
A contract including extra requirement fields, such as "name" would fail validation. This PR removes any extra fields from the validated contract to prevent services with these extra fields from getting rejected by the contract validation.

Change-type: patch

### DIFF
--- a/src/lib/contracts.ts
+++ b/src/lib/contracts.ts
@@ -149,15 +149,20 @@ const ContainerContract = t.intersection([
 			t.null,
 			t.undefined,
 			t.array(
-				t.intersection([
-					t.type({
-						type: t.string,
-					}),
-					t.partial({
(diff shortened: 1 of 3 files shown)


src/lib/contracts.ts
```diff
@@ -149,15 +149,20 @@ const ContainerContract = t.intersection([
...
+				t.exact(
+					t.intersection([
+						t.type({
+							type: t.string,
+						}),
+						// Allow searching the most common contract matchers
+						t.partial({
+							slug: t.union([t.null, t.undefined, t.string]),
+							version: t.union([t.null, t.undefined, t.string]),
+							data: t.record(t.union([t.string, t.number]), t.any),
```
pipex, line 162: Added the `data` field to support future contract requirements

Issue #2367, opened 2024-08-27
Supervisor reverts to the first locally pushed version when restarted in local mode


pipex:
Replication 

- Download https://github.com/balena-io-examples/balena-nodejs-hello-world
- Start a local livepush session balena push <ip address>, wait for the device to settle
- Stop the livepush session, and make a small code change 
- Start a new live push, check that your changes are applied. You'll also see the image id change
- Restart the supervisor, the supervisor will log something like this 
```
Aug 27 18:31:50 68568ed balena-supervisor[10222]: [debug]   Replacing container for service balena-hello-world because of config changes:
Aug 27 18:31:50 68568ed balena-supervisor[10222]: [debug]     Non-array fields:  {"added":{},"deleted":{},"updated":{"image":"sha256:dda4ef12f2dd49f1d7a8c7fa2f2ca35a132a43c65380893737a531a80ea9027b"}}
Aug 27 18:31:50 68568ed balena-supervisor[10222]: [debug]   Replacing container for service balena-hello-world because of config changes:
Aug 27 18:31:50 68568ed balena-supervisor[10222]: [debug]     Non-array fields:  {"added":{},"deleted":{},"updated":{"image":"sha256:dda4ef12f2dd49f1d7a8c7fa2f2ca35a132a43c65380893737a531a80ea9027b"}}
```

The problem is that the supervisor only generates a `saveImage` step if the image tag does not exist in the database, we need to compare that the image exists on the database with the same docker id and tag here https://github.com/balena-os/balena-supervisor/blob/3049b03cba82df70dae7ba0b5fce5786f01939f9/src/compose/application-manager.ts#L640

This does not affect cloud releases because a change always results in a new image tag, while in local releases, the image tag is re-used for every push

pipex:
Fixed in #2543

PR #2423, opened 2025-06-15
Save local image in database on every change

pipex:
The local docker image tag always remains the same, but the image docker id changes on every new push. Before this, the supervisor would revert to the first pushed image.

This is a WIP as it still lacks tests

Change-type: patch
Relates-to: #2367


cwbusacker:
This PR should be able to be closed because #2543 resolves the basis for this change.

pipex:
Fixed on #2543

PR #2424, opened 2025-06-18
Do not block calls to dbus reboot/shutdown

pipex:
This should give the supervisor API time to respond to queries before the device goes down.

Change-type: patch


cywang117 (commented):
This LGTM - still good to merge @pipex ?

pipex:
Here is the context on [Zulip](https://balena.zulipchat.com/#narrow/channel/403752-channel.2Fsupport-help/topic/502.20response.20on.20supervisor.2Fv1.2Freboot.2C.20but.20device.20reboots/near/524727495), let's wait to hear how the user testing goes

PR #2426, opened 2025-06-23
Add support for async OS updates

pipex:
This adds a new service to the supervisor composition. This new service `helios`, proxies communications between this supervisor, the API and app containers.  The new service will little by little take over features of this supervisor until we can drop it entirely. 

The new service is written in Rust for better memory safety and lower resource usage. It will also feature a new planning engine based on [mahler](https://github.com/balena-io-modules/mahler-rs), which removes the need for hard-coding workflows (no more application manager and infer next steps), but can automatically derive a plan from the target and task definitions. The new engine also features better logging and is easier to test. 

## Testing the new service

Run the following commands in your hostOS, replacing the correct architecture in the `supervisor_release` variable

```sh
# Set necessary variables
api_key=$(cat /mnt/boot/config.json | jq -r .deviceApiKey)
uuid=$(cat /mnt/boot/config.json | jq -r .uuid)
api_endpoint=$(cat /mnt/boot/config.json | jq -r .apiEndpoint)

# Supervisor releases
amd64_supervisor=4018627
aarch64_supervisor=4018812
armv7_supervisor=4018628
# Make sure to use the right id for your architecture here
supervisor_release="${aarch64_supervisor}"

supervisor_img=$(\
  curl -X GET \
    "$api_endpoint/v7/release($supervisor_release)?\$expand=contains__image(\$expand=image(\$select=is_stored_at__image_location))" \
    -H  "Content-Type: application/json" \
    -H "Authorization: Bearer $api_key" | jq -r .d[0].contains__image[0].image[0].is_stored_at__image_location \
)

# Patch the device to the draft release
curl -X PATCH -H "Authorization: Bearer $api_key" -H  "Content-Type: application/json" "$api_endpoint/v6/device?\$filter=uuid%20eq%20'$uuid'" -d "{\"should_be_managed_by__supervisor_release\": $supervisor_release}"

# Update the supervisor to the target image
update-balena-supervisor -i "$supervisor_img"
```

Change-type: major

## Release Notes

This release is the first step to migrating the supervisor to a new codebase. It introduces a new service `helios` that acts as a proxy between the existing `balena-supervisor` container, the balena API and other containers. This service will slowly take the supervisor responsibilities until the old service can be replaced entirely. 

The new service is written in Rust for better memory safety and lower resource usage (should use around 5mb of memory). It is just a regular container, and doesn't require any special host privileges or host networking access. To get around the need for special permissions, a second container `helios-api` is included with this release. That service is just a relay of requests to a Unix domain socket where the main service listens for requests.

The new service also provides a long requested feature for the on-device agent service. The ability to perform OS updates without the need of Cloudlink access (async OS updates). Devices can now be pinned to a new target OS release and the agent will receive the new OS release through the target state on the next poll and install the new release as it would do with any other app and wait until locks are cleared in order to boot to the new OS. The service keeps track of failed attempts to abort the process after a certain number of tries.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,8 @@
 version: '2.3'
 
+x-helios-image: &helios-image
+  image: ghcr.io/balena-io/helios:0.25.4
+
 services:
   # This service can only be named `main` or `balena-supervisor` as older
   # devices will keep relying on the `/v6/supervisor_release` endpoint that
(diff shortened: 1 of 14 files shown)


docker-compose.yml
```diff
@@ -14,3 +17,66 @@ services:
...
+    labels:
+      # the database version, can be bumped to re-create the state
+      io.balena.private.db-version: 0
+  cache:
+  runtime:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      # XXX: this directory needs to exist on the host before the container is started
```
kb2ma, line 81: What is the meaning of the "XXX" at the beginning of this comment?

pipex: It's a warning. Depending on the code editor the line will be highlighted

src/lib/firewall.ts
```diff
@@ -242,6 +313,21 @@ export const applyFirewallMode = async function applyFirewallMode(
 								action: iptables.RuleAction.Append,
 								target: 'REJECT',
 							}),
+					)
+					.forChain('DOCKER-USER', (chain) => {
```
kb2ma, line 317: Do we have any concerns about a race condition with user generated rules in this chain? We have seen race conditions in the past with the firewall.

pipex: This doesn't remove any existing rules from the `DOCKER-USER` chain. It's still possible that the rules in the `BALENA-SUPERVISOR` chain conflict with some of the user rules but unfortunately using the `DOCKER-USER` chain is the only option to control access to the API port. 

I would expect that in the future, the supervisor doesn't need to configure the firewall and this is done by the OS. Something like `development mode => ports open` and `production mode => ports closed`

kb2ma:
If the Supervisor uses helios-api for outgoing requests, is it possible for an external actor to impersonate the Supervisor via the User-Agent header, and is that a concern ([code](https://github.com/balena-io/helios/blob/89d30edeea777459e92190fbfa902244e411c835/helios-legacy/src/proxy.rs#L113))?

It looks like the rule below requires that the incoming request is from Cloudlink, correct?
```
Chain  BALENA-SUPERVISOR         (1       references)                                                                                                                                            
num    pkts                      bytes    target                    prot  opt       in                out               source     destination                                                   
1      0                         0        ACCEPT                    tcp   --        resin-vpn         *                 0.0.0.0/0  0.0.0.0/0    tcp          dpt:48484                           
2      0                         0        REJECT                    tcp   --        *                 *                 0.0.0.0/0  0.0.0.0/0    tcp          dpt:48484              reject-with icmp-port-unreachable
```

kb2ma (approved):
Extensive code review, discussion and update testing.

PR #2431, opened 2025-10-01
Update dependencies (2025-10)

cywang117:
19 vulnerabilities -> 4 vulnerabilities

The 4 vulnerabilities are due to our use of `request` NPM package, to patch them we'd have to switch to a maintained request lib such as `got` which we already use for target state.

Bumping @balena/lint resulted in a quite a few linter errors as it appears the latest @balena/lint is more strict. I've fixed these by either ignoring them to keep the existing behavior the same, or modifying the code to satisfy linting rules where the behavior would not have changed with the linting fixes.

Also, a small bugfix with clearing individual memoized values, and a patch for building sqlite3 from source in Docker under newer python3 versions that have removed necessary `distutils` lib.

### DIFF
--- a/src/lib/api-keys.ts
+++ b/src/lib/api-keys.ts
@@ -221,11 +221,11 @@ async function generateKey(
 		// remove the cached lookup for the key
 		const [apiKey] = secrets;
 		if (apiKey != null) {
-			await getApiKeyByKey.delete(apiKey.key);
+			getApiKeyByKey.delete(apiKey.key);
 		}
 
 		// remove the cached value for this lookup
-		await getApiKeyForService.delete(appId, serviceName);
+		getApiKeyForService.delete(appId, serviceName);
 
 		// return a new API key
 		return await createNewKey(appId, serviceName, scopes);
--- a/package-lock.json
+++ b/package-lock.json
@@ -2315,17 +2315,18 @@
       }
     },
     "node_modules/@types/lodash": {
-      "version": "4.17.14",
-      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.17.14.tgz",
-      "integrity": "sha512-jsxagdikDiDBeIRaPYtArcT8my4tN1og7MtMRquFT3XNA6axxyHDRUemqDz/taRDdOUn0GnGHRCuff4q48sW9A==",
+      "version": "4.17.20",
+      "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.17.20.tgz",
+      "integrity": "sha512-H3MHACvFUEiujabxhaI/ImO6gUrd8oOurg7LQtS7mbwIXA/cUqWrvBsaeJ23aZEPk1TAYkurjfMbSELfoCXlGA==",
       "dev": true,
       "license": "MIT"
     },
     "node_modules/@types/memoizee": {
-      "version": "0.4.11",
-      "resolved": "https://registry.npmjs.org/@types/memoizee/-/memoizee-0.4.11.tgz",
-      "integrity": "sha512-2gyorIBZu8GoDr9pYjROkxWWcFtHCquF7TVbN2I+/OvgZhnIGQS0vX5KJz4lXNKb8XOSfxFOSG5OLru1ESqLUg==",
-      "dev": true
+      "version": "0.4.12",
+      "resolved": "https://registry.npmjs.org/@types/memoizee/-/memoizee-0.4.12.tgz",
+      "integrity": "sha512-EdtpwNYNhe3kZ+4TlXj/++pvBoU0KdrAICMzgI7vjWgu9sIvvUhu9XR8Ks4L6Wh3sxpZ22wkZR7yCLAqUjnZuQ==",
+      "dev": true,
+      "license": "MIT"
     },
     "node_modules/@types/methods": {
       "version": "1.1.4",
@@ -2398,15 +2399,16 @@
       "dev": true
     },
     "node_modules/@types/request": {
-      "version": "2.48.12",
-      "resolved": "https://registry.npmjs.org/@types/request/-/request-2.48.12.tgz",
-      "integrity": "sha512-G3sY+NpsA9jnwm0ixhAFQSJ3Q9JkpLZpJbI3GMv0mIAT0y3mRabYeINzal5WOChIiaTEGQYlHOKgkaM9EisWHw==",
+      "version": "2.48.13",
+      "resolved": "https://registry.npmjs.org/@types/request/-/request-2.48.13.tgz",
+      "integrity": "sha512-FGJ6udDNUCjd19pp0Q3iTiDkwhYup7J8hpMW9c4k53NrccQFFWKRho6hvtPPEhnXWKvukfwAlB6DbDz4yhH5Gg==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "@types/caseless": "*",
         "@types/node": "*",
         "@types/tough-cookie": "*",
-        "form-data": "^2.5.0"
+        "form-data": "^2.5.5"
       }
     },
     "node_modules/@types/rewire": {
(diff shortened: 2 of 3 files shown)


src/compose/utils.ts
```diff
@@ -247,6 +247,7 @@ export function getHealthcheck(
 ): ServiceHealthcheck {
 	// get the image info healtcheck
 	const imageServiceHealthcheck = dockerHealthcheckToServiceHealthcheck(
+		// eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
pipex, line 252: Could you add a comment explaining that this is because `Config.Healthcheck` is not defined on the dockerode type, but it may be provided by the engine?. I believe that's why the `_.get` return type is `undefined`

src/compose/volume.ts
```diff
@@ -48,12 +48,12 @@ class VolumeImpl implements Volume {
 		config = {} as Partial<ComposeVolumeConfig>,
 	) {
 		const filledConfig: VolumeConfig = {
-			driverOpts: config.driver_opts || {},
-			driver: config.driver || 'local',
+			driverOpts: config.driver_opts ?? {},
+			driver: config.driver || 'local', // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
```
pipex: Why? Can `config.driver` be `false` or some other falsy value?

src/device-api/v2.ts
```diff
@@ -546,9 +546,9 @@ router.get('/v2/cleanup-volumes', async (req: AuthorizedRequest, res) => {
 router.post('/v2/journal-logs', (req, res) => {
 	const all = checkTruthy(req.body.all);
 	const follow = checkTruthy(req.body.follow);
-	const count = checkInt(req.body.count, { positive: true }) || undefined;
+	const count = checkInt(req.body.count, { positive: true }) ?? undefined;
 	const unit = req.body.unit;
-	const format = req.body.format || 'short';
+	const format = req.body.format || 'short'; // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
```
pipex: Same here, why is null coalescing not right here?

src/device-state/index.ts
```diff
@@ -474,10 +480,12 @@ export async function shutdown({
 			switch (reboot) {
 				case true:
 					logger.logSystemMessage('Rebooting', {}, 'Reboot');
+					// eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
pipex: I think we can remove the assignment here, both dbus function return `void`

src/host-config/proxy.ts
```diff
@@ -37,6 +37,7 @@ const baseBlock = {
 	redirector: 'iptables',
 };
 
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
 export class RedsocksConf {
 	public static stringify(config: RedsocksConfig): string {
```
pipex: I think this can be a `const` object
```suggestion
export const RedsocksConf = {
public static stringify(config: RedsocksConfig): string {
```

src/lib/api-keys.ts
```diff
@@ -66,6 +65,7 @@ export const isScoped = (
 	resources: Partial<ScopedResources>,
 	scopes: Scope[],
 ) =>
+	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	scopes.some((scope) =>
 		scopeChecks[scope.type](resources, scope as unknown as any),
```
pipex: Since scope checks return a `Resolvable` the typecheck seems like it's catching a bug and this should be something like
```suggestion
const mappedScopes = await Promise.all(scopes.map((scope) => scopeChecks[scope.type](resources, scope as unknown as any));
return mappedScopes.some((scope) => scope);
```

But perhaps the `ScopeCheck` type is wrong?

src/lib/docker-utils.ts
```diff
@@ -95,6 +95,7 @@ export function normaliseImageName(image: string) {
 	const repository = [registry, imageName].filter((s) => !!s).join('/');
 
 	if (!digest) {
+		// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
pipex: Why disable this lint? Tag can be nullish right?

src/lib/docker-utils.ts
```diff
@@ -117,6 +118,7 @@ export function getRepoAndTag(image: string): { repo: string; tag?: string } {
 // Same as getRepoAndTag but joined with ':' for searching
 export function getImageWithTag(image: string) {
 	const { repo, tag } = getRepoAndTag(image);
+	// eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
pipex: Why disable this lint?

src/lib/docker-utils.ts
```diff
@@ -318,7 +321,8 @@ export async function fetchDeltaWithProgress(
 	}
 
 	logFn(`Delta applied successfully`);
-	return id!;
+	// @ts-expect-error id is assigned in the try block
```
pipex: I don't like this, I think we should probably refactor the function to avoid this, probably the easiest is just return the id at lines 225 and 271

src/logging/index.ts
```diff
@@ -149,6 +146,7 @@ export async function attach(
 	}
 
 	return Bluebird.using(lock(containerId), async () => {
+		// eslint-disable-next-line @typescript-eslint/await-thenable, @typescript-eslint/no-confusing-void-expression
 		await logMonitor.attach(containerId, async (message) => {
 			await log({ ...message, serviceId });
 		});
```
pipex, line 151:
```suggestion
	return Bluebird.using(lock(containerId), async () => {
		logMonitor.attach(containerId, async (message) => {
			await log({ ...message, serviceId });
		});
		return Promise.resolve();
	});
```

src/logging/monitor.ts
```diff
@@ -236,6 +234,7 @@ class LogMonitor {
 		this.lastSentTimestamp = timestamp;
 	}
 
+	// eslint-disable-next-line @typescript-eslint/require-await
 	private async handleHostServiceRow(
```
pipex: Since the function doesn't really await the systemHook, the `async` is not needed so we can probably remove it

src/local-mode.ts
```diff
@@ -158,7 +158,7 @@ export class LocalModeManager {
 	private async collectOwnResources(): Promise<EngineSnapshot> {
 		try {
 			return this.collectContainerResources(
-				this.containerId || SUPERVISOR_CONTAINER_NAME_FALLBACK,
+				this.containerId || SUPERVISOR_CONTAINER_NAME_FALLBACK, // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
```
pipex: Why is nullish coalescing not the right choice??

src/network.ts
```diff
@@ -80,14 +80,14 @@ export const startConnectivityCheck = _.once(
...
-		const port = parseInt(parsedUrl.port!, 10);
+		const port = parseInt(parsedUrl.port ?? '80', 10);
 		const blink = await getBlink();
 
 		customMonitor(
 			{
 				host: parsedUrl.hostname ?? undefined,
 				port: port || (parsedUrl.protocol === 'https' ? 443 : 80),
-				path: parsedUrl.path || '/',
+				path: parsedUrl.path || '/', // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
```
pipex: Why not use nullish coalescing? A comment explaining it would be useful

src/compose/utils.ts
cywang117: Done:
```
    // This eslint disable is required because dockerode doesn't define Config.Healthcheck,
	// and can be removed once that type is added to @types/dockerode
```

src/compose/volume.ts
cywang117: This and the other lines that you commented on where I disabled `prefer-nullish-coalescing` don't make much sense in hindsight, I updated them to use `??`. Sorry for the noise!

src/device-state/index.ts
cywang117: Done!

src/host-config/proxy.ts
cywang117: Done!

src/lib/api-keys.ts
cywang117: ScopeCheck is wrong, it's a `Resolveable<T>` (`T | Promise<T>`) but there are no scope checks declared in api-keys that return `Promise<T>`. I changed `ScopeCheck` to return `boolean` and removed this `eslint-disable`

src/lib/docker-utils.ts
cywang117: This one is frustrating because we do return `id` in all cases unless an error is thrown, but TypeScript complains regardless. I changed `let id` to be initialized as type `string` or `undefined`, and added a runtime check to satisfy TS at the end before returning id:
```
	// id should always be assigned in all cases unless an error is thrown,
	// but TypeScript complains unless we explicitly confirm id is assigned
	if (!id) {
		throw new Error('Failed to get image ID after delta apply');
	}
	return id;
```
[see change here](https://github.com/balena-os/balena-supervisor/pull/2431/commits/fcab25d0e4d791938d205103c14031c9f48fb746#diff-f0fe0c6d5818f5b7a792d00302ffe0610d8f4164fd789213eef0025b46f8a6feR200)

PR #2432, opened 2025-10-01
Support io.balena.features.host-os.board-rev feature label

cywang117:
When `io.balena.features.host-os.board-rev` label is specified, adds BALENA_HOST_OS_BOARD_REV env var to containers that specified the label.

See: https://balena.fibery.io/Work/Improvement/Add-io.balena.features.host-os-release-label-on-supervisor-3203
Change-type: minor


src/compose/utils.ts
```diff
@@ -406,6 +406,12 @@ export async function addFeaturesFromLabels(
 				Capabilities: [['gpu']],
 				Options: {},
 			} as Dockerode.DeviceRequest),
+		'io.balena.features.host-os-release': async () => {
+			const osVersion = await getOSSemver(constants.hostOSVersionPath);
+			if (osVersion) {
+				setEnvVariables('OS_VERSION', osVersion);
```
pipex: I made a mistake when writing the improvement. There is already a `BALENA_HOST_OS_VERSION` variable. What we need is a `BALENA_HOST_OS_BOARD_REV` variable that includes the contents of `BALENA_BOARD_REV` in `/etc/os-release`. I left a comment on the improvement as well.

pipex: We should name the feature `io.balena.features.host-os.board-rev` just to be consistent

cywang117: Done and also updated improvement!

PR #2434, opened 2025-10-06
fix(2433): Fix redaction of secrets in env variables logging

### LINKED ISSUE #2433: Balena supervisor logging sensitive information.
# Bug: Service comparison logging exposes sensitive environment variables

## Description

There's a bug in the service comparison logging code at `src/compose/service.ts:820` that causes sensitive environment variables to be logged. The code incorrectly casts the result of `detailedDiff` as `ServiceConfig`, but `detailedDiff` returns an object with a different structure, so the redaction logic doesn't work.

## Current Code

```typescript
const diffObj = diff(thisOmitted, otherOmitted) as ServiceConfig;
if (diffObj.environment != null) {
    diffObj.environment = _.mapValues(
        diffObj.environment,
        () => 'hidden',
    );
}
log.debug('  Non-array fields: ', JSON.stringify(diffObj));
```

## The Issue

The `detailedDiff` function from `deep-object-diff` returns:
```typescript
{
  added: {...},
  deleted: {...},
  updated: {...}
}
```

The current code assumes it returns a `ServiceConfig` object and tries to access `diffObj.environment` directly, which won't exist in the actual structure. Environment changes would be under `diffObj.added.environment`, `diffObj.deleted.environment`, or `diffObj.updated.environment`.

**This means the environment redaction logic never runs, and sensitive environment variables are logged in plain text.**

## Impact

- **Security issue**: Sensitive environment variables (API keys, passwords, tokens, etc.) are logged in plain text
- The environment redaction logic doesn't work as intended

tmellapic:
The code was incorrectly casting the result of detailedDiff as ServiceConfig, which prevented environment variable redaction from working. This caused sensitive environment variables (API keys, passwords, tokens) to be logged in plain text in debug logs.

Fixed by:
- Creating a proper DetailedDiff interface for the return type
- Correctly redacting environment variables from added/deleted/updated sections
- Adding tests to verify environment variables are properly redacted

Change-Type: patch

# Description

  Fixes a security vulnerability where sensitive environment variables were being logged in plain text during service configuration comparisons.

  The code at `src/compose/service.ts:820` was incorrectly casting the result of `detailedDiff()` as `ServiceConfig`. However, `detailedDiff()` from the `deep-object-diff` library returns an object with `added`, `deleted`,
  and `updated` properties, not a `ServiceConfig` object. This meant the environment variable redaction logic (`diffObj.environment`) never executed, causing sensitive data (API keys, passwords, tokens, etc.) to be logged
  in plain text in debug logs.

  **Fixed by:**
  - Creating a proper `DetailedDiff<T>` interface to correctly type the return value of `detailedDiff()`
  - Updating the redaction logic to iterate through all three sections (`added`, `deleted`, `updated`) and redact environment variables from each
  - Adding comprehensive tests to verify environment variables are properly redacted from debug logs
  - Ensuring proper TypeScript typing with `Partial<ServiceConfig> | undefined` instead of `any`

  Fixes #2433

  # Type of change

  - [x] Bug fix (non-breaking change which fixes an issue)
  - [x] Security fix (addresses a vulnerability where sensitive data was exposed)

  **Change-Type: patch**

  # How Has This Been Tested?

  1. **Unit tests added**: A new test `should redact environment variables from debug logs when configs differ` was added to `test/unit/compose/service.spec.ts` that:
     - Creates two services with different environment variables containing sensitive values
     - Captures debug logs during `isEqualConfig()` comparison
     - Verifies that sensitive values are NOT present in logs
     - Verifies that the redaction marker `'hidden'` IS present in logs

  2. **Test execution**: All existing tests pass, including the new test:
     ```bash
     npm test

  3. Code verification:
    - TypeScript compilation succeeds with proper typing
    - Linting passes (balena-lint)
    - Code formatted with prettier

  The fix ensures that when service configurations differ and debug logging is enabled, environment variables are replaced with 'hidden' before being logged, preventing sensitive credential exposure.

**I'm working now to try to run this on one our devices to ensure it definitely works**


cywang117 (commented):
/deploy (to trigger GH actions tests)

test/unit/compose/service.spec.ts
```diff
@@ -658,6 +659,66 @@ describe('compose/service: unit tests', () => {
...
+				{ appName: 'test' } as any,
+			);
+
+			// Capture debug logs - log.debug may already be a spy from test setup
+			const debugCalls: string[] = [];
+			const originalFn = log.debug;
+			log.debug = (...args: any[]) => {
+				debugCalls.push(args.join(' '));
+				return originalFn.apply(log, args);
+			};
```
cywang117: You should be able to query the spy calls directly with `(log.debug as SinonSpy)`. See Spy API under https://sinonjs.org/releases/latest/spies/

tmellapic: Thanks for the pointer, adapted

tmellapic:
Hi @cywang117. Do you have any hints as to why the build is failing?

cywang117:
@tmellapic Could you rebase please? Looks like our rebase tool doesn't handle rebases from forks yet

PR #2439, opened 2025-11-13
Await provisioning on error before continuing

pipex:
The `provisionOrRetry` call was not being awaited to avoid memory leaks due to the recursion, however that can cause other issues if the api-binder is started without properly provisioning. 

This re-adds the await but using a loop to avoid the memory leak.

[#channel/support-help > Expired(?) device API key but no expiry set @ 💬](https://balena.zulipchat.com/#narrow/channel/403752-channel.2Fsupport-help/topic/Expired.28.3F.29.20device.20API.20key.20but.20no.20expiry.20set/near/555308672)

Change-type: patch

### DIFF
--- a/test/integration/state-engine.spec.ts
+++ b/test/integration/state-engine.spec.ts
@@ -249,13 +249,12 @@ describe('state engine', () => {
 			containerIds,
 		);
 
-		expect(await docker.getNetwork('123_default').inspect())
-			.to.have.property('IPAM')
-			.to.deep.equal({
-				Config: [{ Gateway: '192.168.91.1', Subnet: '192.168.91.0/24' }],
-				Driver: 'default',
-				Options: {},
-			});
+		const customNet = await docker.getNetwork('123_default').inspect();
+		expect(customNet).to.have.property('IPAM');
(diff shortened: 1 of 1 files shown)


test/integration/state-engine.spec.ts
```diff
@@ -249,13 +249,13 @@ describe('state engine', () => {
...
-			.to.have.property('IPAM')
-			.to.deep.equal({
-				Config: [{ Gateway: '192.168.91.1', Subnet: '192.168.91.0/24' }],
-				Driver: 'default',
-				Options: {},
-			});
+		const customNet = await docker.getNetwork('123_default').inspect();
+		expect(customNet).to.have.property('IPAM');
+		expect(customNet.IPAM).to.have.property('Config');
+		expect(customNet.IPAM.Config).to.have.lengthOf(1);
```
Page-: Fwiw you can combine these like
```suggestion
		expect(customNet.IPAM).to.have.property('Config').that.has.lengthOf(1);
```
and I think it gives slightly better error messages on failures

Page- (approved):

Issue #2440, opened 2025-12-01
Supervisor doesn't honour download-then-kill update strategy for metadata only updates


robthein:
## Summary

If any service has a metadata only update, the supervisor will acquire the lock for all services, even if there are images to download.

This goes against the [download-then-kill ](https://docs.balena.io/learn/deploy/release-strategy/update-strategies/#download-then-kill) update strategy as the user's app is no longer able to assume control of the lock since the supervisor obtains it from the start of the download, and doesn't release it until the containers are restarted and the update is applied.

## Expected behaviour

Locks should only be acquired after all images are downloaded, reflecting the default `download-then-kill` update strategy as documented [here](https://docs.balena.io/learn/deploy/release-strategy/update-strategies/#download-then-kill).

## Steps to reproduce

1. Deploy a multi-service application where some services require an image download and some services are unchanged (to force a metadata only update)
2. Observe the supervisor logs - the `Take update locks` indicates the lock hoarding

PR #2441, opened 2025-12-01
fix: premature locking for metadata-only changes

robthein:
# Description

The Supervisor currently takes the lock for all services when there are metadata-only updates, even when there are images to be downloaded for other services.

This means user apps are not able to acquire the lock during this period until the images finish downloading and the services restart, which goes against the [download-then-kill](https://docs.balena.io/learn/deploy/release-strategy/update-strategies/#download-then-kill) update strategy.

This change will make it respect the `download-then-kill` update strategy by checking if images for other services need downloading first and then only acquiring the lock when services need to restart to apply the update.

Fixes [#2440 ](https://github.com/balena-os/balena-supervisor/issues/2440)

# Type of change

Patch - bug fix

# How Has This Been Tested?

I tested this by making changes locally and syncing them to our device using `npm run sync`. 

1. Deploy a release for a multi-service application where some services have new images that need downloading, and some services remain unchanged.
2. Inspect when the lockfile gets created using something like `watch -n 0.5 "ls -la /tmp/balena/"`
3. The lockfile `/tmp/balena/updates.lock` should only be created at the end of the update when it restarts the services, whereas before the lockfile would be created from the start of the download and remain until the end.

Previously the Supervisor would trigger the `Take update locks` log and proceed to download images for services where the configuration has changed. So another way to verify is to check this isn't getting logged anymore.

I also added a unit test and adjusted the existing one.

### DIFF
--- a/src/compose/app.ts
+++ b/src/compose/app.ts
@@ -688,28 +688,29 @@ class AppImpl implements App {
 				context.availableImages,
 			);
 
-			if (
-				!needsSpecialKill &&
-				target != null &&
-				current.isEqualConfig(target, context.containerIds)
-			) {
-				// Update service metadata or start/stop a service
-				return this.generateContainerStep(
-					current,
-					target,
-					context.appsToLock,
-					context.targetApp.services,
-					servicesLocked,
-					dependenciesMetForKill,
-					context.rebootBreadcrumbSet,
-					context.bootTime,
-				);
-			}
-
 			let strategy: string;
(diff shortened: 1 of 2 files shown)


cywang117 (commented):
/deploy

cywang117:
Thanks @robthein , LGTM. I'll approve once the branch is rebased. Nice work!

cywang117 (approved):
LGTM

PR #2442, opened 2025-12-02
Extra firmware volume

cywang117:
- Write extra firmware volume location to config.json on startup
- Create extra-firmware system volume on startup regardless of feature label presence (the path will just be empty until users install the firmware block and optionally copy firmware into the volume). Also note this is done before writing the location to config.json, so the volume path is available to the OS service by the time the OS service is notified via config.json
- Configure services to use extra-firmware volume if `io.balena.features.extra-firmware` is present
- Purge extra-firmware volume when calling /v1/purge via API or dashboard

Additionally tested on device in local mode with a service with the label specified, and verified working

Change-type: minor
See: https://balena.fibery.io/Work/Project/linux-firmware-updates-1840/Specification-RFC-Extra-Firmware-Volume-4254

### DIFF
--- a/src/compose/volume-manager.ts
+++ b/src/compose/volume-manager.ts
@@ -2,16 +2,15 @@ import _ from 'lodash';
 import path from 'path';
 import type { VolumeInspectInfo } from 'dockerode';
 
-import { isNotFoundError, InternalInconsistencyError } from '../lib/errors';
+import { isNotFoundError } from '../lib/errors';
 import { safeRename } from '../lib/fs-utils';
 import { pathOnData } from '../lib/host-utils';
 import { docker } from '../lib/docker-utils';
 import * as LogTypes from '../lib/log-types';
-import log from '../lib/supervisor-console';
 import * as logger from '../logging';
 import { ResourceRecreationAttemptError } from './errors';
 import type { VolumeConfig } from './types';
-import { Volume } from './volume';
+import { Volume, VolumeNameParsingError } from './volume';
 
 export interface VolumeNameOpts {
 	name: string;
@@ -32,8 +31,9 @@ export async function getAll(): Promise<Volume[]> {
 			const volume = Volume.fromDockerVolume(volumeInfo);
 			volumesList.push(volume);
 		} catch (err) {
-			if (err instanceof InternalInconsistencyError) {
-				log.debug(`Found unmanaged or anonymous Volume: ${volumeInfo.Name}`);
+			if (err instanceof VolumeNameParsingError) {
+				// Ignore volume as it's an unmanaged (not part of an app that's managed by the Supervisor),
+				// an anonymous volume (a volume declared via VOLUME directive in Dockerfile), or the extra firmware volume.
 			} else {
 				throw err;
 			}
--- a/src/compose/volume.ts
+++ b/src/compose/volume.ts
@@ -1,10 +1,10 @@
 import type Docker from 'dockerode';
 import isEqual from 'lodash/isEqual';
 import omitBy from 'lodash/omitBy';
+import { TypedError } from 'typed-error';
 
 import * as constants from '../lib/constants';
 import { docker } from '../lib/docker-utils';
-import { InternalInconsistencyError } from '../lib/errors';
 import * as LogTypes from '../lib/log-types';
 import type { LabelObject } from '../types';
 import * as logger from '../logging';
@@ -17,6 +17,8 @@ import type {
 } from './types';
 
 export type Volume = VolumeIface;
+
+export class VolumeNameParsingError extends TypedError {}
 
 class VolumeImpl implements Volume {
 	private constructor(
@@ -124,14 +126,14 @@ class VolumeImpl implements Volume {
 	} {
 		const match = name.match(/(\d+)_(\S+)/);
 		if (match == null) {
-			throw new InternalInconsistencyError(
+			throw new VolumeNameParsingError(
 				`Could not detect volume data from docker name: ${name}`,
 			);
 		}
 
 		const appId = parseInt(match[1], 10);
 		if (isNaN(appId)) {
-			throw new InternalInconsistencyError(
+			throw new VolumeNameParsingError(
 				`Could not detect application id from docker name: ${match[1]}`,
 			);
 		}


test/unit/compose/volume.spec.ts
```diff
@@ -314,4 +318,19 @@ describe('compose/volume: unit tests', () => {
...
 		});
 	});
+
+	describe('extra firmware volume handling', () => {
+		it('should throw ExtraFirmwareVolumeError when parsing extra-firmware volume from Docker', () => {
+			expect(() =>
+				Volume.fromDockerVolume({
+					Driver: 'local',
+					Name: EXTRA_FIRMWARE_VOLUME_NAME,
+					Mountpoint: `/var/lib/docker/volumes/${EXTRA_FIRMWARE_VOLUME_NAME}/_data`,
```
ycardaillac: @acostach @cywang117, by reading your code I'm not too clear on what's going to the config.json.
On the OS side we are waiting for `<volume_name>/_data`, and I suspect that you are only giving us the `<volume_name>`, personally I'd rather have the full path so that the engine dependency is completely on the supervisor side, and we basically only work with directories, and no clever stuff related to Docker.
See: https://github.com/balena-os/meta-balena/pull/3754#discussion_r2585224724

cywang117: We discussed this internally and came to the conclusion that the OS scripts will be updated to infer `/var/lib/docker/volumes/$VOLUME_NAME/_data` from `$VOLUME_NAME`, as that path could change in the future and having the Supervisor write it to config.json in full would necessitate a reboot from the user to get to firmware at the updated path.

src/compose/volume-manager.ts
```diff
@@ -32,7 +33,9 @@ export async function getAll(): Promise<Volume[]> {
 			const volume = Volume.fromDockerVolume(volumeInfo);
 			volumesList.push(volume);
 		} catch (err) {
-			if (err instanceof InternalInconsistencyError) {
+			if (err instanceof ExtraFirmwareVolumeError) {
```
pipex: I don't think we need this, `addFeaturesFromLabels` is called after at the end of the `fromComposeObject` after the volume normalization has taken place. See how the `supervisor0` network is added to the services on the same function.

src/compose/volume.ts
```diff
@@ -34,6 +38,11 @@ class VolumeImpl implements Volume {
 			driverOpts: inspect.Options ?? {},
 		};
 
+		// We don't want to manage the extra firmware volume as part of state funnel
```
pipex: I don't think this module should know about the extra firmware volume. I suppose you want to avoid the supervisor reading the extra volume as part of the supervised volume list. Isn't that accomplished by not setting the supervised label? Let me know if there is another reason for this

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,77 @@
...
+	await create();
+
+	// Configure OS to look for extra firmware volume
+	await configureVolume(configJson);
+};
+
+const configureVolume = async (configJson: ConfigJsonConfigBackend) => {
+	const os: { [key: string]: any } = (await configJson.get('os')) ?? {};
+	if (typeof os !== 'object') {
+		throw new InternalInconsistencyError(
```
pipex: I don't think this qualifies as an `InternalConsistencyError`, that should be only to validate that internal calls are correct.

I also don't think a misconfigured config.json should crash the supervisor here, we probably need a warning or maybe issue a system error (that shows up in the dashboard)

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,77 @@
...
+			'Malformed config.json: os is not an object',
+		);
+	}
+	os.kernel ??= {};
+	if (typeof os.kernel !== 'object') {
+		throw new InternalInconsistencyError(
```
pipex: Same here

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,77 @@
...
+			'Malformed config.json: os.kernel is not an object',
+		);
+	}
+	if (typeof os.kernel.extraFirmwareVol === 'string') {
+		// Only configure location if not already configured
+		if (os.kernel.extraFirmwareVol === EXTRA_FIRMWARE_VOLUME_NAME) {
+			return;
+		} else {
+			// Other value detected, log warning and overwrite
+			log.warn(
```
pipex: This is probably an info level message, the volume name should be updatable

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+export const initialize = async (configJson: ConfigJsonConfigBackend) => {
+	log.info('Configuring extra firmware volume');
+	// Create volume - volume will be empty unless user includes
+	// appropriate linux-firmware block in their docker-compose.
+	// The OS is configured to use the volume path on host for firmware,
+	// but it's on the user to include the firmware block as well as reboot.
+	await create();
```
pipex, line 46: It makes me uneasy that we create the volume here. I think if this crashes it will bring the supervisor down right?? How about doing it via a step like `ensureSupervisorNetwork` does? 

https://github.com/balena-os/balena-supervisor/blob/master/src/compose/application-manager.ts#L214

src/compose/volume.ts
cywang117: Good point, I removed the supervised label from the extra-firmware volume and also removed the ExtraFirmwareVolumeError, which made things simpler.

src/compose/volume-manager.ts
cywang117: Removed ExtraFirmwareVolumeError and added extra-firmware volume initialization as a composition step, `configureExtraFirmwareVolume`, in d864dc18a68944bca18a91878cc9e0e548d707b7

src/lib/extra-firmware.ts
cywang117: I've switched to logging warnings and returning in d864dc18a68944bca18a91878cc9e0e548d707b7, which means the Supervisor does not crash when there's a malformed `os` object in config.json

src/lib/extra-firmware.ts
cywang117: Done in d864dc18a68944bca18a91878cc9e0e548d707b7

src/lib/extra-firmware.ts
cywang117: Done! d864dc18a68944bca18a91878cc9e0e548d707b7

src/compose/volume-manager.ts
```diff
@@ -33,7 +34,11 @@ export async function getAll(): Promise<Volume[]> {
 			volumesList.push(volume);
 		} catch (err) {
 			if (err instanceof InternalInconsistencyError) {
-				log.debug(`Found unmanaged or anonymous Volume: ${volumeInfo.Name}`);
+				// Only log unmanaged or anonymous volumes that
```
pipex: I would just remove the log line, it's just noise as unmanaged volumes are not an issue.

We should also change the error type here  https://github.com/balena-os/balena-supervisor/blob/a7e0eb63fd723b09620a49269b6e48e776ff38b9/src/compose/volume.ts#L38 to something explicit about parsing the volume name to avoid the use of the `InternalConsistencyError`

src/compose/application-manager.ts
```diff
@@ -213,6 +214,10 @@ export async function inferNextSteps(
 		} else {
 			steps.push({ action: 'ensureSupervisorNetwork' });
 		}
+	} else if (
+		(await extraFirmware.isInitialized(config.configJsonBackend)) === false
+	) {
+		steps.push({ action: 'configureExtraFirmwareVolume' });
```
pipex: Should we name the action `ensureExtraFirmwareVolume` for consistency?

src/compose/application-manager.ts
```diff
@@ -213,6 +214,10 @@ export async function inferNextSteps(
 		} else {
 			steps.push({ action: 'ensureSupervisorNetwork' });
 		}
+	} else if (
```
pipex: The `ensureSupervisorNetwork` step checks for services using the supervisor-api feature and kills them if `supervisorNetworkReady` is false. I'm not sure how it can happen that the supervisor network gets removed while there are container still attached to it. 

I don't think that can happen with volumes (so your code is correct here), but we might want to do some testing

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,99 @@
...
+import log from './supervisor-console';
+import { docker } from './docker-utils';
+import { isNotFoundError } from './errors';
+
+export const EXTRA_FIRMWARE_VOLUME_NAME = 'extra-firmware';
+
+// Extra-firmware is configured if both the volume exists
+// and the OS is configured to look for it.
+// configJson is passed using dependency injection for easier testing.
+export const isInitialized = async (configJson: ConfigJsonConfigBackend) => {
```
pipex: I think it's clearer to use functions rather than consts unless there is a specific reason you chose the latter

```suggestion
export async function isInitialized(configJson: ConfigJsonConfigBackend) {
```

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,99 @@
...
+		// initialization so we don't get stuck in an initialization loop.
+		if (!isNotFoundError(e)) {
+			log.error(`Error getting extra firmware volume: ${(e as Error).message}`);
+			return null;
+		}
+	}
+	return inConfigJson && volumeExists;
+};
+
+export const initialize = async (configJson: ConfigJsonConfigBackend) => {
```
pipex:
```suggestion
export async function initialize(configJson: ConfigJsonConfigBackend) {
```

src/compose/application-manager.ts
```diff
@@ -213,6 +214,10 @@ export async function inferNextSteps(
...
+		(await extraFirmware.isInitialized(config.configJsonBackend)) === false
```
pipex:
```suggestion
		!(await extraFirmware.isInitialized(config.configJsonBackend))
```

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,99 @@
...
+		const os = ((await configJson.get('os')) as Dictionary<any>) ?? {};
+		inConfigJson = os?.kernel?.extraFirmwareVol === EXTRA_FIRMWARE_VOLUME_NAME;
+	} catch {
+		// If there was an error getting the OS configuration,
+		// the only reason is that config.json is corrupted,
+		// so log error and skip initialization.
+		log.error(
+			'Error reading config.json, skipping extra firmware initialization',
+		);
+		return null;
```
pipex: A failure to read config.json should probably throw an error and maybe it's ok to even crash the supervisor as it is a non-recoverable error. We don't want to try to modify config.json at that point as it will probably make things worst.

I think that is how we deal with other config.json related errors, but might be worth a check.

Returning null here doesn't seem right

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,99 @@
...
+	try {
+		await docker.getVolume(EXTRA_FIRMWARE_VOLUME_NAME).inspect();
+		volumeExists = true;
+	} catch (e: unknown) {
+		// If volume doesn't exist, proceed with initialization, but if
+		// there was some other error getting the volume, skip
+		// initialization so we don't get stuck in an initialization loop.
+		if (!isNotFoundError(e)) {
+			log.error(`Error getting extra firmware volume: ${(e as Error).message}`);
+			return null;
```
pipex: I think it's OK for this to fail as well. We don't want initialization to be skipped if there is an engine error.

src/compose/application-manager.ts
pipex: I know the function can return `null` but I think that is not the right approach, on one side the `isInitialized` name suggests a boolean return, but also, a null return means the `configureExtraFirmwareVolume` step gets skipped  which means the container referencing the volume will get created but the volume won't exist yet.

src/compose/application-manager.ts
cywang117: > I'm not sure how it can happen that the supervisor network gets removed while there are container still attached to it.

I did some testing and the Engine should prevent this, so the step killing services does seem unnecessary.

src/compose/volume-manager.ts
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2442/commits/d8a840dbd26491ca3890e12ed9f32da428a12a77#diff-8b73dce5dc89202d16b05e9a58d0991038d9761113de6c097c4507cb0aec0edbR34) and changed error to `VolumeNameParsingError`

src/compose/application-manager.ts
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2442/commits/f4c5c015a9ee590751241721161cef211947dfad#diff-6f72acdfeb2708addb5c2b3f088c8207fc3470a55cbb5780788bd42d1cdd9672R218)

src/lib/extra-firmware.ts
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2442/commits/f4c5c015a9ee590751241721161cef211947dfad#diff-d01416735464a8ad83d3bbd0c6889567b421549ce24af0c3eec179f1996351f0R11) and further down

src/compose/application-manager.ts
cywang117: Done [here](https://github.com/balena-os/balena-supervisor/pull/2442/commits/f4c5c015a9ee590751241721161cef211947dfad#diff-6f72acdfeb2708addb5c2b3f088c8207fc3470a55cbb5780788bd42d1cdd9672R217), and now both config.json read errors and volume errors that aren't `404 not found` will throw `InternalInconsistencyError`

src/lib/extra-firmware.ts
```diff
@@ -0,0 +1,99 @@
...
+			'Error reading config.json, skipping extra firmware initialization',
+		);
+	}
+	let volumeExists = false;
+	try {
+		await docker.getVolume(EXTRA_FIRMWARE_VOLUME_NAME).inspect();
+		volumeExists = true;
+	} catch (e: unknown) {
+		// If volume doesn't exist, proceed with initialization, but if
+		// there was some other error getting the volume, skip
```
pipex: I think you need to update this comment

cywang117: Done:
```
        let inConfigJson = false;
	try {
		const os = ((await configJson.get('os')) as Dictionary<any>) ?? {};
		inConfigJson = os?.kernel?.extraFirmwareVol === EXTRA_FIRMWARE_VOLUME_NAME;
	} catch (e: unknown) {
		// If there was an error getting the OS configuration,
		// the only reason is that config.json is corrupted.
			`Error reading config.json while initializing extra firmware volume: ${(e as Error).message}`,
		);
	}
	let volumeExists = false;
	try {
		await docker.getVolume(EXTRA_FIRMWARE_VOLUME_NAME).inspect();
		volumeExists = true;
	} catch (e: unknown) {
		// If volume doesn't exist, proceed with initialization, but if
		// there was some other error getting the volume, there may be
		// an issue in the Engine or OS so we want to throw here.
		if (!isNotFoundError(e)) {
				`Error getting extra firmware volume: ${(e as Error).message}`,
			);
		}
	}
	```

pipex:
LGTM!

PR #2501, opened 2026-04-01
Add a `PATCH/v2/device/tags` endpoint to tell the supervisor device tags to report

Page-:
https://balena.fibery.io/Work/Project/Supervisor-API-for-device-tags-1031

Related: https://github.com/balena-io/open-balena-api/pull/2226

Change-type: minor

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1201,6 +1201,39 @@ Response:
 }
 ```
 
+> **Introduced in supervisor v17.8**
+
+Schedule device tags to be set in the balena API, persisting the schedule across restarts.
+Note that whilst this endpoint will work when the device does not have an available
+connection to the balena API the changes will only be reported once a connection is
+reestablished.
+
+The body should be a JSON object of tag key-value pairs which will be reported to the
+balena API if they differ from the last reported key-value pairs, saving the device from
+unnecessary requests and bandwidth usage. 
+
+```json
+{
+	"tagKey1": "tagValue1",
+	"tagKey2": "tagValue2",
+	...
+}
+```
+
+From an app container:
+```bash
+$ curl -X PATCH --header "Content-Type:application/json" \
+	--data '{"foo": "bar"}' \
+	"$BALENA_SUPERVISOR_ADDRESS/v2/device/tags?apikey=$BALENA_SUPERVISOR_API_KEY"
+```
+
+When successful, responds with 202 accepted and a JSON object:
+```json
+{
+	"status": "success",
+}
+```
+
 #### Device VPN Information
 
 > **Introduced in supervisor v11.4**
--- a/src/api-binder/tags.ts
+++ b/src/api-binder/tags.ts
@@ -1,0 +1,221 @@
+import * as db from '../db';
+import * as config from '../config';
+import { log } from '../lib/supervisor-console';
+import { createHandleRetry, withBackoff } from '../lib/backoff';
+import { InternalInconsistencyError, StatusError } from '../lib/errors';
+import { getDefaultHeaders } from '../lib/request';
+import { setTimeout } from 'timers/promises';
+import { once } from 'lodash';
+import type { SchemaReturn } from '../config/schema-type';
+
+type TagsPatchV3Body = {
+	[uuid: string]: {
+		[tagKey: string]: string;
+	};
+};
+type TagsPatchRecord = TagsPatchV3Body[string];
+
+const reportedTags: TagsPatchRecord = {};
+const targetTags: TagsPatchRecord = {};
+
+let reportConfigs:
+	| Awaited<
+			ReturnType<
+				typeof config.getMany<
+					| 'apiEndpoint'
+					| 'apiRequestTimeout'
+					| 'deviceApiKey'
+					| 'appUpdatePollInterval'
+					| 'uuid'
+				>
+			>
+	  >
+	| undefined;
+
+export const initialized = once(async () => {
+	await db.initialized();
+	await config.initialized();
+
+	let reported: SchemaReturn<'reportedTags'>;
+	let target: SchemaReturn<'targetTags'>;
+
+	({
+		reportedTags: reported,
+		targetTags: target,
+		...reportConfigs
+	} = await config.getMany([
+		'apiEndpoint',
+		'apiRequestTimeout',
+		'deviceApiKey',
+		'appUpdatePollInterval',
+		'uuid',
+		'reportedTags',
+		'targetTags',
+	]));
+
+	if (reported != null) {
+		for (const key of Object.keys(reported)) {
+			reportedTags[key] = reported[key];
+		}
+	}
+	if (target != null) {
+		let isDifferent = false;
+		for (const key of Object.keys(target)) {
+			targetTags[key] = target[key];
+			if (reportedTags[key] !== targetTags[key]) {
+				isDifferent = true;
+			}
+		}
+		if (isDifferent) {
+			triggerReport();
+		}
+	}
+});
+
+export const setTags = async (tags: TagsPatchRecord) => {
+	let hasChanged = false;
+	for (const [key, value] of Object.entries(tags)) {
+		if (targetTags[key] !== value) {
+			if (targetTags[key] == null && reportedTags[key] === value) {
+				// If the tag isn't already queued for update and we're setting to the currently reported value the we can just skip over the update
+				continue;
+			}
+			targetTags[key] = value;
+			hasChanged = true;
+		}
+	}
+	if (hasChanged) {
+		triggerReport();
+		await config.set({ targetTags });
+	}
+};
+
+async function report(tagsDiff: TagsPatchRecord) {
+	const { apiEndpoint, apiRequestTimeout, deviceApiKey } = reportConfigs!;
+
+	if (!apiEndpoint) {
+		throw new InternalInconsistencyError(
+			'No apiEndpoint available for patching device tags',
+		);
+	}
+
+	const endpoint = new URL('/device/v3/tags', apiEndpoint);
+	const defaultHeaders = await getDefaultHeaders();
+
+	const res = await fetch(endpoint, {
+		method: 'PATCH',
+		headers: {
+			...defaultHeaders,
+			Authorization: `Bearer ${deviceApiKey}`,
+			'Content-Type': 'application/json',
+		},
+		body: JSON.stringify({
+			[reportConfigs!.uuid!]: tagsDiff,
+		} satisfies TagsPatchV3Body),
+		signal: AbortSignal.timeout(apiRequestTimeout),
+	});
+
+	if (res.status < 200 || res.status >= 300) {
+		const retryAfter = res.headers.get('retry-after');
+		let jsonBody;
+		try {
+			jsonBody = await res.json();
+		} catch {
+			// ignore
+		}
+		throw new StatusError(
+			res.status,
+			jsonBody != null ? JSON.stringify(jsonBody, null, 2) : undefined,
+			retryAfter ? parseInt(retryAfter, 10) : undefined,
+		);
+	}
+}
+
+/**
+ * Collects current state and reports. Diffs report content with
+ * previous report and does not send an empty report.
+ * Should only be called by `reportCurrentStateWithBackoff` which wraps with backoff
+ */
+const reportCurrentState = async () => {
+	const tagsDiff: TagsPatchRecord = {};
+	for (const key of Object.keys(targetTags)) {
+		if (reportedTags[key] !== targetTags[key]) {
+			tagsDiff[key] = targetTags[key];
+		} else {
+			// If the values match then remove it, saving memory and ensuring the `Object.keys(targetTags).length` checks work as expected.
+			// Ideally we shouldn't hit/need this case but it's a final safeguard against potential unexpected behavior.
+			delete targetTags[key];
+		}
+	}
+
+	if (Object.keys(tagsDiff).length === 0) {
+		return;
+	}
+
+	await report(tagsDiff);
+	lastReportTime = performance.now();
+	for (const key of Object.keys(tagsDiff)) {
+		reportedTags[key] = targetTags[key];
+		if (targetTags[key] === tagsDiff[key]) {
+			delete targetTags[key];
+		}
+	}
+	await config.set({ reportedTags, targetTags });
+
+	log.info('Reported device tags to the cloud');
+};
+/**
+ * Reports current state with backoff for failures.
+ *
+ * Does *not* validate time elapsed since last report.
+ */
+async function reportCurrentStateWithBackoff() {
+	// Create a report that will backoff on errors
+	const reportWithBackoff = withBackoff(reportCurrentState, {
+		maxDelay: reportConfigs!.appUpdatePollInterval,
+		minDelay: 15000,
+		onFailure: createHandleRetry('Device tags'),
+	});
+
+	// Run in try block to avoid throwing any exceptions
+	try {
+		await reportWithBackoff();
+	} catch (e) {
+		log.error(e);
+	}
+}
+
+// How often can we report our state to the server in ms
+const maxReportFrequency = 10 * 1000;
+let lastReportTime = -Infinity;
+let reportLoopInProgress = false;
+const triggerReport = () => {
+	if (reportLoopInProgress) {
+		return;
+	}
+	reportLoopInProgress = true;
+	void (async () => {
+		try {
+			while (true) {
+				try {
+					if (performance.now() - lastReportTime > maxReportFrequency) {
+						await reportCurrentStateWithBackoff();
+					}
+				} finally {
+					const delay =
+						maxReportFrequency - (performance.now() - lastReportTime);
+					// Wait until we want to report again
+					await setTimeout(delay);
+
+					// If there's no updated target tags then we can stop the loop until a change comes in.
+					if (Object.keys(targetTags).length === 0) {
+						// eslint-disable-next-line no-unsafe-finally -- We very intentionally want to end the loop here.
+						return;
+					}
+				}
+			}
+		} finally {
+			reportLoopInProgress = false;
+		}
+	})();
+};
--- a/src/device-api/v2.ts
+++ b/src/device-api/v2.ts
@@ -28,6 +28,7 @@ import type { AuthorizedRequest } from '../lib/api-keys';
 import { fromV2TargetState } from '../lib/legacy';
 import * as actions from './actions';
 import { v2ServiceEndpointError } from './messages';
+import { setTags } from '../api-binder/tags';
 
 export const router = express.Router();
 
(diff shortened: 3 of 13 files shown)


src/api-binder/tags.ts
```diff
@@ -0,0 +1,209 @@
...
+				>
+			>
+	  >
+	| undefined;
+
+export const initialized = once(async () => {
+	await db.initialized();
+	await config.initialized();
+
+	const r = await db.models('device_tags').select().first();
```
Page-: It probably makes more sense for this to make use of the config mechanism

src/api-binder/tags.ts
```diff
@@ -0,0 +1,209 @@
...
+					| 'apiRequestTimeout'
+					| 'deviceApiKey'
+					| 'appUpdatePollInterval'
+					| 'uuid'
+				>
+			>
+	  >
+	| undefined;
+
+export const initialized = once(async () => {
```
pipex, line 35: This function is only called by tests

src/device-api/v2.ts
```diff
@@ -508,6 +509,40 @@ router.get('/v2/device/tags', async (_req, res) => {
 	}
 });
 
+router.post('/v2/device/tags', async (req, res) => {
```
pipex: This is a POST, but the API method is a PATCH, which might lead to confusion when stuff does not get removed on the backend.

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+	if (reported != null) {
+		for (const [key] of Object.keys(reported)) {
+			reportedTags[key] = reported[key];
+		}
+	}
+	if (target != null) {
+		for (const [key] of Object.keys(target)) {
+			targetTags[key] = target[key];
+			if (reportedTags[key] !== targetTags[key]) {
+				triggerReport();
```
pipex: This doesn't return so it will call `triggerReport` for each tag that is different. I know the function returns immediately if there is one in progress but it's still a bit strange

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+			if (targetTags[key] == null && reportedTags[key] === value) {
+				// If the tag isn't already queued for update and we're setting to the currently reported value the we can just skip over the update
+				continue;
+			}
+			targetTags[key] = value;
+			hasChanged = true;
+			if (reportedTags[key] !== value) {
+				// Only explicitly trigger the report if the reported value is different. If it's the same then it will only
+				// change/need to be reported if an update is in progress and in that case it will trigger a new report if necessary.
+				triggerReport();
```
pipex: Same here, this triggers a report for each unreported tag

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+			}
+		}
+	}
+});
+
+export const setTags = async (tags: TagsPatchRecord) => {
+	let hasChanged = false;
+	for (const [key, value] of Object.entries(tags)) {
+		if (targetTags[key] !== value) {
+			if (targetTags[key] == null && reportedTags[key] === value) {
```
pipex, line 79: this seems unnecessary, the `reportCurrentState` already diffs the current with the target, so there is duplicated logic. This function in my view should just queue the tags and write to disk + trigger report if there are changes

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+	const getStateAndReport = async () => {
+		const tagsDiff: TagsPatchRecord = {};
+		for (const key of Object.keys(targetTags)) {
+			if (reportedTags[key] !== targetTags[key]) {
+				tagsDiff[key] = targetTags[key];
+			}
+		}
+
+		if (Object.keys(tagsDiff).length === 0) {
+			return;
```
pipex: shouldn't it clear/update target tags if this is the case?

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+const triggerReport = () => {
+	if (reportLoopInProgress) {
+		return;
+	}
+	reportLoopInProgress = true;
+	void (async () => {
+		try {
+			while (true) {
+				try {
+					if (performance.now() - lastReportTime > maxReportFrequency) {
```
pipex, line 201: The difference is probably negligible, but you might want to store the result of `performance.now() - lastReportTime` and use that in the `try` and the `finally` blocks

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+						await reportCurrentState();
+					}
+				} finally {
+					const delay =
+						maxReportFrequency - (performance.now() - lastReportTime);
+					// Wait until we want to report again
+					await setTimeout(delay);
+
+					// If there's no updated target tags then we can stop the loop until a change comes in.
+					if (Object.keys(targetTags).length === 0) {
```
pipex, line 211: Should this happen earlier, even perhaps at the top of the loop?

src/api-binder/tags.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * Collects current state and reports with backoff. Diffs report content with
+ * previous report and does not send an empty report.
+ *
+ * Does *not* validate time elapsed since last report.
+ */
+async function reportCurrentState() {
+	const getStateAndReport = async () => {
```
pipex: this function is doing a lot, is not only diffing the target with the already reported tags, but it is also cleaning up already sent tags and updating the database. The name doesn't really reflect that

pipex (commented):
I find the logic a bit hard to parse. I understand some of the decisions you made, that consider that new target tags could be added while you are sending a diff report, however it is not evident from the code and if there are any bugs in the logic they will be difficult to catch. There are also no tests for the module logic, and tests will be difficult to write since we are relying on global variables. 

I wonder if we could use some data structure or functional approach for the queuing/dequeuing logic that can be tested separately.

src/api-binder/tags.ts
```diff
@@ -0,0 +1,209 @@
...
+	| undefined;
+
+export const initialized = once(async () => {
+	await db.initialized();
+	await config.initialized();
+
+	const r = await db.models('device_tags').select().first();
+	const reported = JSON.parse(r.reported);
+	const target = JSON.parse(r.target);
+	for (const [key] of Object.keys(reported)) {
```
cywang117:
```suggestion
	for (const key of Object.keys(reported)) {
```
`[key]` will extract the first char from each key in Object.keys

src/api-binder/tags.ts
```diff
@@ -0,0 +1,209 @@
...
+	// Create a report that will backoff on errors
+	const reportWithBackoff = withBackoff(getStateAndReport, {
+		maxDelay: reportConfigs!.appUpdatePollInterval,
+		minDelay: 15000,
+		onFailure: createHandleRetry('Device tags'),
+	});
+
+	// Run in try block to avoid throwing any exceptions
+	try {
+		console.error('------ reporting with backoff');
```
cywang117: Leftover debug log?

src/device-api/v2.ts
```diff
@@ -483,6 +484,40 @@ router.get('/v2/device/tags', async (_req, res) => {
...
+		}
+		const body = req.body as unknown;
+		if (body == null || typeof body !== 'object') {
+			throw new BadRequestError('Invalid tags body');
+		}
+		for (const [key, value] of Object.entries(body)) {
+			if (typeof key !== 'string' || typeof value !== 'string') {
+				throw new BadRequestError('Invalid tags body');
+			}
+		}
```
cywang117: Based on the validation here I'm confused, is `req.body.tags` meant to contain the tags, or `req.body` directly?

src/device-api/v2.ts
```diff
@@ -483,6 +484,40 @@ router.get('/v2/device/tags', async (_req, res) => {
...
+
+		await setTags(body as Record<string, string>);
+
+		return res.status(202).json({
+			status: 'success',
+		});
+	} catch (e: any) {
+		if (e instanceof BadRequestError) {
+			res.status(e.statusCode).json({
```
cywang117, line 513:
```suggestion
			return res.status(e.statusCode).json({
```

src/device-api/v2.ts
```diff
@@ -483,6 +484,40 @@ router.get('/v2/device/tags', async (_req, res) => {
...
+				status: 'failed',
+				message: e.message,
+			});
+		}
+		res.status(500).json({
```
cywang117, line 519:
```suggestion
		return res.status(500).json({
```
optional suggestion as the previous `return` should take care of it

src/api-binder/tags.ts
Page-: Yeah.. I was trying to follow the supervisor pattern but it is more tricky/confusing to have these `initialized` functions that have to be explicitly called rather than running initialization code at module initialization :/
I've added it to the list of other modules to initialize on startup now

src/api-binder/tags.ts
Page-: The point of this is that we don't need to write to the db if we're setting to the current value, if I remove this check we'll write to the db every time even if the actual report to the API would end up being skipped, ie it's a necessary part of implementing
> This function in my view should just queue the tags and write to disk + trigger report if there are changes

based on my understanding of what you mean, am I misunderstanding your meaning or was there something else?

src/api-binder/tags.ts
Page-: Yeah the idea was that having that logic in the `triggerReport` function itself would guarantee it's always enforced and also make things simpler by not needing to reimplement the "has already triggered" logic in each place that might trigger a report

src/api-binder/tags.ts
Page-: This is in the `finally` because the idea is that it checks if we need to continue iterating or not and we should always need to report at least once. That said I realize that as the code evolved it would be possible to use this to replace the `while(true)` now

src/api-binder/tags.ts
Page-: No, because `lastReportTime` will change between here and the `finally` if we actually report the current state which means the result will be different and could cause unexpecte behavior

src/api-binder/tags.ts
Page-: Yes, in theory it should never be the case that a `targetTags` entry exists that matches `reportedTags` but I've put a safeguard in place to ensure it gets synced up

src/device-api/v2.ts
Page-: Yeah this was me getting confused between trying to match the supervisor api but also trying to match the balena api and having parts of both. I think the `tags` part can just be dropped in the request body as its primary use in the response body of `GET /v2/device/tags` is because it has to sit alongside the `status` property which isn't relevant here. Fixed

src/device-api/v2.ts
Page-: I think it's bad form to return a value when the true intention is an empty/`void` return, just because it can end up causing headaches/confusion later on. I've added a return in the `if` because it 100% needs to be there

src/device-api/v2.ts
Page-: Good catch

src/device-api/v2.ts
Page-: I've changed it to a PATCH

src/api-binder/tags.ts
Page-: I've separated this out to the module level as `reportCurrentState` and renamed this one to `reportCurrentStateWithBackoff` as all it actually does is wrap that function in `withBackoff` and call it, and the only reason we really have this wrapper is because `withBackoff` needs to freshly wrap each time (which makes no difference in its current uses as that would happen/be necessary regardless, but in this case could otherwise be skipped

src/api-binder/tags.ts
Page-: Thinking more on this it might indeed be better to move that specific logic out as it's not that big of a lift (one place already has to have it anyway..) and it means `triggerReport` won't have to handle delaying in order to let all the sync calls complete before doing the first report

test/integration/device-state/tags.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+			expect(batchedTagsPatchBody).to.deep.equal({
+				[uuid!]: {
+					d: '4',
+					e: '5',
+					f: '6',
+				},
+			});
+		});
+
+		it('should not report unchanged tags', async function () {
```
Page-, line 106: It would be possible to report all tags even if only 1 has changed without being that much additional work but I think it would actually end up being more confusing if sometimes unchanged tags get re-reported and others they don't

src/api-binder/tags.ts
pipex: that's fair, is just that there are multiple places in the module where these comparison are made and I was looking for some way to simplify

src/api-binder/tags.ts
Page-: Good spot, this would have been tricky to debug I suspect

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
...
+{
+	"tagKey1": "tagValue1",
+	"tagKey2": "tagValue2",
+	...
+}
+```
+
+From an app container:
+```bash
+$ curl -X PATCH --header "Content-Type:application/json" \
```
thgreasi, line 1225: No-strong-opinion-NIT: Should this be a PUT, so that it's clearer that this endpoint upserts the tags?

src/device-api/v2.ts
```diff
@@ -483,6 +484,40 @@ router.get('/v2/device/tags', async (_req, res) => {
...
+	try {
+		if (req.body == null || typeof req.body.tags !== 'object') {
+			throw new BadRequestError('Invalid tags body');
+		}
+		const body = req.body as unknown;
+		if (body == null || typeof body !== 'object') {
+			throw new BadRequestError('Invalid tags body');
+		}
+		for (const [key, value] of Object.entries(body)) {
+			if (typeof key !== 'string' || typeof value !== 'string') {
```
cywang117: NIT: `key` should always be of a type string, so this check shouldn't ever trigger

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
 }
 ```
 
+> **Introduced in supervisor $TODO**
```
cywang117: Just a placeholder note here to update `$TODO` before merging

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
...
+
+Scedule device tags to be set in the balena API, persisting the schedule across restarts.
```
cywang117:
```suggestion
Schedule device tags to be set in the balena API, persisting the schedule across restarts.
```

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
...
+Note that whilst this endpoint will work when the device does not have an available
+connection to the balena API the changes will not be reported until a connection is
```
cywang117:
```suggestion
connection to the balena API, the changes will not be reported until a connection is
```

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
...
+	"tagKey2": "tagValue2",
+	...
+}
+```
+
+From an app container:
+```bash
+$ curl -X PATCH --header "Content-Type:application/json" \
+	--data '{"foo": "bar"}' \
+	"$BALENA_SUPERVISOR_ADDRESS/v2/device/tags?apikey=$BALENA_SUPERVISOR_API_KEY""
```
cywang117:
```suggestion
	"$BALENA_SUPERVISOR_ADDRESS/v2/device/tags?apikey=$BALENA_SUPERVISOR_API_KEY"
```

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
...
+```
+
+Response:
+```none
+OK
+```
```
cywang117: Based on the code, this response shouldn't ever occur so should be removed

docs/API.md
```diff
@@ -1201,6 +1201,44 @@ Response:
...
+
+Response:
```
cywang117: Should document the status code as `202`

src/api-binder/tags.ts
```diff
@@ -4,7 +4,7 @@ import { log } from '../lib/supervisor-console';
 import { createHandleRetry, withBackoff } from '../lib/backoff';
 import { InternalInconsistencyError, StatusError } from '../lib/errors';
 import { getDefaultHeaders } from '../lib/request';
-import { setTimeout } from 'timers/promises';
+import { setTimeout, setImmediate } from 'timers/promises';
```
cywang117: This commit's changes are cancelled out by 3a05340d, so could be dropped entirely

test/integration/device-state/tags.spec.ts
```diff
@@ -21,47 +22,142 @@ describe('device-tags', () => {
 	});
 
 	describe('Setting tags', () => {
-		it('accepts a tag internally and attempt to update balena-api', async function (this: Mocha.Context) {
```
cywang117: Should add a nock cleanup step in an afterEach to avoid leaking

src/api-binder/tags.ts
```diff
@@ -54,13 +54,13 @@ export const initialized = once(async () => {
...
 	if (reported != null) {
-		for (const [key] of Object.keys(reported)) {
+		for (const key of Object.keys(reported)) {
 			reportedTags[key] = reported[key];
 		}
 	}
 	if (target != null) {
 		let isDifferent = false;
-		for (const [key] of Object.keys(target)) {
+		for (const key of Object.keys(target)) {
```
cywang117: These fixes could be squashed into 25f29d33

src/api-binder/tags.ts
pipex: Is this still something you want to do before merge?

docs/API.md
Page-: Yup, looks like a copy/paste thing where I based most on one example but then copied the correct response from a different one but forgot to forget the original response I copied 🤦

test/integration/device-state/tags.spec.ts
Page-: I didn't see any nock cleanup steps anywhere else, but specifically for this code it uses `.once()` so they'll automatically be cleaned up (unless the test fails, at which point we would be in an unknown/unexpected state anyway and could break in any number of ways)

docs/API.md
Page-: Looking at this again I rephrased it with less negatives so it should be clearer, however I'm pretty sure adding a comment there would be incorrect as it's a continuation of a single statement rather than combining two related ones

src/device-api/v2.ts
Page-: Yup, since it's a JSON object the keys should only ever be strings, and `Object.entries` only returns string keyed entries on top of that.. I think this came from an earlier version where it was complaining about potentially symbols or stuff, either way I've changed it now

docs/API.md
Page-: After discussion I think it's a PATCH because it's updating a subset of the device tag resource rather than replacing them all, if it was eg `/v2/device/tag/:tagKey` then that would make sense as a PUT

pipex (approved):

PR #2508, opened 2026-04-08
Convert code to async/await

Page-:
This is how new code would be written and can result in better stack traces via zero cost async stack traces and less closure overhead

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?


src/compose/service-manager.ts
```diff
@@ -681,7 +681,6 @@ function waitToKill(service: Service, timeout: number | string) {
 		`Waiting for handover to be completed for service: ${service.serviceName}`,
 	);
 
-	return wait().then(() => {
-		log.success(`Handover complete for service ${service.serviceName}`);
-	});
+	await wait();
```
Page-, line 684: I believe this might also stop a memory leak from the nested return chain but I'm not 100% sure if async/await by itself breaks that chain - either way it shouldn't make it worse

src/config/backends/backend.ts
```diff
@@ -43,13 +43,13 @@ export abstract class ConfigBackend {
 
 	// Is a reboot required for the given config options?
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
-	public async isRebootRequired(_opts: ConfigOptions): Promise<boolean> {
-		return Promise.resolve(true);
+	public isRebootRequired(_opts: ConfigOptions): Resolvable<boolean> {
```
Page-, line 46: With async/await `Resolvable` vs `Promise` can be treated equivalently but makes implementing code a bit simpler via not needing to wrap everything

src/lib/fs-utils.ts
```diff
@@ -59,32 +59,39 @@ export async function mkdirp(pathName: string): Promise<void> {
...
-					.then(() => fs.utimes(file, time, time))
-			: e,
-	);
+		if (isENOENT(e)) {
+			const fd = await fs.open(file, 'w');
+			await fd.close();
+			// If date is custom we need to change the file atime and mtime
+			await fs.utimes(file, time, time);
+		} else {
+			// TODO: This is preserving previous behavior but was probably intended to be a throw?
```
Page-, line 90: I left this behavior the same as before but I'm pretty sure the previous behavior was a bug, hence this comment

test/lib/mocked-device-api.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint @typescript-eslint/require-await: 0 */
+/* eslint-disable @typescript-eslint/require-await */
```
Page-, line 1: I switched to `eslint-disable` because I think it's a little clearer and easier to search for

src/migrations/20171129013519_legacy.js
```diff
@@ -153,8 +141,8 @@ exports.up = function (knex) {
 		dropTable('image'),
 		dropTable('container'),
 	]);
-};
+}
 
-exports.down = function () {
-	return Promise.reject(new Error('Not implemented'));
-};
+export function down() {
```
Page-, line 146: I standardized all of these `down` functions to match. Over time it looks like they've been modernized but the old ones left as-is, and then sometimes it went back to an older version (presumably due to copy/pasting from an older migration) so standardizing them all should avoid that in future and solves the promise vs async/await vs sync throw situation at the same time

src/migrations/20171129013519_legacy.js
```diff
@@ -7,64 +7,57 @@
...
-exports.up = function (knex) {
-	const addColumn = function (table, column, type) {
-		return knex.schema.hasColumn(table, column).then((exists) => {
-			if (!exists) {
-				return knex.schema.table(table, (t) => {
-					return t[type](column);
-				});
-			}
-		});
+export async function up(knex) {
```
Page-, line 10: I standardized the exports for migrations files to match the modern format, which included standardizing them to the modern async/await format the newer ones use

src/config/functions.ts
```diff
@@ -79,42 +81,39 @@ export const fnSchema = {
...
+			deviceApiKey: conf.deviceApiKey,
+			apiEndpoint: conf.apiEndpoint,
+			apiRequestTimeout: conf.apiRequestTimeout,
+			registered_at: conf.registered_at,
+			deviceId: conf.deviceId,
+			supervisorVersion: conf.version,
+			osVersion: conf.osVersion,
+			osVariant: conf.osVariant,
+			macAddress: conf.macAddress,
+		};
```
pipex, line 116: Why not 

```suggestion
		return config.getMany([
			'uuid',
			'applicationId',
			'apiKey',
			'deviceApiKey',
			'deviceArch',
			'deviceType',
			'apiEndpoint',
			'apiRequestTimeout',
			'registered_at',
			'deviceId',
			'version',
			'osVersion',
			'osVariant',
			'macAddress',
		]);
```

src/compose/images.ts
```diff
@@ -569,24 +566,22 @@ const inspectByURI = async (imageName: string) =>
...
-				: Promise.reject(
-						new StatusError(
-							404,
-							`Failed to find an image matching ${imageName}`,
-						),
-					),
-		);
+		.select();
+	const [img] = images.filter(
+		(i): i is typeof i & Required<Pick<typeof i, 'dockerImageId'>> =>
```
pipex, line 575: That is quite a type definition to avoid using `img.dockerImageId!` below

pipex: Just commenting on a fact, not suggesting to remove it

src/compose/service-manager.ts
pipex: I'm pretty sure it doesn't but you are right it doesn't make it worse. It seems that basically no-one uses handover which is probably why this has not been an issue

src/device-api/actions.ts
```diff
@@ -35,12 +35,14 @@ import * as extraFirmware from '../lib/extra-firmware';
...
 export const runHealthchecks = async (
-	healthchecks: Array<() => Promise<boolean>>,
+	healthchecks: Array<() => Resolvable<boolean>>,
 ) => {
 	const HEALTHCHECK_FAILURE = 'Healthcheck failed';
 
 	try {
-		const checks = await Promise.all(healthchecks.map((fn) => fn()));
+		const checks = await Promise.all(
+			healthchecks.map(async (fn) => await fn()),
```
pipex, line 44: Isn't this the same as before? It seems that in some cases we are using `return fn()` when `fn()` is async. Is there a benefit to returning `await fn()`?

src/lib/fs-utils.ts
pipex: Yeah, I think it should be a `throw`, I think it's ok to fix that here

pipex (commented):
Just left a few comments @Page-, but they are quite minor. Overall this looks OK to merge. I tested it on a device and made sure to test a few extra code paths.

src/config/functions.ts
Page-: So the primary reason was I didn't want to actually change any behavior unexpectedly (if you hide whitespace changes you'll see this is unchanged), but for this specific case that change would break due to:
```ts
			provisioningApiKey: conf.apiKey,
			...
			supervisorVersion: conf.version,
```
Obviously that could be updated/changed downstream as well but I think it's out of scope of this PR

pipex: You are right, I looked at the list but missed those differences. Keeping it this way is fine

src/compose/images.ts
Page-: Yeah.. it's basically the accurate typing for what is being done which unfortunately doesn't get inferred by default. That said I think there's a better approach to this with the best of both worlds. I'll push that as a separate commit

src/device-api/actions.ts
Page-: So due to the change to `Resolvable` the linter will otherwise complain with
```
  44:4  error  Unexpected iterable of non-Promise (non-"Thenable") values passed to promise aggregator  @typescript-eslint/await-thenable
```
But I've reverted that `Resolvable` change since it's only actually for tests and am handling it in tests instead.

Fwiw adding the `async (...) => await ...` does mean this line will be included in the async stack trace whereas without it then I don't think it will be included

src/compose/service-manager.ts
```diff
@@ -681,7 +681,7 @@ function waitToKill(service: Service, timeout: number | string) {
 		`Waiting for handover to be completed for service: ${service.serviceName}`,
 	);
 
-	return wait().then(() => {
-		log.success(`Handover complete for service ${service.serviceName}`);
-	});
+	// TODO: This is likely to cause a (slow) memory leak whilst the service is waiting to handover
```
Page-, line 684: This isn't new, I just spotted it during this PR and added the note so it is easier to find/fix in future

pipex:
LGTM

PR #2510, opened 2026-04-10
Firewall: avoid `exports.` to improve type-checking

Page-:
`exports` is `any` so referencing it means no typings are applied in cases you would normally expect them to be and means that it's easy to introduce bugs due to expecting typescript to be able to catch them when in fact it's gone via a commonjs indirection. As an added benefit it also makes the code more ESM ready

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/src/lib/firewall.ts
+++ b/src/lib/firewall.ts
@@ -149,12 +149,14 @@ async function applyFirewall(
 	);
 
 	// apply the firewall rules...
-	await exports.applyFirewallMode(firewallMode ?? '');
+	await applyFirewallMode(firewallMode ?? '');
 }
 
 export const ALLOWED_MODES = ['on', 'off', 'auto'];
 
-export async function applyFirewallMode(mode: string) {
+export const applyFirewallMode = async function applyFirewallMode(
+	mode: string,
+) {
 	// only apply valid mode...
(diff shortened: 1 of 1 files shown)


src/lib/firewall.ts
```diff
@@ -149,12 +149,14 @@ async function applyFirewall(
...
 
 	// apply the firewall rules...
-	await exports.applyFirewallMode(firewallMode ?? '');
+	await applyFirewallMode(firewallMode ?? '');
 }
 
 export const ALLOWED_MODES = ['on', 'off', 'auto'];
 
-export async function applyFirewallMode(mode: string) {
+export const applyFirewallMode = async function applyFirewallMode(
```
Page-, line 157: This needed to change to `export const ...` because of an interaction between how typescript generates the output and how the tests try to override it; with `export function` it uses it as `await applyFirewallMode(...)` which cannot be overridden external but with `export const` it uses it as `await (0, exports.applyFirewallMode)(...)` which can be overridden externally. Realistically the tests overriding it is invalid based on the typing (and hence has to have casts/overrides to get past tsc) so these small differences causing unexpected interactions isn't actually that surprising, but it's the reason for this specific change

pipex:
LGTM

PR #2512, opened 2026-04-28
Update contrato to v0.14

pipex:
The newer version has stricter type validation and is more strictly typed.

This change preserves previous decisions about contract validation to avoid breaking existing releases.

Change-type: patch

### DIFF
--- a/test/unit/lib/contracts.spec.ts
+++ b/test/unit/lib/contracts.spec.ts
@@ -102,7 +102,7 @@ describe('lib/contracts', () => {
 				requires: [
 					{
 						type: 'sw.l4t',
-						version: '32.2',
+						version: '^32.2',
 					},
 					{
 						type: 'hw.device-type',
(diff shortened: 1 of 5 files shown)


test/unit/lib/contracts.spec.ts
```diff
@@ -102,7 +102,7 @@ describe('lib/contracts', () => {
 				requires: [
 					{
 						type: 'sw.l4t',
-						version: '32.2',
+						version: '^32.2',
```
pipex, line 105: Contrato now converts version requirements to the correct nomenclature.

test/unit/lib/contracts.spec.ts
```diff
@@ -127,32 +127,6 @@ describe('lib/contracts', () => {
 				}),
 			).to.not.throw();
 		});
-
-		it('should reject if a contract has an unsupported type', () => {
```
pipex, line 131: This is not really needed. Unknown version requirements will be unsatisfied by definition, so it's redundant to validate them separately.

test/unit/lib/contracts.spec.ts
```diff
@@ -660,7 +609,6 @@ describe('lib/contracts', () => {
 			`${semver.major(OS_VERSION)}.${semver.minor(OS_VERSION)}`,
 			`${semver.major(OS_VERSION)}.${semver.minor(OS_VERSION)}.*`,
 			`${semver.major(OS_VERSION)}.${semver.minor(OS_VERSION)}.${semver.patch(OS_VERSION)}`,
-			`${semver.major(OS_VERSION)}.${semver.minor(OS_VERSION)}.${semver.patch(OS_VERSION)}+rev*`,
```
pipex, line 663: `rev*` is not valid metadata according to the semver spec

test/unit/lib/contracts.spec.ts
```diff
@@ -457,31 +431,6 @@ describe('lib/contracts', () => {
...
-					commit: 'd0',
-					serviceName: 'service',
-					contract: {
-						type: 'sw.container',
-						name: 'user-container',
-						slug: 'user-container',
-						requires: [
-							{
-								type: 'hw.device-type',
-								name: 'raspberrypi3',
```
pipex, line 471: `name` is not a valid container requirement matcher, this would fail to validate. This is not important as `parseContract` removes this property anyway

PR #2513, opened 2026-04-29
Fix the network connectivity check

pipex:
The connectivity check function was receiving a `path` argument, added on #820. The function, that just connects to a socket was looking for a unix socket instead of connecting to the remote host, which means it was always failing.

Change-type: patch


cywang117 (approved):
LGTM

PR #2514, opened 2026-04-29
Trim whitespaces around dtparam input values

cywang117:
Slightly more defensive, and prevents boot config apply loop from surrounding whitespaces such as in
`BALENA_HOST_CONFIG_dtparam=" "foo=1","bar=2""`

Do note that whitespaces are invalid as part of an input, and inputs like these are user error.

Encountered on support where user upgraded from v14.x.x SV to v17.x.x. The behavior change originates from v16.0.0 which adds multiline dtparam handling. Pre-v16 Supervisors handled any whitespaces gracefully.

Change-type: patch

### DIFF
--- a/src/config/backends/config-txt.ts
+++ b/src/config/backends/config-txt.ts
@@ -150,9 +150,9 @@ export class ConfigTxt extends ConfigBackend {
 								// We make sure to put the base param in the right overlays
 								// since RPI doesn't seem to be too strict about the ordering
 								// when it comes to these base params
-								baseParams.push(value);
+								baseParams.push(param);
 							} else {
-								currParams.push(value);
+								currParams.push(param);
 							}
 						}
 					} else if (key === 'dtoverlay') {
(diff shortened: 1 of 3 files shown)


src/config/backends/config-txt.ts
```diff
@@ -150,9 +150,9 @@ export class ConfigTxt extends ConfigBackend {
 								// We make sure to put the base param in the right overlays
 								// since RPI doesn't seem to be too strict about the ordering
 								// when it comes to these base params
-								baseParams.push(value);
+								baseParams.push(param);
```
pipex, line 153: this is fixing a bug right? I'm surprised we haven't heard from this

cywang117: Yeah, I think the bug would only ever manifest with a user-edited config.txt containing multiple dtparams on a line, such as
```
dtparam=audio=on,i2c_arm=on
```

The Supervisor writes back one dtparam per line, so this shouldn't occur for the majority of our users. But if a user were to preload a config.txt in the above valid format with multiple dtparams on one line, `getBootConfig` would return `["audio=on,i2c_arm=on", "audio=on,i2c_arm=on"]` instead of the correct `["audio=on","i2c_arm=on"]`

cywang117: I added an integration test to ensure this fix works as expected, and separated it out from the whitespace commit for clarity

pipex (approved):

PR #2515, opened 2026-05-05
Invalidate target state cache when creating device config variables

pipex:
There seems to be an edge condition where the API returns a 304 from the target state endpoint even after creating new device config variables. This can result in the supervisor deleting the local configuration in config.txt leaving the device in a malfunctioning state.

This change adds an `invalidate` function to the target state poll module. This invalidates the in-memory target state cache, forcing the supervisor to request the full target state. We call this function after creating a new device config variable to tell the poll module to expect a new target.

Change-type: patch

### DIFF
--- a/src/api-binder/index.ts
+++ b/src/api-binder/index.ts
@@ -358,6 +358,7 @@ async function reportInitialEnv(
 				resource: 'device_config_variable',
 				body: envVar,
 			});
+			await TargetState.invalidate();
 		}
 	}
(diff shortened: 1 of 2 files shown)


pipex:
This doesn't solve the issue since it relates to the backend returning a stale state which we identify with the 304 but invalidating the cache doesn't guarantee that the target state will be up-to-date, even if it returns 200

PR #2519, opened 2026-05-15
Pin back to Node 22

cywang117:
SV on RPi Zero was working in 17.6.10 but not 17.6.11 which bumped Node to 24. 

Node 24's V8 engine has issues with `rpi`/armv6 arch, throwing `Fatal error in , line 0 unreachable code`.

We still support RPi Zero devices in production, roughly 3k at time of this commit.

Change-type: patch
See: https://balena.fibery.io/Work/Improvement/Unblock-the-raspberrypi-pipeline-4203

### DIFF
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -1,8 +1,8 @@
 ARG ARCH=%%BALENA_ARCH%%
 ARG FATRW_VERSION=0.2.21
-ARG NODE="nodejs~=24"
+ARG NODE="nodejs~=22"
 ARG NPM="npm~=11"
-ARG ALPINE_VERSION="3.23"
+ARG ALPINE_VERSION="3.22"
 
 ###################################################
 # Build the supervisor dependencies
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -60,7 +60,7 @@ externalModules.push(
 	new RegExp(
 		'^(' +
 			_.reject(maybeOptionalModules, requiredModules)
-				.map(RegExp.escape)
+				.map(_.escapeRegExp)
 				.join('|') +
 			')(/.*)?$',
 	),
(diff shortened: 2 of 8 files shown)


Dockerfile.template
```diff
@@ -1,8 +1,8 @@
 ARG ARCH=%%BALENA_ARCH%%
 ARG FATRW_VERSION=0.2.21
-ARG NODE="nodejs~=24"
+ARG NODE="nodejs~=22"
 ARG NPM="npm~=11"
-ARG ALPINE_VERSION="3.23"
+ARG ALPINE_VERSION="3.22"
```
Page-, line 5: Is reverting alpine actually necessary? I'm sure node 22 could work on the newer alpine and make sure it still gets its LTS updates

pipex: Newer versions of alpine sometimes stop providing older packages. We can test if that's the case here though

pipex: it doesn't work, but we could install node 22 on alpine 3.23 using `--repository=http://dl-cdn.alpinelinux.org/alpine/v3.22/main`

cywang117: I don't think using `--repository` is worth it, given system libs in the 3.23 image may differ than 3.22 image against which node 22 is built. This PR is a temporary measure until we figure out what to do with the `rpi`/armv6 architecture and Supervisor releases

pipex: "Nothing is more permanent than a temporary fix" :)

I'll let @Page- chime in since he wrote the original comment

webpack.config.js
```diff
@@ -60,7 +60,7 @@ externalModules.push(
 	new RegExp(
 		'^(' +
 			_.reject(maybeOptionalModules, requiredModules)
-				.map(RegExp.escape)
+				.map(_.escapeRegExp)
```
Page-, line 63: These `RegExp.escape` uses are the only thing that should need reverting for node 22, everything else should work just fine there

cywang117: Thanks, removed the TS downgrade

pipex (approved):

PR #2520, opened 2026-05-15
tests: Run compose tests against all platforms

klutchell:
Bake and run compose tests against all platforms
to confirm there are no runtime failures on ARM architectures.

Change-type: patch
See: https://balena.fibery.io/Work/Improvement/Unblock-the-raspberrypi-pipeline-4203

---
**Closing note (2026-08-14):** Parking this for now. Baking and running the compose tests across platforms doesn't reproduce the node24/armv6 runtime failure, because the arm/v6 container — while built correctly — still runs on a 64-bit compose host. The last commit (QEMU_CPU emulation) was an attempt to force V8 down the real code paths; closing until we can validate reproduction on real 32-bit hardware. Branch `kyle/docker-test-all` is preserved. Ref: Unblock-the-raspberrypi-pipeline-4203.

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -31,3 +31,22 @@ jobs:
         balena_os/amd64-supervisor
       custom_finalize_matrix: balena_os/aarch64-supervisor,
         balena_os/amd64-supervisor
+      # linux/arm/v6 pinned to x86_64 runners — not for performance, for fidelity.
+      # ARM64 hosts can execute 32-bit ARM binaries natively in AArch32 mode,
+      # which exposes the host's ARMv7+ features (NEON, idiv, advanced atomics)
+      # through HWCAP and /proc/cpuinfo. V8's runtime CPU detection then emits
+      # modern instructions — which work on the host but would trap on real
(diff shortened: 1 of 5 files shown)


pipex:
If these tests are passing, it means it's not really catching the node24/armv6 issue we observed no?

klutchell:
> If these tests are passing, it means it's not really catching the node24/armv6 issue we observed no?

Yeah, at least not in this setup where the arm/v6 container is built and run on an arm64 docker host. I suspect the container is being built correctly, but because the compose test is still on a 64-bit host it doesn't run into issues.

I can verify by building the container locally and try running it on my host, vs the pi.

.github/workflows/flowzone.yml
```diff
@@ -31,3 +31,12 @@ jobs:
         balena_os/amd64-supervisor
       custom_finalize_matrix: balena_os/aarch64-supervisor,
         balena_os/amd64-supervisor
+      # Force QEMU emulation for arm/v6/v7 on x86 for V8 testing with QEMU_CPU in entry.sh
```
klutchell: This was able to reproduce the issue, but it is super slow (as expected) so not ideal to merge as-is.

PR #2528, opened 2026-05-29
Restore service to single container configuration pre v18 update

pipex:
We are re-working our release strategy around the new multi-container supervisor after some initial customer concerns. This removes the new services from the supervisor release composition, reverting commit e62c021b0c065b8f66164fe661754d7f796206c2.

For devices already on v18, this removes any overrides created by the new helios service and restarts the supervisor to restore its single-container operation. The extra services and their images will get removed on the next target state apply

## Release Notes

This release rolls back the changes in v18, that introduced a new service `helios` as part of the Balena Supervisor composition. This roll back comes after some customers voiced their questions and concerns with some of the implementation details. We will be addressing those concerns before re-releasing this multi-container supervisor. An announcement and communications will precede the release. 

For any users that already had updated to v18.0.0, this release removes the extra service restoring single-service functionality.

### DIFF
--- a/src/supervisor.ts
+++ b/src/supervisor.ts
@@ -19,6 +19,10 @@ import * as firewall from './lib/firewall';
 import * as constants from './lib/constants';
 import * as uname from './lib/uname';
 
+// FIXME: remove when we release supervisor v19
+import * as servicesManager from './compose/service-manager';
+import * as supervisorMetadata from './lib/supervisor-metadata';
+
 const startupConfigFields: config.ConfigKey[] = [
 	'uuid',
 	'listenPort',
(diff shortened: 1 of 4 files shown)


src/supervisor.ts
```diff
@@ -32,6 +36,42 @@ const startupConfigFields: config.ConfigKey[] = [
...
+// FIXME: remove when we release supervisor v19
+async function recoverFromOverrides() {
+	const { listenPortOverride, apiEndpointOverride } = await config.getMany([
+		'listenPortOverride',
+		'apiEndpointOverride',
+	]);
+	if (listenPortOverride != null || apiEndpointOverride != null) {
+		log.info('Removing supervisor overrides');
+		await db.models('config').del().where({ key: 'listenPortOverride' });
+		await db.models('config').del().where({ key: 'apiEndpointOverride' });
```
cywang117: IMO these should happen after killing the helios services, in case killing the services throws for whatever reason.

pipex: yes, you are correct, nice catch

cywang117 (approved):
lgtm!

PR #2529, opened 2026-06-01
Report the supervisor core service to backend

pipex:
The supervisor cannot update itself so its core service is removed from the target state before applying steps. However we do want the service to be reported so it shows up in the new `Supervisor services` table on the dashboard.

Change-type: patch

### DIFF
--- a/src/lib/supervisor-metadata.ts
+++ b/src/lib/supervisor-metadata.ts
@@ -1,7 +1,4 @@
-export type SupervisorMetadata = {
-	uuid: string;
-	serviceName: string;
-};
+const SUPERVISOR_SVC_NAMES = ['main', 'balena-supervisor', 'core'];
 
 /**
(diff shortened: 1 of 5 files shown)


src/lib/supervisor-metadata.ts
```diff
@@ -1,8 +1,9 @@
 export type SupervisorMetadata = {
```
cywang117: This type is no longer used after the deletion below so can be removed

cywang117 (commented):
This LGTM, let me know when ready for a final pass 👍🏼

src/lib/supervisor-metadata.ts
pipex: nice catch, forgot about that type

cywang117 (approved):
LGTM!

PR #2532, opened 2026-06-05
Pass correct value for restart policy 'no'

pipex:
The supervisor was passing an empty string '' when receiving `restart_policy: no`, this now passes the direct name which is required by engine v25

Change-type: patch

### DIFF
--- a/src/compose/utils.ts
+++ b/src/compose/utils.ts
@@ -95,10 +95,6 @@ export function createRestartPolicy(name?: string): string {
 		return 'always';
 	}
 
-	if (name === 'no') {
-		return '';
-	}
-
 	return name;
 }
 


pipex:
Tested with balenaOS 6.12.3 and 7.2.0

cywang117 (approved):

schrockwell:
This change broke my deploy since balenaOS 7.x, so I'm sharing the knowledge in case anyone comes across the same issue.

I had a `docker-compose.yml` that specified `restart: no`, and this PR caused the supervisor to constantly reboot it. The supervisor always detected a config change upon container start (see log below), kicking it over to restart it, forever. I could not stop the container because the supervisor was holding on to the lock the entire time.

```
Replacing container for service (service name) because of config changes:
  Non-array fields: {"updated":{"restart":""}}
```

The fix was simple: change `docker-compose.yml` to have `restart: on-failure`.

pipex:
@schrockwell could you try with `restart: 'no'`? I think in yaml `no` is interpreted as `false`, which results in an empty string. I believe perhaps that was the reason of the pre-existing code to this change

PR #2533, opened 2026-06-17
Docs: clarify behavior for `PATCH /v2/device/tags` with device vs user changes

Page-:
https://balena.fibery.io/Work/Project/1031

Change-type: patch

# Description

Fixes # (issue)

# Type of change

# How Has This Been Tested?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1203,14 +1203,24 @@ Response:
 
 > **Introduced in supervisor v17.8**
 
-Schedule device tags to be set in the balena API, persisting the schedule across restarts.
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -1210,7 +1210,11 @@ reestablished.
 
 The body should be a JSON object of tag key-value pairs which will be reported to the
```
Page-, line 1212: Should also add more info on how retries work

shaunmulligan (approved):
Looks good

PR #2540, opened 2026-07-27
Use `/tmp` as host runtime dir

pipex:
The path used as the directory for the `local` volume driver needs to exist before starting the container. On reboot, if helios starts before the supervisor, the `/tmp/balena-supervisor` path won't exist resulting in the container failing to start before setting the restart policy and will never be started again.

This makes `/tmp` the host runtime directory, preventing the issue. Note that this also means that some paths that lived under `/tmp/balena-supervisor` now will be created under `/tmp`, e.g. `/tmp/helios.sock`, `/tmp/balenahup`

Change-type: patch

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,9 +40,9 @@ services:
       io.balena.features.host-os.board-rev: '1'
     environment:
       - DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket
-      # Needed for locks and balenahup. This needs to match the device
+      # Needed for locks and balenahup. This needs to match the path
       # used for the `runtime` volume
-      - BALENA_HOST_RUNTIME_DIR=/tmp/balena-supervisor
+      - BALENA_HOST_RUNTIME_DIR=/tmp
       # Only send error logs to dashboard
       - BALENA_LOGS_LEVEL=error
     volumes:
@@ -91,5 +91,4 @@ volumes:
     driver_opts:
       type: none
       o: bind
-      # XXX: this directory needs to exist on the host before the container is started
-      device: /tmp/balena-supervisor
+      device: /tmp


kb2ma:
Why use the top level `/tmp` for the Supervisor temporary directory rather than a sub-directory like `/tmp/balena-supervisor`? A subdirectory provides a more controlled, Supervisor-specific environment.

For example on my workstation, I see directories like these:
```
kbee@gazelle-balena:~/tmp$ ls -l /tmp
drwxrwxr-x  2 kbee kbee 4096 Jul 28 13:35 etcher
drwxrwxr-x  3 kbee kbee 4096 Jul  6 19:53 gimp
drwxrwxr-x  3 kbee kbee 4096 Jul 15 12:44 node-compile-cache
drwx------ 10 root root 4096 Jul 26 19:11 snap-private-tmp
drwx------  3 root root 4096 Jul  6 08:28 systemd-private-6c346939a27242628527dd2779176936-chrony.service-wmXBhq
drwx------  3 root root 4096 Jul  6 08:29 systemd-private-6c346939a27242628527dd2779176936-fwupd.service-qoaAYa
drwx------  3 root root 4096 Jul  6 08:27 systemd-private-6c346939a27242628527dd2779176936-systemd-resolved.service--rw-------  1 kbee kbee    0 Jul 22 20:36 tmpufcsyrz0.png
```

kb2ma:
Here is what I see on a device running the draft Supervisor for this PR. So now there are two temp directories -- `/tmp/balena-supervisor` for the Supervisor and `/tmp` for Helios, right?. Will Helios ever need to read/write to `/tmp/balena-supervisor`?

```
root@4c7101c:~# ls -l /tmp/
total 0
drwxr-xr-x 3 root root 100 Jul 29 00:15 balena-supervisor
srwxr-xr-x 1 root root   0 Jul 29 00:15 helios.sock
drwx------ 3 root root  60 Jul 29 00:14 systemd-private-0c243658420e4f2fbe6b14438c132e08-ModemManager.service-jH92Cf
drwx------ 3 root root  60 Jul 29 00:14 systemd-private-0c243658420e4f2fbe6b14438c132e08-bluetooth.service-e8r2NV
drwx------ 3 root root  60 Jul 29 00:15 systemd-private-0c243658420e4f2fbe6b14438c132e08-chronyd.service-dlE8aE
drwx------ 3 root root  60 Jul 29 00:14 systemd-private-0c243658420e4f2fbe6b14438c132e08-openvpn.service-sogsOq
drwx------ 3 root root  60 Jul 29 00:14 systemd-private-0c243658420e4f2fbe6b14438c132e08-systemd-logind.service-7p80mw

root@4c7101c:~# ls -l /tmp/balena-supervisor/
total 8
drwxr-xr-x 4 root root   80 Jul 29 00:15 services
-rw-r--r-- 1 root root 1679 Jul 29 00:20 state-report-cache
-rw-r--r-- 1 root root  492 Jul 29 00:15 target-state-config
```

pipex:
The way the container gets access to the host `/tmp` needs that any path under `/tmp/run` on the container must match the mountpoint for the runtime directory on the host, that is a path `/host/tmpdir/a/b/c` on the host must be retrievable on `/tmp/run/a/b/c` on the container. I looked for a way to keep the previous path but I couldn't think of a robust enough solution. I also didn't want to hardcode the `balena-supervisor` directory to ensure backwards compatibility


User apps don't have access to host directories so I don't think `/tmp` is not necessarily less controlled than `/tmp/balena-supervisor`

pipex:
> Will Helios ever need to read/write to /tmp/balena-supervisor?

Probably to read user app locks, but the directory will remain in control of the supervisor

kb2ma:
> I looked for a way to keep the previous path but I couldn't think of a robust enough solution.

I'm confused. Why is this a problem now when we have been using `/tmp/balena-supervisor` for v19?

kb2ma:
> Probably to read user app locks, but the directory will remain in control of the supervisor

At some point in the future I guess helios can add another variable to define this path (`/tmp/balena-supervisor`).

pipex:
> > I looked for a way to keep the previous path but I couldn't think of a robust enough solution.
> 
> I'm confused. Why is this a problem now when we have been using `/tmp/balena-supervisor` for v19?

because this changes the mount point configured on the volume to /tmp. You can look at 
 [how helios takes locks](https://github.com/balena-io/helios/blob/main/helios-state/src/tasks/app.rs#L234) to see how this invariant affects the code

That being said we were hardcoding the path in the volume before so maybe there isn’t a backwards compatibility issue, maybe I could limit writes to /tmp/balena-supervisor. I’ll test some things tomorrow

kb2ma (approved):
Able to test draft release. The `helios.sock` file is in `/tmp`. System seems healthy.

PR #2542, opened 2026-07-31
Add kill step for handover services without target

pipex:
This makes sure that a leftover container from a handover gets removed. Before this change, the supervisor would get stuck in a `noop` loop and the state would never settle.

Change-type: patch


kb2ma (approved):
Tested on RPi4 device with an app using hand-over strategy. Old container removed as expected.

PR #2543, opened 2026-08-03
Fix stale image entries causing an old image to be used after reboot

### LINKED ISSUE #2538: Duplicate/stale image DB rows after repeated local-mode pushes can cause wrong image to run after reboot
## Summary

After repeated `balena push <device-ip>` to the same local-mode device, the supervisor's `image` table accumulates **duplicate rows for the same service with different `dockerImageId` values**, and the stale row is sometimes the *only* one left that no longer matches what's actually tagged in the engine. On a device reboot (or any reconciliation that needs to resolve the image name back to a `dockerImageId`, e.g. via `inspectByDigest()`), the supervisor can pick the wrong/stale entry and (re)create a container against an old image.

This looks like a gap left behind by #1616 / PR #1749 ("Use tags instead of dockerIds to track supervised images in docker") — that PR fixed the *removal* side of image bookkeeping, but the *insert* side still creates a new row on every rebuild instead of updating the existing one, because of how `markAsSupervised()`/`upsertModel()` build their match condition.

## Where this comes from

`markAsSupervised()` upserts via:

```ts
// src/compose/images.ts
async function markAsSupervised(image: Image): Promise<void> {
	const formattedImage = format(image);
	await db.upsertModel(
		'image',
		formattedImage,
		// TODO: Upsert to new values only when they already match? This is likely a bug
		// and currently acts like an "insert if not exists"
		formattedImage,
	);
}
```

```ts
// src/db.ts
export async function upsertModel(modelName, obj, id, trx) {
	const k = trx ?? db;
	const n = await k(modelName).update(obj).where(id);
	if (n === 0) {
		return k(modelName).insert(obj);
	}
}
```

Because `id` here is the *entire new row* (including the brand-new `dockerImageId`), the `WHERE` clause can never match the previous row for that service once its content — and therefore its `dockerImageId` — has changed. Every rebuild inserts a new row rather than updating the old one.

This matters most for **local mode pushes**, because the CLI always reuses identical `name`/`appId`/`serviceName`/`commit` metadata across every push (`balena-cli/src/utils/device/deploy.ts`: `LOCAL_APPNAME = 'localapp'`, `LOCAL_RELEASEHASH = '10ca12e1ea5e'`, image always tagged `local_image_<serviceName>:latest`). So the *only* thing that changes between pushes is `dockerImageId` — exactly the field that defeats the upsert match.

Later, `inspectByDigest()` resolves an image name back to a `dockerImageId` by taking whichever DB row comes back first from an **unordered** query:

```ts
// src/compose/images.ts
const inspectByDigest = async (imageName: string) => {
	...
	for (const img of images) {
		if (img.dockerImageId != null) {
			// Assume that all db entries will point to the same dockerImageId, so use
			// the first one. If this assumption is false, there is a bug with cleanup
			return await docker.getImage(img.dockerImageId).inspect();
		}
	}
	...
};
```

That comment's stated assumption doesn't hold once duplicate rows exist.

## Evidence from a real fleet (supervisor v17.2.1 – v17.8.2)

Querying the `image` table (read-only) across 14 local-mode devices in one fleet: **11 of 14 had duplicate rows with conflicting `dockerImageId`s** for at least one service. On several devices, *neither* surviving row matched the `dockerImageId` currently tagged `local_image_<service>:latest` in the engine for some services — meaning any reconciliation that needs to resolve that name has no correct entry to fall back on. Example (service `rsync`, one affected device):

```
id=320  dockerImageId=02b32f50f475  <- stale
id=338  dockerImageId=9ea9061ebc1f  <- stale
id=380  dockerImageId=b6d6770a65f0  <- currently correct (matches engine tag)
```

Devices ranged from supervisor 17.2.1 to 17.8.2; the same `markAsSupervised`/`inspectByDigest` code (function-for-function) is present unchanged on the current `master` branch as of 2026-07-23.

## Steps to reproduce

1. Put a device in local mode.
2. `balena push <device-ip>` a compose project with at least one built (non-external) service.
3. Change the Dockerfile/source so the image content changes, and `balena push <device-ip>` again. Repeat a few times.
4. Inspect the supervisor's `image` table (`/data/database.sqlite` inside the `balena_supervisor`/`core` container) — observe multiple rows for the same service `name`, with different `dockerImageId`s, none of them removed.
5. Optionally: reboot the device and check whether the running container's image matches what was actually tagged most recently, versus what `docker ps`/`inspect` shows after boot.

## Suggested fix

- In `markAsSupervised`, match the existing row by identity fields that are stable across rebuilds (e.g. `name` + `appId`/`appUuid` + `serviceId`/`serviceName`) rather than the full row including `dockerImageId`, so the upsert actually updates in place.
- In `inspectByDigest`, order by something recency-based (e.g. `id DESC`) rather than relying on incidental row order, or better, once (1) is fixed there should only ever be one row to find.

cwbusacker:
# Description

A service rebuilt under the same image name leaves its `image` database entry pointing at the superseded docker id. That entry is then used to resolve the image when a container has to be recreated, so the device can come back up running an old image — most visibly after a reboot. This is easy to hit in local mode, where every `balena push` rebuilds under the same `local_image_<service>:latest` name.

Thanks @pipex for pushing back on the original scope — the first version of this PR only addressed duplicate rows, which as you noted is not the core issue. This now fixes the stale entry itself.

Fixes #2538

## The stale entry

`saveAndRemoveImages` decides whether to emit a `saveImage` step with:

```ts
const isNotSaved = !availableWithoutIds.some((img) => _.isEqual(img, targetImage));
```

`availableWithoutIds` is the database rows with **`dockerImageId` and `id` omitted**. When a rebuild changes only the built content, every compared field is identical, so `isNotSaved` is `false`, no `saveImage` step is emitted, `markAsSupervised` is never reached, and the entry keeps pointing at the old id indefinitely.

The information needed to detect this is already in the function. Target services resolve their `dockerImageId` from the engine (`app.ts`, via `inspectByName`), so `targetImageDockerIds[name]` holds the **current** id while `availableImages` holds the **stored** one. The new check compares them:

```ts
const isStale =
    isAvailableOnTheEngine &&
    availableImages.some(
        (availableImage) =>
            imageManager.isSameImage(availableImage, targetImage) &&
            availableImage.dockerImageId !== engineDockerImageId,
    );
```

This is purely additive — the two existing conditions are untouched.

## Why the upsert change is still here

The `markAsSupervised` match key is a prerequisite for the above, not an alternative to it. It previously matched on the entire row, including `dockerImageId`, so it could never find the row it was meant to update once that column changed.

With the staleness check in place, a `saveImage` step now fires on **every** rebuild. Against the old full-row match that would insert a brand new row each time, turning an occasional duplicate into one per rebuild. Matching on `name`/`appUuid`/`serviceName`/`commit` makes the re-save update in place instead.

As a side effect this also fixes duplicate rows arising from another path: the CLI derives `serviceId`/`imageId` from each service's position in the composition, so adding or removing a service shifts them for every service after it, which previously also defeated the full-row match. Reproduction and fleet data are in #2538.

## Changes

- `saveAndRemoveImages` flags an image for saving when the stored docker id no longer matches what the name resolves to on the engine.
- `markAsSupervised` matches on identity rather than the full row, so a re-save updates in place.
- `inspectByDigest` orders by `id desc`. With the above no new duplicates accumulate, so this only affects rows already on disk — happy to drop it if you would rather keep the diff minimal.
- The migration from the first revision has been removed, per review.

# Type of change

Change-type: patch

# How Has This Been Tested?

- New integration test in `application-manager.spec.ts` asserting a `saveImage` step is inferred when the stored docker id differs from the engine's. Verified passing against this change.
- New integration test in `images.spec.ts` covering the positional `serviceId`/`imageId` shift, verified to fail without the upsert change (`expected 2 to equal 1`) and pass with it.
- Full containerized suite (`npm run test:compose`) run green on the earlier revision of this branch.
- Reproduced end to end on a real device running an unmodified Supervisor: three services produced three clean rows, and adding a fourth ahead of them produced duplicate rows with byte-identical docker ids, confirming a content change is not required to trigger it.

### DIFF
--- a/src/compose/application-manager.ts
+++ b/src/compose/application-manager.ts
@@ -775,11 +775,27 @@ function saveAndRemoveImages(
 
 			// The image is not on the database but we know it exists on the
 			// engine because we could find it through inspectByName
-			const isAvailableOnTheEngine = !!targetImageDockerIds[targetImage.name];
+			const engineDockerImageId = targetImageDockerIds[targetImage.name];
+			const isAvailableOnTheEngine = !!engineDockerImageId;
+
+			// The stored Docker ID can become stale in the database when rebuilt under the same
+			// name (e.g. local mode push), causing the stale image to be used for recreation.
+			// See balena-os/balena-supervisor#2538.
+			const isStaleInDb =
+				isAvailableOnTheEngine &&
+				availableImages.some(
+					(availableImage) =>
+						availableImage.name === targetImage.name &&
+						availableImage.appUuid === targetImage.appUuid &&
+						availableImage.serviceName === targetImage.serviceName &&
+						availableImage.commit === targetImage.commit &&
+						availableImage.dockerImageId !== engineDockerImageId,
+				);
 
 			return (
 				(isActuallyAvailable && isNotSaved) ||
-				(!isActuallyAvailable && isAvailableOnTheEngine)
+				(!isActuallyAvailable && isAvailableOnTheEngine) ||
+				isStaleInDb
 			);
 		},
 	);
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -573,12 +573,14 @@ const inspectByDigest = async (imageName: string) => {
 		.models('image')
 		.where('name', 'like', `%${digest}`)
 		.orWhere({ name: imageName }) // Default to looking for the full image name
+		// markAsSupervised should now keep this to a single row per image, but
+		// order by the most recently written row as a fallback for any
+		// pre-existing duplicates left over from before that fix.
+		.orderBy('id', 'desc')
 		.select();
 
 	for (const img of images) {
 		if (img.dockerImageId != null) {
-			// Assume that all db entries will point to the same dockerImageId, so use
-			// the first one. If this assumption is false, there is a bug with cleanup
 			return await docker.getImage(img.dockerImageId).inspect();
 		}
 	}
@@ -735,9 +737,15 @@ async function markAsSupervised(image: Image): Promise<void> {
 	await db.upsertModel(
 		'image',
 		formattedImage,
-		// TODO: Upsert to new values only when they already match? This is likely a bug
-		// and currently acts like an "insert if not exists"
-		formattedImage,
+		// Match the declared image identity, not the full row: Docker and positional IDs
+		// can change after rebuilds or composition changes, causing duplicates instead of
+		// updates. See balena-os/balena-supervisor#2538.
+		{
+			name: formattedImage.name,
+			appUuid: formattedImage.appUuid,
+			serviceName: formattedImage.serviceName,
+			commit: formattedImage.commit,
+		},
 	);
 }
 
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -573,12 +573,14 @@ const inspectByDigest = async (imageName: string) => {
 		.models('image')
 		.where('name', 'like', `%${digest}`)
 		.orWhere({ name: imageName }) // Default to looking for the full image name
+		// markAsSupervised should now keep this to a single row per image, but
+		// order by the most recently written row as a fallback for any
+		// pre-existing duplicates left over from before that fix.
+		.orderBy('id', 'desc')
 		.select();
 
 	for (const img of images) {
 		if (img.dockerImageId != null) {
-			// Assume that all db entries will point to the same dockerImageId, so use
-			// the first one. If this assumption is false, there is a bug with cleanup
 			return await docker.getImage(img.dockerImageId).inspect();
 		}
 	}
@@ -735,9 +737,20 @@ async function markAsSupervised(image: Image): Promise<void> {
 	await db.upsertModel(
 		'image',
 		formattedImage,
-		// TODO: Upsert to new values only when they already match? This is likely a bug
-		// and currently acts like an "insert if not exists"
-		formattedImage,
+		// Match on the fields that identify "the same declared image for this
+		// service" rather than the full row. Matching on the full row (which
+		// includes dockerImageId) meant a rebuild that only changed the built
+		// content - same name/service/commit, new dockerImageId - could never
+		// find the row it should update, and every rebuild inserted a new row
+		// instead. That left stale rows behind indefinitely, and a later lookup
+		// (e.g. inspectByDigest) could resolve back to one of them instead of
+		// the current image. See balena-os/balena-supervisor#2538.
+		{
+			name: formattedImage.name,
+			appUuid: formattedImage.appUuid,
+			serviceName: formattedImage.serviceName,
+			commit: formattedImage.commit,
+		},
 	);
 }
 
--- a/src/compose/images.ts
+++ b/src/compose/images.ts
@@ -738,13 +738,14 @@ async function markAsSupervised(image: Image): Promise<void> {
 		'image',
 		formattedImage,
 		// Match on the fields that identify "the same declared image for this
-		// service" rather than the full row. Matching on the full row (which
-		// includes dockerImageId) meant a rebuild that only changed the built
-		// content - same name/service/commit, new dockerImageId - could never
-		// find the row it should update, and every rebuild inserted a new row
-		// instead. That left stale rows behind indefinitely, and a later lookup
-		// (e.g. inspectByDigest) could resolve back to one of them instead of
-		// the current image. See balena-os/balena-supervisor#2538.
+		// service" rather than the full row. Matching on the full row meant that
+		// any change to the remaining metadata - notably the positional
+		// serviceId/imageId, which shift for every service after one is added to
+		// or removed from a composition - could never find the row it was meant
+		// to update, so a new row was inserted alongside the old one. That left
+		// stale rows behind indefinitely, and a later lookup (e.g.
+		// inspectByDigest) could resolve back to one of them instead of the
+		// current image. See balena-os/balena-supervisor#2538.
 		{
 			name: formattedImage.name,
 			appUuid: formattedImage.appUuid,
--- a/src/migrations/M00013.js
+++ b/src/migrations/M00013.js
@@ -1,0 +1,34 @@
+// markAsSupervised() previously matched an existing `image` row using the
+// entire new row (dockerImageId included), so a rebuild that only changed the
+// built content could never find the row it should update and inserted a new
+// one instead. That left stale duplicate rows behind indefinitely. This
+// removes them, keeping only the most recently written row for each
+// (name, appUuid, serviceName, commit) group.
+// See balena-os/balena-supervisor#2538.
+export async function up(knex) {
+	const duplicateKeys = await knex('image')
+		.select('name', 'appUuid', 'serviceName', 'commit')
+		.groupBy('name', 'appUuid', 'serviceName', 'commit')
+		.havingRaw('count(*) > 1');
+
+	for (const key of duplicateKeys) {
+		const rows = await knex('image')
+			.where({
+				name: key.name,
+				appUuid: key.appUuid,
+				serviceName: key.serviceName,
+				commit: key.commit,
+			})
+			.orderBy('id', 'desc')
+			.select('id');
+
+		const idsToDelete = rows.slice(1).map((row) => row.id);
+		if (idsToDelete.length > 0) {
+			await knex('image').whereIn('id', idsToDelete).del();
+		}
+	}
+}
+
+export function down() {
+	throw new Error('Not implemented');
+}
--- a/test/integration/compose/application-manager.spec.ts
+++ b/test/integration/compose/application-manager.spec.ts
@@ -1739,6 +1739,52 @@ describe('compose/application-manager', () => {
 			.that.deep.includes({ name: 'main-image' });
 	});
 
+	it('should infer that an image should be saved if the stored docker id no longer matches the engine', async () => {
+		const targetService = await createService(
+			{ image: 'main-image' },
+			// The image name now resolves to a new id on the engine, as happens
+			// when a service is rebuilt under the same name by a local mode push
+			{ options: { imageInfo: { Id: 'sha256:new' } } },
+		);
+		const targetApps = createApps(
+			{
+				services: [targetService],
+				networks: [DEFAULT_NETWORK],
+			},
+			true,
+		);
+		const { currentApps, availableImages, downloading, containerIdsByAppId } =
+			createCurrentState({
+				services: [],
+				networks: [DEFAULT_NETWORK],
+				// The database has the image under the same name and with identical
+				// metadata, so nothing else marks it as needing a save, but it still
+				// points at the superseded id
+				images: [
+					{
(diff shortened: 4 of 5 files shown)


cywang117:
Hi @cwbusacker , this branch is on top of Supervisor 17.8.4. Could you rebase on top of latest (19.0.4) please, before we perform smoke tests on our side? Thanks

cwbusacker:
> Hi @cwbusacker , this branch is on top of Supervisor 17.8.4. Could you rebase on top of latest (19.0.4) please, before we perform smoke tests on our side? Thanks

@cywang117 Done.

src/compose/images.ts
```diff
@@ -735,9 +737,21 @@ async function markAsSupervised(image: Image): Promise<void> {
...
+		// Match on the fields that identify "the same declared image for this
+		// service" rather than the full row. Matching on the full row meant that
+		// any change to the remaining metadata - notably the positional
+		// serviceId/imageId, which shift for every service after one is added to
+		// or removed from a composition - could never find the row it was meant
+		// to update, so a new row was inserted alongside the old one. That left
+		// stale rows behind indefinitely, and a later lookup (e.g.
+		// inspectByDigest) could resolve back to one of them instead of the
+		// current image. See balena-os/balena-supervisor#2538.
+		{
```
pipex, line 743: This by itself doesn't help with the stale image id issue, `markAsSupervised` is only called if the position of the service in the composition changes. For a new image build, the function is not called and the database entry keeps pointing to the old image id, which is what causes the issue on reboot.

src/migrations/M00013.js
```diff
@@ -0,0 +1,34 @@
+// markAsSupervised() previously matched an existing `image` row using the
+// entire new row (dockerImageId included), so a rebuild that only changed the
+// built content could never find the row it should update and inserted a new
+// one instead. That left stale duplicate rows behind indefinitely. This
+// removes them, keeping only the most recently written row for each
+// (name, appUuid, serviceName, commit) group.
+// See balena-os/balena-supervisor#2538.
+export async function up(knex) {
```
pipex: Migrations are only for schema changes as they introduce a breaking point in the supervisor update path. There is already some cleanup that happens at supervisor startup during the application manager initialization

pipex:
I don't think that the duplicate images is the core issue, although is not ideal, local mode was never meant as a permanent state and leaving local mode cleans images and the database on the next supervisor start. 

This probably needs to be updated to actually handle the problem with stale images which is probably related to `saveAndRemoveImages` as mentioned on #2538

cwbusacker:
> I don't think that the duplicate images is the core issue, although is not ideal, local mode was never meant as a permanent state and leaving local mode cleans images and the database on the next supervisor start.
> 
> This probably needs to be updated to actually handle the problem with stale images which is probably related to `saveAndRemoveImages` as mentioned on #2538

I will expand the scope of this PR then to handle the stale images.

Update: This PR now fully handles the stale image issue. With this scope change, I believe this resolves Issue https://github.com/balena-os/balena-supervisor/issues/2367 and makes PR https://github.com/balena-os/balena-supervisor/pull/2423 no longer applicable.

src/migrations/M00013.js
cwbusacker: Deleted.

src/compose/images.ts
cwbusacker: You are correct. By itself, this does not fix the stale image id issue, but it did stand as a pre-requisite to preventing the stale image id. 

Fully Resolved in https://github.com/balena-os/balena-supervisor/pull/2543/commits/28ab0841065e491317faf348597960a8186998ee

src/compose/application-manager.ts
```diff
@@ -775,11 +775,27 @@ function saveAndRemoveImages(
...
-			const isAvailableOnTheEngine = !!targetImageDockerIds[targetImage.name];
+			const engineDockerImageId = targetImageDockerIds[targetImage.name];
+			const isAvailableOnTheEngine = !!engineDockerImageId;
+
+			// The image name is in the database, but the docker id stored against it
+			// is no longer the one that name resolves to on the engine. This happens
+			// when a service is rebuilt under the same image name, e.g. by a local
+			// mode push: the metadata is unchanged, so the checks above see nothing
+			// to do, yet the entry still points at the superseded image and would be
+			// used to recreate the container. See balena-os/balena-supervisor#2538.
```
cywang117: This comment is very verbose, could you make it more concise?

src/compose/application-manager.ts
```diff
@@ -775,11 +775,27 @@ function saveAndRemoveImages(
...
+			const isStale =
```
cywang117: Given this is a check of database entry staleness, `isStaleInDb` or similar would be better for conciseness

src/compose/application-manager.ts
```diff
@@ -775,11 +775,27 @@ function saveAndRemoveImages(
...
+				isAvailableOnTheEngine &&
+				availableImages.some(
+					(availableImage) =>
+						imageManager.isSameImage(availableImage, targetImage) &&
+						availableImage.dockerImageId !== engineDockerImageId,
```
cywang117:
```suggestion
                        availableImage.name === targetImage.name &&
						availableImage.appUuid === targetImage.appUuid &&
						availableImage.serviceName === targetImage.serviceName &&
						availableImage.commit === targetImage.commit &&
```

`isSameImage` and the dockerImageId check flag rows that the resulting `saveImage` step can't actually fix. `isSameImage` matches on digest, so it also picks up rows from other releases of an unchanged service, but `markAsSupervised` only updates the row matching all of `name`, `appUuid`, `serviceName`, and `commit`, so the condition outlives `saveImage`. Matching the same fields `markAsSupervised` does addresses the condition better IMO. Tested on a device and the suggested change also works and does not introduce regressions in non-local mode

src/compose/images.ts
```diff
@@ -735,9 +737,22 @@ async function markAsSupervised(image: Image): Promise<void> {
...
+		// Match on the fields that identify "the same declared image for this
+		// service" rather than the full row. Matching on the full row could never
+		// find the row it was meant to update whenever any other column changed -
+		// the dockerImageId of a rebuild, or the positional serviceId/imageId,
+		// which shift for every service after one is added to or removed from a
+		// composition - so a new row was inserted alongside the old one instead.
+		// Re-saving an image that is already in the database therefore has to
+		// match this way to update in place: without it, the stale entry check in
+		// application-manager would append a duplicate row on every rebuild.
+		// See balena-os/balena-supervisor#2538.
```
cywang117: This comment is way too verbose, can you make it more concise?

cywang117 (commented):
Thanks for your contribution! Tested on a device and all looks good, just some nits.

Additionally

> Should the duplicate rows already on affected devices be cleaned up?

Local mode passes `skipRemoval` so I don't think this is necessary.

src/compose/images.ts
cwbusacker: Done. Is that better?

src/compose/application-manager.ts
cwbusacker: Done. Is that better?

src/compose/application-manager.ts
cwbusacker: Done.

src/compose/application-manager.ts
cwbusacker: Agreed, and applied in https://github.com/balena-os/balena-supervisor/pull/2543/commits/83ad46abeceaea8b5e7671465cd00953d9471f01. Matching the same four fields
`markAsSupervised` keys on means any row flagged as stale is exactly the row the
`saveImage` step rewrites, so the condition clears after one step instead of
persisting and re-emitting on every pass. 

However, I assumed that you did not mean to delete line 793 as that would defeat the whole purpose of this PR and also cause a syntax error with a danging '&&'. So I kept the
`availableImage.dockerImageId !== engineDockerImageId` comparison at the end of the
chain.

cwbusacker:
> Thanks for your contribution! Tested on a device and all looks good, just some nits.
> 
> Additionally
> 
> > Should the duplicate rows already on affected devices be cleaned up?
> 
> Local mode passes `skipRemoval` so I don't think this is necessary.

@cywang117 happy to contribute! I have resolved all of your nits to the best of my
ability. Feel free to make any other changes to this branch if my resolutions are
not satisfactory.

> Local mode passes `skipRemoval` so I don't think this is necessary.

Agreed, dropping it. I think there's an even stronger reason than retention: `upsertModel`
is `.update(obj).where(id)`, so it updates *every* row matching the key. On a device with
pre-existing duplicates, the first `saveImage` rewrites `dockerImageId` on all of them,
making the leftovers inert rather than merely unlikely to be selected. The
`orderBy('id', 'desc')` in `inspectByDigest` covers what's left — rows spanning different
commits that share a digest.

One gap I'd note rather than fix: a duplicate row for a service later removed from the
composition never sees a `saveImage`, and in local mode never sees a `removeImage`
either, so it keeps a stale id indefinitely. Narrow, and it clears on leaving local mode.

I've removed the "Open question" section from the description since it's settled.

cywang117:
Thanks @cwbusacker , almost there. Could you amend the new commits generated from the review comments into the original code changes that they modified?

cwbusacker:
> Thanks @cwbusacker , almost there. Could you amend the new commits generated from the review comments into the original code changes that they modified?

Done https://github.com/balena-os/balena-supervisor/pull/2543/commits/cebb879390c024935652fd4a98e308ba43d5a0f4 and https://github.com/balena-os/balena-supervisor/pull/2543/commits/1f8e4683947e441e9b42cbf5443d033dfb776b36

cwbusacker:
@cywang117 @pipex, not exactly sure what your processes look like, but I think `policy-bot: master` is waiting for an approval before starting to run the deploy process. Either need a LGTM or a GitHub approval before it will run checks.

cywang117:
@cwbusacker We're looking into an issue with allowing external GHA runs in our pipeline, and will merge this once the issue is resolved

cywang117:
Hey @cwbusacker , we resolved the aforementioned issue. Could you rebase please, after which I'll trigger GHA and we can get this merged? Thanks for your patience!

cwbusacker:
> Hey @cwbusacker , we resolved the aforementioned issue. Could you rebase please, after which I'll trigger GHA and we can get this merged? Thanks for your patience!

Done!

cywang117:
`npm run prettify && npm run lint` should take care of the GH check failures

cwbusacker:
There doesn't appear to be a `prettify` script in the `package.json`. But I tried `npm lint-fix` and amended my commit that needed linting.

cwbusacker:
Same tests failed? Did my formatting changes not get in? :thinking:

pipex:
yes, it seems to be the same issue

```
Error: File src/compose/application-manager.ts hasn't been formatted with prettier
```

you are right that `npm run lint-fix` is the correct command. Could you check again that you pushed the formatting fixes?

cwbusacker:
Ok. Appears that I must've forgotten to stage the change before amending the commit. SOrry about that.

cywang117 (approved):
LGTM!

PR #2544, opened 2026-08-04
Update helios to v0.25.28

pipex:
The changes in balena-io/helios#149 cleanup the takeover breadcrumb in case of a failed takeover check to avoid unnecessary restarts.

Change-type: patch
Relates-to: balena-io/helios#149


kb2ma (approved):
Tested on live device; works as expected. Supervisor starts only once on reboot.

PR #2546, opened 2026-08-06
Prevent the supervisor from managing extension containers volumes

alexgg:
Relates to: https://balena.fibery.io/Work/Project/Hostapp-overlay-extensions---supervisor-integration-2333

### DIFF
--- a/src/compose/volume-manager.ts
+++ b/src/compose/volume-manager.ts
@@ -128,7 +128,9 @@ export async function removeOrphanedVolumes(
 		.map((m) => m.Name!)
 		.uniq()
 		.value();
-	const volumeNames = (dockerVolumes.Volumes ?? []).map((v) => v.Name);
+	const volumeNames = (dockerVolumes.Volumes ?? [])
+		.filter((v) => v.Labels?.['io.balena.image.class'] !== 'overlay')
+		.map((v) => v.Name);
 
 	const volumesToRemove = _.difference(
 		volumeNames,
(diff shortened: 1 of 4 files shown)


pipex:
Is this really necessary @alexgg? The extensions volumes [don't fit the naming convention expected](https://github.com/balena-os/balena-supervisor/blob/9b3c694674ffe0ed2ebfe34a00cdaf24121dab64/src/compose/volume.ts#L127) by the supervisor, which means [they will be ignored](https://github.com/balena-os/balena-supervisor/blob/9b3c694674ffe0ed2ebfe34a00cdaf24121dab64/src/compose/volume-manager.ts#L34) and treated as third party volumes

alexgg:
> Is this really necessary @alexgg? 

@pipex there are two changes:

An application calling into `/v2/cleanup-volumes` in the HUP window:

`removeOrphanedVolumes()` calls `docker.listVolumes()` directly and removes all volumes not referenced by a container or in the target state, so not being supervisor managed does not matter.

An extension volume whose container is gone (managed by the OS extension manager) is unreferenced until the extension is redeployed and calling `/v2/cleanup-volumes` in that window will delete it and break HUP.


For the second change, extension volumes are named `ext_<service>_<digest>_<sanitized-path>`, and the current regex matches `<digits>_` anywhere.  For example a volume `ext_svc_<digest>_var_lib_foo2_data` parses as appId 2 and would match. The caret `^` is what avoids this.

pipex:
Yeah, you are right @alexgg, sorry. I only looked superficially at the changes on friday. This looks Ok to merge

pipex (approved):
Nice one

PR #2548, opened 2026-08-10
Update debian docker image dependency to stable-slim

pipex:
For some reason `bookwork-slim` is no longer generating linux/arm/v5 images which is causing supervisor releases to fail to build

Change-type: patch


pipex:
Tested on a balenaOS device, log reporting works as expected

cywang117 (approved):
LGTM!

PR #2550, opened 2026-08-17
Add support for unix sockets as the apiEndpoint

pipex:
This allows the `apiEndpointOverride` to point to a file (socket), rather than provide a URL. When running alongside helios, this should allow the supervisor to communicate directly to that service, rather than going through the `service-relay`. 

This also limits API override communication to the poll/report, and ignores it for other communication with the API. This makes sure that provisioning works as before

Depends-on: balena-io/helios#153

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '2.3'
 
 x-helios-image: &helios-image
-  image: ghcr.io/balena-io/helios:0.25.28
+  image: ghcr.io/balena-io/helios:0.26.3
 
 services:
   # This service can only be named `main`, `balena-supervisor` or `core`
@@ -85,7 +85,7 @@ volumes:
     labels:
       # the cache version, can be bumped to force re-creation of the volume
       # on a new release
-      io.balena.private.version: "0"
+      io.balena.private.version: '1'
   runtime:
     driver: local
     driver_opts:
--- a/src/api-binder/report.ts
+++ b/src/api-binder/report.ts
@@ -1,4 +1,3 @@
-import url from 'url';
 import type { CoreOptions } from 'request';
 import { performance } from 'perf_hooks';
 import { setTimeout } from 'timers/promises';
@@ -17,6 +16,7 @@ import { shallowDiff, prune, empty } from '../lib/json';
 import { pathOnRoot } from '../lib/host-utils';
 import { touch, writeAndSyncFile } from '../lib/fs-utils';
 import pTimeout from 'p-timeout';
+import { resolveURL } from '../lib/api-helper';
 
 let lastReport: DeviceState = {};
 let lastReportTime = -Infinity;
@@ -53,9 +53,9 @@ async function report({ body, opts }: StateReport) {
 		);
 	}
 
-	const endpoint = url.resolve(
+	const endpoint = resolveURL(
 		apiEndpointOverride ?? apiEndpoint,
-		`/device/v3/state`,
+		'/device/v3/state',
 	);
 	const request = await getRequestInstance();
 
--- a/src/lib/api-helper.ts
+++ b/src/lib/api-helper.ts
@@ -17,6 +17,7 @@ import log from './supervisor-console';
 import memoizee from 'memoizee';
 import url from 'url';
 import type { BalenaModel } from 'balena-sdk';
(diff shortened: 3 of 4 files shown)


src/lib/api-helper.ts
```diff
@@ -24,14 +24,12 @@ export const getBalenaApi = memoizee(
...
-				'apiEndpoint',
-				'apiEndpointOverride',
-				'currentApiKey',
-			]);
+		const { apiEndpoint, currentApiKey } = await config.getMany([
+			'apiEndpoint',
+			'currentApiKey',
+		]);
 
-		const baseUrl = url.resolve(apiEndpointOverride ?? apiEndpoint, '/v7/');
```
pipex, line 34: The API configured for the helper is used for provisioning and a couple of local API methods. It doesn't need to go through the override

src/api-binder/report.ts
```diff
@@ -53,9 +53,9 @@ async function report({ body, opts }: StateReport) {
 		);
 	}
 
-	const endpoint = url.resolve(
+	const endpoint = resolveURL(
```
kb2ma, line 56: Should we be surprised that it's not necessary to specify use of a unix socket in this case, like in the update to poll()?

src/lib/api-helper.ts
```diff
@@ -265,3 +264,19 @@ export const provision = async (
...
 	}
 };
+
+/**
+ * Resolve a URL with respect to an endpoint
+ *
+ * If using unix domain sockets, formats the URL/path with a path relative to the root path assuming http
+ */
+export function resolveURL(endpoint: string, path: string): string {
+	// if the starts with `/`, assume a unix domain socket
```
kb2ma: Seems like a reasonable assumption. This comment doesn't quite read right though. Maybe remove "the"?

src/lib/api-helper.ts
kb2ma: Does it make sense to put this comment in the source for clarity? Seems like it would make a good function comment since this generic sounding function is exported.

kb2ma: So this comment/code means that these requests are sent directly to the cloud API? Does this mean that the price of not always using service-relay is that there are these few cases that now talk directly to the cloud? IOW, it's a kind of optimization.

src/api-binder/report.ts
pipex: No, because they use different libraries to make requests (got vs request). Don't remember why but changing that is outside the scope of this PR

src/lib/api-helper.ts
pipex: More of a optimization is a robustness issue, the pine client here is used for provisioning and for a few of the API methods that use the balena API, like [tags](https://docs.balena.io/reference/supervisor/supervisor-api#device-tags).  

There is no benefit to intercepting those requests and it could be potentially dangerous in the case of provisioning. 

For helios operation, intercepting communication with the state endpoint is really all we need.

I'll add a comment in the code

docker-compose.yml
```diff
@@ -85,7 +85,7 @@ volumes:
     labels:
       # the cache version, can be bumped to force re-creation of the volume
       # on a new release
-      io.balena.private.version: "0"
+      io.balena.private.version: '1'
```
pipex, line 88: Bumped this version in case any user rolled back the supervisor, this should force a fresh target state poll

src/lib/api-helper.ts
kb2ma: Is there some way to check provisioning with this updated code? The Supervisor must be incorporated into the image, correct?

pipex: FWIW, this change makes provisioning the same it was in v17, so there should not be any surprises. If you want to test it, you can remove the database, the helios containers and replace config.json using one with a provisioning key. You should a new device pop up on the dashboard on supervisor restart

kb2ma:
Tested several features, and all worked as expected.
* change device var
* update OS
* pin app
* move to new fleet

The only open question is a test for provisioning.

src/lib/api-helper.ts
kb2ma: That worked, thanks.

kb2ma (approved):
Re-provisioning test worked.

PR #2553, opened 2026-08-24
Use the 10.114.105.x subnet for supervisor services

pipex:
Supervisor services were using the docker default 172.17.x range for their network configuration. This could cause an issue if that range is used for other devices on the same network. This moves to the `10.114.x` subnet, mirroring what we do for the `supervisor0` network, which should reduce the probability of clashing.

Change-type: patch


cywang117 (approved):
Tested and LGTM!

PR #2557, opened 2026-08-26
Migrate Flowzone caller off pull_request_target

cywang117:
Reapplies 7f92059c, reverted in 96a908df. The duplicate release issue is fixed in flowzone 23.1.6.

Change-type: patch


pipex:
LGTM!

PR #2562, opened 2026-09-09
Tmp bind mounts

shaunmulligan:
Compose volumes entries that bind mount a host path are no longer dropped when the source is strictly under host `/tmp`, in either short or long syntax, mount propagation options are rejected, and anything rejected is logged. 

This lets containers reach host-created sockets such as `/tmp/argus_socket` from the NVIDIA runtime hostapp extension. 

This will need companion changes to @balena/compose-parser, the CLI and the builder allow these binds through.


src/lib/host-utils.ts
```diff
@@ -41,6 +41,9 @@ export const pathOnState = withBase(constants.stateMountPoint);
 export const pathExistsOnState = async (p: string) =>
 	await exists(pathOnState(p));
 
+// Returns a host path inside the supervisor's /tmp directory
+export const pathInSupervisorTmp = withBase(constants.supervisorTmpDir);
```
shaunmulligan, line 45: This is an optimization since we seem to be hardcoding the `tmp/balena-supervisor`, but happy to remove it as its not strictly need.

src/compose/utils.ts
```diff
@@ -619,6 +620,40 @@ export function compareArrayFields<T extends Dictionary<unknown>>(
...
+	constants.supervisorTmpDir,
+	constants.legacySupervisorTmpDir,
+];
+const within = (p: string, dir: string) => p === dir || p.startsWith(dir + '/');
+
+export const bindMountRule = `only host paths under ${bindRoot} with an optional ro/rw mode are allowed`;
+
+// Lexical check only; symlinks are not resolved
+export function isAllowedBindSource(source: string): boolean {
+	const p = path.posix.normalize(source).replace(/\/$/, '');
```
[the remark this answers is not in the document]
shaunmulligan: Yes, the lexical check does not stop a symlink planted under /tmp, and the reserved-directory exclusion falls to the same trick. We are keeping it as is, for two reasons.

The check cannot be made race-free from the supervisor. The engine resolves the bind source on the host at container-create time and has no nofollow option. Whatever we resolve beforehand, a compromised container that already has write access to the parent directory can swap the symlink between our check and the create. 

The trust model is the same as every Docker bind mount: the compose author is trusted with the host paths they name, and a fleet owner already has privileged and the `io.balena.features.*` labels. The `/tmp` restriction and the reserved-directory exclusion are guardrails against mistakes and against the low-effort case, not a security boundary.

If we want to defend against a compromised container planting symlinks, that has to be an engine change and a separate discussion.

src/compose/utils.ts
```diff
@@ -619,6 +620,40 @@ export function compareArrayFields<T extends Dictionary<unknown>>(
...
+	);
+}
+
+// Short: `source:target[:ro|rw]`. Long: `{type: 'bind', source, target, readOnly?}`
+export function isAllowedBindMount(volume: string | LongBind): boolean {
+	if (typeof volume === 'string') {
+		const [source, target, mode, ...rest] = volume.split(':');
+		return (
+			rest.length === 0 &&
+			!!target &&
```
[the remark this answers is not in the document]
shaunmulligan: Addressed in e008a84603dc74b93389cc0c296e0857874adb72

test/unit/compose/service.spec.ts
```diff
@@ -1476,6 +1492,193 @@ describe('compose/service: unit tests', () => {
...
+			]) {
+				expect(ComposeUtils.isAllowedBindMount(v as any), JSON.stringify(v)).to
+					.be.false;
+			}
+		});
+
+		it('should pass /tmp binds through to the container config unchanged', async () => {
+			const longBind = {
+				type: 'bind',
+				source: '/tmp/argus_socket',
```
shaunmulligan, line 1584: Not sure how we feel about having the argus_socket use case encoded in the tests? We could make it more generic if this feels wrong?

src/compose/utils.ts
```diff
@@ -619,6 +620,40 @@ export function compareArrayFields<T extends Dictionary<unknown>>(
 	}
 }
 
+const bindRoot = '/tmp';
+const reservedBindDirs = [
```
pipex: This blocks us from bind mounting `/tmp/balena-supervisor` for helios, which we could use to avoid this volume and have helios work directly with `/tmp/balena-supervisor` 

https://github.com/balena-os/balena-supervisor/blob/eeaf3d9c83ced4422b510bf54878989f9357eacb/docker-compose.yml#L95-L100

I understand the reason for doing it, but if the user wants to mess with the `/tmp/balena-supervisor` directory then they could already do so, so I'm inclined to remove it

src/compose/utils.ts
```diff
@@ -639,19 +639,21 @@ export function isAllowedBindSource(source: string): boolean {
 	);
 }
 
+const isAllowedBind = (source: string, target?: string) =>
```
pipex, line 633: Please squash this commit to the previous once before merging

pipex (commented):
Just a small observation

src/compose/utils.ts
shaunmulligan: Agreed, have removed it 👍

src/compose/utils.ts
shaunmulligan: done

test/unit/compose/service.spec.ts
cywang117: This seems fine to me 👍🏼

cywang117 (approved):
This LGTM, just a note that the PR description is outdated in that `/tmp/balena-supervisor` dir isn't blocked.

PR #2563, opened 2026-09-09
Compose runtime annotations

shaunmulligan:
Adds support for runtime and annotation fields in legacy supervisor to support hostapp-extensions work.


src/lib/contracts.ts
```diff
@@ -64,12 +64,18 @@ interface ServiceWithContract extends ServiceCtx {
...
 	'sw.os',
 	'sw.kernel',
 	'hw.device-type',
 	'arch.sw',
 ];
+
+// The compose spec level this supervisor implements. The builder attaches
+// a `sw.spec` requirement (slug `compose`) to services that use compose
+// features beyond the baseline, e.g. `runtime` or `annotations`
+export const composeSpecVersion = '2.0.0';
```
shaunmulligan, line 78: This I am unsure of, I assume it should be the v2 of compose spec, but needs clarification.

src/compose/service.ts
```diff
@@ -610,6 +611,7 @@ class ServiceImpl implements Service {
 				(opt: string) => !unsupportedSecurityOpt(opt),
 			),
 			usernsMode: container.HostConfig.UsernsMode ?? '',
+			runtime: container.HostConfig.Runtime ?? 'runc',
```
pipex: Let's not hardcode the `runc` runtime here, we should not pass a runtime if not given and the engine will use its default configuration

src/compose/service.ts
```diff
@@ -364,6 +364,13 @@ class ServiceImpl implements Service {
 		}
 		config.sysctls = _.mapValues(config.sysctls, String);
 
+		if (Array.isArray(config.annotations)) {
+			config.annotations = Object.fromEntries(
+				_.map(config.annotations, (v) => _.split(v, '=')),
```
pipex: nit: Can you do this without using `lodash`, I know we use it all over the place but we've been trying to avoid adding new uses?

pipex: From the definition in types/service.rs, annotations is a dict, so why are we splitting by `=` anyway? Also what if the string has multiple `=`

pipex (commented):
Some minor comments

shaunmulligan:
Tested the old-engine behaviour on a Raspberry Pi 5 running balenaOS 6.12.3 (balenaEngine v20.10.27, API 1.41). The engine accepts the create request and silently drops `HostConfig.Annotations`; the inspect output has no `Annotations` key. The supervisor then sees target `{"com.example.test":"1"}` vs current `{}` on every apply and recreates the service continuously, roughly one kill/install cycle every 6s with no backoff. Services without annotations on the same release are unaffected (0 recreates over 3 minutes). Log excerpt: 
```
13:55:39.749 [event]  Event: Service installed  "serviceName":"annotated"
13:55:41.831 [event]  Event: Service started    "serviceName":"annotated"
13:55:43.193 [debug]  Replacing container for service annotated because of config changes:
13:55:43.194 [debug]    Non-array fields:  {"added":{"annotations":{"com.example.test":"1"}},"deleted":{},"updated":{}}
13:55:46.049 [event]  Event: Service installed  "serviceName":"annotated"
13:55:46.567 [event]  Event: Service started    "serviceName":"annotated"
13:55:47.859 [debug]  Replacing container for service annotated because of config changes:
13:55:47.859 [debug]    Non-array fields:  {"added":{"annotations":{"com.example.test":"1"}},"deleted":{},"updated":{}}
13:55:51.942 [event]  Event: Service installed  "serviceName":"annotated"
```

On balenaEngine v25 (API 1.44) the field is stored and the service is stable.

This is just an edge case I think we will need to document.

src/compose/service.ts
shaunmulligan: Removed. The supervisor passes no runtime when the compose file has none. The engine writes its default runtime name into HostConfig.Runtime at create time, so a container created without one inspects as runc. To keep the comparison converging, the supervisor reads the engine default once from /info and treats a runtime equal to it as unset on both the target and current side. Verified on an Orin (v25, nvidia registered) and a Pi 5 (v20.10): no recreates for services on the default or on nvidia, and an explicit runtime: runc is also stable.

src/compose/service.ts
shaunmulligan: Done, no lodash. Array handling removed.

src/compose/service.ts
```diff
@@ -370,6 +370,11 @@ class ServiceImpl implements Service {
 		}
 		config.sysctls = _.mapValues(config.sysctls, String);
 
+		// Annotations are a map, like labels. Values are sent as strings
+		config.annotations = Object.fromEntries(
+			Object.entries(config.annotations ?? {}).map(([k, v]) => [k, String(v)]),
```
cywang117, line 375: NIT: `{}` in `config.annotations ?? {}` is redundant with `annotations: {}` in `_.defaults` on line 424 below

src/compose/service.ts
```diff
@@ -416,6 +421,7 @@ class ServiceImpl implements Service {
 
 		service.config = _.defaults(config, {
 			portMaps,
+			annotations: {},
```
cywang117, line 424: NIT: This is redundant with the aforementioned line 375

src/lib/contracts.ts
cywang117, line 78: This Supervisor version doesn't implement the entirety of compose spec v2, so we can't set this just yet. I think our original plan with this was to only declare it as a Supervisor/Helios provision once all spec v2 fields (i.e. newer compose fields) are implemented either in Supervisor or Helios. IMO, the correct place to start declaring this spec provision is in a version of the Supervisor that uses Helios with complete userapps support, which is in the future.

As of now, the builder is still on the legacy compose parser, which doesn't inject a spec requirement. In the next 2 cycles, a newer version of the CLI will target a newer builder endpoint which declares the spec requirement. So you could test this and get around the sw.spec restriction by staying on an older CLI version

cywang117 (commented):
Nice work, left some comments

src/lib/contracts.ts
pipex: My only concern with this is that the same composition including annotations will work with the older CLI version but not the new one which is kind of strange, but I think it's an acceptable compromise. I agree that we should bump the spec contract until helios.

